PR #13, opened 2019-04-29
INTLY-1802: Save product config into CR status

matskiv:
JIRA: https://issues.jboss.org/browse/INTLY-1802

This change enables us to save config of each product into CR status.
Example:
```
status:
  product_config:
    amqstreams:
      NAMESPACE: prefix-amq-streams
```


philbrookes:
discussed this with Oleg, and we decided against this approach.

PR #16, opened 2019-06-10
Use integreatly source

sedroche:


pkg/controller/installation/marketplace/manager.go
```diff
@@ -14,20 +14,20 @@ import (
 var providerLabel = "opsrc-provider"
 
 type operatorSources struct {
-	Redhat marketplacev1.OperatorSource
+	Integreatly marketplacev1.OperatorSource
 }
 
 func GetOperatorSources() *operatorSources {
 	return &operatorSources{
```
sedroche: We should really pull these from the cluster. They are hardcoded atm, I can update it in this PR if you want @philbrookes

philbrookes: if it's not too much work, then pulling from the cluster is fine. If it's a lot of work we can leave it hard-coded for now.

PR #18, opened 2019-06-12
Just keycloak install

sedroche:
@philbrookes 

This is adding the keycloak apis and creating keycloak service using the marketplace (without the realm for now)


pkg/controller/installation/installation_controller.go
```diff
@@ -36,11 +37,12 @@ func Add(mgr manager.Manager) error {
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 	conf := controllerruntime.GetConfigOrDie()
 	coreClient, err := kubernetes.NewForConfig(conf)
+	serverClient, err := pkgclient.New(conf, pkgclient.Options{})
```
sedroche: https://github.com/operator-framework/operator-sdk/blob/master/doc/user/client.md#non-default-client

See -> https://github.com/integr8ly/integreatly-operator/pull/18/files#diff-329c149b9583e9a4fcdb670a2ba866f0R178

pkg/controller/installation/products/config/rhsso.go
```diff
@@ -0,0 +1,27 @@
+package config
+
+import "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+
+type RHSSO struct {
+	config ProductConfig
+}
+
+func newRHSSO(config ProductConfig) *RHSSO {
```
philbrookes: you might be intending to address this later, but we will also need an `AddClient(name string, definition string)` and a `getClient(name string) string`, and a `SetOauthURL(url string)` and `GetOauthURL() string`.

This allows, for example 3scale, to have: 
```
rhssoConfig, err := configmanager.readRHSSO()
clientDetails := rhssoConfig.GetClient("3scale")
oauthUrl := rhssoConfig.GetOauthURL()
```
so that it can configure itself to use the cluster RHSSO

pkg/controller/installation/products/rhsso/reconciler.go
```diff
@@ -0,0 +1,190 @@
...
+func NewReconciler(client pkgclient.Client, serverClient pkgclient.Client, coreClient *kubernetes.Clientset, configManager config.ConfigReadWriter, instance *v1alpha1.Installation, clusterHasOLM bool) (*Reconciler, error) {
+	config, err := configManager.ReadRHSSO()
+	if err != nil {
+		return nil, err
+	}
+	if config.GetNamespace() == "" {
+		config.SetNamespace(instance.Spec.NamespacePrefix + defaultInstallationNamespace)
+	}
+	var mpm marketplace.MarketplaceInterface
+	if clusterHasOLM {
```
philbrookes: as we are not considering 3.x any more (for component installs), we can remove this if statement and assume it is always OLM.

pkg/controller/installation/types.go
```diff
@@ -40,7 +40,7 @@ func newWorkshopType() *Type {
 	return &Type{
 		products: []v1alpha1.ProductName{v1alpha1.ProductAMQStreams},
 		productOrder: map[int][]v1alpha1.ProductName{
-			1: {v1alpha1.ProductAMQStreams},
+			1: {v1alpha1.ProductRHSSO, v1alpha1.ProductAMQStreams},
```
philbrookes: I think RHSSO should be in phase 1, and AMQ Streams in phase 2. RHSSO should also be added to the managed install type.

pkg/controller/installation/products/config/rhsso.go
```diff
@@ -0,0 +1,27 @@
...
+	return &RHSSO{config: config}
+}
+
+func (a *RHSSO) GetNamespace() string {
```
philbrookes: can we change the `a *RHSSO` to `r *RHSSO` as this could get confusing if this class grows in complexity.

pkg/apis/addtoscheme_integreatly_v1alpha1.go
```diff
@@ -2,6 +2,7 @@ package apis
 
 import (
 	"github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+	aerogearv1 "github.com/integr8ly/integreatly-operator/pkg/apis/aerogear/v1alpha1"
```
philbrookes: can we not just import these from the keycloak-operator?

Like:
```
import kcv1alpha1 "github.com/integr8ly/keycloak-operator/pkg/apis/v1alpha1"
```

Rather than duplicating them in both repos.

pkg/controller/installation/products/config/rhsso.go
philbrookes: maybe that client definition should be an struct rather than a string actually

pkg/controller/installation/products/config/rhsso.go
philbrookes: this applies to all the methods on this struct.

pkg/apis/addtoscheme_integreatly_v1alpha1.go
sedroche: There is an issue with incompatible operator sdks between keycloak operator and the integreatly operator.

I have opened an issue for this -> https://github.com/integr8ly/integreatly-operator/issues/19

pkg/controller/installation/products/rhsso/reconciler.go
sedroche: :+1:

pkg/controller/installation/types.go
sedroche: :+1:

pkg/controller/installation/products/config/rhsso.go
sedroche: :+1: thumb

pkg/controller/installation/products/rhsso/reconciler.go
sedroche: https://github.com/integr8ly/integreatly-operator/pull/18/commits/90a12657dfd39c6569608da602134dcebf73d002

pkg/controller/installation/types.go
sedroche: https://github.com/integr8ly/integreatly-operator/pull/18/commits/cc802cf0bce3f08ba61474a88af4c338a6c55228

pkg/controller/installation/products/config/rhsso.go
sedroche: https://github.com/integr8ly/integreatly-operator/pull/18/commits/cd425b8b5a95b7b51ddfac0de8e1d92c3110ae68

PR #21, opened 2019-06-14
Add Keycloak realm creation

sedroche:
- Creating the keycloak realm
- adding awaiting operator phase which awaits the installplan being ready
- use server client to create components to avoid "no matches for kind" error


sedroche:
@aidenkeating Build was fixed. Thank you.

PR #26, opened 2019-06-25
Make operator admin of created namespaces

philbrookes:
- make operator have admin role in any namespaces it creates 
- add labels to everything we create

### DIFF
--- a/pkg/controller/installation/marketplace/manager.go
+++ b/pkg/controller/installation/marketplace/manager.go
@@ -57,6 +57,9 @@ func (m *MarketplaceManager) CreateSubscription(os marketplacev1.OperatorSource,
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: ns,
 			Name:      pkg,
+			Labels: map[string]string{
+				"integreatly": "yes",
+			},
 		},
 	}
 	err := m.client.Get(context.TODO(), pkgclient.ObjectKey{Name: sub.Name, Namespace: sub.Namespace}, sub)
@@ -69,6 +72,9 @@ func (m *MarketplaceManager) CreateSubscription(os marketplacev1.OperatorSource,
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "installed-" + os.Labels[providerLabel] + "-" + ns,
 			Namespace: "openshift-marketplace",
+			Labels: map[string]string{
+				"integreatly": "yes",
+			},
 		},
 		Spec: marketplacev1.CatalogSourceConfigSpec{
 			DisplayName:     os.Spec.DisplayName,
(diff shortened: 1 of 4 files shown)


pkg/controller/installation/products/rhsso/reconciler.go
```diff
@@ -101,8 +105,32 @@ func (r *Reconciler) handleAwaitingNSPhase() (v1alpha1.StatusPhase, error) {
...
+			RoleRef: v12.RoleRef{
+				Name:     "admin",
+				Kind:     "ClusterRole",
+				APIGroup: "rbac.authorization.k8s.io",
+			},
+			Subjects: []v12.Subject{
+				{
+					Kind:      "ServiceAccount",
+					Name:      "integreatly-operator",
+					Namespace: "005-test",
```
aidenkeating: Should this be a var of some kind? Same above.

pkg/controller/installation/products/amqstreams/reconciler.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/integr8ly/integreatly-operator/pkg/controller/installation/marketplace"
 	"github.com/integr8ly/integreatly-operator/pkg/controller/installation/products/config"
 	coreosv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	v12 "k8s.io/api/rbac/v1"
```
matskiv: IMHO `rbacv1` name would be better

pkg/controller/installation/products/rhsso/reconciler.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/integr8ly/integreatly-operator/pkg/controller/installation/marketplace"
 	"github.com/integr8ly/integreatly-operator/pkg/controller/installation/products/config"
 	coreosv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	v12 "k8s.io/api/rbac/v1"
```
matskiv: Same here, IMHO `rbacv1` name would be better

pkg/controller/installation/products/rhsso/reconciler.go
philbrookes: Whoops! Yes it should

pkg/controller/installation/products/amqstreams/reconciler.go
philbrookes: sure, I'll rename it

pkg/controller/installation/marketplace/manager.go
```diff
@@ -57,6 +57,9 @@ func (m *MarketplaceManager) CreateSubscription(os marketplacev1.OperatorSource,
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: ns,
 			Name:      pkg,
+			Labels: map[string]string{
+				"integreatly": "yes",
```
matskiv: We already use `integreatly-middleware-service=true` label in 3.x installation:
https://github.com/integr8ly/installation/search?q=integreatly-middleware-service&unscoped_q=integreatly-middleware-service
Maybe we should stick to that?

PR #27, opened 2019-07-01
Update README with OLM installation

sedroche:
@maleck13


maleck13:
Thanks will give it a go

PR #31, opened 2019-07-11
add mocks and basic tests

philbrookes:


pkg/controller/installation/installation_controller.go
```diff
@@ -140,6 +140,7 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 				log.Info("Error updating Installation resource. Requeue and retry.")
 				return reconcile.Result{
 					Requeue: true,
+					RequeueAfter: time.Second * 10,
```
maleck13: seems arbitrary?

pkg/controller/installation/installation_controller.go
```diff
@@ -154,6 +155,7 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
```
maleck13: If this is used every where should be a constant. Is this needed for test reasons?

pkg/controller/installation/installation_controller.go
philbrookes: yeah, we can change it, it just has to have a value, or else it does backoff logic, and eventually doesn't requeue it for ages.

pkg/controller/installation/installation_controller.go
philbrookes: as above, it just has to have a value. It's unrelated to tests, just noticed it while poking around with @sedroche

PR #35, opened 2019-07-12
Intly 2764 add unit tests

philbrookes:
Unit tests for codeready reconciler.

Reconciler needed some small refactoring to make it testable.

### DIFF
--- a/pkg/controller/installation/products/codeready/reconciler_test.go
+++ b/pkg/controller/installation/products/codeready/reconciler_test.go
@@ -3,6 +3,7 @@ package codeready
 import (
 	"errors"
 	"github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+	"github.com/integr8ly/integreatly-operator/pkg/controller/installation/marketplace"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"testing"
 
@@ -24,6 +25,7 @@ func TestCodeready(t *testing.T) {
 		FakeConfig           *config.ConfigReadWriterMock
 		FakeK8sClient        *k8sclient.Clientset
 		FakeControllerClient client.Client
+		FakeMPM				 marketplace.MarketplaceInterface
 	}{
 		{
 			Name:                 "test no phase without errors",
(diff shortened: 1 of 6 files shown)


pkg/controller/installation/products/codeready/reconciler_test.go
```diff
@@ -2,19 +2,39 @@ package codeready
...
 	"github.com/sirupsen/logrus"
 	pkgclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	kafkav1 "github.com/integr8ly/integreatly-operator/pkg/apis/kafka.strimzi.io/v1alpha1"
+	marketplacev1 "github.com/operator-framework/operator-marketplace/pkg/apis/operators/v1"
+	corev1 "k8s.io/api/core/v1"
 )
 
 func TestCodeready(t *testing.T) {
+	testScheme := runtime.NewScheme()
```
maleck13: why do we need to do this?

maleck13: If we need to do this, feels like it should by in an init or separate method?

pkg/controller/installation/products/codeready/reconciler_test.go
```diff
@@ -62,6 +83,380 @@ func TestCodeready(t *testing.T) {
...
+				ReadCodeReadyFunc: func() (ready *config.CodeReady, e error) {
+					return config.NewCodeReady(config.ProductConfig{}), nil
+				},
+				ReadRHSSOFunc: func() (*config.RHSSO, error) {
+					return config.NewRHSSO(config.ProductConfig{}), nil
+				},
+			},
+		},
+		{
+			Name:                 "test awaiting subscrit on failed RHSSO config",
```
maleck13: subscrit ?

pkg/controller/installation/products/codeready/reconciler_test.go
```diff
@@ -62,6 +83,380 @@ func TestCodeready(t *testing.T) {
...
+					Name:      "openshift",
+					Namespace: "rhsso",
+				},
+			}, &chev1.CheCluster{ObjectMeta: metav1.ObjectMeta{Name: defaultCheClusterName, Namespace: defaultInstallationNamespace}}),
+			FakeConfig: &config.ConfigReadWriterMock{
+				ReadCodeReadyFunc: func() (ready *config.CodeReady, e error) {
+					return config.NewCodeReady(config.ProductConfig{}), nil
+				},
+				ReadRHSSOFunc: func() (*config.RHSSO, error) {
+					return config.NewRHSSO(config.ProductConfig{
```
maleck13: this could be abstracted. Seems to be repeated

pkg/controller/installation/products/codeready/reconciler_test.go
philbrookes: yeah, this was a real arm ache, but without it the `client.Create` or `client.Get` etc would fail with unknown type.

pkg/controller/installation/products/codeready/reconciler_test.go
philbrookes: "subscrit on" = "subscription" lol, I'll update it!

PR #36, opened 2019-07-13
Initial pass at fuse installation

maleck13:
This still needs tests and a lot of cleaning up etc but  wanted to put it up as it shows a couple of new things

Two commits 1 vendor and 1 code

- 1 change ns creation to be common (sets up owner ref so will delete the ns when installation is deleted
- 2 reconcile calls everything every time so if you delete the ns it will get reset up
-3 will manage the subscription and catalogsourceconfig (IE delete and recreating as needed)

Still needs tests adding.

If you want to see it in action run locally and create an installation.
Delete the openshift-fuse ns and it will eventually be recreated and everything installed again

**Verification**

Create a new namespace
Create the operator source from integreatly
Deploy the integreatly operator
create the example installation in this repo in the same ns as the operator

After sso is installed there should be a ns integreatly-fuse created. Once installation is complete you should be able to login to fuse via the url in the ns. 
In the syndesis server the env var for 3scale service discovery should be set

Delete the installation CR
The Fuse ns should be removed.


pkg/controller/installation/types.go
```diff
@@ -38,10 +38,10 @@ func InstallationTypeFactory(installationType string) (error, *Type) {
 
 func newWorkshopType() *Type {
 	return &Type{
-		products: []v1alpha1.ProductName{v1alpha1.ProductRHSSO, v1alpha1.ProductAMQStreams},
+		products: []v1alpha1.ProductName{v1alpha1.ProductFuse},
 		productOrder: [][]v1alpha1.ProductName{
-			{v1alpha1.ProductRHSSO},
-			{v1alpha1.ProductCodeReadyWorkspaces, v1alpha1.ProductAMQStreams},
+			{},
+			{v1alpha1.ProductFuse},
```
maleck13: will reset

pkg/controller/installation/products/fuse/reconciler.go
```diff
@@ -0,0 +1,212 @@
...
+	reconciledPhase, err = r.handleAwaitingOperator(ctx)
+	if err != nil{
+		return v1alpha1.PhaseFailed, errors.Wrap(err, " failed to reconcile subscription for fuse ")
+	}
+	reconciledPhase, err = r.reconcileCustomResource(ctx,inst)
+	if err != nil{
+		return v1alpha1.PhaseFailed, errors.Wrap(err, " failed to reconcile fuse custom resource ")
+	}
+
+	logrus.Infof("End of reconcile Phase : ", reconciledPhase)
```
aidenkeating: We can use `r.logger` here to ensure any reconciler-specific labels are set

deploy/crds/examples/installation.cr.yaml
```diff
@@ -5,3 +5,4 @@ metadata:
 spec:
   type: workshop
   namespacePrefix: openshift-
+  createNamespaces: true
```
philbrookes: This is no longer required, maybe you need to rebase?

pkg/controller/installation/products/fuse/reconciler.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		logrus.Infof("error in handleAwaitingOperator ", err.Error())
+		if errors2.IsNotFound(err) {
+			logrus.Infof("error in handleAwaitingOperator is not found error ")
+			if sub != nil{
+				logrus.Infof("time since created %v", time.Now().Sub(sub.CreationTimestamp.Time).Seconds())
+			}
+			if sub != nil && time.Now().Sub(sub.CreationTimestamp.Time) > time.Second * 60{
+				// delete subscription so it is recreated
+				logrus.Info("removing subscription as no install plan ready yet will recreate")
```
aidenkeating: `r.logger` can be used here

pkg/controller/installation/products/fuse/reconcile_test.go
```diff
@@ -0,0 +1,19 @@
+package fuse
+
+import "testing"
+
+func TestReconcileNamespace(t *testing.T) {
+	cases := []struct {
+		Name string
+	}{
+		{
+			"test ns is created if enabled and does not exist",
```
philbrookes: great coverage ;)

pkg/controller/installation/products/fuse/reconciler.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	if fuseConfig.GetNamespace() == "" {
+		fuseConfig.SetNamespace(instance.Spec.NamespacePrefix + defaultInstallationNamespace)
+	}
+	if err = fuseConfig.Validate(); err != nil {
+		return nil, errors.Wrap(err, "keycloak codeReadyConfig is not valid")
+	}
+
+	mpm := marketplace.NewManager(client, rc)
```
philbrookes: This has to be passed in now, otherwise we can't mock this mpm instance. You can see how I've done it in my unit test PR.

pkg/controller/installation/products/fuse/reconciler.go
```diff
@@ -0,0 +1,212 @@
...
+		marketplace.GetOperatorSources().Integreatly,
+		r.Config.GetNamespace(),
+		defaultSubscriptionName,
+		marketplace.IntegreatlyChannel,
+		[]string{r.Config.GetNamespace()},
+		v1alpha12.ApprovalAutomatic)
+	if err != nil && !errors2.IsAlreadyExists(err) {
+		return v1alpha1.PhaseFailed, errors.Wrap(err, fmt.Sprintf("could not create subscription in namespace: %s", r.Config.GetNamespace()))
+	}
+	return r.handleAwaitingOperator(ctx)
```
philbrookes: `handleAwaitingOperator` is called after this function returns in the `Reconcile` function, was it intentional to call it twice?

pkg/controller/installation/products/fuse/reconciler.go
```diff
@@ -0,0 +1,212 @@
...
+			Components:syn.ComponentsSpec{
+				Server:syn.ServerConfiguration{
+					Features:syn.ServerFeatures{
+						ExposeVia3Scale:true,
+					},
+				},
+			},
+		},
+	}
+	if err := r.client.Get(ctx,pkgclient.ObjectKey{Name: cr.Name},&cr); err != nil{
```
philbrookes: Can't we just try to create and ignore an `IsAlreadyExists` error, so we only have to connect to the API once?

pkg/controller/installation/products/fuse/reconciler.go
```diff
@@ -0,0 +1,212 @@
...
+		if errors2.IsNotFound(err){
+			if err := r.client.Create(ctx,&cr); err != nil && !errors2.IsAlreadyExists(err){
+				return v1alpha1.PhaseFailed, errors.Wrap(err, "failed to create a syndesis cr when reconciling cutom resource")
+			}
+			return v1alpha1.PhaseNone,nil
+		}
+		return v1alpha1.PhaseFailed, errors.Wrap(err, "failed to get a syndesis cr when reconciling cutom resource")
```
philbrookes: Failing the whole installation for this seems a bit excessive. This could be triggered because the openshift API is being restarted for example.

pkg/controller/installation/products/fuse/reconciler.go
```diff
@@ -0,0 +1,212 @@
...
+	r.logger.Infof("checking installplan is created for subscription %s in namespace: %s", defaultSubscriptionName, r.Config.GetNamespace())
+	ip,sub, err := r.mpm.GetSubscriptionInstallPlan(defaultSubscriptionName, r.Config.GetNamespace())
+	if err != nil {
+		logrus.Infof("error in handleAwaitingOperator ", err.Error())
+		if errors2.IsNotFound(err) {
+			logrus.Infof("error in handleAwaitingOperator is not found error ")
+			if sub != nil{
+				logrus.Infof("time since created %v", time.Now().Sub(sub.CreationTimestamp.Time).Seconds())
+			}
+			if sub != nil && time.Now().Sub(sub.CreationTimestamp.Time) > time.Second * 60{
```
philbrookes: it gives up waiting for the installplan after a minute? I've seen it work after a longer pause than that, I'd be more inclined to put this at around 5 minutes, and in the fuse config.

pkg/controller/installation/products/fuse/reconciler.go
```diff
@@ -0,0 +1,212 @@
...
+
+				}); err != nil{
+					// not going to fail here will retry
+					r.logger.Error("failed to delete sub after install plan was not available for more than 20 seconds . Ignoring will retry ", err)
+				}
+			}
+			r.logger.Debugf(fmt.Sprintf("installplan resource is not found in namespace: %s", r.Config.GetNamespace()))
+			return v1alpha1.PhaseAwaitingOperator, nil
+		}
+		return v1alpha1.PhaseFailed, errors.Wrap(err, fmt.Sprintf("could not retrieve installplan in namespace: %s", r.Config.GetNamespace()))
```
philbrookes: I wouldn't fail here either.

pkg/controller/installation/products/fuse/reconciler.go
```diff
@@ -0,0 +1,208 @@
...
+	ns, err := r.nsReconciler.Reconcile(ctx, ns, inst)
+	if err != nil {
+		return v1alpha1.PhaseFailed, errors.Wrap(err, "failed to reconcile fuse namespace "+r.Config.GetNamespace())
+	}
+	if ns.Status.Phase == v1.NamespaceTerminating {
+		r.logger.Debugf("namespace %s is terminating, maintaining phase to try again on next reconcile", r.Config.GetNamespace())
+		return v1alpha1.PhaseAwaitingNS, nil
+	}
+	if ns.Status.Phase != v1.NamespaceActive {
+		return v1alpha1.PhaseAwaitingNS, nil
```
dimitraz: Just a question to understand how this works, does returning here not cause it to go into the next phase?

pkg/controller/installation/products/fuse/reconciler.go
```diff
@@ -0,0 +1,208 @@
...
+		return v1alpha1.PhaseFailed, errors.Wrap(err, " failed to reconcile fuse custom resource ")
+	}
+
+	logrus.Info("End of reconcile Phase : ", reconciledPhase)
+
+	return v1alpha1.StatusPhase(v1alpha1.PhaseCompleted), nil
+}
+
+func (r *Reconciler) reconcileNamespace(ctx context.Context, inst *v1alpha1.Installation) (v1alpha1.StatusPhase, error) {
+	if !inst.Spec.CreateNamespaces {
```
dimitraz: I think this can be removed since the namespace wouldn't be created if it existed

pkg/controller/installation/products/fuse/reconciler.go
maleck13: This isnt specific to fuse.

maleck13: moved to config package

pkg/controller/installation/products/fuse/reconciler.go
maleck13: hmm but how would we decide when to fail then? seems we would need some kind of back off logic. Perhaps we can leave it as is for now? If the reconcile logic is always called even when marked as failed it would give a chance to recover. So marking as failed might mean a temporary failure

pkg/controller/installation/products/fuse/reconciler.go
maleck13: see above

pkg/controller/installation/products/fuse/reconciler.go
maleck13: yep. Removed now

philbrookes:
This build is failing because you've pushed the branch straight to integreatly repo, instead of to your own fork. Move it to a fork and open a new PR and it will be corrected.

PR #38, opened 2019-07-16
fix issue with stale serverclient

philbrookes:
the `r.client` has issues with a stale cache, so I've removed it, and pass it in from the installation_controller, with a fresh client created on each reconcile.

### DIFF
--- a/pkg/controller/installation/products/amqstreams/reconciler.go
+++ b/pkg/controller/installation/products/amqstreams/reconciler.go
@@ -24,7 +24,7 @@ var (
 	defaultInstallationNamespace = "amq-streams"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/controller/installation/products/amqstreams/reconciler.go
```diff
@@ -24,15 +24,15 @@ var (
...
 	if err != nil {
 		return nil, err
 	}
 	if config.GetNamespace() == "" {
 		config.SetNamespace(instance.Spec.NamespacePrefix + defaultInstallationNamespace)
 	}
-	return &Reconciler{client: client,
+	return &Reconciler{
 		coreClient:    coreClient,
 		restConfig:    rc,
```
sedroche: Can you check to see if the `*rest.Config` is still needed? I think that was related to the `serverClient`

philbrookes: nice, yeah, removed that!

PR #41, opened 2019-07-16
Intly 2329 3scale

sedroche:


pkg/controller/installation/products/threescale/reconciler.go
```diff
@@ -0,0 +1,594 @@
...
+		r.namespace,
+		packageName,
+		"integreatly",
+		[]string{r.namespace},
+		coreosv1alpha1.ApprovalAutomatic)
+	if err != nil && !k8serr.IsAlreadyExists(err) {
+		return v1alpha1.PhaseFailed, err
+	}
+
+	ip, err := r.mpm.GetSubscriptionInstallPlan(packageName, r.namespace)
```
dimitraz: This error is never checked?

pkg/controller/installation/installation_controller.go
```diff
@@ -220,7 +212,7 @@ func (r *ReconcileInstallation) processStage(instance *v1alpha1.Installation, pr
 		if !(phase == string(v1alpha1.PhaseCompleted)) {
 			incompleteStage = true
 		}
-		reconciler, err := products.NewReconciler(r.context, v1alpha1.ProductName(product), r.client, r.restConfig, r.coreClient, configManager, instance)
```
philbrookes: the context has to be the first argument.

pkg/controller/installation/products/rhsso/reconciler.go
```diff
@@ -16,37 +16,66 @@ import (
...
 
-func NewReconciler(coreClient *kubernetes.Clientset, configManager config.ConfigReadWriter, instance *v1alpha1.Installation, mpm marketplace.MarketplaceInterface) (*Reconciler, error) {
+var CustomerAdminUser = &aerogearv1.KeycloakUser{
+	KeycloakApiUser: &aerogearv1.KeycloakApiUser{
+		Enabled:       true,
+		Attributes:    aerogearv1.KeycloakAttributes{},
+		UserName:      "customer-admin",
+		EmailVerified: true,
+		Email:         "customer-admin@example.com",
+		RealmRoles: []string{
```
philbrookes: these `realmRoles` don't actually do anything, might as well remove them.

pkg/controller/installation/products/rhsso/reconciler.go
```diff
@@ -16,37 +16,66 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
 	pkgclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var (
-	defaultInstallationNamespace = "rhsso"
-	keycloakName                 = "rhsso"
-	keycloakRealmName            = "openshift"
+	DefaultRhssoNamespace = "rhsso"
```
philbrookes: I don't think this should be public, other reconcilers can get this value from the RHSSO config.

pkg/controller/installation/products/rhsso/reconciler.go
```diff
@@ -16,37 +16,66 @@ import (
...
+	customerAdminPassword = "Password1"
+	keycloakName          = "rhsso"
+	KeycloakRealmName     = "openshift"
```
philbrookes: let's expose this through the config too?

pkg/controller/installation/products/threescale/asserts.go
```diff
@@ -0,0 +1,86 @@
...
+	oauthClient "github.com/openshift/client-go/oauth/clientset/versioned/typed/oauth/v1"
+	coreosv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	pkgclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func assertInstallationSuccessfullyReconciled(installation *v1alpha1.Installation, fakeSigsClient *SigsClientInterfaceMock, fakeThreeScaleClient *ThreeScaleInterfaceMock, fakeOauthClient oauthClient.OauthV1Interface) error {
+	ctx := context.TODO()
```
philbrookes: we shouldn't be summoning up contexts from nowhere anymore, let's pass in the root context

philbrookes: oh wait, this is to do with unit tests? In that case it's fine!

philbrookes: let's rename the file `_test.go`?

pkg/controller/installation/products/threescale/clients.go
```diff
@@ -0,0 +1,133 @@
+package threescale
```
philbrookes: let's rename the file `_test.go`

pkg/controller/installation/products/threescale/clients.go
```diff
@@ -0,0 +1,133 @@
...
+	oauthClient "github.com/openshift/client-go/oauth/clientset/versioned/typed/oauth/v1"
+	coreosv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/testing"
+	"net/http"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func GetClients(preReqObjects []runtime.Object, scheme *runtime.Scheme, appsv1PreReqs map[string]*appsv1.DeploymentConfig) (*config.Manager, *SigsClientInterfaceMock, appsv1Client.AppsV1Interface, oauthClient.OauthV1Interface, *ThreeScaleInterfaceMock, marketplace.MarketplaceInterface, error) {
```
philbrookes: does this one need to be public?

pkg/controller/installation/products/threescale/objects.go
```diff
@@ -0,0 +1,127 @@
+package threescale
```
philbrookes: this is test related code too right? So let's add _test.go to the filename?

pkg/controller/installation/products/threescale/reconciler.go
```diff
@@ -0,0 +1,592 @@
...
+						AWSRegion: string(bucket.Data["AWS_REGION"]),
+						AWSCredentials: v1.LocalObjectReference{
+							Name: s3SecretName,
+						},
+					},
+				},
+			},
+		},
+	}
+	err = serverClient.Get(ctx, pkgclient.ObjectKey{Name: apim.Name, Namespace: r.namespace}, apim)
```
philbrookes: Should we check this before loading in the secret?

pkg/controller/installation/products/threescale/reconciler.go
```diff
@@ -0,0 +1,592 @@
...
+			return v1alpha1.PhaseFailed, err
+		}
+	}
+
+	urlSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "credential-rhsso",
+		},
+	}
+	err = serverClient.Get(ctx, pkgclient.ObjectKey{Name: urlSecret.Name, Namespace: r.installation.Spec.NamespacePrefix + rhsso.DefaultRhssoNamespace}, urlSecret)
```
philbrookes: Let's change this to use `r.ConfigManager.ReadRHSSO().GetNamespace()`

pkg/controller/installation/installation_controller.go
sedroche: It is passed to `reconcile`, it doesn't seem to be needed here

pkg/controller/installation/products/rhsso/reconciler.go
sedroche: :+1:

pkg/controller/installation/products/rhsso/reconciler.go
sedroche: Yeah my bad. I'll update this

pkg/controller/installation/products/threescale/clients.go
sedroche: probably not. Ill update if it doesn't

pkg/controller/installation/installation_controller.go
philbrookes: oh, I thought it was moved, didn't realise it was just removed!

pkg/controller/installation/products/threescale/objects.go
sedroche: Have add `_test.go` to all these

pkg/controller/installation/products/threescale/reconciler.go
sedroche: I want to just reconcile on the objects themselves. if secret is not there, reconcile it. if ApiManager is not there reconcile it. etc.

philbrookes: if I'm reading the code right, it goes:
1. load in the secret data
2. create the apimanager object in memory
3. see if apimanager exists in api
4. create apimanager if it doesn't exist

I'm suggesting we change the order to:
1. see if apimanager exists, and if not
2.1. load in the secret data
2.2. create the apimanager object in memory
2.3. create apimanager if it doesn't exist

sedroche: I think we should avoid reconciling objects based on whether another object is there.
We should check if the object exists then reconcile that object.

PR #50, opened 2019-07-18
allow optional installation of products

maleck13:
provides a flag for use in development to choose which products to install ``` --products=rhsso,codeready-workspaces```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -29,6 +29,7 @@ import (
 var (
 	metricsHost       = "0.0.0.0"
 	metricsPort int32 = 8383
+	products    []string
 )
 var log = logf.Log.WithName("cmd")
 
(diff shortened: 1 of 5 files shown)


cmd/manager/main.go
```diff
@@ -47,6 +48,8 @@ func main() {
 	// controller-runtime)
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 
+	pflag.StringSliceVarP(&products, "products", "p", []string{"all"}, "--products=rhsso,fuse")
```
aidenkeating: Can we store `"all"` as a var somewhere, as we use it twice here and it could be used more in the future?

maleck13: yes will do it in my next pr

PR #51, opened 2019-07-18
Intly 2314 uninstall

philbrookes:
- Add owner refs to all created objects so they are deleted when the installation is deleted.
- Pass a context through from the controller, so it can cancel the context when the installation is deleted and prevent any executing reconcilers from recreating the deleted items.
- Update unit tests
- switch to ownerutil package for owner ref management.

## Verification
- Delete any existing integreatly operatorsources
- Create operator-source pointing at `philbrookes`
- Create test namespace
- Deploy integreatly-operator to that namespace from operatorHub
- Create installation CR
- Wait for installation to finish succesfully
- Delete the CR
- Wait and see all created objects are removed
- Create installation CR again
- See installation succeed again.


pkg/controller/installation/installation_controller.go
```diff
@@ -40,13 +40,23 @@ func Add(mgr manager.Manager, products []string) error {
...
 		return &ReconcileInstallation{client: mgr.GetClient(), scheme: mgr.GetScheme()}
 	}
-	return &ReconcileInstallation{client: mgr.GetClient(), scheme: mgr.GetScheme(), coreClient: coreClient, restConfig: restConfig, productsToInstall: products}
+	return &ReconcileInstallation{
+		client:            mgr.GetClient(),
+		scheme:            mgr.GetScheme(),
+		coreClient:        coreClient,
+		restConfig:        restConfig,
+		productsToInstall: products,
+		context:           ctx,
```
maleck13: It is not considered good practice to pass context via a struct property. The general accepted pattern is to pass it as the first argument in a function call
``` Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. ```  https://golang.org/pkg/context/

File: pkg/controller/installation/installation_controller.go [hunk 2]
```diff
@@ -86,20 +96,30 @@ type ReconcileInstallation struct {
 	coreClient        *kubernetes.Clientset
 	restConfig        *rest.Config
 	productsToInstall []string
+	context           context.Context
+	cancel            context.CancelFunc
 }
 
 // Reconcile reads that state of the cluster for a Installation object and makes changes based on the state read
 // and what is in the Installation.Spec
 func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
maleck13: for instance it would normally passed here ```(ctx, request)```

pkg/controller/installation/products/amqstreams/reconciler.go
```diff
@@ -45,17 +47,17 @@ type Reconciler struct {
 	mpm           marketplace.MarketplaceInterface
 }
 
-func (r *Reconciler) Reconcile(inst *v1alpha1.Installation, serverClient pkgclient.Client) (v1alpha1.StatusPhase, error) {
+func (r *Reconciler) Reconcile(inst *v1alpha1.Installation, serverClient pkgclient.Client, ctx context.Context) (v1alpha1.StatusPhase, error) {
```
maleck13: generally considered good practice to have context as the first argument

pkg/controller/installation/products/fuse/reconciler.go
```diff
@@ -56,8 +57,7 @@ func NewReconciler(coreClient kubernetes.Interface, configManager config.ConfigR
 	}, nil
 }
 
-func (r *Reconciler) Reconcile(inst *v1alpha1.Installation, serverClient pkgclient.Client) (v1alpha1.StatusPhase, error) {
-	ctx := context.TODO()
+func (r *Reconciler) Reconcile(inst *v1alpha1.Installation, serverClient pkgclient.Client, ctx context.Context) (v1alpha1.StatusPhase, error) {
```
maleck13: same here ctx as first arg is the standard pattern

pkg/controller/installation/installation_controller.go
philbrookes: I have to store it here, for it to be the same context across multiple reconciles. If I create it in the Reconcile function it would be a different context on each reconcile and the call to cancel wouldn't cancel the other reconciles context.

pkg/controller/installation/installation_controller.go
philbrookes: I don't call this method, it's called from the operator-sdk code.

pkg/controller/installation/products/amqstreams/reconciler.go
philbrookes: sure, I'll move the arg list

pkg/controller/installation/installation_controller.go
```diff
@@ -86,20 +96,30 @@ type ReconcileInstallation struct {
...
 			return reconcile.Result{}, nil
 		}
 		return reconcile.Result{}, err
 	}
 
+	// if the CR is being deleted,
+	// cancel this context to kill all
+	// ongoing requests to the API and exit
+	if instance.DeletionTimestamp != nil {
+		r.cancel() //cancel context
```
sedroche: Could you explain why we need to do this?

philbrookes: This sends the cancel to the context which kills any upcoming API requests from any other reconciles that may be in progress.

The issue this is solving is:
--> reconcile starts
--> RHSSO is reconciled
-----> CR is deleted
-----> Cluster clears all items owned by the CR
--> next products are reconciled and recreate all the objects they see are now missing

pkg/controller/installation/installation_controller.go
philbrookes: it has to be stored at some point, right? Whatever owns the context needs to store it... 

The installation_controller is what owns this context.

PR #52, opened 2019-07-19
Ensure AMQ Online config is continuously reconciled

aidenkeating:
Currently, the host of the AMQ Online config is never set, meaning
that if other products ever depend on this it will not be available.

This change ensures that the host is set on the config once it
becomes available.

Validation:
- Run the operator, either from fresh or with AMQ Online already setup
- Wait for the config reconciliation to end
- Check the config ConfigMap in the Namespace the operator watches
- Ensure HOST is now set for AMQ Online


aidenkeating:
@JameelB Mind taking a look?

PR #57, opened 2019-07-21
Refactor away need for each product to reconcile ns and subscription

maleck13:
Removes the need for each product to reconcile ns and subscription as these are common concerns.


pkg/resources/reconciler.go
```diff
@@ -0,0 +1,137 @@
...
+		if !errors2.IsNotFound(err) {
+			return v1alpha1.PhaseFailed, errors.Wrap(err, fmt.Sprintf("could not retrieve namespace: %s", ns.Name))
+		}
+		prepareNS(ns, inst)
+		if err = client.Create(ctx, ns); err != nil {
+			return v1alpha1.PhaseFailed, errors.Wrap(err, fmt.Sprintf("could not create namespace: %s", ns.Name))
+		}
+		return v1alpha1.PhaseNone, nil
+	}
+	fmt.Println(" ReconcileNamespace ", ns)
```
aidenkeating: Can we remove this?

pkg/resources/reconciler.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		return v1alpha1.PhaseFailed, errors.Wrap(err, fmt.Sprintf("could not retrieve installplan and subscription in namespace: %s", namespace))
+	}
+
+	logrus.Debugf("installplan phase is %s", ip.Status.Phase)
+	if ip.Status.Phase != v1alpha12.InstallPlanPhaseComplete {
+		logrus.Infof("%s install plan is not complete yet ", subscriptionName)
+		return v1alpha1.PhaseAwaitingOperator, nil
+	}
+	logrus.Infof("fuse online install plan is complete. Installation ready ")
```
aidenkeating: This should probably be general, instead of mentioning `fuse`

pkg/resources/reconciler.go
```diff
@@ -0,0 +1,137 @@
...
+		return v1alpha1.PhaseFailed, errors.Wrap(err, "failed to update the ns definition ")
+	}
+	if ns.Status.Phase != v1.NamespaceActive {
+		return v1alpha1.PhaseAwaitingNS, nil
+	}
+	return v1alpha1.PhaseNone, nil
+}
+
+func (r *Reconciler) ReconcileSubscription(ctx context.Context, subscriptionName, namespace string, client pkgclient.Client) (v1alpha1.StatusPhase, error) {
+	logrus.Infof("reconciling subscription %s from channel %s in namespace: %s", subscriptionName, "integreatly", namespace)
```
aidenkeating: We can replace the hardcoded `"integreatly"` with `marketplace.IntegreatlyChannel`here

maleck13:
noticed minor bug will fix before merge

PR #58, opened 2019-07-22
[WIP] Intly 2337 launcher installation

austincunningham:


pkg/controller/installation/products/launcher/reconciler.go
```diff
@@ -0,0 +1,190 @@
...
+
+func (r *Reconciler) handleAwaitingOperator(ctx context.Context, client pkgclient.Client) (v1alpha1.StatusPhase, error) {
+	r.logger.Infof("checking installplan is created for subscription %s in namespace: %s", defaultSubscriptionName, r.Config.GetNamespace())
+	ip, sub, err := r.mpm.GetSubscriptionInstallPlan(defaultSubscriptionName, r.Config.GetNamespace())
+	if err != nil {
+		if k8serr.IsNotFound(err) {
+			if sub != nil {
+				logrus.Infof("time since created %v", time.Now().Sub(sub.CreationTimestamp.Time).Seconds())
+			}
+			// should be triggered ever 3 minuets if no install plan found
```
aidenkeating: This should probably be uncommented or removed

deploy/crds/installation.crd.yaml
```diff
@@ -35,8 +35,18 @@ spec:
...
+              type: string
+            selfSignedCerts:
+              type: boolean
+            githubOauthClientId:
+              type: string
+            githubOauthClientSecret:
+              type: string
           required:
             - type
+            - githubOauthClientId
```
aidenkeating: These GitHub credentials aren't currently required in the installer. We could follow the same pattern of allowing for placeholder values until the CR is updated to be more flexible? Allowing for the GitHub OAuth app to be created after the cluster has been provisioned.

pkg/controller/installation/types.go
```diff
@@ -52,6 +54,7 @@ func newManagedType() *Type {
 		productOrder: [][]v1alpha1.ProductName{
 			{v1alpha1.ProductRHSSO},
 			{v1alpha1.ProductCodeReadyWorkspaces, v1alpha1.ProductFuse},
+			{v1alpha1.ProductLauncher, v1alpha1.ProductCodeReadyWorkspaces},
```
aidenkeating: CodeReady is included twice here, and Launcher is in a third stage. Should Launcher be included in the second stage on L56?

JameelB: These will be changed as the PR is still in progress. Launcher will be included in the second stage and the third one should be removed

deploy/crds/installation.crd.yaml
JameelB: Agree with the placeholder, the github oauth secret should be updated on reconcile. I think having a placeholder here would work and that would prevent the Launcher console giving an error due to missing oauth client id and secret

pkg/controller/installation/products/launcher/reconciler.go
JameelB: This will be removed as I don't think we'd need this in Launcher

austincunningham:
closing in favor of https://github.com/integr8ly/integreatly-operator/pull/61

PR #69, opened 2019-07-30
INTLY-2377: Add Openshift 4 as IDP for RHSSO

matskiv:
JIRA: https://issues.jboss.org/browse/INTLY-2377
Adding OAuthClient creation and configuration for OpenShift 4 IDP for RH SSO.

Complementary manifests PR: https://github.com/integr8ly/manifests/pull/40
I pushed manifests from the above PR into my "omatskiv" application registry namespace on quay.

Verification steps:
- create operator source with "registryNamespace: omatskiv"
- create testing NS and create Installation CR in it
- run operator from this PR (optionally, to minimize installation time and size, pass `"--products=rhsso,codeready-workspaces"` parameter when running operator locally)
- once codeready gets provisioned, try logging into it


pkg/controller/installation/products/rhsso/reconciler.go
```diff
@@ -237,3 +229,58 @@ func (r *Reconciler) exportConfig(ctx context.Context, serverClient pkgclient.Cl
 	}
 	return nil
 }
+
+func (r *Reconciler) setupOpenshiftIDP(ctx context.Context, kcr *aerogearv1.KeycloakRealm, serverClient pkgclient.Client) error {
+	_, err := r.oauthv1Client.OAuthClients().Get(oauthId, metav1.GetOptions{})
+	if err != nil && k8serr.IsNotFound(err) {
+		oauthClient := &oauthv1.OAuthClient{
```
maleck13: No need to do this now but in my PR there is a ReconcileOauthClient method

pkg/controller/installation/products/rhsso/reconciler.go
```diff
@@ -197,7 +183,13 @@ func (r *Reconciler) handleProgressPhase(ctx context.Context, serverClient pkgcl
 			return v1alpha1.PhaseFailed, errors.Wrap(err, "failed to write rhsso config")
 		}
 
-		r.logger.Info("keycloak has successfully processed the keycloakRealm")
+		err = r.setupOpenshiftIDP(ctx, kcr, serverClient)
+		if err != nil {
+			logrus.Errorf("Failed to setup Openshift IDP %v", err)
```
maleck13: rather than logging here, which can create extra noise I would prefer we wrap the error and return it up the call stack

matskiv:
@maleck13 I updated the PR as requested, can you take another look?

PR #75, opened 2019-08-06
creating solution explorer manifest

philbrookes:
create manifest configmap for solution explorer

## Verification
- Create test namespace
- Create installation CR (selfSignedCerts: true)
- Run this code in that namespace
- Once solution explorer goes to phase complete see a manifest secret in the solution explorer namespace containing the values of all completed products.


philbrookes:
bah, wrong jira number in branch name

PR #76, opened 2019-08-07
first pass at an integreatly selectorSyncSet

philbrookes:
SelectorSyncSet tested and deployed through OSD.

Cannot really verify this, but I have personally verified it with the OSD team, and over bluejeans with @maleck13 and @sedroche


sedroche:
One question about the s3 secrets?
https://github.com/integr8ly/integreatly-operator/blob/master/deploy/s3-secrets.yaml

philbrookes:
@sedroche, oops merged before I saw that. I created those secrets manually, before creating the installation CR manually.

PR #78, opened 2019-08-08
add support for dedicated like setup

maleck13:
**Verification**

Against an OS 4 cluster create a namespace ```mynamespace``` with an admin user
then run
```
cd scripts
./dedicated-setup.sh
```
once the dedicated operator is up and running
logout of your cluster and login back in using the ```test Identity Provider```
login with the customer-admin user
The password for this user is outputted from the script
You should only be able to see the mynamespace project

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,21 @@ The username is `admin`, and the password can be retrieved by running:
 oc get dc sso -n integreatly-rhsso -o jsonpath='{.spec.template.spec.containers[0].env[?(@.name=="SSO_ADMIN_PASSWORD")].value}'
 ```
 
+## Setting up your cluster to be OSD like
+
+To setup your cluster to have dedicated admins run the ```dedicated-setup.sh``` script
+```
+cd scripts
+./dedicated-setup.sh
+
+```
+
+If you want to remove the opertor run
+
+```bash
+cd scripts
+./dedicated-cleanup.sh
+```
 
 ## Deploying to a Cluster with OLM
 
--- a/scripts/dedicated-setup.sh
+++ b/scripts/dedicated-setup.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+
+
+PASSWORD=$(openssl rand -base64 12)
+RELEASE=release-4.1
+BASE_HREF=https://raw.githubusercontent.com/openshift/dedicated-admin-operator/${RELEASE}
+
+echo Targeting $(oc whoami --show-server)
+sleep 1
+
+
+echo Setting up htpasswd IDP
+
+if [[ ! -f "htpasswd" ]]; then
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -74,6 +74,14 @@ The username is `admin`, and the password can be retrieved by running:
 oc get dc sso -n integreatly-rhsso -o jsonpath='{.spec.template.spec.containers[0].env[?(@.name=="SSO_ADMIN_PASSWORD")].value}'
 ```
 
+## Setting up your cluster to be OSD like
+
+To setup your cluster to have dedicated admins run the ```dedicated-setup.sh``` script
```
dimitraz: You could also point out the cleanup step

sedroche:
Is it the case that this removes any previous htpasswd users (opentlc-mgr) in pds?

Can dedicated-admin create another user and bump to admin?

scripts/dedicated-setup.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
+
+
+PASSWORD=$(openssl rand -base64 12)
```
sedroche: Can this be something simple? (good old Password1)

README.md
```diff
@@ -74,6 +74,14 @@ The username is `admin`, and the password can be retrieved by running:
...
+```
+cd scripts
```
sedroche: Make command would be nice and save using changing directory

maleck13:
> Is it the case that this removes any previous htpasswd users (opentlc-mgr) in pds?
> 
> Can dedicated-admin create another user and bump to admin?

No dedicated admin can't do that. You would need to add another user to the htpassword file and then login with system:admin and grant it cluster-admin. 

I think this is less of an issue currently as we can get our own clusters now

README.md
matskiv: You can run `./scripts/dedicated-setup.sh` to save a `cd`

scripts/dedicated-setup.sh
maleck13: I'd prefer to not use that. The script outputs the password

README.md
sedroche: More the make command as per how we do other things in the repo. Not a huge issue

maleck13: your wish is my command :)

maleck13:
updated with ```make setup/dedicated ``` and readme change

sedroche:
> No dedicated admin can't do that. You would need to add another user to the htpassword file and then login with system:admin and grant it cluster-admin.
> 
> I think this is less of an issue currently as we can get our own clusters now

Just for remote clusters where I don't think we can login to a remote cluster with system:admin. In this scenario unless you are currently logged in you are stuck on dedicated admin perms.

PR #80, opened 2019-08-12
allow any auth user to login to fuse

maleck13:
**Verification**

Create operator source from https://raw.githubusercontent.com/integr8ly/manifests/3ec60ab0af1dc80a30ba14b10fc268c85bfb3dac/operator-source.yml

```
oc create -f https://raw.githubusercontent.com/integr8ly/manifests/3ec60ab0af1dc80a30ba14b10fc268c85bfb3dac/operator-source.yml
```

Wait for the custom operators to show up

Provision the integreatly operator

create the installation 
```
oc create -f https://raw.githubusercontent.com/integr8ly/integreatly-operator/master/deploy/crds/examples/installation.cr.yaml -n <ns>
```

- Fuse should be installed
- Once fuse is installed setup the the htpassword provider
```
./scripts/setup-htpass-idp.sh
```

Copy the url of the fuse console

Logout and log in to the fuse console using the customer-admin user created earlier. 
You should be able to login without issue.


mikenairn:
Followed the verification instructions as described and was able to login to the fuse console as the "customer-admin" user.

@maleck13 Are the travis build failures something we need look into?

maleck13:
@mikenairn no they look to be a problem with an external service

PR #81, opened 2019-08-13
Intly 2798: Dedicated admin to 3scale admin

sedroche:
Jira: https://issues.jboss.org/browse/INTLY-2798

3scale and rhsso part of the above ticket.
Openshift users should be automatically added to 3scale and rhsso and openshift users that are in the dedicated-admins group should be admins in 3scale and rhsso also.

## Verification
1. Add an OperatorSource that points to my app registry
```yaml
kind: OperatorSource
metadata:
  name: integreatly-operators
spec:
  authorizationToken: {}
  registryNamespace: sedroche
  type: appregistry
```
2. Create an Installation CR and wait for the installation to succeed.
3. Go to the 3scale admin console (3scale namespace -> routes -> `https://3scale-admin.<routingSubdomain>`) and using the ` Authenticate through Red Hat Single Sign-On` option, sign in using an openshift user. (The sign-in should not need any extra configuration or ask for any extra verification)
4. If the openshift user is not part of the `dedicated-admins` group it should have limited options within the 3scale admin console
5. To test you can create the `dedicated-admins` group by:
```bash
oc adm groups new dedicated-admins
```
then add and remove users by:
```bash
oc adm groups add-users dedicated-admins <username>
```
and
```bash
oc adm groups remove-users dedicated-admins <username>
```
6. Adding and deleting openshift users should be reflected in the 3scale and rhsso users.

7. Adding and removing users from the `dedicated-admins` group should change roles in rhsso and 3scale


pkg/controller/installation/products/rhsso/reconciler.go
```diff
@@ -163,9 +176,37 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, inst *v1alpha1.Ins
 		},
 	}
 	ownerutil.EnsureOwner(kcr, inst)
-	err = serverClient.Create(ctx, kcr)
-	if err != nil && !k8serr.IsAlreadyExists(err) {
-		return v1alpha1.PhaseFailed, errors.Wrap(err, "failed to create keycloak realm")
+	_, err = controllerutil.CreateOrUpdate(ctx, serverClient, kcr, UpdateNoOp)
```
matskiv: Why switch to `CreateOrUpdate` if we are doing nothing on update?

sedroche: I want the update as I want to update:
```go
CreateOnly:                        false,
DeleteUsers:                       true,
```
and in the case of an already existing CR (as in an upgrade scenario) the CR wouldn't get updated.

I believe that `CreateOrUpdate` does an equality check and updates based on the passed obj anyway but let me check

pkg/controller/installation/products/rhsso/reconciler.go
```diff
@@ -294,3 +335,89 @@ func containsIdentityProvider(providers []*aerogearv1.KeycloakIdentityProvider,
...
+				ClientRoles: map[string][]string{
+					"account": {
+						"manage-account",
+						"view-profile",
+					},
+					"realm-management": {
+						"manage-users",
+						"manage-identity-providers",
+						"view-realm",
+					},
```
matskiv: I think that we don't need to give `realm-management` roles to any user now.
Customer-admin / dedicated-admins will be able to manage users in openshift or in the IDP that's configured in openshift. I think that was the only reason for giving them access.

sedroche: Ok. I'll update this :+1:

pkg/controller/installation/products/threescale/reconciler.go
```diff
@@ -430,6 +443,78 @@ func (r *Reconciler) reconcileUpdatingAdminDetails(ctx context.Context, serverCl
...
+			res, err := r.tsClient.SetUserAsAdmin(tsUser.UserDetails.Id, *accessToken)
+			if err != nil || res.StatusCode != http.StatusOK {
+				return v1alpha1.PhaseFailed, err
+			}
+		} else {
+			res, err := r.tsClient.SetUserAsMember(tsUser.UserDetails.Id, *accessToken)
+			if err != nil || res.StatusCode != http.StatusOK {
+				return v1alpha1.PhaseFailed, err
+			}
+		}
```
matskiv: The code above will "spam" http requests on each reconcile. Can we check tsUser.UserDetails.Role to see if it really needs to be updated?

pkg/controller/installation/products/threescale/reconciler.go
```diff
@@ -356,25 +363,31 @@ func (r *Reconciler) reconcileRHSSOIntegration(ctx context.Context, serverClient
...
 	}
+	if tsIsNotFoundError(err) {
+		site := rhssoConfig.GetHost() + "/auth/realms/" + rhssoRealm
+		res, err := r.tsClient.AddAuthenticationProvider(map[string]string{
+			"kind":                              "keycloak",
+			"name":                              rhssoIntegrationName,
+			"client_id":                         clientId,
+			"client_secret":                     clientSecret,
+			"site":                              site,
+			"skip_ssl_certificate_verification": "true",
```
matskiv: This should probably use `selfSignedCerts` value from InstallationSpec.

sedroche: That value is set on the http client:

https://github.com/integr8ly/integreatly-operator/blob/38c490ff568f2ad201c692136cad6d9ddae1fd3e/pkg/controller/installation/products/reconciler.go#L69-L75

maleck13:
@matskiv I agree as we are now turning all dedicated admins in the cluster into "effective" customer-admin users, I see no need for the specific customer-admin user. That said I think we can create a follow on to remove this user rather than needing to do it as part of this PR?

PR #86, opened 2019-08-19
add precommit git hook to check for unformatted files

JameelB:
## What
Adds a precommit git hook to check if any staged .go files, ignoring `vendor/` files, are not formatted correctly. This should prevent build failures caused by unformatted files. 

https://golang.org/misc/git/pre-commit


JameelB:
@philbrookes @maleck13 wdyt?

PR #88, opened 2019-08-21
ensure fuse online integration limit is set correctly

aidenkeating:
currently, we set the integration limit of fuse online to -1, this
results in no integration being able to be deployed as the actual
value for specifying unlimited integrations is 0.

this updates the default limit value to 0 to allow unlimited
integrations to be created in the shared fuse instance in a
cluster.

the code around this in fuse can be seen:
- https://github.com/syndesisio/syndesis/blob/38cbf5f90e4ca61c547f92c861be36fc377f0f6c/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/user/UserConfigurationProperties.java#L31
- https://github.com/syndesisio/syndesis/blob/ba937b8043a18a90019b5b354f19ff3b10ae033b/app/server/controller/src/main/java/io/syndesis/server/controller/integration/IntegrationPublishValidator.java#L49-L50

verification:
- have an openshift 4 cluster
- ensure that a basic integration cannot be created in the current
  fuse deployment, use a timer -> log integration.
- run this commit of the integreatly operator against a cluster and
  ensure that the fuse online that is provisioned can now create
  more than one integration.


sedroche:
Can we use `createOrUpdate`? We need to ensure upgrades will work too i.e with an existing CR.

e.g.
https://github.com/integr8ly/integreatly-operator/blob/684af094ad628433fa5fbbb98673f5dee4ed376d/pkg/controller/installation/products/rhsso/reconciler.go#L132-L143

JameelB:
@aidenkeating Can we close this PR? We now set the integration limit to `0` https://github.com/integr8ly/integreatly-operator/blob/master/pkg/controller/installation/products/fuse/reconciler.go#L242

aidenkeating:
@JameelB Closing

PR #89, opened 2019-08-23
Solution explorer openshift 4 support

aidenkeating:
Add the `OPENSHIFT_VERSION` env var to the solution explorer


pkg/controller/installation/products/solutionexplorer/reconciler.go
```diff
@@ -164,10 +165,11 @@ func (r *Reconciler) ReconcileCustomResource(ctx context.Context, inst *v1alpha1
...
 			Parameters: map[string]string{
-				param_oauth_client:         defaultSubNameAndPkg,
-				param_sso_route:            ssoConfig.GetHost(),
-				param_openshift_host:       inst.Spec.MasterURL,
-				param_openshift_oauth_host: oauthURL,
+				paramOauthClient:        defaultSubNameAndPkg,
+				paramSSORoute:           ssoConfig.GetHost(),
+				paramOpenShiftHost:      inst.Spec.MasterURL,
+				paramOpenShiftOauthHost: oauthURL,
+				paramOpenShiftVersion:   "4",
```
sedroche: Can we use `controllerutil.CreateOrUpdate` to update Cr spec to ensure upgrades are handled correctly

version/version.go
```diff
@@ -1,5 +1,5 @@
 package version
 
 var (
-	Version = "1.6.0"
+	Version = "1.7.0"
```
sedroche: Why is this minor bump?

aidenkeating: We're not fixing a bug here, we're enabling OpenShift 4 support in the webapp by adding an env var. I think a minor bump makes more sense here.

Happy to move to a patch if you feel strongly about it though

PR #90, opened 2019-08-26
Watch users and add to rhmi-developers group

sedroche:
Jira: https://issues.jboss.org/browse/INTLY-2836

- All users in an integreatly cluster should be part of the `rhmi-developers` group.
- integreatly-operator should create the `rhmi-developers` group if it isn't there and update the groups Users list accordingly.
- Users should be able to be added and removed from the `dedicated-admins` group and get escalated privileges.

### Verification
1. Deploy integreatly-operator using my OperatorSource
```yaml
kind: OperatorSource
metadata:
  name: integreatly-operators
spec:
  authorizationToken: {}
  registryNamespace: sedroche
  type: appregistry
```
This should create the `rhmi-developers` group and add any users to the group.
```bash
$ oc describe group rhmi-developers
Name:           rhmi-developers
Created:        22 minutes ago
Labels:         <none>
Annotations:    <none>
Users:          opentlc-mgr
```
2. Add an extra user to the cluster if you don't have one
```bash
htpasswd -c -b htpasswd testuser Password1
```
and add that user/pass into the htpasswd secret in openshift-config namespace

3. Create a namespace `test1`
4. Logout and login with testuser
5. Check permissions can be added and removed to/from the groups

```bash
$ oc adm policy add-role-to-group view rhmi-developers -n test1
// testuser should see the test1 namespace but not have create perms

$  oc adm groups new dedicated-admins
$ oc adm policy add-role-to-group edit dedicated-admins -n test1
// dedicated-admins should be have create perms in that namespace

$ oc adm groups add-users dedicated-admins testuser
// testuser should have create perms on the namespace

$ oc adm groups remove-users dedicated-admins testuser
// testuser should revert back to just view perms
```


sedroche:
@maleck13 Is this how you are imagining this?

pkg/controller/user/user_controller.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	or, err := controllerutil.CreateOrUpdate(ctx, c, rhmiGroup, func(existing runtime.Object) error {
+		users := &usersv1.UserList{}
+		err := c.List(ctx, &client.ListOptions{}, users)
+		if err != nil {
+			return err
+		}
+
+		g := existing.(*usersv1.Group)
+		g.Users = mapUserNames(users)
```
sedroche: I am thinking we don't need to check if they are dedicated-admins as users can be part of rhmi-developers and dedicated-admins groups

maleck13:
> @maleck13 Is this how you are imagining this?

So I had been thinking if the user is not in the the dedicated-admins group then add them the rhmi-developer group. Do we know if there is any possible problem with being in both groups. Wondering about a conflicting perm (IE dedicated admins has edit on a resource where rhmi-developers has view).

sedroche:
> > @maleck13 Is this how you are imagining this?
> 
> So I had been thinking if the user is not in the the dedicated-admins group then add them the rhmi-developer group. Do we know if there is any possible problem with being in both groups. Wondering about a conflicting perm (IE dedicated admins has edit on a resource where rhmi-developers has view).

I don't see a problem with that example, since dedicated-admins should have the escalated privileges.
If dedicated-admins is a superset of rhmi-developers then they can be in both. We would only need to check membership of dedicated-admins If there are permissions rhmi-developers will have that dedicated-admins will not

maleck13:
Ok that seems reasonable. So all users of the cluster will be rhmi developers if you are a dedicated admin you get extra stuff

sedroche:
> Ok that seems reasonable. So all users of the cluster will be rhmi developers if you are a dedicated admin you get extra stuff

Exactly

PR #91, opened 2019-08-26
[INTLY-2813] Clean up oauth clients on cr deletion

dimitraz:
### Description
Cleans up oauth clients and other cluster level objects when an `Installation` custom resource is deleted. This is done with the use of a new finalizer reconciler which accepts an anonymous function that handles the execution of the finalization logic. 

Each product implements its own finalizer and handles its own finalization logic and the deletion of its finalizer from the CR.

### Verification steps

- Create the operator source which points to my app registry:
```yml
kind: OperatorSource
metadata:
  finalizers:
    - finalizer.operatorsources.operators.coreos.com
  name: integreatly-operators
spec:
  authorizationToken: {}
  registryNamespace: dimitra
  type: appregistry
```
- Create a new namespace to test in, and create the secrets required for 3scale and launcher
- Create a subscription to the Integreatly operator from the Operator Hub in the namespace just created, then create an installation custom resource
```yml
kind: Installation
metadata:
  name: example-installation
spec:
  type: workshop
  namespacePrefix: integreatly-
  routingSubdomain: apps.cluster-pbrookes-1b0d.pbrookes-1b0d.openshiftworkshop.com
  masterUrl: https://console-openshift-console.apps.cluster-pbrookes-1b0d.pbrookes-1b0d.openshiftworkshop.com
  selfSignedCerts: true
```
- Once all the products have completed, verify that 3 new finalizers have been added to the custom resource yaml 
- Delete the installation. Verify that each product finalizer (`finalizer.rhsso.integreatly.org`, `finalizer.3scale.integreatly.org`, `finalizer.webapp.integreatly.org`) is removed from the cr once its corresponding oauth client has been deleted (verify that the following oauth clients are successfully cleaned up: `rhsso`, `3scale`, `integreatly-solution-explorer`)
- Verify that once these finalizers have been removed from the installation cr, the installation itself is also deleted and the product namespaces are cleaned up


pkg/resources/finalizers.go
```diff
@@ -0,0 +1,86 @@
...
+	if _, ok := FinalizeProducts[finalizer]; !ok {
+		FinalizeProducts[finalizer] = product
+	}
+	return nil
+}
+
+// RemoveOauthClient deletes an oauth client owned by a product and removes its finalizer from
+// the installation custom resource
+func RemoveOauthClient(ctx context.Context, inst *v1alpha1.Installation, client pkgclient.Client, oauthClient oauthClient.OauthV1Interface, finalizer string, oauthId string) error {
+	if contains(inst.GetFinalizers(), finalizer) {
```
sedroche: This finalizer code is repeated in the `ReconcileFinalizer` so it can be removed here.
This function should just delete the Oauth Client

aidenkeating:
👀

pkg/controller/installation/products/rhsso/reconciler.go
```diff
@@ -107,6 +116,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, inst *v1alpha1.Installation,
 		return phase, err
 	}
 
+	phase, err = r.ReconcileOauthClient(ctx, inst, &oauthv1.OAuthClient{
```
sedroche: I would leave this as part of the `setupOpenshiftIDP` function if it is all related to openshift idp setup

pkg/controller/installation/installation_controller.go
```diff
@@ -138,6 +129,41 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
 
+	// If the CR is being deleted, cancel the current context
+	// and attempt to clean up the products with finalizers
+	if instance.DeletionTimestamp != nil {
+		// Cancel this context to kill all ongoing requests to the API
+		// and use a new context to handle deletion logic
+		r.cancel()
+
+		// Clean up the products which have finalizers associated to them
+		var merr error
```
aidenkeating: Is there a problem with initialising `merr` here so we don't need to perform the `nil` check each time we want to add to it. We can avoid the repeated casting too. Then we'd just need to check the length of the `errors` list in `merr` before deciding whether to return it or `nil` at the end.

pkg/controller/installation/installation_controller.go
```diff
@@ -160,37 +186,37 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
 	if err != nil {
 		// The 'Update' function can error if the resource has been updated by another process and the versions are not correct.
 		if k8serr.IsConflict(err) {
 			// If there is a conflict, requeue the resource and retry Update
-			log.Info("Error updating Installation resource status. Requeue and retry.")
+			logrus.Info("Error updating Installation resource status. Requeue and retry.", err)
 			return reconcile.Result{
 				Requeue:      true,
-				RequeueAfter: time.Second * 10,
+				RequeueAfter: time.Second,
```
aidenkeating: Just curious, what was the reasoning for reducing this to a second? Should the other 10 second requeue on L189 also be reduced?

dimitraz: No reason, was just testing and forgot to change it back! Thanks

pkg/controller/installation/installation_controller.go
dimitraz: :+1:

PR #94, opened 2019-08-29
fix bad email format on osd

philbrookes:


dimitraz:
@philbrookes just needs another version bump

PR #96, opened 2019-09-02
[WIP] INTLY-2711 Add Integreatly manifests

dimitraz:
```
**Note**: Do not merge yet, this is pending Delorean work
```

### Description 
Moves integreatly manifests from the manifests repo to the operator repo. 

The `make gen/csv` command generates the latest csv, using the latest versions of:
- `operator.yaml`
- `role.yaml` and `clusterrole.yaml` 
- The crd

The previous version of the csv is used as a base for the new csv (required fields such as spec.keywords and spec.description are copied over from this csv). The `replaces` field in the new csv is set to the version of the previous csv.

### How
For this to work, the following version numbers must be maintained:
- The current version (`TAG`) and the previous version (`PREVIOUS_TAG`) in the Makefile
- The quay.io image version in the `operator.yaml` file
- The package manifest file 

```sh
operator-sdk olm-catalog gen-csv 
    --csv-version $(TAG) 
    --from-version $(PREVIOUS_TAG)
```
The resulting folder structure will look something like the following:

```sh
deploy/
    olm-catalog/
        integreatly-operator/
            1.8.1/
                 installation.crd.yaml
                 integreatly-operator.v1.8.1.clusterserviceversion.yaml
        integreatly-operator.package.yaml
    csv-config.yaml # config file which allows us to specify the paths to our roles and clusterroles
```


philbrookes:
I'm not sure that we want all the history of the manifests in this repo, it'd be better to tag the repo and modify the files in each tag, I imagine the dir structure should look like this:
```
manifests/
    installation.crd.yaml
    integreatly-operator.clusterserviceversion.yaml
```
And whatever version is specified in the clusterserviceversion file is what the tag in the repo this should be committed into.

This way we can find any manifest for any version, without having maintain eventually hundreds of files.

philbrookes:
I'm not sure about the package file, if that should be here or not, I currently think it probably shouldn't be here. 

I also am not sure that we want this Makefile trying to update an app registry, or travis either.

Eventually the stuff in the Makefile and Travis file will be replaced by the deloreon project, so I'd be inclined to leave this out. @maleck13 @mikenairn any thoughts on this?

maleck13:
Yeah I think moving them back for now is good. We have a task to auto gen these resources for a future sprint, DeLorean should then take care of the rest

deploy/olm-catalog/integreatly-operator/1.8.1/integreatly-operator.v1.8.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,362 @@
...
+    alm-owner-integreatly: integreatly-operator
+    operated-by: integreatly-operator
+  maintainers:
+  - email: integreatly-support@redhat.com
+    name: Integr8ly
+  maturity: alpha
+  minKubeVersion: 1.11.0
+  provider:
+    name: Integr8ly
+  replaces: integreatly-operator.v1.8.0
```
mikenairn: @philbrookes @maleck13 Do either of you have any more information about how this should work?

The tooling @dimitraz is using to generate these files looks to be assuming that you have all versions of the manifest, at least the previous version, in this repo so that it can use that to generate the new one. It also sets this replaces value. If it is expecting the previous manifest here, should we just have all versions here?

What's the best practise here for maintaining these? It's hard to find any examples that do it in the same way, but on the productised versions that i could find, all manifest versions were just kept in the operator repo.

dimitraz: @mikenairn Yeah, I found one productised example that seemed to keep all the versions within the operator repo. If we keep on using the `gen-csv` command this shouldn't be very difficult to maintain either. 

If we don't need the replaces field, we can use a generic integreatly-operator csv template that the command can use to base off of and remove the replaces field on the generated csv, but that isn't necessarily the intended usage for the tool. I think it depends on what the requirements are for delorean

dimitraz:
@austincunningham Yep, I have a separate PR for that since I decided to refactor the Readme quite a bit

aidenkeating:
@dimitraz fyi, a 1.8.1 change has been merged to master (fuse online updates)

dimitraz:
@aidenkeating As far as I know, we're going to manually maintain the manifests for a little while longer in the manifests repo until it is fully automated by Delorean, and I'm not entirely sure when this will get merged (it might need more discussion).  

So bump the version there and I'll update my PR to 1.8.2

JameelB:
@dimitraz Can we close this PR? Changes from this has already been added/merged to master in this PR https://github.com/integr8ly/integreatly-operator/pull/156

dimitraz:
@JameelB Yeah, will do

PR #98, opened 2019-09-05
Refactor readme

dimitraz:
### Description 
Update the readme to use updated `Makefile` commands for less tedious copying and pasting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,120 +2,104 @@
 
 A Kubernetes Operator based on the Operator SDK for installing and reconciling Integreatly products.
 
-## Current status
+### Project status: _alpha_ 
 
-This is a PoC / alpha version. Most functionality is there but it is higly likely there are bugs and improvements needed
+This is a proof of concept/alpha version. Most functionality is present but it is higly likely there are bugs and improvements needed.
+
+### Installed products
+Currently the operator installs the following products:
+- AMQ Online
+- AMQ Streams
+- Codeready
+- Fuse
+- Launcher
+- Nexus
+- RHSSO (both a cluster instance and a user instance)
+- 3scale
+- Integreatly solution explorer
 
 ## Prerequisites
-- [Go (v1.12+)](https://golang.org/dl/) 
+- [operator-sdk](https://github.com/operator-framework/operator-sdk) version v0.10.0.
+- [go](https://golang.org/dl/) version 1.12+
 - [moq](https://github.com/matryer/moq)
+- [oc](https://docs.okd.io/3.11/cli_reference/get_started_cli.html#cli-reference-get-started-cli) version v3.11+
+- Access to an Openshift v4.1.0+ cluster
+- A user with administrative privileges in the OpenShift cluster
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -2,120 +2,117 @@
...
+- RHSSO 
+- 3scale
 
 ## Prerequisites
-- [Go (v1.12+)](https://golang.org/dl/) 
+- [operator-sdk](https://github.com/operator-framework/operator-sdk) version v0.10.0.
+- [go](https://golang.org/dl/) version 1.12+
 - [moq](https://github.com/matryer/moq)
+- [oc](https://docs.okd.io/3.11/cli_reference/get_started_cli.html#cli-reference-get-started-cli) version v3.11+
+- Access to an Openshift v3.11.0+ cluster
```
matskiv: We also have a dependency on OLM, so maybe say `Access to an Openshift v3.11.0+ cluster with OLM`
Or it might be safer to say v4.1.0+ cluster, because I doubt anyone tested it on 3.11.

dimitraz: Good point, I'll change that :+1:

README.md
```diff
@@ -2,120 +2,118 @@
...
+### Installed products
+Currently the operator installs the following products:
+- AMQ Online
+- AMQ Streams
+- Codeready
+- Fuse
+- Launcher
+- Nexus
+- RHSSO 
+- 3scale
```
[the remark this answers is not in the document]
austincunningham: Should we call user-sso out as a product ??

README.md
```diff
@@ -2,120 +2,118 @@
...
-```
+Some products require certain credentials to be present in the namespace before installation can proceed: 
+* 3scale requires AWS credentials for backups to an S3 bucket. The bucket should have all public access turned off
+* RHSSO requires Github OAuth credentials to create a Github Identity Provider for Launcher (see [here](https://github.com/integr8ly/installation/#51-create-github-oauth-to-enable-github-authorization-for-launcher) for creating a Github OAuth app)
 
-Create the Namespace/Project for the Integreatly Operator to watch:
-```sh
-oc new-project <namespace>
-```
+**Note:** If these secrets aren't created, the integreatly preflight checks will fail
```
austincunningham: not sure if the preflight will stop without github secrets, I know Launcher doesn't have it as a prerequisite

JameelB: I'm not sure that the 3scale secrets are also included in the preflight. However, installation will fail. Rather than stating that the preflight checks will fail here, we can change this to the installation will fail? wdyt?

dimitraz: The preflight checks look for the s3 secrets in the namespace before the installation of the products: https://github.com/integr8ly/integreatly-operator/blob/6d2ee2fb9478ef08b5200349b627c0f3484226bf/pkg/controller/installation/installation_controller.go#L227

If they're not present, the checks fail. Will make a PR for adding the Github secret to that list as well

philbrookes: it won't we should update it so that it does though

PR #99, opened 2019-09-05
reconcile pull secret for fuse and 3scale

philbrookes:
- Prepare for using GA version of 3scale operator
- Abstract docker pull secret reconciliation
- Update unit tests


deploy/operator.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       containers:
         - name: integreatly-operator
           # Replace this with the built image name
-          image: quay.io/integreatly/integreatly-operator:v1.8.1
+          image: quay.io/integreatly/integreatly-operator:v1.8.2
```
austincunningham: The manifest is still pointing at v1.8.0 , is there a version gone missing somewhere https://github.com/integr8ly/manifests/blob/master/Makefile#L2

philbrookes: maybe, I'll catch it up when I PR

philbrookes:
@austincunningham for verification you can check that the sample secret is copied into the 3scale namespace and the fuse namespace

deploy/operator.yaml
austincunningham: :+1:

PR #103, opened 2019-09-06
INTLY-2784 - create bootstrap phase

matskiv:
JIRA: https://issues.jboss.org/browse/INTLY-2784

I've added new installation phase for object creation(and reconcile) that needs to be done before any products are installed.
Currently the only purpose of this phase is to generate OAutchClient secrets and store them in a Secret. They are consumed by the products that need OAuthClient with a secret.
This will allow us to have randomly generated secret of a client, and be able to recreate OAuthClient with the original secret if need be. Also, I included recovery mechanism, in case `oauth-client-secrets` Secret is deleted - it will be restored by reading OAuthClient objects.
I added ns prefix to OAuthClients as well.

Verification steps:
Install integreatly using this PR.
In operator NS check `oauth-client-secrets` Secret. It should contain three keys(3scale, rhsso and rhssouser) with randomly generated secret values.
For each of this products: 3scale, RHSSO and User SSO check following:
- OAuthClient with ns prefix was created
- spec.secret value contains randomly generated value, that's identical to one in `oauth-client-secrets` Secret
- you can log into the product console with a openshift user(use customer-admin because of unrelated permission restrictions)


matskiv:
I fixed one bug caught by unit tests, updated the tests and rebased.
I will try to add some new tests on Monday, but this is ready for review and verification now.

philbrookes:
Hey, this took a while, but I got there, it all looks good except in 3scale oauth client:
```
kind: OAuthClient
apiVersion: oauth.openshift.io/v1
metadata:
  name: integreatly-solution-explorer
  ...<snip>...
secret: test
```

But in the secret 3scale is set to: `3sUjSLCViXbEEJmLZc7PUbDL1btqtHlm` looks like some placeholder code survived?

philbrookes:
I think this is probably the problem: 
https://github.com/integr8ly/integreatly-operator/pull/103/files#diff-e57ecf748f2846acee838386782f0fb2R129

matskiv:
@philbrookes I forgot to remove the secret from webapp client. It is not used in web app auth flow. Will do it now.

philbrookes:
@matskiv Cool, I will re-verify now

PR #116, opened 2019-09-19
handle upgrades of component operators

philbrookes:
switch installplan approval to manual, and automate approval of installplans for versions at or below supported level.

This operator also ensures that it's own subscription is set to manual approval.

## Jiras:
- https://issues.jboss.org/browse/INTLY-3189
- https://issues.jboss.org/browse/INTLY-3188
- https://issues.jboss.org/browse/INTLY-3187

## Verification
This is a bit tricky. We can verify by running the install, and seeing that the operators are installed and running and the approval is set to manual.

To verify upgrades will require more work though, I did it by installing everything, then creating a 0.10.0 "release" of nexus in my app registry, seeing it was not upgraded once available in the UI and that the operator logged: "integreatly-nexus has an upgrade installplan above the maximum allowed version". Then editing the operator version for nexus to 0.10.0 in the operator, re-running it, and seeing the upgrade occur.


pkg/controller/installation/products/amqstreams/reconciler.go
```diff
@@ -24,6 +24,7 @@ var (
 	defaultInstallationNamespace = "amq-streams"
 	defaultSubscriptionName      = "integreatly-amq-streams"
 	clusterName                  = "integreatly-cluster"
+	operatorVersion              = "1.1.0"
```
maleck13: I really think this should be controlled by 1 file which could be easily overridden by DeLorean

pkg/resources/installPlan.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/sirupsen/logrus"
+	pkgclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
+)
+
+func upgradeApproval(ctx context.Context, client pkgclient.Client, ip *v1alpha1.InstallPlan, maxVersion *Version) error {
+	if ip.Spec.Approved == false && len(ip.Spec.ClusterServiceVersionNames) > 0 {
+		logrus.Infof("getting version for %s", ip.Spec.ClusterServiceVersionNames[0])
+		// convert "keycloak.1.8.2" into "1.8.2"
+		verStr := strings.SplitN(ip.Spec.ClusterServiceVersionNames[0], ".", 2)[1]
```
maleck13: this seems a bit gross. How do we know it will always be at index 0?

pkg/resources/version.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+type Version struct {
+	Major int
+	Minor int
+	Patch int
+}
+
+func NewVersion(version string) (*Version, error) {
+	r, _ := regexp.Compile(`([0-9]+)\.([0-9]+)(\.|\-)([0-9]+)`)
```
maleck13: did we see versions with ```-``` in them?

pkg/resources/version_test.go
```diff
@@ -0,0 +1,55 @@
...
+func TestVersion(t *testing.T) {
+	scenarios := []struct {
+		Name            string
+		TestVersion     string
+		ExpectedVersion *Version
+		Verifier        func(version *Version, err error, t *testing.T)
+	}{
+		{
+			Name:        "test valid version",
+			TestVersion: "1.2.3",
```
maleck13: prob could improve on these test cases

pkg/resources/installPlan.go
philbrookes: because we make them, it is really gross. I could not see a cleaner way to get the version number though, as the version in this object is the likes of "v1alpha1".

pkg/resources/version.go
philbrookes: yes I have seen version like "1.9-7".

pkg/controller/installation/products/amqonline/reconciler.go
```diff
@@ -25,6 +25,7 @@ const (
 	defaultInstallationNamespace = "amq-online"
 	defaultSubscriptionName      = "integreatly-amq-online"
 	defaultConsoleSvcName        = "console"
+	operatorVersion              = "0.28.0"
```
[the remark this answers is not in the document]
philbrookes: sure, I have asked @mikenairn if he has any preferences for where these strings are stored, care to comment here, @mikenairn ?

pkg/controller/installation/products/amqstreams/reconciler.go
philbrookes: discussed in another comment.

pkg/resources/version_test.go
philbrookes: there's not a lot of complexity in that version class, these tests cover a good and a bad version string. What other tests do you imagine would be helpful?

maleck13: something with a ```-``` in it as it looks to be the case we expect that type of version?

maleck13: something with more than 3 points in its release ```1.2.2.2323123```

maleck13:
@philbrookes do we want to consider always allow patch versions. IE if a new patch version is pushed and we push the button to go ahead, it will work without the need for changes to the operator

philbrookes:
@maleck13 that's a cool idea, I'll implement it.

maleck13:
@philbrookes actual should our logic not just be
 if patch version change ok to upgrade once manual approval given on integreatly operator. 
 if minor version change ok to go ahead if next version is only 1 minor version greater than the current version.
 major version should match current major version in code so 1.x.x

matskiv:
Minor version can have some significant differences, so I would be careful with that.
What use case would warrant one minor version upgrade without integreatly operator upgrade?

maleck13:
> Minor version can have some significant differences, so I would be careful with that.
> What use case would warrant one minor version upgrade without integreatly operator upgrade?

My thought process here is that, the upgrade cannot be rolled out without manual approval. If we auto test via DeLoren the new update and all is good. With this approach, there is no code change required. We approve the upgrade in the clusters and it is done. I guess the trigger here would need to be a rebuild of the same version of the integreatly operator. 
So if we have to publish a new version of the operator anyway and If we can automate the process of updating the versions, then having the explicit versions is preferable

pkg/controller/subscription/subscription_controller.go
```diff
@@ -0,0 +1,93 @@
...
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	if instance.Spec.InstallPlanApproval != v1alpha1.ApprovalManual {
+		instance.Spec.InstallPlanApproval = v1alpha1.ApprovalManual
+		r.client.Update(context.TODO(), instance)
```
matskiv: Error handling is missing here. Once added we notice that additional permissions need to be added to operators role. Error:
```
"error":"subscriptions.operators.coreos.com \"integreatly\" is forbidden: 
User \"system:serviceaccount:001-integreatly:integreatly-operator\" 
cannot update resource \"subscriptions\" in API group \"operators.coreos.com\" 
```

matskiv:
I have a few comments/questions:
1) There is one more cluster wide permission required - `update` verb for `installplans` resource in `operators.coreos.com` group.


2) I noticed that once we create a Subscription resource for a product, we don't update it, so if someone was to change approval to automatic, integreatly-operator won't do anything about that. Do we want to ensure that it always stays manual, or is it unreasonable to  expect that someone will mess with the subscription?


3) I think we should ensure a backup is ran and finished successfully before we approve any upgrade. I created a follow up JIRA, we can discuss there - https://issues.jboss.org/browse/INTLY-3686


4) (Minor) I think we could clean up auto-generated comments in `pkg/controller/subscription/subscription_controller.go ` and maybe add comment of our own, saying that this will only reconcile subscription of the integreatly-operator (it took me a moment to realize that 😅 ).

matskiv:
I was able to verify this with the additional permissions that I mention above. 👍

pkg/resources/version_test.go
philbrookes: added a bunch of tests now

pkg/controller/subscription/subscription_controller.go
philbrookes: I have made a new manifest that adds this permission, but I intended to open that PR once this one was merged. I imagined this was being verified by running the operator locally.

matskiv: @philbrookes But there are also cluster roles in deploy/ folder, but I noticed they are slightly out of date already, so we can fix that in one go later...
What do you think of error handling here, is it worth adding?

maleck13:
@matskiv I think we should manage the subscriptions and ensure they are how we expect. Running backup makes sense. Will speak with Aiden about this as it will be managed by AWS services. Could you create a follow on Jira for the subscription change and comments update. The cluster level perm could you add it to https://issues.jboss.org/browse/INTLY-2945

PR #117, opened 2019-09-19
fix finalizer bug

philbrookes:
fixing a bug where the finalizers are never removed if the installation object is deleted while the operator is not running.

## Verification
- Install stack using this operator
- stop the operator
- delete the installation CR
- start the operator
- see the uninstall function normally.

### DIFF
--- a/pkg/controller/installation/products/rhssouser/reconciler_test.go
+++ b/pkg/controller/installation/products/rhssouser/reconciler_test.go
@@ -413,6 +413,31 @@ func TestReconciler_fullReconcile(t *testing.T) {
 		},
 	}
 
+	installation := &v1alpha1.Installation{
+		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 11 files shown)


pkg/controller/installation/products/rhssouser/reconciler_test.go
```diff
@@ -413,6 +413,31 @@ func TestReconciler_fullReconcile(t *testing.T) {
 		},
 	}
 
+	installation := &v1alpha1.Installation{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:       "installation",
+			Namespace:  defaultRhssoNamespace,
+			Finalizers: []string{"finalizer.user-sso.integreatly.org"},
```
maleck13: do we want a test here that adds more than one finalizer? Perhaps it exists elsewhere?

dimitraz:
:eyes:

PR #127, opened 2019-09-26
INTLY-3288 - Add UPS Install

KevFan:
## Description
Using the Unified Push Server (UPS) Operator, install UPS as part of the integreatly install

The associated pull request for adding the UPS cluster service version:
* https://github.com/integr8ly/manifests/pull/70

## Verification
* Create a operator source that points to my app registry:
```
apiVersion: operators.coreos.com/v1
kind: OperatorSource
metadata:
  name: integreatly-operators
  namespace: openshift-marketplace
spec:
  authorizationToken: {}
  displayName: Integreatly Operators
  endpoint: 'https://quay.io/cnr'
  publisher: Integreatly Publisher
  registryNamespace: kevfan
  type: appregistry
```
* Browse to the operator hub and verifiy that a Unifiedpush Operator with version `0.2.0` is present
* Install Integreatly using the following Installation.cr as an example point to your cluster:
```
apiVersion: integreatly.org/v1alpha1
kind: Installation
metadata:
  name: example-installation
spec:
  type: workshop
  namespacePrefix: integreatly-
  routingSubdomain: apps.cluster-id.openshiftworkshop.com
  masterUrl: https://console-openshift-console.apps.cluster-id.openshiftworkshop.com
  apiUrl: https://api.cluster-id.openshiftworkshop.com
  selfSignedCerts: true
```
* Verify that the `integreatly-ups` namespace is created
* Verify that the `unifiedpush-operator` pod is created
* Verify that the `ups` custom resource install is created and resources are porvisioned correctly
* Verifiy that the installation custom resource yaml correctly updated the phase and exposed UPS route
* Visit exposed ups route to ensure it works
  * Note: You may get a internal 500 error when trying to login due to https://github.com/aerogear/unifiedpush-operator/pull/75 and the manifest may not be updated to use a image tag with this change (can be resolved manually by updating oauth-proxy image in image stream to `quay.io/openshift/origin-oauth-proxy:4.2.0` )

PR #128, opened 2019-09-27
Add basic e2e test.

mikenairn:
Tests that all product operators deploy successfully when creating a new workshop installation.

Run with:

```make test/e2e```

### CODE REVIEW

File: test/e2e/integreatly_test.go [hunk 1]
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func waitForProductDeployment(t *testing.T, f *framework.Framework, ctx *framework.TestCtx, product, deploymentName string) error {
+	namespace := intlyNamespacePrefix + product
+	t.Logf("Checking %s:%s", namespace, deploymentName)
+
+	start := time.Now()
+	err := e2eutil.WaitForDeployment(t, f.KubeClient, namespace, deploymentName, 1, deploymentRetryInterval, deploymentTimeout)
+	if err != nil {
+		return err
```
tremes: I am just curious what the message will look like here. I mean if it won't be too general.

mikenairn: Will output something like:

```
integreatly_test.go:143: timed out waiting for the condition   
```
Which you can figure out in context:

```
....
wait_util.go:55: Waiting for availability of tutorial-web-app-operator deployment
wait_util.go:55: Waiting for availability of tutorial-web-app-operator deployment   
integreatly_test.go:143: timed out waiting for the condition 
```

But we can also wrap the error to make it clearer, Will do that.

tremes: OK thanks. I think it could be sufficient but maybe worth checking in some next similar places.

cmd/manager/main.go
```diff
@@ -79,11 +84,15 @@ func main() {
...
+	if isLeaderElectionEnabled() {
+		// Become the leader before proceeding
+		err = leader.Become(ctx, "integreatly-operator-lock")
+		if err != nil {
+			log.Error(err, "")
+			os.Exit(1)
+		}
+	} else {
+		log.Info("Warning: Leader election is disabled")
 	}
```
mikenairn: This is required for CI. To become a leader it needs to be running in a cluster, and would skip it if being run locally like we do in CI tests. Since the CI is running on a kubernetes cluster though, this thinks it is running in a cluster and fails the tests. This just allows us to disable it when executing it locally in CI.

test/e2e/integreatly_test.go
```diff
@@ -0,0 +1,141 @@
...
+	for product, deploymentName := range products {
+		err = waitForProductDeployment(t, f, ctx, product, deploymentName)
+		if err != nil {
+			break
+		}
+	}
+	//These test only that the operators themselves have came up. Further testing should be done to verify the individual parts of each product come up also (What the operator is creating).
+
+	return err
+}
```
mikenairn: This needs to be expended to do some additional testing of each operators function. Currently it will check that all product operator deployments succeed.

test/e2e/integreatly_test.go
```diff
@@ -0,0 +1,141 @@
...
+		"3scale":               "3scale-operator",
+		"amq-online":           "enmasse-operator",
+		"amq-streams":          "amq-streams-cluster-operator",
+		"codeready-workspaces": "codeready-operator",
+		"fuse":                 "syndesis-operator",
+		"launcher":             "launcher-operator",
+		"nexus":                "nexus-operator",
+		"solution-explorer":    "tutorial-web-app-operator",
+		"user-sso":             "keycloak-operator",
+	}
```
mikenairn: Since https://github.com/integr8ly/integreatly-operator/pull/127, this should be checking UPS also, but this hasn't been merged yet https://github.com/integr8ly/manifests/pull/70, so will likely fail.

@maleck13 @philbrookes What is the dev flow you expect when adding or updating products? I'd expect master here to always work against the integreatly app registry, but if we don't merge back manifest changes for product updates first potentially it won't, at least without pulling from a different registry.

I'd also like to update these tests to fail early if tests are missing (i.e. We install ups in a workshop, but don't have a test for it here), but we would end up in dead lock if we expect to merge manifest PRs after the operator one. Should product  update manifest PRs be merged ahead of any integreatly operator PR maybe?

tremes: What about the possible namespace prefix? We usually use `openshift-` AFAIK.

test/e2e/integreatly_test.go
```diff
@@ -0,0 +1,141 @@
...
+		},
+	}
+	// use TestCtx's create helper to create the object and add a cleanup function for the new object
+	err = f.Client.Create(goctx.TODO(), workshopInstallation, &framework.CleanupOptions{TestContext: ctx, Timeout: installationCleanupTimeout, RetryInterval: installationCleanupRetryInterval})
+	if err != nil {
+		return err
+	}
+
+	//Auth Stage - verify operators deploy
+	err = waitForProductDeployment(t, f, ctx, "rhsso", "keycloak-operator")
```
tremes: The "rhsso", "keycloak-operator" params refer to namespace's name? If so there could be the problem with the namespace prefix.

tremes:
Are all the ` vendor/github.com/spf13/cobra/*`, ` vendor/github.com/BurntSushi/*` files required?

test/e2e/integreatly_test.go
mikenairn: waitForProductDeployment adds the prefix here https://github.com/integr8ly/integreatly-operator/pull/128/files/8ce92eda791272eb72fa63dbb8f875648ee01056#diff-e3e74a4fcd6e24a416beb0a2b8f02762R48, and is set in the test here https://github.com/integr8ly/integreatly-operator/pull/128/files/8ce92eda791272eb72fa63dbb8f875648ee01056#diff-e3e74a4fcd6e24a416beb0a2b8f02762R28

test/e2e/integreatly_test.go
mikenairn: > What about the possible namespace prefix? We usually use openshift- AFAIK.
@tremes Prefix gets used elsewhere as required, this is just listing the products.

mikenairn:
> Are all the vendor/github.com/spf13/cobra/*, vendor/github.com/BurntSushi/*, etc. files required?

@tremes Guess so, must be dependencies of the test framework.

test/e2e/integreatly_test.go
tremes: OK I see. Sorry I didn't see it.

test/e2e/integreatly_test.go
maleck13: I think it is fine to merge manifest changes first. followed by the operator. Ideally they should be merged pretty much at the same time

mikenairn:
> This looks like a great start. Can we capture the improvements we want to make in a Jira (IE more tests around the products spinning up etc)

@maleck13 Yes, intend to create some follow on Jiras for improvements.

PR #138, opened 2019-10-09
feat: adds mdc install 

jjaferson:
This PR adds the installation of the mobile developer console as a product of the integreatly operator.

https://issues.jboss.org/browse/INTLY-3290

Verification steps:
1) Make sure the MDC operator manifest file has been pushed to the integreatly registry in [quay.io](https://quay.io/application/) 
2) Check if the MDC operator is available on the operator hub under `Catalog >> Operator Hub`
3) Run the integreatly-operator locally installing only the MDC operator: 
`operator-sdk up local --namespace=<integreatly_operator_namespace> --operator-flags=--products=mdc`
4) Verify if the `integreatly-mdc` namespace is created
5) Check if there is a subscription of the MDC operator in the integreatly-mdc namespace under `Catalog >> Operator Management` and if the status is Up to date
6) In the `integreatly-mdc` namespace check if there are two pods created, one for the MDC operator and another for the MDC service

### DIFF
--- a/pkg/controller/installation/products/fuse/reconciler.go
+++ b/pkg/controller/installation/products/fuse/reconciler.go
@@ -83,7 +83,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, inst *v1alpha1.Installation,
 	if err != nil || phase != v1alpha1.PhaseCompleted {
 		return phase, err
 	}
-	phase, err = r.ReconcilePullSecret(ctx, r.Config.GetNamespace(), inst, serverClient)
+	phase, err = r.ReconcilePullSecret(ctx, r.Config.GetNamespace(), defaultFusePullSecret, inst, serverClient)
 	if err != nil || phase != v1alpha1.PhaseCompleted {
 		return phase, err
 	}
--- a/pkg/controller/installation/products/threescale/reconciler.go
+++ b/pkg/controller/installation/products/threescale/reconciler.go
@@ -94,10 +94,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, in *v1alpha1.Installation, p
 		return phase, err
 	}
 
-	phase, err = r.ReconcilePullSecret(ctx, r.Config.GetNamespace(), in, serverClient)
+	phase, err = r.ReconcilePullSecret(ctx, r.Config.GetNamespace(), "", in, serverClient)
 	if err != nil || phase != v1alpha1.PhaseCompleted {
 		return phase, err
 	}
+
 	version, err := resources.NewVersion(v1alpha1.OperatorVersion3Scale)
 	if err != nil {
 		return v1alpha1.PhaseFailed, errors.Wrap(err, "invalid version number for launcher")
--- a/pkg/resources/pullSecret.go
+++ b/pkg/resources/pullSecret.go
@@ -1,0 +1,54 @@
+package resources
+
+import (
+	"github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+	"golang.org/x/net/context"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	corev1 "k8s.io/api/core/v1"
+	pkgclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	DefaultOriginPullSecretName      = "samples-registry-credentials"
+	DefaultOriginPullSecretNamespace = "openshift"
+)
+
+// Gets the default pull secret for pulling container images from registry
+func GetDefaultPullSecret(inst *v1alpha1.Installation, client pkgclient.Client, context context.Context) (corev1.Secret, error) {
+	if inst.Spec.PullSecret.Name == "" {
+		inst.Spec.PullSecret.Name = DefaultOriginPullSecretName
+	}
+	if inst.Spec.PullSecret.Namespace == "" {
+		inst.Spec.PullSecret.Namespace = DefaultOriginPullSecretNamespace
+	}
+
+	openshiftSecret := corev1.Secret{}
+
+	err := client.Get(context, types.NamespacedName{Name: inst.Spec.PullSecret.Name, Namespace: inst.Spec.PullSecret.Namespace}, &openshiftSecret)
+
+	return openshiftSecret, err
+}
+
+// Copys the default pull secret to a target namespace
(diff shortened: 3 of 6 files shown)


deploy/crds/mdc_v1alpha1_mobiledeveloperconsole_cr.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: mdc.aerogear.org/v1alpha1
```
philbrookes: I don't think we should have this file

pkg/apis/integreatly/v1alpha1/installation_types.go
```diff
@@ -29,30 +29,32 @@ var (
...
-	VersionLauncher            ProductVersion = "0.1.2"
-	VersionUps                 ProductVersion = "2.3.2"
+	VersionAMQOnline              ProductVersion = "1.2.2"
+	VersionAMQStreams             ProductVersion = "1.1.0"
+	VersionCodeReadyWorkspaces    ProductVersion = "1.2.0.GA"
+	VersionFuseOnOpenshift        ProductVersion = "master"
+	VersionNexus                  ProductVersion = "3.16"
+	VersionLauncher               ProductVersion = "0.1.2"
+	VersionUps                    ProductVersion = "2.3.2"
+	VersionMobileDeveloperConsole ProductVersion = "1.1.11"
```
philbrookes: this should only be added this way, if there is no way to find the version dynamically, is that the case?

pkg/controller/installation/products/fuseonopenshift/reconciler.go
```diff
@@ -1,4 +1,4 @@
-package fuseonopenshift
+	package fuseonopenshift
```
philbrookes: misplaced tab here (surprised that this compiled)

pkg/controller/installation/products/mobiledeveloperconsole/reconciler.go
```diff
@@ -0,0 +1,162 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: r.Config.GetNamespace(),
+			Name:      resourceName,
+		},
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: mdc.SchemeGroupVersion.String(),
+			Kind:       mdc.MobileDeveloperConsoleKind,
+		},
+		Spec: mdc.MobileDeveloperConsoleSpec{
+			OAuthClientID:     "mobile-developer-console",
```
philbrookes: this doesn't look right? Should these values be pulled from the github oauth secret we already create?

pkg/controller/installation/products/mobiledeveloperconsole/reconciler.go
```diff
@@ -0,0 +1,162 @@
...
+	r.logger.Info("checking mdc pods are running")
+
+	pods := &corev1.PodList{}
+	err := client.List(ctx, &pkgclient.ListOptions{Namespace: r.Config.GetNamespace()}, pods)
+	if err != nil {
+		return v1alpha1.PhaseFailed, errors.Wrap(err, "failed to list mdc in mdc namespace")
+	}
+
+	// expecting 2 pods in total
+	if len(pods.Items) < 2 {
```
philbrookes: if this is how it has to be, then so be it, but it'd be preferable to check it was ready from the status object of the CR we created. If that is not possible, we should feed that back to the operator development team so that they can add it.

scripts/htpasswd
```diff
@@ -0,0 +1,2 @@
+customer-admin:$apr1$ln7wLhog$oeUZwbqBc7AC0gK9R2bQJ0
```
philbrookes: I don't think we want this file either?

mikenairn:
/hold

Can we make sure that the e2e test is updated to include what is being added here, and the e2e test runs OK before merging, see https://github.com/integr8ly/integreatly-operator/commit/ac77d63366c81ab4044e207b76d09f12706f769f#diff-e3e74a4fcd6e24a416beb0a2b8f02762

pkg/controller/installation/products/mobiledeveloperconsole/reconciler.go
jjaferson: I have removed this

scripts/htpasswd
jjaferson: removed

pkg/controller/installation/products/mobiledeveloperconsole/reconciler.go
jjaferson: I've added the openshift oauth

mikenairn:
@jjaferson Can you squash the commits. Is there a manifest update for this and has that been merged? We can trigger the e2e tests after you do these things.

jjaferson:
> @jjaferson Can you squash the commits. Is there a manifest update for this and has that been merged? We can trigger the e2e tests after you do these things.

Yep, I'll work on squash those commits. @philbrookes is working on verify the manifest file, I'll ping you once it's merged.

philbrookes:
manifest stuff is merged

deploy/crds/examples/installation.cr.yaml
```diff
@@ -7,4 +7,4 @@ spec:
   namespacePrefix: integreatly-
   routingSubdomain: apps.example.com
   masterUrl: http://console.apps.example.com
-  selfSignedCerts: true
+  selfSignedCerts: true
```
mikenairn: Probably should be reverted

pkg/apis/integreatly/v1alpha1/installation_types.go
```diff
@@ -30,19 +30,20 @@ var (
...
+	ProductRHSSOUser              ProductName = "rhssouser"
+	ProductCodeReadyWorkspaces    ProductName = "codeready-workspaces"
+	ProductFuse                   ProductName = "fuse"
+	ProductFuseOnOpenshift        ProductName = "fuse-on-openshift"
+	Product3Scale                 ProductName = "3scale"
+	ProductNexus                  ProductName = "nexus"
+	ProductLauncher               ProductName = "launcher"
+	ProductUps                    ProductName = "ups"
+	ProductMonitoring             ProductName = "monitoring"
+	ProductMobileDeveloperConsole ProductName = "mdc"
```
JameelB: :+1: Matches the product id in the [webapp](https://github.com/integr8ly/tutorial-web-app/blob/master/src/product-info.js#L50).

PR #143, opened 2019-10-14
working on RHPDS

philbrookes:
bumping version as new manifest is required to add RHPDS installation type to the CRD.

We cannot use S3 for 3scale in RHPDS, and they do not currently have RWX PVs, so it is not installed.

We cannot use github oauth in RHPDS, so launcher has been removed. 

The preflight checks for the s3 secret and the github oauth secret do not execute when running an RHPDS installation type.

Related PR to manifests *Do not merge until this PR is  merged*: https://github.com/integr8ly/manifests/compare/master...philbrookes:INTLY-3194?expand=1


deploy/crds/installation.crd.yaml
```diff
@@ -35,6 +35,7 @@ spec:
               enum:
                 - workshop
                 - managed
+                - RHPDS
```
mikenairn: @philbrookes What's the difference between rhpds and workshop?

mikenairn: Should probably have read the description, but does it have to be called rhpds?

PR #125, opened 2019-09-24
INTLY-3286: Add installed_services param to the webapp dc

JameelB:
## What
Adds an `INSTALLED_SERVICES` param that contains the installed products information i.e. dashboard url, version. This ensures that the webapp has access to the product routes so that it can add these links to the *Applications* section in the solution explorer's landing page.
- [x] Add `INSTALLED_SERVICES` param to the webapp CR.
- [x] Move the solution explorer installation to the very last stage.
    - This should ensure that all the products have been successfully installed first
- [x] Add unit tests
- [x] Bump operator version

## Verification Steps
1. Install the Integreatly operator and create an installation CR.
   - Install using an operator source pointing to `jameelb` app registry or run locally
2. **Verify**: All products installed by Integreatly should be completed successfully.
3. **Verify**: Solution Explorer should be installed as the last stage
   - Check the installation CR and ensure that there is a stage called `solutionexplorer`.
   - Ensure that the phase of the solution explorer is `completed` with the correct host and version.
4. **Verify**: `INSTALLED_SERVICES` param is specified in the webapp CR in the solution explorer namespace.
   - Ensure that it doesn't contain AMQ online
   - Ensure that it doesn't contain products that doesn't have a dashboard URL.
   - Ensure that the product names matches the [product id](https://github.com/integr8ly/tutorial-web-app/blob/master/src/product-info.js) that the webapp expects it to be.

PR #144, opened 2019-10-15
Add new make target to install and wait for installation

mikenairn:
* Add new make target to install and wait for installation cluster/deploy/integreatly-installation-cr.yml
* Add new make target to install operator source cluster/prepare/osrc
* Make installation type and name configurable in deploy/integreatly-installation-cr.yml

/cc @philbrookes @StanleyKaleta 

This will allow us to use the same logic to test up status in both jenkins jobs and prow

```
make cluster/deploy/integreatly-installation-cr.yml
```

Will wait until the installation completes or fails.

Added fix for https://github.com/integr8ly/integreatly-operator/pull/125/commits/f5c35c9206a3a2ab802ffec2f760c83a9cebbed0#r335021486

/cc @JameelB

### DIFF
--- a/test/e2e/integreatly_test.go
+++ b/test/e2e/integreatly_test.go
@@ -103,9 +103,18 @@ func integreatlyWorkshopTest(t *testing.T, f *framework.Framework, ctx *framewor
 		"launcher":              "launcher-operator",
 		"middleware-monitoring": "application-monitoring-operator",
 		"nexus":                 "nexus-operator",
-		"solution-explorer":     "tutorial-web-app-operator",
 		"user-sso":              "keycloak-operator",
 		"ups":                   "unifiedpush-operator",
+	}
+	for product, deploymentName := range products {
+		err = waitForProductDeployment(t, f, ctx, product, deploymentName)
+		if err != nil {
+			break
+		}
+	}
+	//SolutionExplorer Stage - verify operators deploy
+	products = map[string]string{
+		"solution-explorer": "tutorial-web-app-operator",
 	}
 	for product, deploymentName := range products {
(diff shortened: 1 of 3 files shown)


test/e2e/integreatly_test.go
```diff
@@ -113,6 +112,16 @@ func integreatlyWorkshopTest(t *testing.T, f *framework.Framework, ctx *framewor
...
+	//SolutionExplorer Stage - verify operators deploy
+	products = map[string]string{
+		"solution-explorer": "tutorial-web-app-operator",
+	}
+	for product, deploymentName := range products {
+		err = waitForProductDeployment(t, f, ctx, product, deploymentName)
+		if err != nil {
+			break
+		}
+	}
```
pawelpaszki: I don't like this change, since it still does not check whether the installation actually completed. I would prefer to create a PR to the e2e test and include your changes to the Makefile (there is also one additional step for solution explorer in the `make/e2e/olm`

pawelpaszki: added your Makefile changes here: https://github.com/integr8ly/integreatly-operator/pull/153 on top of the e2e test changes. Do you wanna close this PR?

mikenairn: The change here was fine since it was just adding the additional stage with the same level of testing as it always had, to allow the test to actually work. Since you have now rolled all these changes (I assume it has all the changes anyway, you wiped out my commit) into your PR https://github.com/integr8ly/integreatly-operator/pull/153, i guess we can close this one.

mikenairn:
These changes were rolled into https://github.com/integr8ly/integreatly-operator/pull/153

/close

PR #146, opened 2019-10-16
bumps version to 1.9.6

damienomurchu:
Bump version to 1.9.6. INTLY-3289


deploy/operator.template.yaml
```diff
@@ -26,7 +26,7 @@ objects:
           containers:
             - name: integreatly-operator
               # Replace this with the built image name
-              image: quay.io/integreatly/integreatly-operator:v1.9.2
+              image: quay.io/integreatly/integreatly-operator:v1.9.6
```
damienomurchu: Just bumping the version in the template to something more recent too

PR #151, opened 2019-10-16
Use CRO to create s3 bucket for 3scale

dimitraz:
https://issues.jboss.org/browse/INTLY-3092

### DIFF
--- a/pkg/controller/installation/products/threescale/reconciler.go
+++ b/pkg/controller/installation/products/threescale/reconciler.go
@@ -206,6 +206,9 @@ func (r *Reconciler) reconcileSMTPCredentials(ctx context.Context, inst *v1alpha
 		},
 	}
 
+	// add owner ref
+	resources.PrepareObject(&smtpCred.ObjectMeta, r.installation)
+
 	_, err := controllerutil.CreateOrUpdate(ctx, serverClient, smtpCred, func(existing runtime.Object) error {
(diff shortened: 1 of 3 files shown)


pkg/controller/installation/products/threescale/reconciler.go
```diff
@@ -332,6 +299,146 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, serverClient pkgcl
...
+			credSec.Data["AWS_ACCESS_KEY_ID"] = blobStorageSec.Data["credentialKeyID"]
+			credSec.Data["AWS_SECRET_ACCESS_KEY"] = blobStorageSec.Data["credentialSecretKey"]
+
+			err = serverClient.Create(ctx, credSec)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to create blob storage aws credentials secret")
+			}
+		}
+
+		// return the file storage spec
```
aidenkeating: From here on, we could do

- read operator sec
- createorupdate namespace sec using data from operator sec

that will flatten this entire section

i'd also split these into two funcs and move the external resource conditional up a level

aidenkeating:
this looks great, going to give it a spin once I have access to my cluster again?

i know you're away tomorrow, are you good with this being merged if verification goes ok while you're away?

dimitraz:
@aidenkeating Sure if you can that would be great!

PR #153, opened 2019-10-25
updated e2e test to check installation completion

pawelpaszki:
## What
updated e2e test to check installation completion and added Makefile changes from this PR: https://github.com/integr8ly/integreatly-operator/pull/144

### DIFF
--- a/test/e2e/integreatly_test.go
+++ b/test/e2e/integreatly_test.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	goctx "context"
 	"fmt"
+	"strings"
 
 	//"sigs.k8s.io/controller-runtime/pkg/client"
 	"testing"
@@ -13,7 +14,10 @@ import (
 
 	framework "github.com/operator-framework/operator-sdk/pkg/test"
 	"github.com/operator-framework/operator-sdk/pkg/test/e2eutil"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
@@ -27,6 +31,11 @@ var (
 	installationCleanupRetryInterval = time.Second * 20
 	installationCleanupTimeout       = time.Minute * 4 //Longer timeout required to allow for finalizers to execute
 	intlyNamespacePrefix             = "intly-"
+	installationName                 = "e2e-managed-installation"
+	bootstrapStage                   = "bootstrap"
+	authenticationStage              = "authentication"
+	productsStage                    = "products"
+	solutionExplorerStage            = "solution-explorer"
 )
 
 func TestIntegreatly(t *testing.T) {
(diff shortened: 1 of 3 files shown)


test/e2e/integreatly_test.go
```diff
@@ -71,14 +74,14 @@ func integreatlyWorkshopTest(t *testing.T, f *framework.Framework, ctx *framewor
 	// create installation custom resource
 	workshopInstallation := &operator.Installation{
```
mikenairn: Should probably rename it since it's no longer a "workshop" installation

test/e2e/integreatly_test.go
```diff
@@ -115,9 +116,33 @@ func integreatlyWorkshopTest(t *testing.T, f *framework.Framework, ctx *framewor
...
+				if strings.Contains(webapp, "completed") {
+					return err
+				} else {
+					time.Sleep(deploymentRetryInterval)
+				}
+			}
+		} else {
+			time.Sleep(deploymentRetryInterval)
+		}
+	}
```
mikenairn: I would have seen these tests going more along the lines of:

- For each stage in the type of installation we are testing (managed, workshop, whatever we might add later)
  - For each operator being installed in this stage
    - Wait with timeout for the operator deployment to succeed(waitForProductDeployment)
  - Wait with timeout for the phase status to be complete (.status.stages.\<phasename\>.phase)

So the current "managed" installation type would be more like:
````
func waitForInstallatonStage() {
	//Implemenet this taking inspiration from https://github.com/operator-framework/operator-sdk/blob/master/pkg/test/e2eutil/wait_util.go#L46-L72
	//Should use a timeout and retryInterval to avoid a test just sitting infinitely waiting for some state
}

//Bootstrap Stage
//verify bootstrap stage completes
err = waitForInstallatonStage(t, f, ctx, installationName, "bootstrap")
if err != nil {
	return err
}

//Auth Stage
//verify operators deploy
err = waitForProductDeployment(t, f, ctx, "rhsso", "keycloak-operator")
if err != nil {
	return err
}
//verify authentication stage completes
err = waitForInstallatonStage(t, f, ctx, installationName, "authentication")
if err != nil {
	return err
}

//Products Stage
//verify operators deploy
for product, deploymentName := range products {
	err = waitForProductDeployment(t, f, ctx, product, deploymentName)
}
if err != nil {
	return err
}
//verify authentication stage completes
err = waitForInstallatonStage(t, f, ctx, installationName, "products")

//Solutions Explorer Stage
//verify operators deploy
err = waitForProductDeployment(t, f, ctx, "solution-explorer", "tutorial-web-app-operator")
if err != nil {
	return err
}
//verify authentication stage completes
err = waitForInstallatonStage(t, f, ctx, installationName, "solution-explorer")
if err != nil {
return err
}
````

This way you are testing each stage consistently and it will be far easier to update as the operator evolves. It will also allow for simpler addition of tests for different installation scenarios should we want to do that, which we likely will if/when we introduce different supported installation configurations.

Comment (pawelpaszki) [reply test/e2e/integreatly_test.go]: wanna have another look? I have updated the code based on your comments

File: test/e2e/integreatly_test.go [hunk 3]
```diff
@@ -115,11 +134,54 @@ func integreatlyWorkshopTest(t *testing.T, f *framework.Framework, ctx *framewor
 			break
 		}
 	}
```
mikenairn: I think we need to check err here before continuing:

```
if err != nil {
    return err
}
```

test/e2e/integreatly_test.go
mikenairn: Looks good, pretty much how i seen it working. :+1:

test/e2e/integreatly_test.go
mikenairn: Looks to me like we break out of the loop if hit an error, but never check it before continuing on to the next part https://github.com/integr8ly/integreatly-operator/blob/fbfb5fe39ade58fb68c621a152c2ddf9ca4a02db/test/e2e/integreatly_test.go#L131-L139

Should we not be returning err at that point if we have identified one of the product operators isn't running?

PR #154, opened 2019-10-30
Fix monitoring version not showing up in CR

davidkirwan:
In the _installations_ CR the monitoring version shows up as blank

```
        monitoring:
          host: ''
          name: monitoring
          status: completed
          version: ''
```

Should now get populated correctly with the version stated in _OperatorVersionMonitoring_:

_pkg/apis/integreatly/v1alpha1/installation_types.go_

### DIFF
--- a/pkg/controller/installation/products/config/monitoring.go
+++ b/pkg/controller/installation/products/config/monitoring.go
@@ -1,6 +1,10 @@
 package config
 
-import "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+import (
+	"errors"
+
+	"github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+)
 
 type Monitoring struct {
 	Config ProductConfig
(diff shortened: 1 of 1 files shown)


pkg/controller/installation/products/config/monitoring.go
```diff
@@ -35,7 +39,7 @@ func (r *Monitoring) GetProductName() v1alpha1.ProductName {
 }
 
 func (r *Monitoring) GetProductVersion() v1alpha1.ProductVersion {
-	return v1alpha1.ProductVersion(r.Config["VERSION"])
+	return v1alpha1.ProductVersion(v1alpha1.OperatorVersionMonitoring)
```
david-martin: This makes sense.
@philbrookes  or @matskiv  Should all other products get the version for https://github.com/integr8ly/integreatly-operator/blob/8859df57ca42278c59821437ca5709f68b0ee707/pkg/apis/integreatly/v1alpha1/installation_types.go#L65-L78 as well?

davidkirwan: One and the same thing with regards the Monitoring version, but I can make it easier to read. Let me push a change there.

philbrookes: We do have a series of version variables too, but prefer to find the version dynamically, so that we don't have to keep it upto date.

davidkirwan:
If we're happy can I get a lgtm so @david-martin @philbrookes please.

PR #155, opened 2019-10-30
Add missing cluster role permissions

mikenairn:
**Verification**

Run the e2e make target against a clean OpenShift 4 cluster, it should pass:
```
make test/e2e
```


mikenairn:
/unshrug

It works :tada:

PR #159, opened 2019-11-01
Separate cluster roles

dimitraz:
### Description 
Separate roles and cluster roles into two separate files for simpler future maintenance. 

```
**Note**: requires operator-sdk version 0.10.0 or greater
```

### Verification 
Run `make gen/csv` and verify that both cluster roles and roles are being pulled into the csv


philbrookes:
As the operator is coded against 0.8.0, I'd prefer our workflows remain compatible with that version of the operator-sdk. We should either update the code to 0.10.0, or leave the roles in a single file.

damienomurchu:
@philbrookes operator-sdk is now on `0.12.0` - we should probably look at updating our code before it get's too painful to update operator-sdk version easily?

philbrookes:
Closing this for now as it is not yet compatible and is out of date. Please re-open once the operator-sdk is updated.

PR #161, opened 2019-11-01
Fix comment in role.yaml

mikenairn:
/cc @philbrookes 

Wasn't causing an issue, but noticed it was wrong


damienomurchu:
will roll that into another PR
/close

PR #165, opened 2019-11-04
Remove launcher

philbrookes:
launcher is no longer required in a managed install


mikenairn:
/hold 

The e2e test will need updated to reflect this change https://github.com/integr8ly/integreatly-operator/blob/master/test/e2e/integreatly_test.go#L137

PR #171, opened 2019-11-05
schedule monitoring first

philbrookes:
All components expect to be able to create monitoring resources, so it must be installed before all other components.

### DIFF
--- a/test/e2e/integreatly_test.go
+++ b/test/e2e/integreatly_test.go
@@ -34,6 +34,7 @@ var (
 	intlyNamespacePrefix             = "intly-"
 	installationName                 = "e2e-managed-installation"
 	bootstrapStage                   = "bootstrap"
+	monitoringStage                  = "monitoring"
 	authenticationStage              = "authentication"
 	productsStage                    = "products"
 	solutionExplorerStage            = "solution-explorer"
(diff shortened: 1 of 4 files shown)


david-martin:
@davidkirwan Can you take a look at this?

mikenairn:
/hold

Lets ensure the e2e passes before merging

test/e2e/integreatly_test.go
```diff
@@ -34,6 +34,7 @@ var (
 	intlyNamespacePrefix             = "intly-"
 	installationName                 = "e2e-managed-installation"
 	bootstrapStage                   = "bootstrap"
+	monitoringStage				     = "monitoring"
```
mikenairn: format?

philbrookes: I did do a make code/fix, weird

mikenairn:
@philbrookes You'll need to add an additional line here also https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L126

@pawelpaszki Will this affect the nightly/jenkins pipeline at all? Has that been modified to use the status checks for the Makefile or does it still do it it's own way?

pawelpaszki:
@mikenairn nightly still uses the old version of install (till I will get another pipeline created specifically for OSD - likely this week)

mikenairn:
/hold cancel

Tested a lot now and has past the e2e test 3 times in a row, no failures, looks good :+1:

PR #175, opened 2019-11-06
Bump operator-marketplace/pkg/apis/operators to v2

matskiv:
JIRA: https://issues.jboss.org/browse/INTLY-3869

Updating to newer version to enable OpenShift 4.2 support.

For verification try installation on 4.2.x cluster.
Optionally test it on 4.1.x as well, but if that fails, we can upgrade.

Ping @philbrookes


philbrookes:
we can merge this with the failing e2e test, as that is being resolved separately.

JameelB:
Will need to add `marketplacev2` to the scheme https://github.com/integr8ly/integreatly-operator/blob/master/pkg/apis/addtoscheme_integreatly_v1alpha1.go.

This won't work for OpenShift 4.1 because `CatalogSourceConfig (v2)` will not be available. I think we'll need an upgrade. 

^ @philbrookes

philbrookes:
going to merge this now. As we cannot update CI to 4.2 until the code-base passes on 4.2.

PR #176, opened 2019-11-07
[INTLY-3093] Update 3scale reconciler to use high availability mode with CRO

dimitraz:
### Description 
Use the `cloud-resource-operator` to deploy AWS resources in managed environments for 3scale

### Verification
- Run the integreatly operator from this branch 
- Run the cloud resource operator
- Create a new `installation` resource:
```yml
kind: Installation
metadata:
  name: example-installation
spec:
  type: managed
  namespacePrefix: integreatly-
  routingSubdomain: example.com
  masterUrl: http://master.example.com
  selfSignedCerts: true
  useExternalResources: true
```
- Create a VPC peering connection and update the two VPC route tables to point to the peering connection 
- Verify that all 3scale pods are running 
- Verify that two elasticache instances were created and are available and one RDS instance was created and is available 
- Verify that normal 3scale actions can be performed (inviting a user, creating a service, etc)


pkg/controller/installation/products/threescale/reconciler.go
```diff
@@ -305,39 +311,165 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, serverClient pkgcl
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      externalPostgresSecretName,
+			Namespace: r.Config.GetNamespace(),
+		},
+		Data: map[string][]byte{},
+	}
+	_, err = controllerutil.CreateOrUpdate(ctx, serverClient, postgresSecret, func(existing runtime.Object) error {
+		username := postgresCredSec.Data["username"]
+		password := postgresCredSec.Data["password"]
+		url := fmt.Sprintf("postgresql://%s:%s@%s:5432/%s", username, password, postgresCredSec.Data["host"], postgresCredSec.Data["database"])
```
austincunningham: /LGTM 
May want to apply a var to `5432`, I know is the default postgres port

aidenkeating: @dimitraz do we not expose the port as part of the secret? if not, we should create a ticket for that

dimitraz: We do, I just missed that. Updated now :+1:

PR #183, opened 2019-11-08
bump rhsso operator version to 1.9.5

philbrookes:
fix various tags in fuse,, until they correct it in their operator, issue reported here:
https://github.com/syndesisio/syndesis/issues/7111


JameelB:
@philbrookes You'll need to add `update` permissions for the `imagestreams` in the rbac file.

JameelB:
Verified on OSD4.2 :+1: 

/lgtm
/approve

PR #198, opened 2019-11-15
prometheus_additional_scrape_config resources

obrienrobert:
## Additional Information

https://issues.jboss.org/browse/INTLY-4024

## Verification

- Authenticate with Openshift
- Create the `integreatly` namespace
- Create the secrets for preflight checks:

```
# Execute from integreatly-operator dir
oc process -f ./deploy/s3-secrets.yaml -p INSTALLATION_NAMESPACE=integreatly -p AWS_ACCESS_KEY_ID=xxxx -p AWS_SECRET_ACCESS_KEY=xxxx -p AWS_BUCKET=xxxx-test -p AWS_REGION=eu-central-1 | oc create -f -
oc create secret generic github-oauth-secret --from-literal=clientId=xxxx --from-literal=secret=xxx -n integreatly
```

- Visit OperatorHub in the Openshift UI, search for integreatly and install it into the `integreatly` namespace

- Scale the replicas to 0 in the operator deployment in the `integreatly` namespace

- Run the operator locally: `operator-sdk up local --namespace "integreatly" --operator-flags "--products=monitoring"`

- Create instance of the installation CR, changing both the `masterURL` and `routingSubdomain` to suit your envrionment:

```
kind: Installation
metadata:
  name: example-installation
spec:
  type: workshop
  namespacePrefix: integreatly-
  routingSubdomain: apps.com
  masterUrl: https://someurl.apps.com
  selfSignedCerts: true
```

- Wait for reconcile of the monitoring product to complete

- Navigate to the `integreatly-middleware-monitoring` namespace

- Verify that the following secrets exists:
  `additional-scrape-configs`
  `integreatly-additional-scrape-configs`

- Reveal the secret values (using the reveal feature in the data section at the end) and ensure that both secrets contain the following 2 job-names with correct, valid values.

   `job_name: 3scale-apicast-pods`
   `job_name: openshift-monitoring-federation`


pkg/controller/installation/products/monitoring/reconciler.go
```diff
@@ -2,7 +2,11 @@ package monitoring
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"github.com/operator-framework/operator-marketplace/pkg/client"
+	v12 "k8s.io/api/core/v1"
```
matskiv: `corev1` would have been better, but we can address this later (I created [a JIRA](https://issues.jboss.org/browse/INTLY-4187) about unifying naming in this operator).

pkg/controller/installation/products/monitoring/reconciler.go
```diff
@@ -218,3 +283,62 @@ func (r *Reconciler) createResource(inst *v1alpha1.Installation, resourceName st
...
+	if err != nil {
+		return v1alpha1.PhaseFailed, err
+	}
+
+	if len(datasources.DataSources) < 1 {
+		return v1alpha1.PhaseFailed, errors.New("cannot obtain prometheus credentials")
+	}
+
+	// Obtain the 3scale config and namespace
+	threeScaleConfig, err := r.ConfigManager.ReadThreeScale()
```
davidkirwan: Following on from the discussion with regard to monitoring resources and where and when they get installed, does this need now to be handled in the threescale reconciler instead?

matskiv: yes, I was just thinking about that.
Let's discuss it with @obrienrobert tomorrow?

pb82: @davidkirwan @matskiv @obrienrobert The monitoring operator does only allow for one additional scrape config secret (see [here](https://github.com/integr8ly/application-monitoring-operator/blob/46c0b4c13fe4d216ae7475f5e8462e647889dd4d/pkg/controller/applicationmonitoring/applicationmonitoring_controller.go#L554)) and we concatenate all the jobs into this one secret. So in my opinion it makes sense to generate it here.

pb82:
@obrienrobert after talking with @matskiv there's something we should change: it's possible that not all products are installed, but currently we always create all the scrape config jobs. Instead we should check if a product's namespace is set (using the config.GetNamespace() function) and only if it returns a non empty string we should add the template to the scrape config secret.

matskiv:
Verified on OSD4.
/approve
/lgtm

PR #201, opened 2019-11-18
INTLY-4131: Create rhmi-admins users group

JameelB:
## What
The following changes ensures that a group `rhmi-admins` is created during the bootstrap stage. This group should contain all users that are expected to be admins in the Integreatly product suite, which includes all users from the `dedicated-admins` group. 

Currently, the way users are set as admins to the Integreatly product suite is by reading users from the admin group. Instead of reading from `dedicated-admins`, the product should get users specified in the `rhmi-admins` group.

By doing this, we can set users as admins to the Integreatly product suite without having to give them `dedicated-admins` permissions.

## Verification Steps
1. Add a user to the `dedicated-admins` group.
2. Run the operator locally
3. Verify that the `rhmi-admins` group is created
   - Go to *Search* > *Groups*
   - The group `rhmi-admins` should be present
4. Verify that users added to `dedicated-admins` group is added to the `rhmi-admins` group
   - Go to the YAML definition of the `rhmi-admins` group
   - Any users included in the `dedicated-admins` group should also be included in the `rhmi-admins` group.
5. Verify that users are successfully reconciled between the `dedicated-admins` and `rhmi-admins` group.
   - Add another user to the `dedicated-admins` group.
   - This user should be added to the `rhmi-admins` group upon bootstrap reconcile.
6. Verify that users added to the `rhmi-admins` group are set as an admin user in the Integreatly product suite.
   - **Visible Namespaces**: 
      - `rhmi-admins` users should only be able to view the Fuse namespace and their own namespaces. 
          - Any users belonging to the `rhmi-admins` group should be added to the `rhmi-admins-view-fuse` role binding.
      - `dedicated-admins` users should be able to see all namespaces. 
      - A user that doesn't belong to any of these two groups should not be able to see any namespaces apart from their own.
   - **RHSSO**: 
      - Go to *Search* > *Keycloak Realm* > *openshift*
      - admin users should have the following extra permissions added to their `clientRoles`: 
         ```
          realm-management:
          - manage-users
          - manage-identity-providers
          - view-realm
         ```
   - **3Scale**:
      - Log in to the 3Scale admin console.
      - Any users belonging to the `rhmi-admins` group should have the role `admin`. 
      - Any users that doesn't belong to this group should have the role `member`

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -276,8 +276,8 @@ rules:
     resources:
       - "rolebindings"
     verbs:
-      - create #needed for creating rolebinding for dedicated-admins to view fuse
-      - update #needed for updating subject list to match changes to dedicated-admins users
+      - create #needed for creating rolebinding for rhmi-admins to view fuse
+      - update #needed for updating subject list to match changes to rhmi-admins users
   # This is used to allow the operator to create and update the rhmi-developers group with users
   - apiGroups:
       - user.openshift.io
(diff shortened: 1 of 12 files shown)


JameelB:
`rhmi-admins` group is no longer needed. Instead, we will use `rhmi-developers` to add general permissions to non-admin users.  

An update with view access permission to Fuse namespace will be added on another PR. 

Closing this PR due to the above reasons.

PR #202, opened 2019-11-18
New keycloak operator

pb82:
Update to use the new [Keycloak Operator](https://github.com/keycloak/keycloak-operator). This required changes in the way realms, users and clients are handled:

in the 'old' keycloak operator clients and users were managed through the realm. When a user or client was updated, the realm was modified. This is no longer the case, users and clients are now their own CRs. Updates to realms are no longer reconciled.

Verifications steps:

1. Install all products using this branch and make sure single sign on still works as expected
1. Make sure user sso still works as expected


pkg/apis/addtoscheme_integreatly_v1alpha1.go
```diff
@@ -9,12 +9,12 @@ import (
 	launcherv1alpha2 "github.com/fabric8-launcher/launcher-operator/pkg/apis/launcher/v1alpha2"
 	grafanav1alpha1 "github.com/integr8ly/grafana-operator/pkg/apis/integreatly/v1alpha1"
 	threescalev1 "github.com/integr8ly/integreatly-operator/pkg/apis/3scale/v1alpha1"
-	aerogearv1 "github.com/integr8ly/integreatly-operator/pkg/apis/aerogear/v1alpha1"
 	enmassev1beta1 "github.com/integr8ly/integreatly-operator/pkg/apis/enmasse/v1beta1"
 	enmassev1beta2 "github.com/integr8ly/integreatly-operator/pkg/apis/enmasse/v1beta2"
 	solutionExplorer "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/tutorial-web-app-operator/pkg/apis/v1alpha1"
 	"github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
 	kafkav1 "github.com/integr8ly/integreatly-operator/pkg/apis/kafka.strimzi.io/v1alpha1"
+	aerogearv1 "github.com/integr8ly/integreatly-operator/pkg/apis/keycloak/v1alpha1"
```
philbrookes: Can we bring these in via dep now?

pb82: Yes I think we can. Will update.

pkg/controller/installation/products/codeready/reconciler.go
```diff
@@ -24,9 +25,9 @@ import (
 
 const (
 	defaultInstallationNamespace = "codeready-workspaces"
-	defaultClientName            = "che-client"
```
davidffrench: Just to confirm, this should have always been called `codeready-client`?

pkg/controller/installation/products/launcher/asserts_test.go
```diff
@@ -50,17 +49,6 @@ func assertInstallationSuccessfullyReconciled(scenario LauncherTestScenario, con
 		return errors.New(fmt.Sprintf("%s ConsoleURL not equal MasterURL ", defaultLauncherName))
 	}
 
-	// RHSSO integration should be configured.
-	rhssoConfig, err := configManager.ReadRHSSO()
```
davidffrench: What is the reason for this being removed?

pkg/controller/installation/products/launcher/reconciler.go
```diff
@@ -215,136 +222,140 @@ func (r *Reconciler) reconcileRHSSOIntegration(ctx context.Context, serverClient
...
-							"jsonType.label":       "String",
-						},
+	r.logger.Infof("The operation result for keycloakclient %s was %s", kcClient.Name, or)
+	return v1alpha1.PhaseCompleted, nil
+}
+
+func getKeycloakClientSpec(launcherUrl string) keycloak.KeycloakClientSpec {
+	return keycloak.KeycloakClientSpec{
+		RealmSelector: &metav1.LabelSelector{
+			MatchLabels: rhsso.GetInstanceLabels(),
```
davidffrench: Which client is this being added too? Is it the same as realm for codeready?

pkg/controller/installation/products/codeready/reconciler.go
pb82: Good question. I can change it to codeready for consistency.

pkg/controller/installation/products/launcher/asserts_test.go
pb82: The keycloak realm is no longer modified, so this is no longer needed.

pkg/controller/installation/products/launcher/reconciler.go
pb82: It's added to the launcher client: https://github.com/integr8ly/integreatly-operator/pull/202/files/04c2082ee23977ddf8ca3b5b7e8107cf9d6b00e3#diff-a52ceb0461f0b7d9c06ce3093fdd81a1R235

pkg/controller/installation/products/launcher/asserts_test.go
davidffrench: Thanks @pb82

PR #204, opened 2019-11-19
INTLY-4131: Ensure rhmi-developers group has view access to Fuse namespace

JameelB:
## What
The following changes ensures that all users belonging to the group `rhmi-developers` can access and view the Fuse namespace. 

## Verification Steps
1. Run the operator locally
2. Login as a non-admin user
    - User should get added to the `rhmi-developers` group
    - Verify that this user gets added as a subject to the `rhmi-developers-view-fuse` role binding located in the Fuse namespace. (This may take a couple of seconds to reconcile)
    - Verify that the role binding references the cluster role `view`
    - User should only be able to see Fuse and their own namespaces.
3. Verify that the user only has view access to the Fuse namespace
    - Click on the Fuse namespace
    - Ensure that the user can view the *Overview*, *YAML* and *Workload* pages.
    - Ensure that the user is not able to update any resources within the namespace

### DIFF
--- a/pkg/controller/installation/products/fuse/reconciler.go
+++ b/pkg/controller/installation/products/fuse/reconciler.go
@@ -3,8 +3,9 @@ package fuse
 import (
 	"context"
 	"fmt"
-	v13 "github.com/openshift/api/image/v1"
 	"strings"
+
+	v13 "github.com/openshift/api/image/v1"
 
 	appsv1 "github.com/openshift/api/apps/v1"
 	v1 "github.com/openshift/api/route/v1"
@@ -30,8 +31,9 @@ import (
 const (
 	defaultInstallationNamespace = "fuse"
 	defaultSubscriptionName      = "integreatly-syndesis"
-	adminGroupName               = "dedicated-admins"
 	defaultFusePullSecret        = "syndesis-pull-secret"
+	developersGroupName          = "rhmi-developers"
+	clusterViewRoleName          = "view"
 )
 
 type Reconciler struct {
@@ -99,7 +101,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, inst *v1alpha1.Installation,
 		return phase, err
 	}
 
-	phase, err = r.reconcileAdminPerms(ctx, serverClient)
+	phase, err = r.reconcileViewFusePerms(ctx, serverClient)
 	if err != nil || phase != v1alpha1.PhaseCompleted {
 		return phase, err
 	}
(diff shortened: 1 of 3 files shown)


JameelB:
/cc @philbrookes @maleck13  Here are the changes for adding view access to the Fuse namespace for the `rhmi-developer` users. 

What exactly should these users be able to see in the namespace? I've added the following permissions so that users can view the overview and workload pages within the Fuse namespace: https://github.com/integr8ly/integreatly-operator/pull/204/files#diff-816e28fbedf30d0dcbe03b9ca6b81896R211-R229

pkg/controller/installation/products/fuse/reconciler.go
```diff
@@ -191,28 +192,41 @@ func (r *Reconciler) reconcileImageVersion(ctx context.Context, install *v1alpha
...
 
-func (r *Reconciler) reconcileAdminPerms(ctx context.Context, client pkgclient.Client) (v1alpha1.StatusPhase, error) {
-	r.logger.Infof("Reconciling permissions for %s group on %s namespace", adminGroupName, r.Config.GetNamespace())
+// Ensures all users in rhmi-developers group have view Fuse permissions
+func (r *Reconciler) reconcileViewFusePerms(ctx context.Context, client pkgclient.Client) (v1alpha1.StatusPhase, error) {
+	r.logger.Infof("Reconciling view Fuse permissions for %s group on %s namespace", developersGroupName, r.Config.GetNamespace())
 
-	roleName := adminGroupName + "-view-fuse"
-	adminViewFuseRole := &rbacv1.ClusterRole{
+	roleName := developersGroupName + "-view-fuse"
```
maleck13: There is an existing cluster ``` view ``` role. Why not create a rolebinding in the namespace to that role?

JameelB: Didn't know that role existed. I'll update this so that it points to that cluster role instead. We can then remove this additional cluster role. Users will be able to view the Fuse CR with this role, is that ok?

JameelB:
@maleck13 Updated this PR to use the `view` cluster role for the fuse role bindings. 

Had to add this [cluster role binding](https://github.com/integr8ly/integreatly-operator/pull/204/files#diff-e5509d76d4a75bd5d2e5cc7fc3020593R27-R39) to the operator's rbac file so that the operator can grant the `view` cluster role permissions to the rhmi-developer users.

^ fyi @philbrookes. Will we need a new release for this rbac change?

KevFan:
Verified locally against OSD cluster

JameelB:
@maleck13 This has been verified. Are you happy with the changes here?

PR #205, opened 2019-11-19
Intly 4153 monitoring unit tests

david-martin:
https://issues.jboss.org/browse/INTLY-4153

These tests are mostly based on the mobilesecurityservice unit tests, but adapted for monitoring where necessary

PR #207, opened 2019-11-20
Include all packages in coverprofile, but omit generated files

david-martin:
A recent PR that added new unit tests caused the coverage % to go down.
https://github.com/integr8ly/integreatly-operator/pull/205
It looks like any package that doesn't have a test file in it is omitted from the coverage report. In that PR, a  test file was added, which caused all files in the monitoring package to be included int he coverage report, and take the overall % of coverage down.

This PR changes the coverprofile so it include all packages (omitting generated files) in an attempt to get a better gauge on where coverage is missing. (It should show any 0% covered packages, which were omitted from results before)

See https://github.com/golang/go/issues/24570 for some discussion on this topic.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,7 @@ anaconda-mode/
 *.test
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+*.out.tmp
 ### Vim ###
 # swap
 .sw[a-p]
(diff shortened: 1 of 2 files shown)


david-martin:
@mikenairn Everythings good here. Needs approved & lgtm.
With this change, the coveralls report now includes any packages with 0% coverage.
For example, monitoring
image
(which is being improved in #205

PR #195, opened 2019-11-15
INTLY-4023 Port all monitoring resources from create_alerts.yml

davidkirwan:
JIRA: https://issues.jboss.org/browse/INTLY-4023

## Description
https://github.com/integr8ly/installation/blob/master/roles/middleware_monitoring_config/tasks/main.yml#L2-L8
Porting of monitoring stack configuration 

```
monitoring_prometheusrules_resource_templates:
- kube_state_metrics_alerts.yml
- kube_state_metrics_3scale_alerts.yml
- kube_state_metrics_fuse_online_alerts.yml
```

```
Breaking alerts out into individual product templates:
    kube_state_metrics_3scale_alerts.yaml
    kube_state_metrics_alerts.yaml
    kube_state_metrics_amqonline_alerts.yaml
    kube_state_metrics_codeready_alerts.yaml
    kube_state_metrics_fuse_online_alerts.yaml
    kube_state_metrics_launcher_alerts.yaml
    kube_state_metrics_monitoring_alerts.yaml
    kube_state_metrics_rhsso_alerts.yaml
    kube_state_metrics_solution_explorer_alerts.yaml
    
Removing references to che/enmasse, replaced with codeready/amqonline
```

## Verification
- Authenticate with Openshift
- Create ``integreatly`` namespace
- Create the secrets for preflight checks:
```
# Execute from integreatly-operator dir
oc process -f ./deploy/s3-secrets.yaml -p INSTALLATION_NAMESPACE=integreatly -p AWS_ACCESS_KEY_ID=xxxx -p AWS_SECRET_ACCESS_KEY=xxxx -p AWS_BUCKET=xxxx-test -p AWS_REGION=eu-central-1 | oc create -f -
oc create secret generic github-oauth-secret --from-literal=clientId=xxxx --from-literal=secret=xxx -n integreatly
```
- Run the task in the Makefile: ``make cluster/prepare/local``
- Run the operator locally: ``operator-sdk up local --namespace "integreatly" --operator-flags "--products=monitoring"``
- Create instance of the installation CR eg:
```
apiVersion: integreatly.org/v1alpha1
kind: Installation
metadata:
  name: example-installation
spec:
  type: workshop
  namespacePrefix: integreatly-
  routingSubdomain: apps.com
  masterUrl: https://someurl.apps.com
  selfSignedCerts: true
```
- Wait for reconcile of the monitoring product to complete
- Check the PrometheusRules have been created in the ``integreatly-middleware-monitoring`` namespace: 
[...]

PR #212, opened 2019-11-21
Intly 4025

pb82:
Add blackbox targets for all products (excluding MSS and MDC).

Verification steps:
1. Run a full installation from this branch
2. Check that blackbox targets for each product are created in the monitoring namespaces
3. Check that all blackbox targets are reconciled into the additional scrape config secret

### DIFF
--- a/templates/monitoring/blackbox/target.yaml
+++ b/templates/monitoring/blackbox/target.yaml
@@ -1,0 +1,12 @@
+apiVersion: applicationmonitoring.integreatly.org/v1alpha1
+kind: BlackboxTarget
+metadata:
+  name: {{ index .ExtraParams "name"  }}
+  namespace: {{ .Namespace }}
+  labels:
+    monitoring-key: {{ .MonitoringKey }}
+spec:
+  blackboxTargets:
+    - service: {{ index .ExtraParams "service"  }}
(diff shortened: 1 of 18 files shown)


davidkirwan:
Screenshot from 2019-11-25 15-44-27

Installed RHSSO and the Monitoring stack:

```
operator-sdk up local --namespace "integreatly" --operator-flags "--products=monitoring,rhsso"
```

Created the `integreatly-rhsso` blackboxtarget.

templates/monitoring/blackbox/target.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: applicationmonitoring.integreatly.org/v1alpha1
+kind: BlackboxTarget
+metadata:
+  name: {{ index .ExtraParams "name"  }}
+  namespace: {{ .Namespace }}
+  labels:
+    monitoring-key: {{ .MonitoringKey }}
```
davidkirwan: Can we switch to:

- `index .Params "name"`
- `index .Params "Namespace"`
- `index .Params "MonitoringKey"`

Save me rebasing and changing in #195

davidkirwan: Or nm, might be less work if I just tackle the change in #195

PR #206, opened 2019-11-20
Use cro to deploy blobstorage for 3scale

dimitraz:
## Description
Use cro to generate blobstorage to be used by 3scale.   

## Verification 
- Create an operator source pointing to the integreatly registry 
```yml
apiVersion: operators.coreos.com/v1
kind: OperatorSource
metadata:
  name: integreatly-operators
  namespace: openshift-marketplace
spec:
  authorizationToken: {}
  displayName: Integreatly Operators
  endpoint: 'https://quay.io/cnr'
  publisher: Integreatly Publisher
  registryNamespace: integreatly
  type: appregistry
```
- Create a configmap in the integreatly namespace for the cloud resource operator which specifies in-cluster only deployments: 
```yml
kind: ConfigMap
apiVersion: v1
metadata:
  name: cloud-resource-config
  namespace: intly
data:
  managed: |
    {"blobstorage":"openshift", "smtpcredentials":"openshift", "redis":"openshift", "postgres":"openshift"}
  workshop: |
    {"blobstorage":"openshift", "smtpcredentials":"openshift", "redis":"openshift", "postgres":"openshift"}
```
- Create a configmap in the integreatly namespace for openshift provider specific configuration:
```yml
kind: ConfigMap
apiVersion: v1
metadata:
  name: cloud-resources-openshift-strategies
data:
  blobstorage: |
    {"production": { "strategy": {} }}
  smtpcredentials: |
    {"production": { "strategy": {} }}
```
- Run the integreatly operator from this branch and ensure the `cloud-resources` stage completes
- Ensure that 3scale is installed correctly
- Ensure that a `BlobStorage` custom resource and the respective secret with connection details to this resource are created in the integreatly namespace
- Ensure that an `s3-credentials` secret is created in the 3scale namespace with the same credentials that are in the blob storage secret

PR #213, opened 2019-11-21
feat: Creates s3 credentials and bucket automatically

jjaferson:
Automates the creation of s3 credentials and bucket 

https://issues.jboss.org/browse/INTLY-4009

## Verification steps

1. Run a full install of the integreatly-operator:
`operator-sdk up local --namespace=<integreatly_operator_namespace>`

2. Verify if the `s3-credentials` and `s3-bucket` secrets are created:
`oc get secret`

3. Ensure installation of all products has been completed successfully


aidenkeating:
@jjaferson What's the reasoning here for not leveraging the Cloud Resource Operator here to create the bucket and provide the credentials, similar to https://github.com/integr8ly/integreatly-operator/pull/206 ?

Of course, we're still blocked on OSD, but the Cloud Resource Operator will create a placeholder secret at the moment until OSD has been unblocked.

jjaferson:
> @jjaferson What's the reasoning here for not leveraging the Cloud Resource Operator here to create the bucket and provide the credentials, similar to #206 ?
> 
> Of course, we're still blocked on OSD, but the Cloud Resource Operator will create a placeholder secret at the moment until OSD has been unblocked.


That makes more sense indeed, I will make the changes and update the PR

jjaferson:
> @jjaferson What's the reasoning here for not leveraging the Cloud Resource Operator here to create the bucket and provide the credentials, similar to #206 ?
> 
> Of course, we're still blocked on OSD, but the Cloud Resource Operator will create a placeholder secret at the moment until OSD has been unblocked.

Actually after looking at using the Cloud Resource Operator to automate the creation of s3 credentials and bucket, I am not sure it will be possible as this happens at bootstrap stage and CRO has not been reconciled yet

pkg/apis/integreatly/v1alpha1/installation_types.go
```diff
@@ -91,12 +91,13 @@ type InstallationSpec struct {
...
-	NamespacePrefix  string         `json:"namespacePrefix,omitempty"`
-	SelfSignedCerts  bool           `json:"selfSignedCerts"`
-	PullSecret       PullSecretSpec `json:"pullSecret"`
+	Type                 string         `json:"type"`
+	RoutingSubdomain     string         `json:"routingSubdomain,omitempty"`
+	MasterURL            string         `json:"masterUrl,omitempty"`
+	NamespacePrefix      string         `json:"namespacePrefix,omitempty"`
+	SelfSignedCerts      bool           `json:"selfSignedCerts"`
+	PullSecret           PullSecretSpec `json:"pullSecret"`
+	UseExternalResources bool           `json:"useExternalResources"`
```
dimitraz: We've actually removed the use of this flag now, we provision resources with cro by default

pkg/controller/installation/products/cloudresources/reconciler.go
```diff
@@ -86,10 +89,46 @@ func (r *Reconciler) Reconcile(ctx context.Context, inst *v1alpha1.Installation,
...
+
 	product.Host = r.Config.GetHost()
 	product.Version = r.Config.GetProductVersion()
 	product.OperatorVersion = r.Config.GetOperatorVersion()
 
 	r.logger.Infof("%s has reconciled successfully", r.Config.GetProductName())
 	return v1alpha1.PhaseCompleted, nil
 }
+
+func (r *Reconciler) reconcileS3BlobStorage(ctx context.Context, installation *v1alpha1.Installation, serverClient pkgclient.Client) (v1alpha1.StatusPhase, error) {
```
dimitraz: Which product is this blob storage for, just out of curiosity?

pkg/apis/integreatly/v1alpha1/installation_types.go
jjaferson: `UseExternalResources`, right? if so I will remove it

pkg/controller/installation/products/cloudresources/reconciler.go
jjaferson: it's for the integreatly operator, are the products now responsible for reconciling s3 credentials and bucket? if so I am not sure if we still need to have an blob storage only for the integreatly operator

pkg/apis/integreatly/v1alpha1/installation_types.go
dimitraz: Yeah :+1:

pkg/controller/installation/products/cloudresources/reconciler.go
dimitraz: Yeah so the individual products would be responsible for creating the custom resources for the databases/caches that they need. The cloud resource operator picks up on the custom resource and provisions the db and the connection secret that the product will use to connect to the resource

jjaferson: Great, so I don't think we need s3 credentials and bucket for the integreatly operator as each individual product creates their own resources. 

ping @philbrookes

PR #216, opened 2019-11-21
Add the monitoring label to all managed namespaces, including the nam…

david-martin:
…espace of the integreatly operator

This is to allow the application monitoring operator & it's components to watch all monitoring resources in all managed namespaces.


david-martin:
/assign @pb82 @davidffrench @davidkirwan 
This will be needed before anything is scraped or shown from product namespaces

pb82:
:eyes:

PR #203, opened 2019-11-19
INTLY-4182: Updating operator-sdk to v0.12.0

davidffrench:
## What
Updating the operator-sdk to v0.12.0. This upgrades `controller-runtime` and `controller-tools` from `0.1.12` to `0.3.0` which includes multiple breaking changes. Also migrating from dep to mod to manage dependencies.

### Breaking changes
1. Mod is unable to pull in enmasse resources from their GitHub repository.
2. client.List() parameters have switched position and opts require a new format
3. Log and Scheme dependencies have change path within the controller-runtime
4. Controller-utils createOrUpdate mutate function no longer takes a runtime object
5. CRD regenerated with new name and structural schema
6. log.logf.ZapLogger() no longer available

## Changes for each team member
1. Upgrade Go version to 1.13
2. Upgrade operator-sdk version to v0.12.0

## Verification Steps
1. E2E tests should pass

PR #222, opened 2019-11-25
use cloud resource operator 0.2.0

aidenkeating:
@dimitraz Mind taking a look?

@philbrookes This contains a bunch of fixes for the cloud resource operator. Is this the pattern that should be followed? Or should a new directory be created for 0.2.0?


aidenkeating:
@davidffrench fyi

davidffrench:
Thanks @aidenkeating . I have also updated the version of the cloud resources operator in my PR - https://github.com/integr8ly/integreatly-operator/pull/203/files#diff-37aff102a57d3d7b797f152915a6dc16R15

grdryn:
This seems old, so I'll close for now -- the current version of the operator here looks to be 0.4.0, but this PR wants to upgrade from 0.1.0 to 0.2.0.

/close

PR #225, opened 2019-11-26
Add permissions for granting fuse only view

KevFan:
## Description
When installing operator via OperatorSource, the operator does not have the necessary permissions for granting the fuse only view from https://github.com/integr8ly/integreatly-operator/pull/204 

## Verification
* Install OperatorSource pointed to my registry
```
kind: OperatorSource
metadata:
  name: integreatly-operators
spec:
  authorizationToken: {}
  registryNamespace: kevfan
  type: appregistry
```
* Install integreatly
* Login as a non-admin user
  * User should get added to the rhmi-developers group
  * Verify that this user gets added as a subject to the rhmi-developers-view-fuse role binding located in the Fuse namespace. (This may take a couple of seconds to reconcile)
  * Verify that the role binding references the cluster role view
  * User should only be able to see Fuse and their own namespaces.
* Verify that the user only has view access to the Fuse namespace
  * Click on the Fuse namespace
  * Ensure that the user can view the Overview, YAML and Workload pages.
  * Ensure that the user is not able to update any resources within the namespace

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -107,6 +107,7 @@ rules:
     verbs:
       - get
       - list
+      - watch
   - apiGroups:
       - ""
     resources:
@@ -361,3 +362,11 @@ rules:
       - update
       - create
   # remove to here
+  # Needed for granting fuse only view
+  - apiGroups:
+      - build.openshift.io
+    resources:
+      - jenkins
+    verbs:
+      - "*"
+  # end permissions for granting fuse only view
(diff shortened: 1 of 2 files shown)


JameelB:
/lgtm
/approve

Verified locally and via Operator Source. Works as expected.

deploy/role.yaml
```diff
@@ -360,4 +361,11 @@ rules:
...
-  # remove to here
\ No newline at end of file
+  # remove to here
+  # Needed for granting fuse only view
+  - apiGroups:
+    - build.openshift.io
+    resources:
+    - jenkins
+    verbs:
+    - "*"
```
philbrookes: can we put an end comment too please. i.e. `# end permissions for granting fuse only view`

PR #237, opened 2019-11-28
INTLY-3031 - Create backup jobs alerts

matskiv:
JIRA: https://issues.jboss.org/browse/INTLY-3031

In this PR I am creating same alerts as we have in 3.x version ([here](https://github.com/integr8ly/installation/blob/master/roles/backup/templates/backup-monitoring-alerts.yml.j2)).
I also updated and cleaned up Readme and Makefile

Verification:
Install to your cluster with at least these products: `monitoring,rhsso,codeready-workspaces,amqonline`
Check that [these alerts](https://github.com/integr8ly/installation/blob/master/roles/backup/templates/backup-monitoring-alerts.yml.j2#L13-L45) x2 (different time and severity) have been created in Prometheus
Check that alerts for CodeReady and AMQ Online backup jobs have been created as well.


README.md
```diff
@@ -121,16 +121,7 @@ spec:
...
-```sh
-make setup/dedicated
-```
-
-If you want to remove the dedicated admin operator, run:
-
-```sh
-make clean/dedicated
-```
+To setup your cluster to have dedicated admins run the `./scripts/setup-htpass-idp.sh` script which installs the dedicated admin operator and creates users.
```
pawelpaszki: I think dedicated admin operator is installed by default, so this script only creates users

damienomurchu:
Confirming all alerts listed [here](https://github.com/integr8ly/installation/blob/master/roles/backup/templates/backup-monitoring-alerts.yml.j2#L13-L45) were created on cluster - there were one alert created for each, and 2 for `JobRunningTimeExceeded`

Not sure how many backup jobs there are for codeready & amqonline, but can confirm the below were created: 
```
CronJobExists_integreatly-codeready-workspaces_codeready-postgres-backup (0 active)
CronJobExists_integreatly-codeready-workspaces_codeready-pv-backup (0 active)
CronJobExists_integreatly-amq-online_enmasse-pv-backup
```

Can you confirm the above were what's expected?

Re `JobRunningTimeExceeded` - both jobs show up with the same name, and wonder if its possible to perhaps update the alert name with the timing, ie `JobRunningTimeExceeded (> 300 seconds)` and `JobRunningTimeExceeded (> 600 seconds)` maybe?

matskiv:
@damienomurchu 
Yes, alerts for codeready & amqonline backup jobs are as expected.

As for the alert name(JobRunningTimeExceeded), as discussed in chat, let's keep it same as on 3.x.

PR #247, opened 2019-12-02
Monitoring resources

pb82:
Update monitoring to v1.0.0

Verification steps:

* Install monitoring using this branch
* Make sure that all operators, prometheus, alertmanager and grafana come up successfully
* Test with other products and make sure that all alerts, dashboards and blackboxtargets are imported


manifests/integreatly-monitoring/monitoring-1.0.0/application-monitoring-operator.v1.0.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,332 @@
...
+        kind: GrafanaDashboard
+        displayName: Integreatly monitoring Grafana dashboard.
+        description: Represents an instance of the Integreatly Grafana dashboard.
+      - name: grafanadatasources.integreatly.org
+        version: v1alpha1
+        kind: GrafanaDataSource
+        displayName: Integreatly monitoring Grafana datasource.
+        description: Represents an instance of the Integreatly Grafana datasource.
+  description: Placeholder description
+  displayName: Application Monitoring Operator
```
davidkirwan: Might need to add `replaces: application-monitoring-operator.v0.0.28`
like: https://github.com/integr8ly/manifests/blob/master/integreatly-monitoring/monitoring-0.0.28/application-monitoring-operator.v0.0.28.clusterserviceversion.yaml#L45

pb82: hmm, good point. But will we ever upgrade from 0.0.28?

davidkirwan: Hmm I guess not! Just looked all the manifests in quay have been yanked.

StevenTobin:
You pulled in my alert test commit, if you add this commit it'll get rid of the 3scale alerts we don't need anymore https://github.com/integr8ly/integreatly-operator/pull/255/commits/40a176f4260fa8445fb7ba35c42e59b30b5267b2

PR #252, opened 2019-12-03
Upgrade AMQ Online

matskiv:
Upgrading to latest GA release as part of my work on https://issues.jboss.org/browse/INTLY-3149

CSV and CRDs copied from: https://github.com/jboss-container-images/amq-online-images/tree/1.3.1.GA/templates/install/olm/amq-online

Verification:
Install this PR with at least these products: `--products=cloud-resources,monitoring,rhsso,amqonline,solution-explorer`
Run `./scripts/setup-htpass-idp.sh` script and make note of the generated password
Open solution explorer and login as test-user
Provision AMQ Online AddressSpace from the SolutionExplorer (just use the UI button from the list of various consoles on the right side of main page, AFAIK it says "Provison")
Verify that provisioning finishes and console for your AddressSpace opens successfully


.gitignore
```diff
@@ -1,5 +1,6 @@
 .idea
 scripts/htpasswd
+.DS_Store
```
jjaferson: Not too important but I think it's a MacOS file, not sure if we need this file to be merged

matskiv: Yeah, it's just a MacOS metadata file for Finder.
I am pretty sure we don't want to add it to the repo, and it's annoying to see it in git changes, so that's why I added it to .gitignore :)

jjaferson: True, never mind it was .gitignore ^^

matskiv:
@jjaferson found an error when verifying.
Replica set for AMQ's keycloak fails:
`Error creating: pods "standard-authservice-598887495d-" is forbidden: unable to validate against any security context constraint: [fsGroup: Invalid value: []int64{185}: 185 is not an allowed group]`

matskiv:
FYI: the problem with fsGroup is known issue, JIRA: https://issues.jboss.org/projects/ENTMQMAAS/issues/ENTMQMAAS-1532

PR #254, opened 2019-12-03
Scale down 3scale postgres pods

dimitraz:
## Description 
With the current version of 3scale (2.6) when high availability mode is enabled the `system-postgresql` pods are still deployed. This causes the 3scale reconciler to never successfully complete since those pods are using credentials meant for the cloud resource operator postgres deployments. 

This is a workaround which scales down these pods and skips over the requirement for this deployment to be ready in order for the reconciler to continue

## Verification 
- Run the integreatly operator from this branch
- Ensure that 3scale completes in the installation cr
- Ensure that the 3scale `system-postgresql` deployment config spec specifies 0 deployments and that no `system-postgresql` pods are running

### DIFF
--- a/pkg/controller/installation/products/threescale/reconciler.go
+++ b/pkg/controller/installation/products/threescale/reconciler.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/http"
 
+	ocv1 "github.com/openshift/api/apps/v1"
+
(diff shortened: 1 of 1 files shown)


pb82:
👀

damienomurchu:
Just verifying this on a cluster now, fyi

cc @pb82

PR #255, opened 2019-12-03
Intly 3770 monitoring e2e

StevenTobin:
Adds a check for alerts which are firing to the e2e test by exec'ing into the prometheus pod to curl the alerts endpoint.

To Verify:

Check out this branch
Run make test/e2e

### DIFF
--- a/templates/monitoring/kube_state_metrics_monitoring_alerts.yaml
+++ b/templates/monitoring/kube_state_metrics_monitoring_alerts.yaml
@@ -7,14 +7,14 @@ metadata:
   namespace: {{ index .Params "Namespace" }}
 spec:
   groups:
-    - name: mointoring.rules
+    - name: monitoring.rules
       rules:
         - alert: MiddlewareMonitoringPodCount
           annotations:
             sop_url: https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md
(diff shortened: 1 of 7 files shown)


templates/monitoring/kube_state_metrics_monitoring_alerts.yaml
```diff
@@ -7,14 +7,14 @@ metadata:
   namespace: {{ index .Params "Namespace" }}
 spec:
   groups:
-    - name: mointoring.rules
+    - name: monitoring.rules
```
davidkirwan: :heart_eyes: that was annoying me

StevenTobin:
Investigating the test failure, it seems to be down to flakiness. An amq-broker pod was OOM killed and failed the second time ( which the bot didnt comment because I was too quick pushing the fix, link: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/integr8ly_integreatly-operator/247/pull-ci-integr8ly-integreatly-operator-master-e2e/500 ) because master has moved on and when I rebased it brought in changes that meant three 3scale alerts were outdated ( the three 3scale db pod checks ) so the e2e failed because alerts were firing, which is a win I suppose.

StevenTobin:
`Throttling: Rate exceeded` I'll hold off on retesting for a while.

PR #256, opened 2019-12-04
New keycloak operator

davidffrench:
Follow on from @pb82 PR https://github.com/integr8ly/integreatly-operator/pull/202

Update to use the new [Keycloak Operator](Keycloak Operator). This required changes in the way realms, users and clients are handled:

in the 'old' keycloak operator clients and users were managed through the realm. When a user or client was updated, the realm was modified. This is no longer the case, users and clients are now their own CRs. Updates to realms are no longer reconciled.

Verifications steps:
1. Install all products using this branch and make sure single sign on still works as expected
2. Make sure user sso still works as expected


davidffrench:
closing in favour of https://github.com/integr8ly/integreatly-operator/pull/266

PR #264, opened 2019-12-09
update codeready backup secret

ciaranRoche:
JIRA - https://issues.redhat.com/browse/INTLY-4387

### DIFF
--- a/pkg/controller/installation/products/codeready/reconciler.go
+++ b/pkg/controller/installation/products/codeready/reconciler.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
 	v1alpha12 "github.com/integr8ly/integreatly-operator/pkg/apis/monitoring/v1alpha1"
 
 	"k8s.io/apimachinery/pkg/runtime"
@@ -39,6 +41,7 @@ const (
 type Reconciler struct {
 	Config        *config.CodeReady
 	ConfigManager config.ConfigReadWriter
+	installation  *v1alpha1.Installation
 	extraParams   map[string]string
 	mpm           marketplace.MarketplaceInterface
 	logger        *logrus.Entry
@@ -59,6 +62,7 @@ func NewReconciler(configManager config.ConfigReadWriter, instance *v1alpha1.Ins
 	return &Reconciler{
 		ConfigManager: configManager,
 		Config:        config,
+		installation:  instance,
 		mpm:           mpm,
 		logger:        logger,
 		Reconciler:    resources.NewReconciler(mpm),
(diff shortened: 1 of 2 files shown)


pkg/controller/installation/products/codeready/reconciler.go
```diff
@@ -496,28 +455,7 @@ func (r *Reconciler) reconcileKeycloakClient(ctx context.Context, serverClient p
...
-		BackendSecret: resources.BackupSecretLocation{Name: r.Config.GetBackendSecretName(), Namespace: r.ConfigManager.GetOperatorNamespace()},
-		Components: []resources.BackupComponent{
-			{
-				Name:     "codeready-postgres-backup",
-				Type:     "postgres",
-				Secret:   resources.BackupSecretLocation{Name: r.Config.GetPostgresBackupSecretName(), Namespace: r.Config.GetNamespace()},
-				Schedule: r.Config.GetBackupSchedule(),
-			},
-			{
-				Name:     "codeready-pv-backup",
```
dimitraz: How do these backups actually work? If its backing up a pv does this need to be skipped when aws is used instead of in-cluster deployments?

ciaranRoche: I think I need to reach out to the codeready team for more context here. 

But I am guessing that with every workspace that is created a PV is provisioned. So i reckon the backups are syncing between the PV and Postgres. 

I will reach out for clarification and let you know

pkg/controller/installation/products/codeready/reconciler.go
```diff
@@ -582,49 +525,61 @@ func (r *Reconciler) createCheCluster(ctx context.Context, kcCfg *config.RHSSO,
 		},
 	}
 
-	ownerutil.EnsureOwner(cheCluster, inst)
+	ownerutil.EnsureOwner(cheCluster, r.installation)
 	if err := serverClient.Create(ctx, cheCluster); err != nil {
```
dimitraz: What do you think about using a `controllerutil.CreateOrUpdate` here instead? That would mean a refactor of this block as well - https://github.com/integr8ly/integreatly-operator/blob/15014750a4a8df7abc2fb06e5963c3c81f9d0552/pkg/controller/installation/products/codeready/reconciler.go#L230-L250

pkg/controller/installation/products/codeready/reconciler.go
```diff
@@ -526,26 +464,24 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, inst *v1alpha
...
-	selfSignedCerts := inst.Spec.SelfSignedCerts
+func (r *Reconciler) createCheCluster(ctx context.Context, kcCfg *config.RHSSO, kr *keycloakv1.KeycloakRealm, serverClient pkgclient.Client) (*chev1.CheCluster, error) {
+	selfSignedCerts := r.installation.Spec.SelfSignedCerts
 
-	// setup external postgres db using cloud-resources-operator default is Openshift
-	cheClusterExternalPostgres, err := r.reconcileExternalPostgres(ctx, inst, serverClient)
+	// get the postgres cloud resources operator cr
+	croSec := &v1.Secret{}
+	postgresName := fmt.Sprintf("codeready-postgres-%s", r.installation.Name)
+	err := serverClient.Get(ctx, pkgclient.ObjectKey{Name: postgresName, Namespace: r.installation.Namespace}, croSec)
```
dimitraz: I think it would be better to first get the postgres custom resource and then get the secret from the secret ref, since the secret name and namespace might differ depending on the cr configuration. 

```suggestion
	err := serverClient.Get(ctx, pkgclient.ObjectKey{Name: postgres.Status.SecretRef.Name, Namespace: postgres.Status.SecretRef.Namespace}, croSec)
```

dimitraz:
Verified with both rds and in-cluster postgres deployments :+1: 

/lgtm
/approve

PR #268, opened 2019-12-11
Organise imports & go vet

grdryn:
JIRA: https://issues.redhat.com/browse/INTLY-4187

This is the first of a couple of PRs to clean up imports and stuff. I'm splitting to more digestable chunks that should realistically be reviewable & mergeable. If I went with my first approach of doing everything in one PR, everyone would pay the price with complex rebases after it got merged.

Click through to the description of each commit here for more info on each. :)

PR #271, opened 2019-12-11
Always use corev1 & metav1; fix context imports

grdryn:
JIRA: https://issues.redhat.com/browse/INTLY-4187

Like #268, this doesn't solve everything. It just does the following:

- [x] Sets `metav1` and `corev1` import aliases consistently
- [x] Fixes a couple of `context` imports with `go tool fix -r context cmd/ pkg/`
- [x] Adds a comment to an exported function to satisfy a warning highlighted by openshift-ci-robot in response to the `/lint` prow command


grdryn:
That prow "lint" command is useful! :) Not sure why there isn't just a status check for lint though :thinking: 

I'll make that `GetNS` function private for now.

/honk

matskiv:
Nice! 👍 
/lgtm
/approve

PR #273, opened 2019-12-12
added to query to fix error

R-Lawton:
The following pr is to verify the following [Jira](https://issues.redhat.com/browse/INTLY-4351)
### Prerequisites:
OSD4 cluster with integreatly based of the pr branch 
### Verification steps
1.Create 2 jobs using the ui (example job works fine)
2.Modify the yaml to include the label `monitoring_key=middleware` as well as include sleep(400) like the following:
```
command:
            - perl
            - '-Mbignum=bpi'
            - '-wle'
            - sleep(400)
            - print bpi(2000)
```
3. Navigate to prometheus alert JobRunningTimeExceeded (300 seconds and 600 seconds)
4. When the job has been running for 5 mins execute the query and the two jobs show be returned.


matskiv:
@R-Lawton I haven't tested this yet, but from the code it seems like one more alert needs same fix. It's next alert in the edited file.

R-Lawton:
@matskiv grand job ill change that now. ill add to the jira the second query to for context

matskiv:
Verified 👍 
/lgtm
/approve

PR #274, opened 2019-12-12
INTLY-4327 - Update 3scale operator to use 2.7 3scale

KevFan:
## Description
Updates the 3scale operator to use 3scale 2.7. 

CSV files were copied from https://github.com/3scale/3scale-operator/tree/3scale-2.7.0/deploy/olm-catalog/3scale-operator with edits to point to the `registry.redhat.io` instead of `registry.stage.redhat.io` and removing the old `0.3.0` csv and the `replaces` label

Jira:
 * https://issues.redhat.com/browse/INTLY-4327

## Verification
* Clone this branch
* Run the operator locally against cluster with at least these products installed:
```
operator-sdk up local --namespace=integreatly --operator-flags=--products=rhsso,rhssouser,cloud-resources,3scale
```
* Ensure 3scale gets installed correctly
* Run the `setup-htpass-idp.sh` script and note the outputed password for `customer-admin`
```
./scripts/setup-htpass-idp.sh
```
* Login in to 3scale using the Single Sign On option using the `htpasswd` provider using the `customer-admin` credentials as noted above
* Input account information while keeping username as `customer-admin`
* Link to existing `customer-admin` accont, using the password: `Password1`
* Ensure user signed it correctly and can complete the tutorial for creating an example endpoint after initial login

### DIFF
--- a/manifests/integreatly-3scale/3scale-0.4.0/3scale-operator.v0.4.0.clusterserviceversion.yaml
+++ b/manifests/integreatly-3scale/3scale-0.4.0/3scale-operator.v0.4.0.clusterserviceversion.yaml
@@ -1,95 +1,95 @@
 apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 metadata:
-  name: 3scale-operator.v1.9-7
+  name: 3scale-operator.v0.4.0
   namespace: placeholder
   annotations:
     repository: https://github.com/3scale/3scale-operator
-    capabilities: Basic Install
+    capabilities: Full Lifecycle
     categories: "Integration & Delivery"
     certified: "false"
     description: 3scale Operator to provision 3scale and publish/manage API
-    containerImage: registry.redhat.io/3scale-amp26/3scale-operator:1.9-7
+    containerImage: registry.redhat.io/3scale-amp2/3scale-rhel7-operator:1.10-4
     createdAt: 2019-05-30T22:40:00Z
     support: Red Hat, Inc.
     tectonic-visibility: ocs
     alm-examples: >-
-      [{"apiVersion":"apps.3scale.net/v1alpha1","kind":"APIManager","metadata":{"name":"example-apimanager"},"spec":{"wildcardDomain":"example.com"}},
-      {"apiVersion":"apps.3scale.net/v1alpha1","kind":"APIManager","metadata":{"name":"example-apimanager-ha"},"spec":{"highAvailability":{"enabled":true},"wildcardDomain":"example.com"}},
-      {"apiVersion":"apps.3scale.net/v1alpha1","kind":"APIManager","metadata":{"name":"example-apimanager-s3"},"spec":{"system":{"fileStorage":{"amazonSimpleStorageService":{"awsBucket":"\u003cbucket-name\u003e","awsCredentialsSecret":{"name":"\u003ccredentials-secret-name\u003e"},"awsRegion":"\u003cregion\u003e"}}},"wildcardDomain":"\u003cdesired-domain\u003e"}},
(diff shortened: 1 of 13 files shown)


manifests/integreatly-3scale/3scale-0.4.0/3scale-operator.v0.4.0.clusterserviceversion.yaml
```diff
@@ -1,95 +1,95 @@
 apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 metadata:
-  name: 3scale-operator.v1.9-7
+  name: 3scale-operator.v0.4.0
```
grdryn: Why is the version here (and in the filenames) v0.4.0, when the actual operator image has a tag of 1.10-4. Should this say v1.10.0 or v1.10.4 or something?

manifests/integreatly-3scale/3scale-0.4.0/apps_v1alpha1_apimanager_crd.yaml
```diff
@@ -37,19 +37,49 @@ spec:
                   type: string
                 openSSLVerify:
                   type: boolean
+                productionSpec:
```
grdryn: I wonder if any of these new CR fields would be useful for us to set? :thinking:

manifests/integreatly-3scale/3scale-0.4.0/3scale-operator.v0.4.0.clusterserviceversion.yaml
KevFan: `0.4.0` is the operator version as specified in `version.go` but the tagged release is `1.10-4`. Not too sure what the convention for these versions are since this is mostly a straight copy from the 3scale repo. I had asked @matskiv about it and he said can just leave it as is :man_shrugging: 
Unsure how this would fit it with DeLorean eventually also if we'd always need to change the versions / filenames to match the release tag version instead :confused:

manifests/integreatly-3scale/3scale-0.4.0/apps_v1alpha1_apimanager_crd.yaml
KevFan: I'm not too sure on this one to be honest, don't know much on 3scale :see_no_evil:

manifests/integreatly-3scale/3scale-0.4.0/3scale-operator.v0.4.0.clusterserviceversion.yaml
grdryn: I've just created this issue to see if we can get clarification on that: https://github.com/3scale/3scale-operator/issues/297

Of course, we don't need to wait for a reply there to continue here. :slightly_smiling_face:

grdryn: I've just created this issue to see if we can get clarification on that: https://github.com/3scale/3scale-operator/issues/297

Of course, we don't need to wait for a reply there to continue here. :slightly_smiling_face:

eguzki: This can be helpful
https://github.com/3scale/3scale-operator/blob/2.7-stable-prod/deploy/olm-catalog/3scale-operator/3scale-operator.v0.4.0.clusterserviceversion.yaml

grdryn:
I've verified that this works as expected, good job @KevFan! Unfortunately I'm not a reviewer or approver yet, so I can't add any signoff.

Like I said, let's discuss at the standup tomorrow what we should do about the 3scale (and other product operator) packages that we've duplicated in here. I think we should depend on them properly, but I don't know if there are reasons we can't do that. Also that doesn't need to be done as part of this issue anyway.

PR #275, opened 2019-12-13
🎨 Replace usage of pkgerr with fmt.Errorf

grdryn:
JIRA: https://issues.redhat.com/browse/INTLY-4187

Go 1.13 has the ability to wrap/unwrap errors in the standard library now, and `github.com/pkg/errors` is no longer needed.

### DIFF
--- a/pkg/controller/installation/products/rhssouser/reconciler.go
+++ b/pkg/controller/installation/products/rhssouser/reconciler.go
@@ -2,9 +2,8 @@ package rhssouser
 
 import (
 	"context"
+	"fmt"
 
-	"github.com/pkg/errors"
-	pkgerr "github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	aerogearv1 "github.com/integr8ly/integreatly-operator/pkg/apis/aerogear/v1alpha1"
@@ -145,7 +144,7 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, inst *v1alpha1.Ins
 		return nil
 	})
 	if err != nil {
-		return v1alpha1.PhaseFailed, errors.Wrap(err, "failed to create/update keycloak custom resource")
+		return v1alpha1.PhaseFailed, fmt.Errorf("failed to create/update keycloak custom resource: %w", err)
 	}
 	r.logger.Infof("The operation result for keycloak %s was %s", kc.Name, or)
 
@@ -173,7 +172,7 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, inst *v1alpha1.Ins
 		return nil
 	})
 	if err != nil {
-		return v1alpha1.PhaseFailed, errors.Wrap(err, "failed to create/update keycloak realm")
+		return v1alpha1.PhaseFailed, fmt.Errorf("failed to create/update keycloak realm: %w", err)
 	}
 	r.logger.Infof("The operation result for keycloakrealm %s was %s", kcr.Name, or)
 
@@ -198,18 +197,18 @@ func (r *Reconciler) handleProgressPhase(ctx context.Context, inst *v1alpha1.Ins
 
 	err = serverClient.Get(ctx, pkgclient.ObjectKey{Name: keycloakRealmName, Namespace: r.Config.GetNamespace()}, kcr)
 	if err != nil {
-		return v1alpha1.PhaseFailed, errors.Wrap(err, "failed to get keycloak realm custom resource")
+		return v1alpha1.PhaseFailed, fmt.Errorf("failed to get keycloak realm custom resource: %w", err)
 	}
 
 	if kcr.Status.Phase == aerogearv1.PhaseReconcile {
 		err = r.exportConfig(ctx, serverClient)
 		if err != nil {
-			return v1alpha1.PhaseFailed, errors.Wrap(err, "failed to write user-sso config")
+			return v1alpha1.PhaseFailed, fmt.Errorf("failed to write user-sso config: %w", err)
 		}
 
 		err = r.setupOpenshiftIDP(ctx, inst, kcr, serverClient)
 		if err != nil {
-			return v1alpha1.PhaseFailed, errors.Wrap(err, "failed to setup Openshift IDP for user-sso")
+			return v1alpha1.PhaseFailed, fmt.Errorf("failed to setup Openshift IDP for user-sso: %w", err)
 		}
 
 		logrus.Infof("Keycloak has successfully processed the keycloakRealm for user-sso")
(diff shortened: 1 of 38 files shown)


pkg/controller/installation/products/threescale/reconciler.go
```diff
@@ -269,12 +268,12 @@ func (r *Reconciler) getOauthClientSecret(ctx context.Context, serverClient pkgc
...
 	err := serverClient.Get(ctx, pkgclient.ObjectKey{Name: oauthClientSecrets.Name, Namespace: r.ConfigManager.GetOperatorNamespace()}, oauthClientSecrets)
 	if err != nil {
-		return "", errors.Wrapf(err, "Could not find %s Secret", oauthClientSecrets.Name)
+		return "", fmt.Errorf("Could not find %s Secret: %w", oauthClientSecrets.Name, err)
 	}
 
 	clientSecretBytes, ok := oauthClientSecrets.Data[string(r.Config.GetProductName())]
 	if !ok {
-		return "", errors.Wrapf(err, "Could not find %s key in %s Secret", string(r.Config.GetProductName()), oauthClientSecrets.Name)
+		return "", fmt.Errorf("Could not find %s key in %s Secret: %w", string(r.Config.GetProductName()), oauthClientSecrets.Name, err)
```
matskiv: @grdryn this is the same case you mentioned to me in PM, `err` can be nil...

pkg/controller/installation/products/rhssouser/reconciler.go
```diff
@@ -262,12 +261,12 @@ func (r *Reconciler) setupOpenshiftIDP(ctx context.Context, inst *v1alpha1.Insta
...
 	err := serverClient.Get(ctx, pkgclient.ObjectKey{Name: oauthClientSecrets.Name, Namespace: r.ConfigManager.GetOperatorNamespace()}, oauthClientSecrets)
 	if err != nil {
-		return pkgerr.Wrapf(err, "Could not find %s Secret", oauthClientSecrets.Name)
+		return fmt.Errorf("Could not find %s Secret: %w", oauthClientSecrets.Name, err)
 	}
 
 	clientSecretBytes, ok := oauthClientSecrets.Data[string(r.Config.GetProductName())]
 	if !ok {
-		return pkgerr.Wrapf(err, "Could not find %s key in %s Secret", string(r.Config.GetProductName()), oauthClientSecrets.Name)
+		return fmt.Errorf("Could not find %s key in %s Secret: %w", string(r.Config.GetProductName()), oauthClientSecrets.Name, err)
```
matskiv: @grdryn I think here is the same issue, `err` can be nil and even if it's not, it's still unrelated to this error message.

test/e2e/integreatly_test.go
```diff
@@ -445,11 +442,11 @@ func checkIntegreatlyNamespaceLabels(t *testing.T, f *framework.Framework, names
 		namespace := &corev1.Namespace{}
 		err := f.Client.Get(goctx.TODO(), client.ObjectKey{Name: intlyNamespacePrefix + namespaceName}, namespace)
 		if err != nil {
-			return errors.Wrap(err, "Error getting namespace: "+namespaceName+" from cluster")
+			return fmt.Errorf("Error getting namespace: %v from cluster: %w", namespaceName, err)
 		}
 		value, ok := namespace.Labels[label]
 		if !ok || value != "true" {
-			return errors.Wrap(err, "Incorrect label on integreatly namespace: "+namespaceName+". Expected: "+label+". Got: "+label+"="+value)
+			return fmt.Errorf("Incorrect %v label on integreatly namespace: %v. Expected: true. Got: %v. %w", label, namespaceName, value, err)
```
matskiv: @grdryn similiar issue here^, `err` object is irrelevant to this error statement

test/e2e/integreatly_test.go
```diff
@@ -460,13 +457,13 @@ func checkOperatorVersions(t *testing.T, f *framework.Framework, namespace strin
...
 	if err != nil {
-		return errors.Wrap(err, "Error getting installation CR from cluster when checking operator versions")
+		return fmt.Errorf("Error getting installation CR from cluster when checking operator versions: %w", err)
 	}
 
 	for product, version := range operatorVersions {
 		clusterVersion := installation.Status.Stages[stage].Products[operator.ProductName(product)].OperatorVersion
 		if clusterVersion != operator.OperatorVersion(version) {
-			return errors.Wrap(err, fmt.Sprintf("Error with version of %s operator deployed on cluster. Expected %s. Got %s", product, version, clusterVersion))
+			return fmt.Errorf("Error with version of %s operator deployed on cluster. Expected %s. Got %s. %w", product, version, clusterVersion, err)
```
matskiv: @grdryn This seems to be the same case as other comments - `err` is irrelevant here.

grdryn:
Thanks for pointing out those other cases @matskiv. I hate the way `err` is always in scope after its first declaration, and the convention to always re-use the name `err`. If we weren't, then the compiler would catch such things I think!

test/e2e/integreatly_test.go
```diff
@@ -478,13 +475,13 @@ func checkOperandVersions(t *testing.T, f *framework.Framework, namespace string
...
 	if err != nil {
-		return errors.Wrap(err, "Error getting installation CR from cluster when checking operand versions")
+		return fmt.Errorf("Error getting installation CR from cluster when checking operand versions: %w", err)
 	}
 
 	for product, version := range operandVersions {
 		clusterVersion := installation.Status.Stages[stage].Products[operator.ProductName(product)].Version
 		if clusterVersion != operator.ProductVersion(version) {
-			return errors.Wrap(err, fmt.Sprintf("Error with version of %s deployed on cluster. Expected %s. Got %s", product, version, clusterVersion))
+			return fmt.Errorf("Error with version of %s deployed on cluster. Expected %s. Got %s. %w", product, version, clusterVersion, err)
```
matskiv: @grdryn This seems to be the same case as other comments - err is irrelevant here.

matskiv:
I dropped a few comments about the issues I found, but everything else looks good. I'll approve once the comments are addressed. :)
Great job @grdryn

grdryn:
@matskiv thanks for the review, I believe all are addressed now. It's a good job the original one caused a test failure so we knew what to look for! :) At the same time, changing to `fmt.Errorf` would have resulted in a new error anyway, even if it was trying to wrap an existing error (nil or not).

grdryn:
@matskiv thanks for the review, I believe all are addressed now. It's a good job the original one caused a test failure so we knew what to look for! :) At the same time, changing to `fmt.Errorf` would have resulted in a new error anyway, even if it was trying to wrap an existing error (nil or not).

test/e2e/integreatly_test.go
grdryn: This one "broke" the e2e test -- it was silently succeeding up to now when it should have been erroring. I've added a new commit to fix that, not sure if there will be additional failures after that.

grdryn:
The tests pass! :balloon: Could you please give this one last lgtm @matskiv ?

PR #280, opened 2019-12-19
INTLY-4661 - update code ready operator to use 2.0.0

KevFan:
## Description
Upgrade the code ready operator version to 2.0.0 GA

Jira
* https://issues.redhat.com/browse/INTLY-4661

## Verification
* Clone this branch
* Run the operator locally against cluster with at least these products installed:
```
operator-sdk up local --namespace=integreatly --operator-flags=--products=rhsso,rhssouser,cloud-resources,codeready-workspaces
```
* Ensure codeready gets installed correctly
* Run the `setup-htpass-idp.sh` script and note the outputed password for `customer-admin`
```
./scripts/setup-htpass-idp.sh
```
* Login in to code ready using the Single Sign On option using the `htpasswd` provider using the `customer-admin` credentials as noted above
* Input account information while keeping username as `customer-admin`
* Link to existing `customer-admin` accont, using the password: `Password1`
* Ensure user signed it correctly and can provision a workspace without visible errors


Makefile
```diff
@@ -89,7 +89,7 @@ test/unit:
 .PHONY: test/e2e
 test/e2e: export GH_CLIENT_ID := 1234
 test/e2e: export GH_CLIENT_SECRET := 1234
-test/e2e: cluster/cleanup cluster/prepare cluster/prepare/configmaps
+test/e2e: cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/prepare/configmaps cluster/prepare/crd deploy/integreatly-installation-cr.yml
```
grdryn: How about have `cluster/cleanup` depend on `cluster/cleanup/crds`, and `cluster/prepare` depend on `cluster/prepare/configmaps` and `cluster/prepare/crd`?

I'm not sure what expectations you folks have around running `cluster/prepare` or `cluster/cleanup, so it may not be appropriate.

Makefile
```diff
@@ -133,9 +132,12 @@ cluster/prepare/configmaps:
 cluster/prepare/osrc:
 	- oc process -p NAMESPACE=$(NAMESPACE) OPERATOR_SOURCE_REGISTRY_NAMESPACE=$(ORG) -f deploy/operator-source-template.yml | oc apply -f - -n openshift-marketplace
 
+.PHONY: cluster/prepare/crd
+cluster/prepare/crd:
```
grdryn: I'd consider changing this to `cluster/prepare/crds`, to correspond to `cluster/cleanup/crds`. WDYT?

Makefile
```diff
@@ -170,15 +172,18 @@ cluster/cleanup/crds:
 	@-oc delete crd installations.integreatly.org
 	@-oc delete crd webapps.integreatly.org
 
+.PHONY: deploy/integreatly-installation-cr.yml
```
grdryn: I'm not sure about the need for this to be .PHONY -- it's a target that corresponds to a file that is created as a result of running it, so it's a more normal target than our usual ones, right?

If we're going to change it to a phony target, it might be worth changing the name to something that's not also a file path, e.g. `deploy/cr`.

Oh gosh, now I'm questioning the sense in having phony targets that look like file paths (slashy) :thinking:

Makefile
```diff
@@ -170,15 +172,18 @@ cluster/cleanup/crds:
...
+deploy/integreatly-installation-cr.yml: export INSTALLATION_PREFIX := rhmi
 deploy/integreatly-installation-cr.yml:
 	@echo "selfSignedCerts = $(SELF_SIGNED_CERTS)"
-	sed "s/INSTALLATION_NAME/$(INSTALLATION_NAME)/g" | \
+	sed "s/INSTALLATION_NAME/$(INSTALLATION_NAME)/g" deploy/crds/examples/integreatly-installation-cr.yaml | \
 	sed "s/INSTALLATION_TYPE/$(INSTALLATION_TYPE)/g" | \
+	sed "s/INSTALLATION_PREFIX/$(INSTALLATION_PREFIX)/g" | \
 	sed "s/SELF_SIGNED_CERTS/$(SELF_SIGNED_CERTS)/g" > deploy/integreatly-installation-cr.yml
-
+	@-oc create -f deploy/integreatly-installation-cr.yml
```
grdryn: Ah, I guess this is why it's a phony target now.

As I see it, there are a couple of things here:
- It creates the file and then runs `oc create`, so even if the file exists, it's a phony target to make sure `oc create` is run even if the file already existed

BUT!

- If the file already exists, then we have to assume that this was already run, so `oc create` was already run; and running `oc create` twice does not bring joy (even if we are ignoring failures)

There are a couple of ideas that could be done to improve on this I think (pick and choose if you think they're good):

- We could move the `oc create` away, and make it a boring ol' non-phony file target again
- If the file isn't actually important, we could remove creation of it, and just pipe to `oc create -f -`
- _Maybe_ we should default to `oc apply` rather than `oc create` in make recipes (not 100% sure on that, just a feeling)

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/RHsyseng/operator-utils v0.0.0-20191024171829-7e918ca09e5e
 	github.com/aerogear/unifiedpush-operator v0.0.0-20191212165307-3ea5e7aa705f
 	github.com/coreos/prometheus-operator v0.29.0
-	github.com/eclipse/che-operator v0.0.0-20190620145712-2f639261d8b5
+	github.com/eclipse/che-operator v0.0.0-20191122191946-81d08d3f0fde
```
grdryn: I can see that this is the commit that corresponds to the 2.0.0.GA branch, but I'm confused why it's still saying `v0.0.0` :thinking: 

@davidffrench thoughts? Is that because the che-operator is not a "module project" yet, or is it because of the `.GA` suffix on the tag, or something else?

davidffrench: @grdryn I am under the impression this happens when the dependency version does not specifically have the format `v`+`x.x.x`. Some dependencies don't include the `v` in their git tags.

Makefile
KevFan: I'm good with having `cluster/cleanup` to depend on `cluster/cleanup/crds` as it looks like the `cluster/cleanup` is only used from the `test/e2e` command

Also good with `cluster/prepare` to depend on `cluster/prepare/configmaps` and `cluster/prepare/crd` for the same reason.

From what I can see `cluster/prepare` and `cluster/cleanup` both are just used in `test/e2e` though am not sure are people using them much outside of this

Makefile
KevFan: Yeah, thats good with me

Makefile
KevFan: Reason why I did this was now both `test/e2e` and `test/e2e/olm` depends on `deploy/integreatly-installation-cr.yml` and needed to create the cr also

`cluster/deploy/integreatly-installation-cr.yml` depends on `deploy/integreatly-installation-cr.yml` beforehand anyhow,  previously with the `oc create` command to create the cr so just wanted to remove the extra need of havin the same `oc create` in the `test/e2e`

> We could move the oc create away, and make it a boring ol' non-phony file target again


Yeah, I'm good with moving the `oc create` away again but I would need to add it back in to `cluster/deploy/integreatly-installation-cr.yml` and also to `test/e2e`

> If the file isn't actually important, we could remove creation of it, and just pipe to oc create -f - 


I think the file is important since it looks like its used in `cluster/cleanup/olm`

> Maybe we should default to oc apply rather than oc create in make recipes (not 100% sure on that, just a feeling)


Not sure on this, haven't used `oc apply` much but I think it should work

go.mod
grdryn: Seems likely. from [here](https://github.com/golang/go/wiki/Modules#modules):

> Modules must be semantically versioned according to semver, usually in the form v(major).(minor).(patch), such as v0.1.0, v1.2.3, or v1.5.0-rc.1. The leading v is required. If using Git, tag released commits with their versions.

grdryn:
@KevFan thinking about it more, trying to cleanup all the Makefile stuff will likely lead down an unnecessary rabbit hole. I'm going to create a new issue to clean up some of the stuff in it, but feel free to make additional changes in this PR if there's anything you want to clean up in it.

I'll lgtm this now, & verify it on a cluster. Let me know if you're making any further changes here & if not & verification goes well, I'll approve. :+1: 

/lgtm

grdryn:
Verified successfully & everything seems to work! :100: 

/approve

PR #283, opened 2020-01-03
💚 Run e2e on image in cluster, not local binary

grdryn:
In https://github.com/integr8ly/integreatly-operator/pull/282 we're
changing the operator to watch all namespaces, and it's not possible
to run the e2e tests with the "--up-local" flag & also set the
namespace variable to an empty string.

----

#

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,11 @@ test/unit:
 test/e2e: export GH_CLIENT_ID := 1234
 test/e2e: export GH_CLIENT_SECRET := 1234
 test/e2e: cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/prepare/configmaps cluster/prepare/crd deploy/integreatly-installation-cr.yml
-	INTEGREATLY_OPERATOR_DISABLE_ELECTION=true operator-sdk --verbose test local ./test/e2e --namespace $(NAMESPACE) --up-local --go-test-flags "-timeout=60m" --debug
+ifdef OPENSHIFT_BUILD_NAMESPACE
+	operator-sdk --verbose test local ./test/e2e --namespace $(NAMESPACE) --go-test-flags "-timeout=60m" --debug --image "registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:integreatly-operator"
+else
+	operator-sdk --verbose test local ./test/e2e --namespace $(NAMESPACE) --go-test-flags "-timeout=60m" --debug
+endif
 
 .PHONY: test/e2e/olm
 test/e2e/olm: export AWS_ACCESS_KEY_ID := 1234
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ COMPILE_TARGET=./tmp/_output/bin/$(PROJECT)
 OPERATOR_SDK_VERSION=0.12.0
 AUTH_TOKEN=$(shell curl -sH "Content-Type: application/json" -XPOST https://quay.io/cnr/api/v1/users/login -d '{"user": {"username": "$(QUAY_USERNAME)", "password": "${QUAY_PASSWORD}"}}' | jq -r '.token')
 TEMPLATE_PATH="$(shell pwd)/templates/monitoring"
+INTEGREATLY_OPERATOR_IMAGE ?= $(REG)/$(ORG)/$(PROJECT):v$(TAG)
 
 define wait_command
 	@echo Waiting for $(2) for $(3)...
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -90,7 +90,11 @@ test/unit:
 test/e2e: export GH_CLIENT_ID := 1234
 test/e2e: export GH_CLIENT_SECRET := 1234
 test/e2e: cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/prepare/configmaps cluster/prepare/crd deploy/integreatly-installation-cr.yml
-	INTEGREATLY_OPERATOR_DISABLE_ELECTION=true operator-sdk --verbose test local ./test/e2e --namespace $(NAMESPACE) --up-local --go-test-flags "-timeout=60m" --debug
+ifdef OPENSHIFT_BUILD_NAMESPACE
```
grdryn: I think what might be better here is to actually set an additional env var on the job in the ci-operator config, e.g. here: https://github.com/openshift/release/blob/master/ci-operator/jobs/integr8ly/integreatly-operator/integr8ly-integreatly-operator-master-presubmits.yaml#L32

This would allow us locally to specify what image to use also.

Makefile
```diff
@@ -69,28 +70,32 @@ code/fix:
...
 .PHONY: image/build/test
 image/build/test:
-	operator-sdk build --enable-tests $(REG)/$(ORG)/$(PROJECT):v$(TAG)
+	operator-sdk build --enable-tests $(INTEGREATLY_OPERATOR_IMAGE)
 
 .PHONY: test/unit
 test/unit:
 	TEMPLATE_PATH=$(TEMPLATE_PATH) ./scripts/ci/unit_test.sh
 
+.PHONY: prow/test/e2e
```
grdryn: I think we need to _either_ get prow to use this target, _or_ remove this and set the `INTEGREATLY_OPERATOR_IMAGE` env var in the prow config.

I'm not sure which is better, or if there are better alternatives. @philbrookes @mikenairn thoughts?

mikenairn:
@grdryn We had an issue on the backlog to make this change here https://issues.redhat.com/browse/INTLY-4210, are you working to that or some other Jira?

Makefile
mikenairn: We can change the prow config to use this target instead https://github.com/openshift/release/blob/master/ci-operator/config/integr8ly/integreatly-operator/integr8ly-integreatly-operator-master.yaml#L57. As per the jira i linked, we planned to do that anyway https://issues.redhat.com/browse/INTLY-4210

Main issue i see with this that we will not get any logs on failure so figuring out why something has failed will become more difficult unless we can gather operator logs and store them in the build artifacts.

grdryn:
/unhold

@mikenairn I think this is how we want it now. Could you take a look when you've got a moment? I'll create a PR to the openshift/release repo after lunch then

Makefile
```diff
@@ -69,26 +70,30 @@ code/fix:
...
-	operator-sdk build --enable-tests $(REG)/$(ORG)/$(PROJECT):v$(TAG)
+	operator-sdk build --enable-tests $(INTEGREATLY_OPERATOR_IMAGE)
 
 .PHONY: test/unit
 test/unit:
 	TEMPLATE_PATH=$(TEMPLATE_PATH) ./scripts/ci/unit_test.sh
 
+.PHONY: test/e2e/prow
+prow/test/e2e: INTEGREATLY_OPERATOR_IMAGE := "registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:integreatly-operator"
+prow/test/e2e: test/e2e
```
mikenairn: Names don't match here, did we want to call this test/e2e/prow or prow/test/e2e?

grdryn: Good spot, fixed :+1:

Makefile
```diff
@@ -69,23 +70,27 @@ code/fix:
...
 image/build/test:
-	operator-sdk build --enable-tests $(REG)/$(ORG)/$(PROJECT):v$(TAG)
+	operator-sdk build --enable-tests $(INTEGREATLY_OPERATOR_IMAGE)
 
 .PHONY: test/unit
 test/unit:
 	TEMPLATE_PATH=$(TEMPLATE_PATH) ./scripts/ci/unit_test.sh
 
+.PHONY: test/e2e/prow
+test/e2e/prow: INTEGREATLY_OPERATOR_IMAGE := "registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:integreatly-operator"
```
mikenairn: Does this need export as well?

grdryn: I wouldn't consider that one a blocker for now, since it's not being used, and there are no downstream targets of this. Since we know we'll need another change after the openshift/release one, we can export this in that also.

PR #286, opened 2020-01-08
☸ Run e2e tests with image deployed on cluster

grdryn:
I've also removed the manual leader election toggle. Leader election
will be skipped if running locally anyway now:

```
{"level":"info","ts":1578480384.2597575,"logger":"leader","msg":"Trying to become the leader."}
{"level":"info","ts":1578480384.2597752,"logger":"leader","msg":"Skipping leader election; not running in a cluster."}
```

There was one other reference to it, which didn't seem like a great
usage of it (it looks like we wanted to check if we were running
locally, rather than if leader election should be disabled).

### DIFF
--- a/pkg/resources/oauthResolver.go
+++ b/pkg/resources/oauthResolver.go
@@ -10,6 +10,8 @@ import (
 	"os"
 
 	"github.com/sirupsen/logrus"
+
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 )
 
 const oauthServerDetails = "%s/.well-known/oauth-authorization-server"
@@ -33,7 +35,7 @@ func (or *OauthResolver) GetOauthEndPoint() (*OauthServerConfig, error) {
 
 	caCert, err := ioutil.ReadFile(rootCAFile)
 	// if running locally, CA certificate isn't available in expected path
-	if os.IsNotExist(err) || os.Getenv("INTEGREATLY_OPERATOR_DISABLE_ELECTION") != "" {
+	if os.IsNotExist(err) && os.Getenv(k8sutil.ForceRunModeEnv) == string(k8sutil.LocalRunMode) {
 		logrus.Warn("GetOauthEndPoint() will skip certificate verification - this is acceptable only if operator is running locally")
 		tr := &http.Transport{
 			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ test/e2e/prow: test/e2e
 test/e2e: export GH_CLIENT_ID := 1234
 test/e2e: export GH_CLIENT_SECRET := 1234
 test/e2e: cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/prepare/configmaps cluster/prepare/crd deploy/integreatly-installation-cr.yml
-	INTEGREATLY_OPERATOR_DISABLE_ELECTION=true operator-sdk --verbose test local ./test/e2e --namespace $(NAMESPACE) --up-local --go-test-flags "-timeout=60m" --debug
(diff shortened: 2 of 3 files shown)


pkg/resources/oauthResolver.go
```diff
@@ -33,7 +35,7 @@ func (or *OauthResolver) GetOauthEndPoint() (*OauthServerConfig, error) {
 
 	caCert, err := ioutil.ReadFile(rootCAFile)
 	// if running locally, CA certificate isn't available in expected path
-	if os.IsNotExist(err) || os.Getenv("INTEGREATLY_OPERATOR_DISABLE_ELECTION") != "" {
+	if os.IsNotExist(err) && os.Getenv(k8sutil.ForceRunModeEnv) == string(k8sutil.LocalRunMode) {
```
grdryn: @matskiv I've changed the logic here slightly -- my understanding is that with the old version we'd be setting InsecureSkipVerify if the file was not found, regardless of whether the operator was running locally or not.

I was additionally thinking of changing it to just:

```go
if os.Getenv(k8sutil.ForceRunModeEnv) != string(k8sutil.LocalRunMode) {
```

and then doing the ReadFile (L36) in the `else` part. WDYT?

grdryn:
/cc @mikenairn @philbrookes @matskiv

The main motivation for this is to get the e2e tests running using the built image, and make sure that works (before/without) the changes in #282.

grdryn:
image


/this-is-fine

pkg/resources/oauthResolver.go
matskiv: I like the switch to `k8sutil.ForceRunModeEnv`, but I am inclined to keeping `os.IsNotExist(err) ||` piece.

I think your change would mean we would always have to set `OSDK_FORCE_RUN_MODE` (value of the `k8sutil.ForceRunModeEnv` variable). That might require some tweaks to local debugging config on everyone's machine.

matskiv: @grdryn and yes, what you are saying in first paragraph of your comment is correct.
AFAIK in normal working cluster that file will always be present, so if it's not there, we must be running locally.

grdryn: @matskiv would you expect people to run locally in another way than `operator-sdk up local` (this is what we've got in the `make code/run` target)? If not, then this env var will be set as expected for this scenario:

https://github.com/operator-framework/operator-sdk/blob/6d51fa1760b8905acd3f855f95f1f7c11d7626d5/cmd/operator-sdk/up/local.go#L139

(I verified this by logging out the value of the env var locally)

matskiv: @grdryn when running the operator from IDE I am building the code and running the binary directly. I think I am not the only one who uses this approach when debugging locally.
I know that one can use `operator-sdk up local --enable-delve` for debugging, but it's not as convenient as one click on the "debug" button in IDE :)

Either way, it's not a deal breaker. We just have to let everyone know that they need to add new variable into their debugging config.

grdryn: @matskiv ah right :thinking: For that case, I think it might never get to this part, because the main.go would fail to become the leader (it does a check for the env var value implicitly, where it used to depend on the `INTEGREATLY_OPERATOR_DISABLE_ELECTION` being set...kind of).

I'd prefer to force everyone who doesn't use the `operator-sdk up local` to manually set the env var, rather than potentially shipping something that "works" but goes into InsecureSkipVerify silently in production because a file was missing in the parent image or something.

Is it easy to set those variables in your IDE for debugging?

matskiv: @grdryn oops, I didn't notice that change in main.go.
Yes, let's go with your proposed solution, it's going to be more reliable.

Changing the variables isn't difficult, we just need to let everyone know :)

PR #291, opened 2020-01-09
honor CR_NAMESPACE, remove unrequired secret

philbrookes:
- CR Creation only triggered if CR_NAMESPACE is set.
- No longer need github-oauth-secret, so removed it.


deploy/operator.yaml
```diff
@@ -24,9 +24,7 @@ spec:
             - name: WATCH_NAMESPACE
               value: ""
             - name: CR_NAMESPACE
-              valueFrom:
-                fieldRef:
-                  fieldPath: metadata.namespace
+              value: "integreatly-operator"
```
matskiv: is this namespace present in e2e environment?

deploy/olm-catalog/integreatly-operator/integreatly-operator-1.15.0/integreatly-operator.v1.15.0.clusterserviceversion.yaml
```diff
@@ -32,7 +32,7 @@ metadata:
     capabilities: Basic Install
     categories: Integreating greatly
     certified: "false"
-    containerImage: quay.io/integreatly/integreatly-operator:v1.14.0
+    containerImage: quay.io/integreatly/integreatly-operator:v1.15.0
```
matskiv: This probably shouldn't be in a "remove github-oauth-secret" commit, since as we discussed, we might need to revert that in the future...

philbrookes: yeah, in reality the release to 1.15.0 should be it's own PR, so I've removed it altogether

PR #292, opened 2020-01-09
INTLY-4780 cloud resource cleanup

aidenkeating:
avoid cloud resource custom resources being recreated and left around by the Integreatly operator.

verification:
- TODO


aidenkeating:
@JameelB @dimitraz would you mind taking a look? I believe I've managed to reproduce the issue. I then ran this change and it appeared to resolve it. However, it'd be good to get an opinion of someone more familiar with the issue.

JameelB:
/approve

Verified these changes on OSD. Tested this by deleting the installation CR at the start of the products stage. Uninstallation completed and any recreated cloud resources instances were cleaned up successfully.

PR #293, opened 2020-01-10
Filter integreatly in subscription controller

matskiv:
Now, that the integreatly operator will only be installed in cluster mode, it will automatically watch all Subsriptions on the cluster.
This PR should ensure that it will only change it's own subscription to Manual.


matskiv:
I fixed the unit tests and added some new ones.

PR #295, opened 2020-01-10
feat: updated tests

psturc:
https://issues.redhat.com/browse/INTLY-4757


mikenairn:
Changes look good. Verified using the test script for the container image against an image i built locally and the one already built on CI, all worked as expected. :+1:

PR #297, opened 2020-01-13
Intly 2932 remove customer admin

briangallagher:
**JIRAs:**
https://issues.redhat.com/browse/INTLY-2932

**What:**
Remove the creation of CustomerAdminUser in Keycloak
Remove "realm-management" permissions as managing users is now done via OSD configuration

**Verification:**
- Run basic install and verify success
- Run go test -v -coverprofile cover.out ./pkg/products/rhsso/...
- Run go test -v -coverprofile cover.out ./pkg/products/threescale/...
- run ./scripts/setup-htpass-idp.sh
- Verify customer-admin does not exist in OpenShift, Keycloak and ThreeScale

### DIFF
--- a/pkg/products/rhsso/reconciler.go
+++ b/pkg/products/rhsso/reconciler.go
@@ -34,7 +34,6 @@ import (
 
 var (
 	defaultRhssoNamespace               = "rhsso"
-	customerAdminPassword               = "Password1"
 	keycloakName                        = "rhsso"
 	keycloakRealmName                   = "openshift"
 	defaultSubscriptionName             = "integreatly-rhsso"
@@ -43,19 +42,6 @@ var (
 	githubOauthAppCredentialsSecretName = "github-oauth-secret"
 	manifestPackage                     = "integreatly-rhsso"
 )
-
-var CustomerAdminUser = &aerogearv1.KeycloakUser{
-	KeycloakApiUser: &aerogearv1.KeycloakApiUser{
-		Enabled:       true,
-		Attributes:    aerogearv1.KeycloakAttributes{},
-		UserName:      "customer-admin",
-		EmailVerified: true,
-		Email:         "customer-admin@example.com",
-		ClientRoles:   getKeycloakRoles(true),
-	},
-	Password:     &customerAdminPassword,
-	OutputSecret: "customer-admin-user-credentials",
-}
 
 type Reconciler struct {
 	Config        *config.RHSSO
@@ -247,9 +233,6 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
 			"metrics-listener",
 		}
 
-		if kcr.Spec.Users == nil {
-			kcr.Spec.Users = []*aerogearv1.KeycloakUser{CustomerAdminUser}
-		}
 		users, err := syncronizeWithOpenshiftUsers(kcr.Spec.Users, ctx, serverClient)
 		if err != nil {
 			return err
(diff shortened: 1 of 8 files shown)


scripts/setup-htpass-idp.sh
```diff
@@ -10,14 +10,11 @@ if [[ ! -f "htpasswd" ]]; then
...
-
 htpasswd -b htpasswd test-user ${PASSWORD}
 echo user added test-user ${PASSWORD}
 
 oc delete secret htpasswd-secret -n openshift-config
 oc create secret generic htpasswd-secret --from-file=htpasswd=htpasswd -n openshift-config
 oc patch oauth cluster --type=merge -p '{ "spec": { "identityProviders": [{ "name": "htpasswd_provider", "challenge": true, "login": true, "mappingMethod": "claim", "type": "HTPasswd", "htpasswd": { "fileData": { "name": "htpasswd-secret" } } }] } }'
 
-oc adm groups add-users dedicated-admins customer-admin
+oc adm groups add-users dedicated-admins test-user
```
matskiv: I don't think we want to make any changes to this script.
Intention was to remove only the hard-coded customer-admin user.
We still need "customer-admin" user in this script to test permissions of a user in "dedicated-admins" group.
And we need "test-user" to test permissions of a user that wasn't added to any group explicitly.

JIRA description might have been a little misleading, sorry for that.

pkg/products/rhsso/reconciler.go
```diff
@@ -530,10 +514,6 @@ func syncronizeWithOpenshiftUsers(keycloakUsers []*aerogearv1.KeycloakUser, ctx
 		return nil, err
 	}
 	for _, kcUser := range keycloakUsers {
-		if kcUser.UserName == CustomerAdminUser.UserName {
-			continue
-		}
-
 		kcUser.ClientRoles = getKeycloakRoles(isOpenshiftAdmin(kcUser, openshiftAdminGroup))
```
pb82: `isOpenshiftAdmin(kcUser, openshiftAdminGroup)` can be removed since now it's always the same roles that get assiged to every user

pkg/products/threescale/reconciler.go
```diff
@@ -815,48 +814,6 @@ func (r *Reconciler) reconcileUpdatingDefaultAdminDetails(ctx context.Context, s
 		return integreatlyv1alpha1.PhaseFailed, err
 	}
 
-	kcUsers := filterUsers(kcr.Spec.Users, func(u *aerogearv1.KeycloakUser) bool {
```
pb82: Looks like the whole function (reconcileUpdatingDefaultAdminDetails) can be removed since it doesn't modify anything anymore.

matskiv: Good catch, it won't do anything now. 
But there is some editing of "system-seed" secret that I don't fully understand, maybe we need that logic..?
Let's create a follow up task, as this can be considered out of scope for this PR. wdyt?

jjaferson:
FYI: I have done those changes requested in the PR review and I am just waiting on the operator installation in my cluster to finish @briangallagher

scripts/setup-htpass-idp.sh
jjaferson: I have reverted it back

jjaferson:
@pb82 @briangallagher @matskiv I have created another PR addressing these changes mentioned in this PR and fixing some issues. https://github.com/integr8ly/integreatly-operator/pull/307

briangallagher:
test

PR #300, opened 2020-01-13
INTLY-4817 - create Fuse view binding for whole Group

matskiv:
JIRA: https://issues.redhat.com/browse/INTLY-4817

We don't need to keep track of individual users in rhmi-developers group to give them view access to Fuse ns. We can create a role binding for the group instead.

P.S: I also added small fix of import name.


KevFan:
Verified on OSD

/approve
/lgtm

PR #301, opened 2020-01-14
Keycloak operator 8.0.1

pb82:


pkg/products/rhsso/reconciler.go
```diff
@@ -215,105 +316,217 @@ func (r *Reconciler) reconcileTemplates(ctx context.Context, installation *integ
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
 
+// workaround: the keycloak operator creates a route with TLS passthrough config
+// this should use the same valid certs as the cluster itself but for some reason the
+// signing operator gives out self signed certs
+// to circumvent this we create another keycloak route with edge termination
```
philbrookes: Is this going to be addressed in a future release of keycloak / the keycloak-operator? If so, can we create a Jira to remove this workaround at that point, so we don't forget?

pkg/products/rhsso/reconciler.go
```diff
@@ -366,35 +572,39 @@ func (r *Reconciler) setupOpenshiftIDP(ctx context.Context, installation *integr
...
-			r.Config.GetHost() + "/auth/realms/openshift/broker/openshift-v4/endpoint",
-		},
-		GrantMethod: oauthv1.GrantHandlerPrompt,
+		Secret:       clientSecret,
+		RedirectURIs: redirectUris,
+		GrantMethod:  oauthv1.GrantHandlerPrompt,
 	}
 
-	_, err = r.ReconcileOauthClient(ctx, installation, oauthc, serverClient)
+	_, err = r.ReconcileOauthClient(ctx, inst, oauthClient, serverClient)
```
philbrookes: We refer to the installation object as `installation` everywhere now, to try to keep our code consistent

pkg/products/rhsso/reconciler.go
```diff
@@ -494,15 +702,15 @@ func getUserDiff(keycloakUsers []*aerogearv1.KeycloakUser, openshiftUsers []user
 
 	var deleted []int
 	for i, kcUser := range keycloakUsers {
-		if kcUser.UserName != CustomerAdminUser.UserName && !OsUserInKc(openshiftUsers, kcUser) {
```
philbrookes: is this a typo? `OsUserInKc` => `IsUserInKc`?

pb82: I think this stands for `Openshift user in Keycloak`

pkg/products/rhsso/reconciler.go
pb82: will update

pkg/products/rhsso/reconciler.go
pb82: Yes, the PR to update the route in the keycloak operator has now been merged, but this is not in the 8.0.1 release: https://github.com/keycloak/keycloak-operator/pull/115

PR #306, opened 2020-01-16
[WIP] RANDOM UPDATE DO NOT MERGE

mikenairn:
Test vendor jobs

/hold

### DIFF
--- a/test/e2e/integreatly_test.go
+++ b/test/e2e/integreatly_test.go
@@ -5,6 +5,8 @@ import (
 	goctx "context"
 	"encoding/json"
 	"fmt"
+	"os"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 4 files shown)


test/e2e/integreatly_test.go
```diff
@@ -549,10 +551,24 @@ func waitForInstallationStageCompletion(t *testing.T, f *framework.Framework, na
...
 	ctx := framework.NewTestCtx(t)
 	defer ctx.Cleanup()
-	err := ctx.InitializeClusterResources(&framework.CleanupOptions{TestContext: ctx, Timeout: cleanupTimeout, RetryInterval: cleanupRetryInterval})
+
+	cleanupResources, _ := strconv.ParseBool(getEnv("CLEANUP_RESOURCES", "true"))
+	cleanupOptions := &framework.CleanupOptions{TestContext: ctx, Timeout: cleanupTimeout, RetryInterval: cleanupRetryInterval}
+	if !cleanupResources {
+		cleanupOptions = nil
+	}
+	err := ctx.InitializeClusterResources(cleanupOptions)
```
mikenairn: @philbrookes @grdryn @pb82 Just wondering how we feel about adding a change like this?

Problem that we are trying to get round here is that since we test the image running in the cluster now, we currently can't get the logs archived for any failed builds because the test suite destroys all the resources before it gets to archive them (Separate stage in the CI pipeline after the "test" stage).

This change stops the test cleaning up resources that it created, so all of them would be archived, and in particular the logs of the integreatly operator would be available for inspection. Example from this PR (deliberately failing) here https://storage.googleapis.com/origin-ci-test/pr-logs/pull/integr8ly_integreatly-operator/306/pull-ci-integr8ly-integreatly-operator-master-e2e/780/artifacts/e2e/pods/integreatly_integreatly-operator-675c8757b8-vsfrd_integreatly-operator.log. This is fine on CI clusters where we never try and run the e2e test twice, but shouldn't be done on clusters where we are testing multiple times.

The thing we are no longer testing though is that the clean up of these resources work (finalizers), which IMO should always work and should be tested.

mikenairn: I wonder if we could have it not cleanup if the test has failed at some point, but do the cleanup if it didn't :thinking:

philbrookes:
Can this PR be closed?

PR #310, opened 2020-01-20
✨ Add Spec.SMTPSecret & Status.SMTPEnabled to CRD

grdryn:
JIRA: https://issues.redhat.com/browse/INTLY-4787

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,12 @@ AUTH_TOKEN=$(shell curl -sH "Content-Type: application/json" -XPOST https://quay
 TEMPLATE_PATH="$(shell pwd)/templates/monitoring"
 INTEGREATLY_OPERATOR_IMAGE ?= $(REG)/$(ORG)/$(PROJECT):v$(TAG)
 
+export SELF_SIGNED_CERTS   ?= true
+export INSTALLATION_TYPE   ?= managed
+export INSTALLATION_NAME   ?= integreatly
+export INSTALLATION_PREFIX ?= redhat-rhmi
+export USE_CLUSTER_STORAGE ?= true
+
 define wait_command
 	@echo Waiting for $(2) for $(3)...
 	@time timeout --foreground $(3) bash -c "until $(1); do echo $(2) not ready yet, trying again in $(4)s...; sleep $(4); done"
@@ -110,7 +116,6 @@ test/e2e: cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/prepare/c
 test/e2e/olm: cluster/cleanup/olm cluster/prepare/olm cluster/prepare/configmaps cluster/deploy/integreatly-installation-cr.yml
 
 .PHONY: cluster/deploy/integreatly-installation-cr.yml
-cluster/deploy/integreatly-installation-cr.yml: export INSTALLATION_NAME := integreatly-operator
 cluster/deploy/integreatly-installation-cr.yml: deploy/integreatly-installation-cr.yml
 	$(call wait_command, oc get Installation $(INSTALLATION_NAME) -n $(NAMESPACE) --output=json -o jsonpath='{.status.stages.bootstrap.phase}' | grep -q completed, bootstrap phase, 5m, 30)
 	$(call wait_command, oc get Installation $(INSTALLATION_NAME) -n $(NAMESPACE) --output=json -o jsonpath='{.status.stages.monitoring.phase}' | grep -q completed, monitoring phase, 10m, 30)
@@ -175,11 +180,6 @@ cluster/cleanup/crds:
 	@-oc delete crd webapps.integreatly.org
 
 .PHONY: deploy/integreatly-installation-cr.yml
-deploy/integreatly-installation-cr.yml: export SELF_SIGNED_CERTS := true
-deploy/integreatly-installation-cr.yml: export INSTALLATION_NAME := integreatly
-deploy/integreatly-installation-cr.yml: export INSTALLATION_TYPE := managed
-deploy/integreatly-installation-cr.yml: export INSTALLATION_PREFIX := redhat-rhmi
-deploy/integreatly-installation-cr.yml: export USE_CLUSTER_STORAGE := true
 deploy/integreatly-installation-cr.yml:
 	@echo "selfSignedCerts = $(SELF_SIGNED_CERTS)"
 	sed "s/INSTALLATION_NAME/$(INSTALLATION_NAME)/g" deploy/crds/examples/integreatly-installation-cr.yaml | \
--- a/Makefile
+++ b/Makefile
@@ -50,8 +50,8 @@ setup/git/hooks:
 	git config core.hooksPath .githooks
 
 .PHONY: code/run
-code/run:
-	@operator-sdk up local --namespace="${NAMESPACE}"
+code/run: code/gen cluster/prepare/smtp
+	@operator-sdk up local --namespace="$(NAMESPACE)"
 
 .PHONY: code/run/service_account
 code/run/service_account: setup/service_account
@@ -124,7 +124,7 @@ cluster/deploy/integreatly-installation-cr.yml: deploy/integreatly-installation-
 	$(call wait_command, oc get Installation $(INSTALLATION_NAME) -n $(NAMESPACE) --output=json -o jsonpath='{.status.stages.solution-explorer.phase}' | grep -q completed, solution-explorer phase, 10m, 30)
 
 .PHONY: cluster/prepare
-cluster/prepare: cluster/prepare/project cluster/prepare/osrc
+cluster/prepare: cluster/prepare/project cluster/prepare/osrc cluster/prepare/smtp
 
 .PHONY: cluster/prepare/project
 cluster/prepare/project:
@@ -145,7 +145,7 @@ cluster/prepare/crd:
 	- oc create -f deploy/crds/*_crd.yaml
 
 .PHONY: cluster/prepare/local
-cluster/prepare/local: cluster/prepare/project cluster/prepare/crd
+cluster/prepare/local: cluster/prepare/project cluster/prepare/crd cluster/prepare/smtp
 	@oc create -f deploy/service_account.yaml
 	@oc create -f deploy/role.yaml
 	@oc create -f deploy/role_binding.yaml
@@ -155,6 +155,15 @@ cluster/prepare/olm: cluster/prepare/project cluster/prepare/osrc
 	oc process -p NAMESPACE=$(NAMESPACE) -f deploy/operator-subscription-template.yml | oc create -f - -n $(NAMESPACE)
 	$(call wait_command, oc get crd installations.integreatly.org, installations.integreatly.org crd, 1m, 10)
 	$(call wait_command, oc get deployments integreatly-operator -n $(NAMESPACE) --output=json -o jsonpath='{.status.availableReplicas}' | grep -q 1, integreatly-operator ,2m, 10)
+
+.PHONY: cluster/prepare/smtp
+cluster/prepare/smtp:
+	@-oc create secret generic rhmi-smtp -n $(NAMESPACE) \
+		--from-literal=host=smtp.example.com \
+		--from-literal=username=dummy \
+		--from-literal=password=dummy \
+		--from-literal=port=587 \
+		--from-literal=tls=true
 
 .PHONY: cluster/cleanup
 cluster/cleanup:
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 A Kubernetes Operator based on the Operator SDK for installing and reconciling Integreatly products.
 
-### Project status: _alpha_ 
+### Project status: _alpha_
 
 This is a proof of concept/alpha version. Most functionality is present but it is highly likely there are bugs and improvements needed.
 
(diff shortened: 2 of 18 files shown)


pkg/apis/integreatly/v1alpha1/installation_types.go
```diff
@@ -91,6 +92,17 @@ type InstallationSpec struct {
...
+	// SMTPSecret is the name of a secret in the installation
+	// namespace containing SMTP connection details. The secret
+	// must contain the following fields:
+	//
+	// host
+	// port
+	// tls
+	// username
+	// password
+	SMTPSecret string `json:"smtpSecret,omitempty"`
```
grdryn: Should I change this to be required (remove the `,omitempty`)?

pkg/apis/integreatly/v1alpha1/installation_types.go
```diff
@@ -101,13 +113,14 @@ type PullSecretSpec struct {
...
 type InstallationStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// INSERT ADDITIONAL STATUS FIELDS - define observed state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
 	Stages           map[StageName]*InstallationStageStatus `json:"stages"`
 	PreflightStatus  PreflightStatus                        `json:"preflightStatus,omitempty"`
 	PreflightMessage string                                 `json:"preflightMessage,omitempty"`
 	LastError        string                                 `json:"lastError"`
+	SMTPEnabled      bool                                   `json:"smtpEnabled,omitempty"`
```
grdryn: Should I change this to be required (remove the `,omitempty`)?

pkg/controller/installation/installation_controller.go
```diff
@@ -363,6 +391,12 @@ func (r *ReconcileInstallation) preflightChecks(installation *integreatlyv1alpha
 
 	result := reconcile.Result{}
 	requiredSecrets := []string{"github-oauth-secret"}
```
grdryn: Not sure how the `github-oauth-secret` gets created. If we've got the ability, it might be good to add the `owner: integreatly` label to it so that it gets watched like the SMTP secret.

pkg/apis/integreatly/v1alpha1/installation_types.go
HubertStefanski: Can the service be enabled at a later date?, if not then I'd say to have this as required, otherwise I don't think it makes that much of a difference

pkg/apis/integreatly/v1alpha1/installation_types.go
HubertStefanski: I think it's fine to leave it with `,omitempty` for now

grdryn: One of the concerns with making these new fields required is that it makes the CRD backwards-incompatible with the previous release, and people wouldn't be able to upgrade without manual intervention.

I think the way that this would generally work is creating a new resource version (say, v1alpha2). That would be a different can of worms that I'd rather not have to deal with if possible! :sweat_smile:

pkg/controller/installation/watch.go
```diff
@@ -0,0 +1,131 @@
+package installation
+
```
grdryn: @philbrookes if you think it would be good, I could also put the contents of `enqueueIntegreatlyOwner.go` in here too (grouping all of the stuff related to watching in this file).

pkg/apis/integreatly/v1alpha1/installation_types.go
HubertStefanski: Well. then it seems like we've a valid reason to leave it as it is :smile:

pkg/controller/installation/watch.go
philbrookes: So we have removed the enqueueIntegreatlyObject now, instead using a system of @matskiv devising, that doesn't require a cluster level operator, if possible to use that, it'd be far preferred, we found that acting at cluster level caused far too many headaches.

@matskiv If I'm thinking correctly, we can delete (you did already delete?) the enqueueforintegreatly?

philbrookes:
As the SMTP credentials missing blocks the install, this is probably breaking the e2e test, you will need to modify the e2e test to create the required secret before they start.

pkg/controller/installation/watch.go
matskiv: I haven't deleted the enqueueforintegreatly, but we can do that, since our new Watches for CRDs are watching only specific namespaces.

Makefile
```diff
@@ -92,21 +97,26 @@ test/unit:
...
+test/e2e/image: export GH_CLIENT_SECRET := 1234
+test/e2e/image: cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/prepare/configmaps cluster/prepare/crd deploy/integreatly-installation-cr.yml
+	operator-sdk --verbose test local ./test/e2e --namespace="$(NAMESPACE)" --go-test-flags "-timeout=60m" --debug --image=$(INTEGREATLY_OPERATOR_IMAGE)
 
 .PHONY: test/e2e
 test/e2e: export GH_CLIENT_ID := 1234
 test/e2e: export GH_CLIENT_SECRET := 1234
 test/e2e: cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/prepare/configmaps cluster/prepare/crd deploy/integreatly-installation-cr.yml
-	operator-sdk --verbose test local ./test/e2e --namespace="$(NAMESPACE)" --go-test-flags "-timeout=60m" --debug --image=$(INTEGREATLY_OPERATOR_IMAGE)
+	operator-sdk --verbose test local ./test/e2e --namespace="$(NAMESPACE)" --go-test-flags "-timeout=60m" --debug
```
grdryn: @philbrookes @mikenairn Now that the operator is no longer cluster-scoped again, how do you feel about this change to make the default `test/e2e` not require a pre-built image? For the moment, prow will still use the image that it builds.

pkg/apis/integreatly/v1alpha1/installation_types.go
grdryn: This is set to true based on the Secret being found, and it will block for the _managed_ Installation type, but won't for the workshop Installation type.

pkg/controller/installation/installation_controller.go
grdryn: We don't need these labels now -- I've changed it to watch all secrets in the namespace.

grdryn:
> As the SMTP credentials missing blocks the install, this is probably breaking the e2e test, you will need to modify the e2e test to create the required secret before they start.

That should be happening here: https://github.com/integr8ly/integreatly-operator/pull/310/files#diff-e3e74a4fcd6e24a416beb0a2b8f02762R661

grdryn:
/test e2e

I think the test is just timing out, so I'll retry for a while. It successfully completed when I ran it locally & it took just under 30 mins.

grdryn:
The e2e tests have passed! :tada:

pkg/controller/installation/installation_controller.go
```diff
@@ -121,6 +140,7 @@ func createInstallationCR(ctx context.Context, serverClient k8sclient.Client) er
 				Type:            string(integreatlyv1alpha1.InstallationTypeManaged),
 				NamespacePrefix: DefaultInstallationPrefix,
 				SelfSignedCerts: false,
+				SMTPSecret:      "rhmi-smtp",
```
aidenkeating: To confirm, if the RHMI operator is left to create it's own Installation CR. The secret will be expected to be called `rhmi-smtp` and will be expected to be in the same namespace as the RHMI operator?

Just confirming for SOP work

aidenkeating:
@grdryn I did rebase this and try it out, it worked and I could send an email via 3scale. I also tried it out with the new CLI and it all worked end to end.

Let me know if/when you're ready for another review/verification.

Makefile
mikenairn: Sorry, only seeing this. Why do we want to do this?

pkg/controller/installation/installation_controller.go
grdryn: Yes :+1: 

Also, currently the secret will always be expected to be in the same namespace as the operator/CR, no matter what it's named.

Makefile
grdryn: Just a thought, so that people running the e2e tests locally don't need to first build the operator image & publish it somewhere. They still could do that though, by using `make test/e2e/image`. Maybe it would be preferable to have the non-image way be `make test/e2e/local` and leave the default be with-image?

It might not be worth the effort though, as I'm not sure how to get the operator logs from the non-image way. I'll remove this change from this PR anyway, as it's just something that could cause a delay in getting it merged.

pkg/apis/integreatly/v1alpha1/installation_types.go
```diff
@@ -103,14 +115,15 @@ type PullSecretSpec struct {
...
 type InstallationStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// INSERT ADDITIONAL STATUS FIELDS - define observed state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
 	Stages             map[StageName]*InstallationStageStatus `json:"stages"`
 	PreflightStatus    PreflightStatus                        `json:"preflightStatus,omitempty"`
 	PreflightMessage   string                                 `json:"preflightMessage,omitempty"`
 	LastError          string                                 `json:"lastError"`
 	SetupGHCredentials bool                                   `json:"setupGHCredentials,omitempty"`
```
grdryn: @jjaferson I noticed that this is new :slightly_smiling_face: 

I think it would be nice if both this and the SMTP flag that I'm adding on the line below, were named consistently (at least I _think_ they have similar purposes). Would there be any issue with me changing this one to be `GitHubEnabled` or `GHCredentialsEnabled` or something like that (suggestions welcome)?

pkg/controller/installation/installation_controller.go
```diff
@@ -324,6 +344,36 @@ func (r *ReconcileInstallation) preflightChecks(installation *integreatlyv1alpha
 	logrus.Info("Running preflight checks..")
 
 	result := reconcile.Result{}
+	eventRecorder := r.mgr.GetEventRecorderFor("Preflight Checks")
+
+	if installation.Spec.Type == string(integreatlyv1alpha1.InstallationTypeManaged) {
+		requiredSecrets := []string{installation.Spec.SMTPSecret}
```
grdryn: @aidenkeating @philbrookes This is now the only secret being checked in pre-flight, since @jjaferson's recent change handles the github-oauth-secret in a different way. I'm not sure if this should now be handled in that other way, or if it's still best here.

pkg/apis/integreatly/v1alpha1/installation_types.go
jjaferson: not at all, I am not very creative when it comes to naming things, I'd go for `GHCredentialsEnabled` but any of them are good to me.

grdryn: Thanks! For now I've changed it to GitHubOAuthEnabled, since that's the thing being enabled.

README.md
```diff
@@ -61,6 +61,27 @@ export GH_CLIENT_SECRET=<client secret>
 make cluster/prepare/local
 ```
 
+* 3Scale requires SMTP credentials to be able to send mail to users.
+
+**Note:** If this secret isn't created, the integreatly preflight checks will fail
```
aidenkeating: To confirm, this will only happen in `managed` installations?

pkg/controller/installation/installation_controller.go
aidenkeating: @grdryn I don't believe any changes to this approach have been confirmed. @philbrookes Are you still ok with this approach?

For now the flow would be:
- RHMI installation halts
- SRE complete SOP to create SMTP Secret
- RHMI installation continues

Eventually, the flow would be:
- RHMI installation halts
- Automated SMTP Service creates Secret on cluster
- RHMI installation continues

README.md
grdryn: @aidenkeating yes, that's a good point. The preflight checks will skip the secret check for the other kind(s) of installation. That reminds me that I haven't tried with a non-`managed` installation.

Also, maybe I should add a dummy secret for the `make code/run`, since that's a managed installation. WDYT?

aidenkeating: @grdryn I think a dummy secret for `make code/run` makes sense as it'll make development simpler. I think people will likely be confused what secret this is if it blocks them all of a sudden.

pkg/products/threescale/reconciler.go
```diff
@@ -316,26 +316,10 @@ func (r *Reconciler) getOauthClientSecret(ctx context.Context, serverClient k8sc
...
-
-	// wait for the smtp credential set cr to reconcile
-	if smtpCred.Status.Phase != types.PhaseComplete {
-		return integreatlyv1alpha1.PhaseAwaitingComponents, nil
-	}
 
 	// get the secret containing smtp credentials
 	credSec := &corev1.Secret{}
-	err = serverClient.Get(ctx, k8sclient.ObjectKey{Name: smtpCred.Status.SecretRef.Name, Namespace: smtpCred.Status.SecretRef.Namespace}, credSec)
+	err := serverClient.Get(ctx, k8sclient.ObjectKey{Name: r.installation.Spec.SMTPSecret, Namespace: r.installation.Namespace}, credSec)
```
aidenkeating: If the error is a does not exist error, I'd think we don't want to fail out here? Especially when the installation is non-`managed`

grdryn: Yeah, I was just looking at this in the context of non-managed. I'll update and make it work.

pkg/controller/installation/installation_controller.go
grdryn: There's maybe another concern here also: the `SMTPSecret` field is optional (because we already do releases of this operator project here on GitHub, so we shouldn't break compatibility & I don't really want to have to support a v1alpha1 and v1alpha2 at the same time (although at v1alpha* I'm not sure if we'd be expected to? Maybe I should just change to v1alpha2 (dropping v1alpha1 immediately) & make it mandatory?).

README.md
grdryn: Dummy secret added.

aidenkeating:
/approve

verified with @austinmartinh on my cluster on OSD. saw a workshop installation complete with @grdryn also

pkg/controller/installation/installation_controller.go
grdryn: Going to leave it as optional for now. Once we're on 4.3, we can look at adding a default value for it.

PR #312, opened 2020-01-21
🚑 Fix dependency on github.com/openshift/api

grdryn:
It looks like the openshift/api project deleted all of their tags, so
now the following error is thrown any time we run e.g. a go mod tidy
in any project that depends on it:

```
go: github.com/operator-framework/operator-sdk@v0.12.1-0.20191112211508-82fc57de5e5b requires
        github.com/operator-framework/operator-lifecycle-manager@v0.0.0-20190605231540-b8a4faf68e36 requires
        github.com/openshift/api@v3.9.1-0.20190424152011-77b8897ec79a+incompatible: invalid pseudo-version: preceding tag (v3.9.0) not found
```

For this change, all I did was set an override in the "replace"
section of the go.mod, and ran go mod tidy (which changed a bunch of
other things...).

See openshift/api#456 for more info on the issue.


grdryn:
Setting `GOPROXY=https://proxy.golang.org` is a sufficient workaround for this for now

PR #314, opened 2020-01-21
split operators and operands

philbrookes:
INTLY-3371 and INTLY-4956


pkg/resources/reconciler.go
```diff
@@ -111,18 +112,17 @@ func (r *Reconciler) ReconcileNamespace(ctx context.Context, namespace string, i
 
 		return integreatlyv1alpha1.PhaseCompleted, nil
 	}
-	// ns exists so check it is our namespace
-	if !IsOwnedBy(ns, inst) && ns.Status.Phase != corev1.NamespaceTerminating {
-		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("existing namespace found with name %v but it is not owned by the integreatly installation and it isn't being deleted", ns.Name)
+
+	PrepareObject(ns, inst)
+	if err := client.Update(ctx, ns); err != nil {
```
matskiv: I am not sure why this was changed. With new behavior, any "left-over" namespaces from a previous Installation instance will be re-purposed for a new installation. This will happen because we will override "ownership" label instead of failing as we did in the past.
btw: if we keep this change, we no longer need `IsOwnedBy` function.

pkg/products/rhsso/reconciler.go
```diff
@@ -124,31 +123,49 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
 		}
-		return integreatlyv1alpha1.PhaseCompleted, nil
+
+		//if both namespaces are deleted, return complete
+		_, operatorNSErr := resources.GetNS(ctx, r.Config.GetOperatorNamespace(), serverClient)
+		_, nsErr := resources.GetNS(ctx, r.Config.GetNamespace(), serverClient)
+		if k8serr.IsNotFound(operatorNSErr) && k8serr.IsNotFound(nsErr) {
+			return integreatlyv1alpha1.PhaseCompleted, nil
+		}
+		return integreatlyv1alpha1.PhaseInProgress, nil
```
matskiv: I think operator will get to this point only if the namespaces are gone, so this couple of lines above are not necessary, and we can leave just `return integreatlyv1alpha1.PhaseCompleted, nil` as it was before.
But if my assumption is incorrect, and there was a bug that caused this change, then we perhaps need to do it in all other products as well?

pkg/products/monitoring/reconciler.go
```diff
@@ -142,13 +142,33 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+				return phase, err
+			}
+		}
+		//if both namespaces are deleted, return complete
+		_, operatorNSErr := resources.GetNS(ctx, r.Config.GetOperatorNamespace(), serverClient)
+		_, nsErr := resources.GetNS(ctx, r.Config.GetNamespace(), serverClient)
+		if k8serr.IsNotFound(operatorNSErr) && k8serr.IsNotFound(nsErr) {
+			return integreatlyv1alpha1.PhaseCompleted, nil
+		}
+		return integreatlyv1alpha1.PhaseInProgress, nil
```
matskiv: Same as I commented on `pkg/products/rhsso/reconciler.go`
IMHO 7 lines above can be replaced with just `return integreatlyv1alpha1.PhaseCompleted, nil`

Makefile
```diff
@@ -177,10 +178,10 @@ cluster/cleanup/crds:
 	@-oc delete crd webapps.integreatly.org
 
 .PHONY: deploy/integreatly-installation-cr.yml
-deploy/integreatly-installation-cr.yml: export SELF_SIGNED_CERTS := true
-deploy/integreatly-installation-cr.yml: export INSTALLATION_NAME := integreatly-operator
+deploy/integreatly-installation-cr.yml: export SELF_SIGNED_CERTS := false
```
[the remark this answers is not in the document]
matskiv: I asked Mike about this, I'll post an update once I get a reply :)

pkg/config/fuseonopenshift.go
```diff
@@ -26,6 +26,10 @@ func (f *FuseOnOpenshift) SetNamespace(newNamespace string) {
 	f.config["NAMESPACE"] = newNamespace
 }
 
+func (f *FuseOnOpenshift) GetOperatorNamespace() string {
+	return f.config["NAMESPACE"] + "-operator"
+}
+
```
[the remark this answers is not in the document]
philbrookes: yeah, I was just fulfilling the interface.

pkg/resources/reconciler.go
philbrookes: This was removed because the installation keeps failing on "namespace exists but is not owned by us", and as we're using a reserved namespace prefix, I don't think this worry is too pressing.

pkg/products/rhsso/reconciler.go
philbrookes: I thought that too, but I was wrong, these reconcilers were reporting complete too early, and not removing both namespaces.

pkg/products/monitoring/reconciler.go
philbrookes: lets just discuss this above.

pkg/config/fuseonopenshift.go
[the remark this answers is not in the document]
matskiv: But `GetOperatorNamespace()` wasn't added to the interface :D
https://github.com/integr8ly/integreatly-operator/blob/5317a8136898538b1b02c27e923477ab2a2a2537/pkg/config/manager.go#L61-L69

pkg/products/rhsso/reconciler.go
matskiv: @philbrookes ok, I think it could be attributed to the last condition here:
https://github.com/integr8ly/integreatly-operator/blob/9e160fdd416762b7a32667539b650f1a93b4802e/pkg/resources/finalizers.go#L44-L51

I think an "admin" permissions for our SA are removed before the NS is gone, and with that^ logic it gives a "false possitive".
To fix this, the `RemoveNamespace` function should ignore IsForbidden errors, and return PhaseCompleted only when GetNS return NotFound error. So something like this:
```
func RemoveNamespace(ctx context.Context, inst *integreatlyv1alpha1.Installation, client k8sclient.Client, namespace string) (integreatlyv1alpha1.StatusPhase, error) {
	ns, err := GetNS(ctx, namespace, client)
	if err != nil {
		if k8serr.IsNotFound(err)
		}
		// Since we are using ProjectRequests and limited permissions,
		// request can return "forbidden" error even when Namespace simply doesn't exist
		if !k8serr.IsForbidden(err) {
			logrus.Error("Error getting a namespace", err)
		}
	} else {
		if ns.Status.Phase == corev1.NamespaceTerminating {
	  		return integreatlyv1alpha1.PhaseInProgress, nil
		}
        }
...
```
Does that make sense?

matskiv: My code above makes it rather complicated.
Perhaps in the RemoveNamespace we should just be checking for the existence of Project resource, instead of `GetNS`. Maybe that would be more reliable and readable?

philbrookes: I'll add something to tech debt for this.

pkg/config/fuseonopenshift.go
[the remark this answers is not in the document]
philbrookes: oh lol

Makefile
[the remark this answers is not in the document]
philbrookes: ah good point! it doesn't seem to be upsetting anything so far.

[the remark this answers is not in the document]
philbrookes: I'll wait for an update from @mikenairn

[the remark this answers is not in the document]
matskiv: @philbrookes Mike said in the chat: "it does not have valid certs afaik"

PR #319, opened 2020-01-23
Further refactoring and syntaxing to INTLY-4907

HubertStefanski:
https://issues.redhat.com/browse/INTLY-4907

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ code/compile:
 
 .PHONY: code/gen
 code/gen: deploy/crds/integreatly.org_installations_crd.yaml
-	find ./ -name *_moq.go -type f -delete -not -path "./vendor/*"
+	find ./ -name *_moq.go -type f -not -path "./vendor/*" -delete
 	operator-sdk generate k8s
 	operator-sdk generate openapi
 	@go generate ./...
--- a/Makefile
+++ b/Makefile
@@ -56,8 +56,8 @@ code/compile:
 	@GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o=$(COMPILE_TARGET) ./cmd/manager
 
 .PHONY: code/gen
-code/gen:
-	find ./ -name *_moq.go -type f -not -path "./vendor/*"  -delete
+code/gen: deploy/crds/integreatly.org_installations_crd.yaml
+	find ./ -name *_moq.go -type f -delete -not -path "./vendor/*"
 	operator-sdk generate k8s
 	operator-sdk generate openapi
 	@go generate ./...
--- a/pkg/controller/installation/types.go
+++ b/pkg/controller/installation/types.go
@@ -122,15 +122,15 @@ func (t *Type) GetStages() []Stage {
 	return t.Stages
 }
 
-func InstallationTypeFactory(installationType string, products []string) (error, *Type) {
+func InstallationTypeFactory(installationType string, products []string) (*Type, error) {
 	//TODO: export this logic to a configmap for each installation type
 	switch installationType {
 	case string(integreatlyv1alpha1.InstallationTypeWorkshop):
-		return nil, newWorkshopType(products)
+		return newWorkshopType(products), nil
 	case string(integreatlyv1alpha1.InstallationTypeManaged):
-		return nil, newManagedType(products)
+		return newManagedType(products), nil
 	default:
-		return errors.New("unknown installation type: " + installationType), nil
+		return nil, errors.New("unknown installation type: " + installationType)
 	}
 }
 
--- a/pkg/controller/installation/types.go
+++ b/pkg/controller/installation/types.go
@@ -122,7 +122,7 @@ func (t *Type) GetStages() []Stage {
 	return t.Stages
 }
 
-func InstallationTypeFactory(installationType string, products []string) (*Type, error) {
+func TypeFactory(installationType string, products []string) (*Type, error) {
 	//TODO: export this logic to a configmap for each installation type
 	switch installationType {
 	case string(integreatlyv1alpha1.InstallationTypeWorkshop):
--- a/pkg/products/monitoring/reconciler.go
+++ b/pkg/products/monitoring/reconciler.go
@@ -392,7 +392,7 @@ func (r *Reconciler) populateParams(ctx context.Context, serverClient k8sclient.
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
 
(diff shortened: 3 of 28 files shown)


pkg/products/monitoring/reconciler.go
```diff
@@ -387,7 +387,7 @@ func (r *Reconciler) populateParams(ctx context.Context, serverClient k8sclient.
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
 
-func CreateBlackboxTarget(name string, target monitoring_v1alpha1.BlackboxtargetData, ctx context.Context, cfg *config.Monitoring, installation *integreatlyv1alpha1.Installation, serverClient k8sclient.Client) error {
+func CreateBlackboxTarget(ctx context.Context, target monitoring_v1alpha1.BlackboxtargetData, name string, cfg *config.Monitoring, installation *integreatlyv1alpha1.Installation, serverClient k8sclient.Client) error {
```
grdryn: I don't think it's necessary to shift the `name` param to after the `target` one. This could instead be:

```suggestion
func CreateBlackboxTarget(ctx context.Context, name string, target monitoring_v1alpha1.BlackboxtargetData, cfg *config.Monitoring, installation *integreatlyv1alpha1.Installation, serverClient k8sclient.Client) error {
```

pkg/products/rhsso/reconciler.go
```diff
@@ -779,7 +779,7 @@ func OsUserInKc(osUsers []usersv1.User, kcUser keycloak.KeycloakAPIUser) bool {
 	return false
 }
 
-func (r *Reconciler) createOrUpdateKeycloakUser(user keycloak.KeycloakAPIUser, inst *integreatlyv1alpha1.Installation, ctx context.Context, serverClient k8sclient.Client) (controllerutil.OperationResult, error) {
+func (r *Reconciler) createOrUpdateKeycloakUser(ctx context.Context, inst *integreatlyv1alpha1.Installation, user keycloak.KeycloakAPIUser, serverClient k8sclient.Client) (controllerutil.OperationResult, error) {
```
grdryn: Similar to above, the `user` param doesn't need to be moved to after the `inst` one I think. So it could instead be:

```suggestion
func (r *Reconciler) createOrUpdateKeycloakUser(ctx context.Context, user keycloak.KeycloakAPIUser, inst *integreatlyv1alpha1.Installation, serverClient k8sclient.Client) (controllerutil.OperationResult, error) {
```

pkg/resources/marketplace/manifestHelper.go
```diff
@@ -66,10 +66,9 @@ func GetFilesFromManifestAsStringList(dir string, regex string, packageYaml stri
 			}
 
 			return err
-		} else { // Otherwise find all matching files and process to a string list
-			if err == nil && libRegEx.MatchString(info.Name()) {
-				return ProcessYamlFile(path, &stringList)
-			}
+		} // Otherwise find all matching files and process to a string list
```
grdryn: Now that this doesnt have an `else`, I think the comment would be better placed on a line by itself:

```suggestion
		}
		// Otherwise find all matching files and process to a string list
```

pkg/resources/pullSecret.go
```diff
@@ -33,8 +33,8 @@ func GetDefaultPullSecret(inst *integreatlyv1alpha1.Installation, client k8sclie
 }
 
 // Copys the default pull secret to a target namespace
-func CopyDefaultPullSecretToNameSpace(nameSpaceToCopy string, nameOfSecret string, inst *integreatlyv1alpha1.Installation, client k8sclient.Client, context context.Context) error {
-	openshiftSecret, err := GetDefaultPullSecret(inst, client, context)
+func CopyDefaultPullSecretToNameSpace(context context.Context, nameOfSecret string, inst *integreatlyv1alpha1.Installation, client k8sclient.Client, nameSpaceToCopy string) error {
```
grdryn:
```suggestion
func CopyDefaultPullSecretToNameSpace(context context.Context, nameSpaceToCopy string, nameOfSecret string, inst *integreatlyv1alpha1.Installation, client k8sclient.Client) error {
```

pkg/controller/installation/types.go
```diff
@@ -122,15 +122,15 @@ func (t *Type) GetStages() []Stage {
 	return t.Stages
 }
 
-func InstallationTypeFactory(installationType string, products []string) (error, *Type) {
+func TypeFactory(installationType string, products []string) (*Type, error) {
```
grdryn: For the name here, it's saying that it's not great to have `installation.InstallationTypeFactory`, and `installation.TypeFactory` would be better, right?

In this case, I can't see any usage of it outside this package, so we only ever say `TypeFactory`. If I'm correct, then I think it would be better to unexport this, and keep the name the same. WDYT?

In essence:
```suggestion
func installationTypeFactory(installationType string, products []string) (*Type, error) {
```

HubertStefanski: Think this was resolved in another PR, the recent rebase already has this implemented it seems

pkg/resources/pullSecret.go
```diff
@@ -33,8 +33,8 @@ func GetDefaultPullSecret(inst *integreatlyv1alpha1.Installation, client k8sclie
 }
 
 // Copys the default pull secret to a target namespace
-func CopyDefaultPullSecretToNameSpace(nameSpaceToCopy string, nameOfSecret string, inst *integreatlyv1alpha1.Installation, client k8sclient.Client, context context.Context) error {
-	openshiftSecret, err := GetDefaultPullSecret(inst, client, context)
+func CopyDefaultPullSecretToNameSpace(context context.Context, nameSpaceToCopy string, inst *integreatlyv1alpha1.Installation, client k8sclient.Client, nameOfSecret string) error {
```
grdryn: For some reason now `nameOfSecret` got moved to the end.

```suggestion
func CopyDefaultPullSecretToNameSpace(context context.Context, nameSpaceToCopy string, nameOfSecret string, inst *integreatlyv1alpha1.Installation, client k8sclient.Client) error {
```

The change on line 20 above also looks to have the `inst` variable moved to the end

HubertStefanski:
@grdryn Mind doing a final review? if it's ready to merge ?

Makefile
```diff
@@ -56,8 +56,8 @@ code/compile:
 	@GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o=$(COMPILE_TARGET) ./cmd/manager
 
 .PHONY: code/gen
-code/gen:
-	find ./ -name *_moq.go -type f -not -path "./vendor/*"  -delete
+code/gen: deploy/crds/integreatly.org_installations_crd.yaml
```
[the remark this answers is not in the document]
HubertStefanski: Yes it is, I added it to this PR when I was generating files, for convenience, The whole update for the MakeFile will come with #326

PR #321, opened 2020-01-23
make sure multiple replicas are deployed for components in 3scale

wei-lee:
https://issues.redhat.com/browse/INTLY-5008


pkg/apis/3scale/v1alpha1/apimanager_types.go
```diff
@@ -2,17 +2,35 @@ package v1alpha1
 
 import (
 	"fmt"
-
 	"github.com/RHsyseng/operator-utils/pkg/olm"
 
-	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
davidffrench: this should stay named `corev1` with our new import naming convention

pkg/apis/3scale/v1alpha1/register.go
```diff
@@ -7,7 +7,7 @@ package v1alpha1
 
 import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/controller-runtime/pkg/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/scheme"
```
davidffrench: this should stay as `"sigs.k8s.io/controller-runtime/pkg/scheme"`

pkg/apis/3scale/v1alpha1/apimanager_types.go
dimitraz: @davidffrench is this convention defined anywhere? Want to make sure I'm also following it correctly

wei-lee: where is that naming convention defined?

This is a copy/paste from the 3scale operator. I tried to add 3scale-operator as a dependency so that we don't need to duplicate the api definitions, but unfortunately it doesn't work. The version of the operator-sdk it's using is just too old. Hopefully with Sergio's work, we will be able to get rid of this from the integreatly operator completely.

Given that this is a copy/paste code and it is something we want to get rid of, I don't think it worth the effort to rename it.

pkg/apis/3scale/v1alpha1/register.go
wei-lee: ok

pkg/apis/3scale/v1alpha1/apimanager_types.go
davidffrench: Good questions. This was recently done by @grdryn in a couple of PRs. @grdryn Is the naming convention documented anywhere?

wei-lee:
@davidffrench I have this PR running on a cluster and you can check the 3scale APIManager CR [here](https://console-openshift-console.apps.weili-43.w3g9.s1.devshift.org/k8s/ns/rhmi-3scale/apps.3scale.net~v1alpha1~APIManager/3scale/yaml) and pod status [here](https://console-openshift-console.apps.weili-43.w3g9.s1.devshift.org/k8s/ns/rhmi-3scale/deploymentconfigs).

wei-lee:
@davidffrench @philbrookes @maleck13 turns out I can get rid of the duplicated 3scale API definitions and just add it as a dependency after all 😄

pkg/apis/3scale/v1alpha1/apimanager_types.go
wei-lee: @davidffrench this file is removed now.

grdryn:
@wei-lee I had made issues for removing all of those duplicated API packages. It looks like you're resolving https://issues.redhat.com/browse/INTLY-4641 here :)

davidffrench:
> @davidffrench @philbrookes @maleck13 turns out I can get rid of the duplicated 3scale API definitions and just add it as a dependency after all 😄

That is fantastic. What was the issue with including it, to begin with, and how did you manage to get around it. The same might be able to be done for the other duplicated API packages.

grdryn:
> That is fantastic. What was the issue with including it, to begin with, and how did you manage to get around it. The same might be able to be done for the other duplicated API packages.

@davidffrench there are issues on @philbrookes' backlog to de-duplicate all of these. I only know of an issue with importing the enmasse types as a Go module (we hit it with the unifiedpush-operator), not sure about the others. I've reported the enmasse issue upstream: https://github.com/EnMasseProject/enmasse/issues/3558

wei-lee:
@davidffrench the problem was the [version of `openshift/client-go`](https://github.com/wei-lee/integreatly-operator/blob/f1578fa67ef057ccbb6b74c0a659ae03ea609457/go.mod#L18) is using an [older version of `k8s/client-go`](https://github.com/wei-lee/integreatly-operator/blob/f1578fa67ef057ccbb6b74c0a659ae03ea609457/go.mod#L31). Since we [replaced the version of `k8s/client-go`](https://github.com/wei-lee/integreatly-operator/blob/f1578fa67ef057ccbb6b74c0a659ae03ea609457/go.mod#L45), we need to [do the same for `openshift/client-go`](https://github.com/wei-lee/integreatly-operator/blob/f1578fa67ef057ccbb6b74c0a659ae03ea609457/go.mod#L39) as well.

PR #325, opened 2020-01-24
Update SSO replicas to 3

davidffrench:
## What
Deploying RHSSO in a HA fashion by deploying with 3 replicas for both the cluster SSO and Application SSO

Removed the alert for number of pods running in the namespace as this is actually taken care of by an alert included in Keycloak now anyway - https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/prometheus_rule.go#L86-L94


davidffrench:
@wei-lee Could you put a `/lgtm` on this PR please

PR #326, opened 2020-01-24
Update Makefile to start optimisiation of code gen and rm obsolete generated files

HubertStefanski:
The purpose of this PR is to 

- Showcase the possibility of running `code/gen` depending on whether the source file has been modified  (If generated file is older than source file then `run code/gen` else do nothing)
- Possibly running `code/gen` in `pre-commit` .githook, cutting down forgetfulness and PR's without generated files, eliminating the need to generate them locally after pulling PR's or master after merge
- Cutting down `code/gen` time to only as much as needed and when it's needed
Example:
case 1: Source file for generation wasn't modified in commit
image
case 2: Source file has been modified and requires a new file to be generated
image
case 3: Generate all files, regardless of modified status
image

 
@grdryn Could perhaps give more context and elaborate?


vendor/github.com/integr8ly/cloud-resource-operator/pkg/resources/cluster_moq.go
```diff
@@ -1,83 +0,0 @@
-// Code generated by moq; DO NOT EDIT.
```
HubertStefanski: Just noticed this, will have to refactor to omit ./vendor in future versions

grdryn:
@HubertStefanski superb description! :clap: :clap: :clap: 

> @grdryn Could perhaps give more context and elaborate?

I think you've got it pretty much nailed -- because `make code/gen` currently takes more than a minute to run, nobody is inclined to run it very much. We _should_ have it as a pre-requisite task to the `make code/run` target (and maybe others) and the `.githooks/pre-commit`, but that would be extremely painful if it took 60+ seconds to run each time. The intention here is that we'll use Make as it was designed, and only run those expensive commands _when necessary_.

deploy/crds/integreatly.org_installations_crd.yaml
```diff
@@ -75,6 +75,7 @@ spec:
                 code after modifying this file Add custom validation using kubebuilder
                 tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html'
           required:
+          - lastError
```
grdryn: CASE In POINT: Here's a good example of someone having forgotten to run `make code/gen` or `operator-sdk generate openapi` before merging changes back to master.

philbrookes:
I think this is great, but also has the potential to fail (e.g. I just don't bother adding the pre-commit hook, or I'm new and unaware of it), so could we add a check to our CI too, similar to the one that checks `go fmt` hasn't modified any files, and do `make code/gen` and check there are no modified files, or if there are then fail the build?

Makefile
```diff
@@ -8,7 +8,7 @@ SHELL=/bin/bash
 PREVIOUS_TAG=1.14.0
 TAG=1.15.0
 PKG=github.com/integr8ly/integreatly-operator
-TEST_DIRS?=$(shell sh -c "find $(TOP_SRC_DIRS) -name \\*_test.go -exec dirname {} \\; | sort | uniq")
+TEST_DIRS?=$(shell sh -c "find $(aouTOP_SRC_DIRS) -name \\*_test.go -exec dirname {} \\; | sort | uniq")
```
philbrookes: is this a typo?

HubertStefanski: It certainly is, It's already been fixed since, going to be pushed in the next commit

grdryn:
@philbrookes Yes, we could add such a check to CI, and that would be a separate issue (and require changes in a separate project). This PR is first addressing a related-but-different problem, where it's expensive to run the generation commands locally (and therefore we don't add them to automatically run in the pre-commit hook, or as a pre-requisite to other locally-facing make targets).

Makefile
```diff
@@ -53,19 +55,24 @@ code/run/service_account: setup/service_account
...
 
 .PHONY: code/compile
-code/compile:
+code/compile: code/gen
 	@GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o=$(COMPILE_TARGET) ./cmd/manager
 
+deploy/crds/integreatly.org_installations_crd.yaml: pkg/apis/integreatly/v1alpha1/installation_types.go
+	operator-sdk generate openapi
+
+pkg/apis/integreatly/v1alpha1/zz_generated.deepcopy.go:	pkg/apis/integreatly/v1alpha1/installation_types.go
```
grdryn: @HubertStefanski maybe we need to duplicate this one to have `zz_generated.openapi.go` as a target also, WDYT?

HubertStefanski: Makes sense, as it's dependent on the same types.go file, Will do!

Makefile
```diff
@@ -53,19 +55,26 @@ code/run/service_account: setup/service_account
...
+code/gen: setup/moq deploy/crds/integreatly.org_installations_crd.yaml pkg/apis/integreatly/v1alpha1/zz_generated.deepcopy.go pkg/apis/integreatly/v1alpha1/zz_generated.openapi.go
 	find ./ -name *_moq.go -type f -not -path "./vendor/*" -delete
-	operator-sdk generate k8s
-	operator-sdk generate openapi
 	@go generate ./...
 
 .PHONY: code/check
 code/check:
 	@diff -u <(echo -n) <(gofmt -d `find . -type f -name '*.go' -not -path "./vendor/*"`)
+	golint
```
grdryn: This is misleading, since it doesn't do anything. It should either be removed or changed to something that we care about like:

```bash
golint ./pkg/... ./cmd/... ./test/... ./version/... | grep -v "or be unexported" | grep -v "comment on exported"
```

Maybe leave it out for now, as it's not really relevant to this PR. WDYT?

Makefile
```diff
@@ -16,6 +16,8 @@ AUTH_TOKEN=$(shell curl -sH "Content-Type: application/json" -XPOST https://quay
 TEMPLATE_PATH="$(shell pwd)/templates/monitoring"
 INTEGREATLY_OPERATOR_IMAGE ?= $(REG)/$(ORG)/$(PROJECT):v$(TAG)
 
+
+
```
grdryn: These two empty lines can be removed.

Makefile
HubertStefanski: I agree, going to change it ASAP

PR #328, opened 2020-01-27
feat: allow editing of GH secret and sync into SSO

jjaferson:
### What:
We can no longer create GitHub OAuth apps for users, but it is needed for GH repo import into CodeReady.
We need to allow customer to specify their own GitHub OAuth app credentials if they wish to use repo import feature.

https://issues.redhat.com/browse/INTLY-4818

### Verification Steps

1) install the operator
2) Ensure secret CR `github-oauth-secret` was created in the operator namespace with dummy values
3) Ensure Role and RoleBinding CRs `github-oauth-secret` were created for dedicated-admin SA in the operator namespace
4) Edit the github credentials in the `github-oauth-secret` secret CR 
5) Verify if the GH credentials were added to IdentityProviders in the KeycloakRealm CR on the rhsso namespace 
eg.: `oc get KeycloakRealm openshift  -n <rhsso-namespace> -o jsonpath='{.spec.realm.identityProviders}'`
6) Ensure the installation CR have a status `SetupGHCredentials` set to true
eg.: `oc get installation example-installation -o jsonpath='{.status.setupGHCredentials}` -n <operator-namespace>


pkg/controller/installation/bootstrapReconciler.go
```diff
@@ -158,6 +165,79 @@ func getConsoleRouteCR(ctx context.Context, serverClient k8sclient.Client) (*rou
...
+		return integreatlyv1alpha1.PhaseFailed, err
+	} else if k8serr.IsNotFound(err) {
+		githubOauthSecretCR.Data = map[string][]byte{
+			"clientId": []byte("dummy"),
+			"secret":   []byte("dummy"),
+		}
+	}
+	githubOauthSecretCR.ObjectMeta.ResourceVersion = ""
+
+	err = resources.CreateOrUpdate(ctx, serverClient, githubOauthSecretCR)
```
matskiv: I think a `serverClient.Create` in an `else if` branch above, would work for us.

btw: I think we will eventually remove our `resources.CreateOrUpdate` function once we switch to `controllerutil.CreateOrUpdate` in our code.

pkg/controller/installation/bootstrapReconciler.go
```diff
@@ -158,6 +165,79 @@ func getConsoleRouteCR(ctx context.Context, serverClient k8sclient.Client) (*rou
...
+			{
+				APIGroups:     []string{""},
+				Resources:     []string{"secrets"},
+				Verbs:         []string{"update", "get"},
+				ResourceNames: []string{r.ConfigManager.GetGHOauthClientsSecretName()},
+			},
+		},
+	}
+
+	err = resources.CreateOrUpdate(ctx, serverClient, secretRoleCR)
```
matskiv: It would be nicer to use `controllerutil.CreateOrUpdate` here.

pkg/controller/installation/bootstrapReconciler.go
```diff
@@ -158,6 +165,79 @@ func getConsoleRouteCR(ctx context.Context, serverClient k8sclient.Client) (*rou
...
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Name: "dedicated-admins",
+				Kind: "Group",
+			},
+		},
+	}
+
+	err = resources.CreateOrUpdate(ctx, serverClient, secretRoleBindingCR)
```
matskiv: Same as above - it would be nicer to use `controllerutil.CreateOrUpdate` here. :)

pkg/controller/installation/bootstrapReconciler.go
matskiv: Or as an alternative to a `serverClient.Get` + conditions, you can do a `serverClient.Create` and ignore "AllreadyExists" errors. :)

matskiv:
I added a few minor comments, but otherwise code looks good.
I will verify this a bit later, once my cluster is freed up.

maleck13:
@matskiv @jjaferson Do we want a watch on this secret. It seems we only use it during the SSO reconciliation. If the secret was updated by a customer how long before they would be able to use Code Ready. As a customer I would expect it to happen almost straight away.

matskiv:
@maleck13 yes good point. I was just verifying this and also found out we need a watch. :)

pkg/controller/installation/bootstrapReconciler.go
```diff
@@ -158,6 +166,76 @@ func getConsoleRouteCR(ctx context.Context, serverClient k8sclient.Client) (*rou
...
+		},
+		Data: map[string][]byte{
+			"clientId": []byte("dummy"),
+			"secret":   []byte("dummy"),
+		},
+	}
+
+	err := serverClient.Create(ctx, githubOauthSecretCR)
+	if err != nil {
+		if !k8serr.IsAlreadyExists(err) {
```
matskiv: Very minor, feel free to ignore: this condition can merged with the one the previous line :)

matskiv:
@jjaferson Code looks good. I am in progress of the verification.
Bumped into one issue though - operators role doesn't have permissions for Role/RoleBinding.
Can you add it please? I added permissions pasted below, and installation moved on.


In deploy/role.yaml, in "integreatly-operator" Role:
```
  # Adding a role and rolebinding to allow GitHub Secret management by dedicated-admins
  - apiGroups:
      - rbac.authorization.k8s.io
    resources:
      - roles
      - rolebindings
    verbs:
      - get
      - list
      - watch
      - create
      - update
      - delete
```

pkg/products/rhsso/reconciler.go
```diff
@@ -683,8 +682,44 @@ func (r *Reconciler) setupGithubIDP(ctx context.Context, kcr *keycloak.KeycloakR
...
+		return fmt.Errorf("Unable to authenticate to the Keycloak API: %s", err)
+	}
+
+	identityProvider, err := authenticated.GetIdentityProvider(githubIdpAlias, kcr.Spec.Realm.DisplayName)
+	if err != nil {
+		return fmt.Errorf("Unable to get Identity Provider from Keycloak API: %s", err)
+	}
+
+	// checking if GH credentials have changed
+	if githubClientID == identityProvider.Config["clientId"] && githubClientSecret == identityProvider.Config["secret"] {
```
matskiv: `githubClientSecret == identityProvider.Config["secret"]` will never be true because Keycloak API always returns `********` for sensitive values, such as this secret. Screenshot with evidence below :)
So I think this condition can be simply removed.

Screenshot 2020-01-30 at 14 27 06

matskiv:
I verified that Secret update has been reflected in Keycloak.
Please rebase and address my last two comments, and then I think we can merge this :) 👍

pkg/products/rhsso/reconciler.go
```diff
@@ -106,23 +108,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 	ns := r.Config.GetNamespace()
 
 	phase, err := r.ReconcileFinalizer(ctx, serverClient, installation, string(r.Config.GetProductName()), func() (integreatlyv1alpha1.StatusPhase, error) {
-		// Check if namespace is still present before trying to delete it resources
-		_, err := resources.GetNS(ctx, ns, serverClient)
-		if !k8serr.IsNotFound(err) {
```
matskiv: Un-installation gets stuck here.
I think this check for the existence of the namespace should be here. What was the motivation for removal?

pkg/products/rhsso/reconciler.go
jjaferson: I didn't know that, well spotted. Thanks

pkg/products/rhsso/reconciler.go
jjaferson: this wasn't removed, perhaps it was some merging conflict

PR #329, opened 2020-01-28
Intly 3413 - Add dedicated admin to user-sso

briangallagher:
### What:
Users added to the dedicated-admin group in OpenShift should be created as admins in the master realm of the usersso Keycloak.
Users deleted from the group should be removed from the master realm.
Users should be able to log in to Keycloak with their OpenShift credentials via openshift v4 Oauth client. 

### Verification Steps
- Install the Operator
- Verify that usersso is installed with a master Realm and default admin user
- run CUSTOM_PWD=Password1 ./scripts/set-htpass-idp.sh
- This will create 2 users, customer-admin and test-user. 
- customer-admin is added to the dedicated-admins group. test-user is not
- Log into OpenShift as customer-admin and test-user. This will create a user and identity record for each. Verify with oc get users and oc get identities
- Log into Keycloak as the admin user.
- Verify that customer-admin exists in the master realm and test-user does not (as it is not in the dedicated-admins group)
- Log out of Keycloak and log back in as customer-admin, using the OpenShift-v4 IDP option
- At this point you will be requested to link the customer-admin accounts. Retrieve the customer-admin password from the Keycloak-User YAML to do so. (There is an [open JIRA to remove this linking](https://issues.redhat.com/browse/INTLY-3957))

**Removing User**
- Remove customer-admin from dedicated-admins group
- oc adm groups remove-users dedicated-admins customer-admin
- Log back into Keycloak as the admin user and verify that the customer-admin has been removed from the master Realm


pkg/products/rhsso/reconciler.go
```diff
@@ -742,18 +747,26 @@ func syncronizeWithOpenshiftUsers(keycloakUsers []keycloak.KeycloakAPIUser, ctx
...
+func isOpenshiftAdmin(kcUser keycloak.KeycloakAPIUser, adminGroup *usersv1.Group) bool {
+	for _, name := range adminGroup.Users {
+		if kcUser.UserName == name {
+			return true
+		}
+	}
+
+	return false
+}
+
```
maleck13: @philbrookes are we currently adding dedicated-admins to the cluster sso? I don't think that is needed right?

pkg/products/rhssouser/reconciler.go
```diff
@@ -225,9 +245,334 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
...
+			exists, err := identityProviderExists(kcClient)
+			if err != nil {
+				return integreatlyv1alpha1.PhaseFailed, errors.Wrap(err, "Error attempting to get existing idp on user sso, master realm")
+			} else if !exists {
+				err = kcClient.CreateIdentityProvider(masterKcr.Spec.Realm.IdentityProviders[0], masterKcr.Spec.Realm.Realm)
+				if err != nil {
+					return integreatlyv1alpha1.PhaseFailed, errors.Wrap(err, "Error creating idp on master realm, user sso")
+				}
+			} else {
+				r.logger.Infof("IDP %s, already exists on master realm", idpAlias)
```
maleck13: This else doesn't seem to really add anything

pkg/products/rhssouser/reconciler.go
```diff
@@ -225,9 +245,334 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
 	}
 	r.logger.Infof("The operation result for keycloakrealm %s was %s", kcr.Name, or)
 
+	// We want to update the master realm by adding an openshift-v4 idp. We can not add the idp until we know the host
+	// url - kc.Status.InternalURL
+	if kc.Status.InternalURL != "" {
```
maleck13: would it simplify things to have ``` if kc.Status.InternalURL == ""{ return } ``` I think we could get rid of the else at the end then also

pkg/products/rhssouser/reconciler.go
```diff
@@ -225,9 +245,334 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
...
+
+		kcClient, err := r.keycloakClientFactory.AuthenticatedClient(*kc)
+		if err != nil {
+			return integreatlyv1alpha1.PhaseFailed, err
+		}
+
+		// Ensure the IDP exists before trying to create via rhsso client.
+		// We have to create via rhsso client as keycloak will not accept changes to the master realm, via cr changes,
+		// after its initial creation
+		if masterKcr.Spec.Realm.IdentityProviders != nil && masterKcr.Spec.Realm.IdentityProviders[0] != nil {
```
maleck13: same here what if we switched this to be  ``` if masterKcr.Spec.Realm.IdentityProviders == nil || masterKcr.Spec.Realm.IdentityProviders[0] == nil {  return }```

grdryn:
/approve

@philbrookes I think something needs to happen for Brian so that prow tasks will run for his PRs. It seems to be a trust thing.

pkg/products/rhsso/reconciler.go
philbrookes: no, that isn't necessary, @briangallagher are we doing that here?

philbrookes: oh, this is probably a hangover from earlier changes, you can remove this admin reconciling stuff from the RHSSO reconciler @briangallagher

pkg/products/rhssouser/reconciler.go
briangallagher: ok, will do

pkg/products/rhssouser/reconciler.go
briangallagher: ok, will do

pkg/products/rhsso/reconciler.go
briangallagher: ok, will do

htpasswd
```diff
@@ -0,0 +1,3 @@
+quay-builder-admin:$apr1$nSoKDg76$I1S86T79V6SctC9tuuqlH/
+customer-admin:$apr1$eyjXVi70$0W7/FOp6Ey4CBDg2/ibuc0
+test-user:$apr1$7pXj.OFD$U7W43u6SJ6vblUO0/uwZI/
```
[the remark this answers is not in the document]
philbrookes: yep, let's do that. I'd really prefer we were testing with real IDPs added through the OSD web console now anyway.

pkg/controller/installation/installation_controller.go
```diff
@@ -94,6 +94,12 @@ func add(mgr manager.Manager, r ReconcileInstallation) error {
 		return err
 	}
 
+	// Watch for changes to groups
+	err = c.Watch(&source.Kind{Type: &usersv1.Group{}}, &EnqueueAllInstallations{})
```
philbrookes: ncie catch!

pkg/products/rhssouser/reconciler.go
```diff
@@ -406,11 +767,11 @@ func (r *Reconciler) setupOpenshiftIDP(ctx context.Context, installation *integr
 
 	oauthc := &oauthv1.OAuthClient{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: r.getOAuthClientName(),
+			Name: r.getOAuthClientName(realm),
 		},
 		Secret: clientSecret,
 		RedirectURIs: []string{
-			r.Config.GetHost() + "/auth/realms/user-sso/broker/openshift-v4/endpoint",
+			host + "/auth/realms/" + realm + "/broker/openshift-v4/endpoint",
```
matskiv: Same as my other comment about the host.
Won't this use a route with wrong certificate?

pkg/products/rhssouser/reconciler.go
```diff
@@ -122,7 +130,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+		if err != nil {
+			return integreatlyv1alpha1.PhaseFailed, err
+		}
+
+		err = resources.RemoveOauthClient(ctx, installation, serverClient, r.oauthv1Client, r.getOAuthClientName("master"))
+		if err != nil {
+			return integreatlyv1alpha1.PhaseFailed, err
+		}
+
+		err = resources.RemoveOauthClient(ctx, installation, serverClient, r.oauthv1Client, r.getOAuthClientName(keycloakRealmName))
```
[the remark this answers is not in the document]
matskiv: New params in `setupOpenshiftIDP` and `getOAuthClientName` won't be required if we are creating only a master realm.

pkg/products/rhssouser/reconciler.go
```diff
@@ -55,10 +61,12 @@ type Reconciler struct {
 	oauthv1Client oauthClient.OauthV1Interface
 	ApiUrl        string
 	*resources.Reconciler
-	recorder record.EventRecorder
+	recorder              record.EventRecorder
+	keycloakClientFactory keycloakCommon.KeycloakClientFactory
 }
 
-func NewReconciler(configManager config.ConfigReadWriter, installation *integreatlyv1alpha1.Installation, oauthv1Client oauthClient.OauthV1Interface, mpm marketplace.MarketplaceInterface, recorder record.EventRecorder, apiURL string) (*Reconciler, error) {
+func NewReconciler(configManager config.ConfigReadWriter, installation *integreatlyv1alpha1.Installation, oauthv1Client oauthClient.OauthV1Interface, mpm marketplace.MarketplaceInterface, recorder record.EventRecorder, apiUrl string, keycloakClientFactory keycloakCommon.KeycloakClientFactory) (*Reconciler, error) {
```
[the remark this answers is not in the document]
briangallagher: To allow a moq be passed in the test @matskiv

pkg/products/reconciler.go
```diff
@@ -99,6 +109,9 @@ func NewReconciler(product integreatlyv1alpha1.ProductName, rc *rest.Config, con
 		tsClient := threescale.NewThreeScaleClient(httpc, installation.Spec.RoutingSubdomain)
 
 		reconciler, err = threescale.NewReconciler(configManager, installation, appsv1, oauthv1Client, tsClient, mpm, recorder)
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
briangallagher: The linter complains with below if I remove. Do we care about lint errors?

pkg/products/reconciler.go:111:15: ineffectual assignment to `err` (ineffassign)
		reconciler, err = threescale.NewReconciler(configManager, installation, appsv1, oauthv1Client, tsClient, mpm, recorder)

pkg/controller/installation/installation_controller.go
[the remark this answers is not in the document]
briangallagher: Will add this,thanks

pkg/products/rhssouser/reconciler.go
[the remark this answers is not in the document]
briangallagher: @matskiv Are you suggesting to remove user-sso Realm completely?

pkg/products/rhsso/reconciler.go
```diff
@@ -567,7 +578,7 @@ func (r *Reconciler) setupOpenshiftIDP(ctx context.Context, installation *integr
 	clientSecret := string(clientSecretBytes)
 
 	redirectUris := []string{
-		r.Config.GetHost() + "/auth/realms/openshift/broker/openshift-v4/endpoint",
+		host + "/auth/realms/openshift/broker/openshift-v4/endpoint",
```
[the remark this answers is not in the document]
briangallagher: There is another JIRA to create the edge route: 
https://issues.redhat.com/browse/INTLY-5093

@matskiv

pkg/products/rhssouser/reconciler.go
briangallagher: Yes, temporarily use self signed certs until  https://issues.redhat.com/browse/INTLY-5093 implemented

pkg/products/rhsso/reconciler.go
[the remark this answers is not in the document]
matskiv: That JIRA is about user SSO though. New route is already there for cluster sso.

pkg/products/reconciler.go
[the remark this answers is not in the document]
matskiv: oh, maybe `err` needs to be declared before `switch`?

pkg/products/rhssouser/reconciler.go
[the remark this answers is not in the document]
matskiv: @briangallagher yes

pkg/products/rhssouser/reconciler.go
```diff
@@ -243,9 +268,342 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
...
+
+	kcClient, err := r.keycloakClientFactory.AuthenticatedClient(*kc)
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, err
+	}
+
+	// Ensure the IDP exists before trying to create via rhsso client.
+	// We have to create via rhsso client as keycloak will not accept changes to the master realm, via cr changes,
+	// after its initial creation
+	if masterKcr.Spec.Realm.IdentityProviders == nil && masterKcr.Spec.Realm.IdentityProviders[0] == nil {
```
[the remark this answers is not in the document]
briangallagher: @matskiv Yes, good catch. typo. Should have been || not &&

pkg/products/reconciler.go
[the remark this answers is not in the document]
briangallagher: The reason this is needed is because err is redeclared in the case statement. If you look at the whole function err is redeclared several times (in several case statements).
It seems err does not really need to be redeclared but the associated return types do. 
for example: oauthv1Client, err := (L98)
It's a bit strange. 

I think checking the err is required otherwise the final return in the function i.e. L123 could return nil for err when there was actually an err returned from threescale.NewReconciler(

pkg/products/rhsso/reconciler.go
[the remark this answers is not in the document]
briangallagher: ah, ok, I see.

pkg/products/reconciler.go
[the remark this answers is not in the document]
matskiv: I would think that if we declare err on higher scope (before switch), this call `oauthv1Client, err := (L98)` would use that variable instead of declaring a local one. But I am not claiming that you are wrong, it's just unexpected for me, and I am no Go expert...
Anyway, I have no problem with leaving these the error checks in. :)

pkg/products/rhssouser/reconciler.go
```diff
@@ -243,9 +268,342 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create/update keycloak realm: %w", err)
+	}
+	r.logger.Infof("The operation result for keycloakrealm %s was %s", kcr.Name, or)
+
+	return kcr, nil
+}
+
+func (r *Reconciler) createOrUpdateKeycloakAdmin(user keycloak.KeycloakAPIUser, inst *integreatlyv1alpha1.Installation, ctx context.Context, serverClient k8sclient.Client) (controllerutil.OperationResult, error) {
+	selector := &keycloak.KeycloakUser{
```
matskiv: `selector` is a bit weird name for KeycloakUser, code inside of controllerutil.CreateOrUpdate was a bit confusing because of that.

pkg/products/rhssouser/reconciler.go
```diff
@@ -243,9 +268,342 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
...
+		},
+	}
+
+	return controllerutil.CreateOrUpdate(ctx, serverClient, selector, func() error {
+		ownerutil.EnsureOwner(selector, inst)
+		selector.Spec.RealmSelector = &metav1.LabelSelector{
+			MatchLabels: getMasterLabels(),
+		}
+		selector.Labels = GetInstanceLabels()
+		selector.Spec.User = user
```
matskiv: We need to add federated identity to ensure the accounts are linked.
We do it here:
https://github.com/integr8ly/integreatly-operator/blob/5db8026f76754e6d73c828cc31f436d4ab0fb961/pkg/products/rhsso/reconciler.go#L751-L757

matskiv: Wait a second, `syncAdminUsersInMasterRealm` sets all the fields for the users, then it gets returned and passed into `createOrUpdateKeycloakAdmin`
I suspect that because of the way how `controllerutil.CreateOrUpdate` works, the fields that are not overridden in CreateOrUpdate func, won't be updated.

matskiv: Ohh, never mind, now I see it - `selector.Spec.User = user`
Please mark this comment thread as resolved. 
Sorry for spam everyone.

pkg/products/rhssouser/reconciler.go
```diff
@@ -243,9 +268,342 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
...
+func (r *Reconciler) createOrUpdateKeycloakAdmin(user keycloak.KeycloakAPIUser, inst *integreatlyv1alpha1.Installation, ctx context.Context, serverClient k8sclient.Client) (controllerutil.OperationResult, error) {
+	selector := &keycloak.KeycloakUser{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("generated-%v", user.UserName),
+			Namespace: r.Config.GetNamespace(),
+		},
+	}
+
+	return controllerutil.CreateOrUpdate(ctx, serverClient, selector, func() error {
+		ownerutil.EnsureOwner(selector, inst)
```
matskiv: Installation resource should not own a KeycloakUser.
Installation CR is namespaced, and resides in a different namespace then KeycloakUser.
Such owner reference is invalid as per Kubernetes design. See [garbage collection docs](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/).
Relevant piece:
```
Note: Cross-namespace owner references are disallowed by design. This means: 1) Namespace-scoped dependents can only specify owners in the same namespace, and owners that are cluster-scoped. 2) Cluster-scoped dependents can only specify cluster-scoped owners, but not namespace-scoped owners.
```

I now see that such invalid owner reference exists in a couple of places in SSO reconcilers.
We can create a follow up ticket, you don't have to fix all of them now.

pkg/products/rhssouser/reconciler.go
matskiv: That change was just merged to master btw :)

matskiv:
Un-installation gets stuck because keycloak-operator waiting for removal of master KeycloakRealm(checked in `isKeycloakResourcesDeleted` method).
Keycloak operator can't handle removal of master realm. This is from it's logs:
```
FAILED removing realm redhat-rhmi-user-sso/master
```

We will have to either remove the finalizer from this KeycloakRealm resource ourselves
or
not create this KeycloakRealm in the first place? As far as I understood, we have to add IDP and everything else via API calls anyway. So creating this KeycloakRealm CR doesn't actually do anything useful for us? @briangallagher please correct me if I am wrong about this.

matskiv:
@briangallagher I probably know why we need KeycloakRealm CR - for the `.Spec.RealmSelector` field on KeycloakUser(s). So I think we will need to remove it's finalizer on uninstall.

pkg/products/reconciler.go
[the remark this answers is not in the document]
briangallagher: I experimented a bit with this. Declaring err before the switch didn't work. 
In the below code oauthv1Client needs to be declared which forces the redeclaration of err.
At which point the original declaration is now out of scope and the new one is in scope for the given case statement. 

`oauthv1Client, err := oauthClient.NewForConfig(rc)`

I'm new to go so happy to be corrected here....

pkg/products/rhssouser/reconciler.go
briangallagher: updated to kcUser

pkg/products/rhssouser/reconciler.go
briangallagher: I'm now creating the edge route before Realms in order to pass edge host URL to Realm.

pkg/products/rhssouser/reconciler.go
matskiv: Follow up issue created here - https://issues.redhat.com/browse/INTLY-5168

deploy/role.yaml
```diff
@@ -199,16 +199,16 @@ rules:
...
       - user.openshift.io
     resources:
       - groups
-    resourceNames: 
-      - dedicated-admins
     verbs:
       - get
+      - list
+      - watch
       # - watch will probably be needed in the near future
```
matskiv: We added "watch" verb, so we can remove the comment on line 212 :)

pkg/products/rhssouser/reconciler.go
```diff
@@ -226,33 +319,241 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
...
+
+	err := serverClient.List(ctx, &allUsers, listOptions...)
+	if err != nil {
+		return err
+	}
+
+	for _, delUser := range deletedUsers {
+
+		for _, user := range allUsers.Items {
+			if user.Spec.User.UserName == delUser.UserName {
```
matskiv: Maybe we should add a condition to check realm selector as well?
I know that right now, there are KeycloakUser(s) with master realm only, but if change this in the future, this could be a nasty bug.

matskiv: Or maybe we can match them with `.ID` ?

pkg/products/rhssouser/reconciler.go
```diff
@@ -211,13 +218,105 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
...
 	r.logger.Infof("The operation result for keycloak %s was %s", kc.Name, or)
 
+	host := r.Config.GetHost()
+
+	r.logger.Infof("Found host %s", host)
+	r.logger.Infof("Found host %s", host)
+	r.logger.Infof("Found host %s", host)
+	r.logger.Infof("Found host %s", host)
+	r.logger.Infof("Found host %s", host)
+	r.logger.Infof("Found host %s", host)
```
matskiv: I guess these^ were added for debugging? :)

pkg/products/rhsso/reconciler.go
```diff
@@ -539,7 +553,7 @@ func (r *Reconciler) exportConfig(ctx context.Context, serverClient k8sclient.Cl
 	// TODO: once the keycloak operator generates a route with a valid certificate, that
 	// should be reverted back to using the InternalURL
 	// r.Config.SetHost(kc.Status.InternalURL)
-	r.Config.SetHost(r.KeycloakHost)
+	// r.Config.SetHost(r.KeycloakHost)
```
matskiv: My guess is, this was commented out temporarily? We can remove it now, right?

matskiv:
@briangallagher I did some testing and found two problems:
a) the automatic browser redirect into Openshif 4 IDP is not configured
b) we can not delete KeycloakUser CR when a user is removed from dedicated-admins group. We should remove just the permissions that are given to dedicated-admins, but the user needs to stay. There could be realms owned by this user, or the user could be given some permissions into other realms, but all that will be lost if we remove the user completely.

pkg/products/rhssouser/reconciler.go
briangallagher: yes, removed :-)

pkg/products/rhsso/reconciler.go
briangallagher: yes

briangallagher:
@matskiv 
a) Is this something we need to do as part of this PR or is a separate piece of work? Do you have any insight into what config is required to achieve this?
b) see your point here, I'll investigate how to change that.

matskiv:
@briangallagher 
a) IMHO it should be done as part of this PR to match what we have in RHMI 1.x. I can explain technical details of it in PM chat.

pkg/products/rhssouser/reconciler.go
briangallagher: We match by ID now where possible

pkg/products/rhssouser/reconciler.go
```diff
@@ -226,33 +315,364 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
...
+func (r *Reconciler) createOrUpdateKeycloakAdmin(user keycloak.KeycloakAPIUser, inst *integreatlyv1alpha1.Installation, ctx context.Context, serverClient k8sclient.Client) (controllerutil.OperationResult, error) {
+	kcUser := &keycloak.KeycloakUser{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("generated-%v", user.UserName),
+			Namespace: r.Config.GetNamespace(),
+		},
+	}
+
+	return controllerutil.CreateOrUpdate(ctx, serverClient, kcUser, func() error {
+		ownerutil.EnsureOwner(kcUser, inst)
```
matskiv: @briangallagher This would result in a invalid OwnerReference, and we already fixed the rest of them on master, so can you please remove this line?

matskiv:
Verified. Everything worked as expected.
Awesome work @briangallagher 

/approve
/lgtm

PR #331, opened 2020-01-28
INTLY-4908 exposed version number metric to prometheus for integreatl…

briangallagher:
…y operator and integreatly product

### What
Create 2 metrics for prometheus: 
1 for the operator version
1 for the product (integreatly) version

### Verify
Run the operator
`curl 127.0.0.1:8383/metrics | grep oper`
Should produce 
`integreatly_version_info{integreatly_version="2.0.0",operator_version="1.15.0"} 0`

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -50,7 +50,7 @@ var (
 var (
 	operatorVersion = prometheus.NewGauge(
 		prometheus.GaugeOpts{
-			Name:        "integreatly_version_info",
+			Name:        "version",
 			Help:        "Integreatly operator information",
 			ConstLabels: prometheus.Labels{
 				"operator_version": version.Version,
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -50,11 +50,11 @@ var (
 var (
 	operatorVersion = prometheus.NewGauge(
 		prometheus.GaugeOpts{
-			Name:        "version",
-			Help:        "Integreatly operator information",
+			Name: "integreatly_version_info",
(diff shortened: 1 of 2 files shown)


cmd/manager/main.go
```diff
@@ -54,7 +54,10 @@ var (
 		prometheus.GaugeOpts{
 			Name:        "integreatly_version_info",
 			Help:        "Integreatly operator information",
-			ConstLabels: prometheus.Labels{"operator_version": version.Version},
+			ConstLabels: prometheus.Labels{
+				"operator_version": version.Version,
+				"integreatly_version" : version.IntegreatlyVersion,
```
david-martin: Can we use "version" as the label key on this metric.
This will make it easier for querying and visualising if all operators use the same metrics.

briangallagher: Yes, will update

david-martin: @briangallagher just reopening this comment thread to ensure the key is changed to `version` instead of `integreatly_version`.
I think there might have been some confusion.

Here's an example from 3scale:
```
{
			Name: "threescale_version_info",
			Help: "3scale Operator version and product version",
			ConstLabels: prometheus.Labels{
				"operator_version": version.Version,
				"version":          product.ThreescaleRelease,
			},
```

and here's how i'd imagine it should look for integreatly:
```
{
			Name: "integreatly_version_info",
			Help: "Integreatly operator information",
			ConstLabels: prometheus.Labels{
				"operator_version": version.Version,
				"version":          version.IntegreatlyVersion,
			},
```

PR #333, opened 2020-01-30
Intly 5085

philbrookes:
INTLY-5108
INTLY-5085

### DIFF
--- a/pkg/apis/integreatly/v1alpha1/installation_types.go
+++ b/pkg/apis/integreatly/v1alpha1/installation_types.go
@@ -85,12 +85,13 @@ type InstallationSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
-	Type             string         `json:"type"`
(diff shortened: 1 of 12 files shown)


pkg/apis/integreatly/v1alpha1/installation_types.go
```diff
@@ -85,12 +85,13 @@ type InstallationSpec struct {
...
-	NamespacePrefix  string         `json:"namespacePrefix"`
-	SelfSignedCerts  bool           `json:"selfSignedCerts"`
-	PullSecret       PullSecretSpec `json:"pullSecret,omitempty"`
+	Type              string         `json:"type"`
+	RoutingSubdomain  string         `json:"routingSubdomain,omitempty"`
+	MasterURL         string         `json:"masterURL,omitempty"`
+	NamespacePrefix   string         `json:"namespacePrefix"`
+	SelfSignedCerts   bool           `json:"selfSignedCerts,omitempty"`
+	PullSecret        PullSecretSpec `json:"pullSecret,omitempty"`
+	UseClusterStorage bool           `json:"useClusterStorage,omitempty"`
```
grdryn: For this and `SelfSignedCerts` above, because they're `bool`, it seems like `false` and "not existing" are equivalent. This means that if they're false in the yaml, then they won't be printed when someone queries the server. There might be other consequences also, like if you checked for non-existence, and wanted to do something different in that case than when it was false. I believe the correct way to do _that_ would be to set the type here to `*bool` rather than `bool`.

PR #336, opened 2020-01-31
INTLY-5109 Create AWS Resource Alerts

davidkirwan:
## JIRA
JIRA: https://issues.redhat.com/browse/INTLY-5109

## Description
Adding Prometheus rules to cover the known RHMI products which use the Cloud Resource Operator to create RDS/Elasticache instances.

## Verification
1.
2.

### DIFF
--- a/pkg/products/codeready/reconciler.go
+++ b/pkg/products/codeready/reconciler.go
@@ -14,6 +14,7 @@ import (
 
 	chev1 "github.com/eclipse/che-operator/pkg/apis/org/v1"
 
+	"github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1"
 	crov1alpha1 "github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1"
 	cro1types "github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1/types"
 	croUtil "github.com/integr8ly/cloud-resource-operator/pkg/resources"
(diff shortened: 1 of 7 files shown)


davidkirwan:
@dimitraz ping

pkg/apis/integreatly/v1alpha1/installation_types.go
```diff
@@ -56,7 +56,7 @@ var (
 	VersionMonitoring          ProductVersion  = "1.0.2"
 	Version3Scale              ProductVersion  = "2.7"
 	VersionUps                 ProductVersion  = "2.3.2"
-	VersionCloudResources      ProductVersion  = "0.7.1"
+	VersionCloudResources      ProductVersion  = "0.8.0"
```
dimitraz: Should this be `0.8.1`?

manifests/integreatly-cloud-resources/cloud-resources-0.8.0/cloud-resources.v0.8.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,374 @@
+apiVersion: operators.coreos.com/v1alpha1
```
dimitraz: Unless this has changed recently, the history of CSVs should only be kept in the operator repo. So I think this version can be deleted

pkg/products/codeready/reconciler.go
```diff
@@ -628,3 +645,123 @@ func getKeycloakClientSpec(cheURL string) keycloak.KeycloakClientSpec {
...
+	ruleName := fmt.Sprintf("availability-rule-%s", instanceID)
+
+	pr := &prometheusv1.PrometheusRule{}
+	selector := client.ObjectKey{
+		Namespace: namespace,
+		Name:      ruleName,
+	}
+
+	if err := serverClient.Get(ctx, selector, pr); err != nil {
+		return errorUtil.Wrapf(err, "exception calling DeleteRDSAvailabilityAlert: %s", ruleName)
```
dimitraz: I think if we do a get on the prometheus rule and it doesn't exist (for example, it might have been deleted earlier), we shouldn't need to return an error

pkg/products/codeready/reconciler.go
```diff
@@ -196,6 +204,15 @@ func (r *Reconciler) reconcileExternalDatasources(ctx context.Context, serverCli
 		return integreatlyv1alpha1.PhaseAwaitingComponents, nil
 	}
 
+	alertNamespace := postgres.Namespace
+	alertResourceID := postgres.Name
+	alertProductName := postgres.Labels["productName"]
+	alertName := alertProductName + "RDSInstanceUnavailable"
+
+	if err = r.CreateRDSAvailabilityAlert(ctx, postgres, serverClient, alertName, alertNamespace, alertResourceID, alertProductName); err != nil {
```
dimitraz: Are we still creating these even when the `type` is workshop and not managed?

pkg/products/codeready/reconciler.go
davidkirwan: I think you're right, but should we print a log statement maybe ?

david-martin:
Closing. Work progressing in #372

PR #340, opened 2020-02-03
Enable to use self-signed certs

jjaferson:
Enables to use self-signed certs on managed installations


jjaferson:
FYI: Waiting for a response from the service now ticket https://redhat.service-now.com/help?id=rh_ticket&table=incident&sys_id=a8aea5dddbfa4090ad88c170ba96194c

philbrookes:
I think this PR should be closed now.

david-martin:
@jjaferson @philbrookes Given that production clusters will have valid certs, I don't think we should change the default value in the auto-created CR (At least not for prod builds).
For stage build, I wonder if there's a better way by passing in some value from the selectorsyncset, or detecting automatically from within the operator.

PR #342, opened 2020-02-03
add edge url for user SSO

philbrookes:
Add edge route for user-sso


matskiv:
@philbrookes e2e test failed because of a legit reason:
`integreatly_test.go:675: Expected 1 routes in redhat-rhmi-user-sso namespace. Found 2`

PR #344, opened 2020-02-04
🎨 Deduplicate custom event handlers

grdryn:
It looks like both @philbrookes and I invented the same wheel in
different ways at the same time. His got merged first, and I didn't
notice when rebasing. This change removes the unneeded wheel.

I've chosen my implementation, because it seems like that's the reason
the mapping interface exists.

----

#


grdryn:
Hmm, might need to remove the listOption in my mapper function actually :thinking:

grdryn:
> Hmm, might need to remove the listOption in my mapper function actually thinking

Done.

PR #347, opened 2020-02-05
change grant method to auto

philbrookes:


david-martin:
/lgtm

This will simplify the auth flow by removing the scope prompt on first login.
The more we can do to simplify the already complex auth architecture & flows, the better.

PR #349, opened 2020-02-06
renamed olm install make target

pawelpaszki:
## What
renamed olm install make target

## Why
because the old make target is misleading

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,8 +112,12 @@ test/e2e/prow: test/e2e
 test/e2e: cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/prepare/configmaps cluster/prepare/crd deploy/integreatly-installation-cr.yml
 	operator-sdk --verbose test local ./test/e2e --namespace="$(NAMESPACE)" --go-test-flags "-timeout=60m" --debug --image=$(INTEGREATLY_OPERATOR_IMAGE)
 
+.PHONY: install/olm
+install/olm: cluster/cleanup/olm cluster/prepare/olm cluster/prepare/configmaps cluster/deploy/integreatly-installation-cr.yml
+
 .PHONY: test/e2e/olm
-test/e2e/olm: cluster/cleanup/olm cluster/prepare/olm cluster/prepare/configmaps cluster/deploy/integreatly-installation-cr.yml
+test/e2e/olm: install/olm
+#ToDo Trigger test suite here
 
 .PHONY: cluster/deploy/integreatly-installation-cr.yml
 cluster/deploy/integreatly-installation-cr.yml: deploy/integreatly-installation-cr.yml


psturc:
@pawelpaszki it should be probably changed [here](https://github.com/openshift/release/blob/master/ci-operator/jobs/integr8ly/integreatly-operator/integr8ly-integreatly-operator-master-presubmits.yaml#L113) as well, otherwise lgtm

pawelpaszki:
@psturc created PR in the openshift/release repo as well. Thanks for spotting it

mikenairn:
Renaming the target as it is seems fine to me, but the intention with it was that it would actually run some tests, hence the original name and having it be called that in the release repo.

Are you actually using this make target in a Jenkins job currently?

pawelpaszki:
The reason for this change is the fact that current name is misleading (as it only installs via olm). I am happy to leave it the way it is, if in the future we will be running tests along with the installation

psturc:
Just my 2 cents here. If there are no tests present atm and none are planned to implement now or in near future, let's rename it, since it could confuse people.
We can always add a new target once we have the tests there, wdyt?

mikenairn:
There are no tests, but tests are definitely planned and will be implemented next sprint as part of the work to get a functional test suite in place.

I'm suggesting sometyhing more along the lines of:

```
.PHONY: install/olm
install/olm: cluster/cleanup/olm cluster/prepare/olm cluster/prepare/configmaps cluster/deploy/integreatly-installation-cr.yml

.PHONY: test/e2e/olm
test/e2e/olm install/olm:
	//ToDo Trigger test suite here
```

And leave the prow test stuff alone.

pawelpaszki:
@psturc added Mike's suggested change. I think it is ok for now.

psturc:
👍 sounds good to me

Makefile
```diff
@@ -112,8 +112,12 @@ test/e2e/prow: test/e2e
 test/e2e: cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/prepare/configmaps cluster/prepare/crd deploy/integreatly-installation-cr.yml
 	operator-sdk --verbose test local ./test/e2e --namespace="$(NAMESPACE)" --go-test-flags "-timeout=60m" --debug --image=$(INTEGREATLY_OPERATOR_IMAGE)
 
+.PHONY: install/olm
+install/olm: cluster/cleanup/olm cluster/prepare/olm cluster/prepare/configmaps cluster/deploy/integreatly-installation-cr.yml
+
 .PHONY: test/e2e/olm
-test/e2e/olm: cluster/cleanup/olm cluster/prepare/olm cluster/prepare/configmaps cluster/deploy/integreatly-installation-cr.yml
+test/e2e/olm install/olm:
+	//ToDo Trigger test suite here
```
mikenairn: Think this should be #ToDo Trigger test suite here

mikenairn:
/override ci/prow/e2e-olm

The test failed but the make target is actually fine, so not related to the changes here. Should look into the olm test failure separately.

PR #352, opened 2020-02-06
Intly 4910

philbrookes:
Things that didn't work:
- Having CRO watch UPS namespace (CRO CSV dictates singleNamespace only)
- Having CRO create secret in UPS namespace (insufficient permissions, see above failure)
- Copying the secret to the UPS project (field names are different inside the secret)

What worked:
- Making a new secret with the same values but different field names in the UPS project

Why there's all this other stuff:
Some of the changes I made seemed like sensible changes to include:
- A helper function for copying a secret around and reconciling it
- Allowing the ReconcileSubscription to supply more than one operandNS


pkg/resources/pullSecret.go
```diff
@@ -16,40 +18,38 @@ const (
 	DefaultOriginPullSecretNamespace = "openshift"
 )
 
-// Gets the default pull secret for pulling container images from registry
-func GetDefaultPullSecret(context context.Context, client k8sclient.Client, inst *integreatlyv1alpha1.Installation) (corev1.Secret, error) {
+// Copys the default pull secret to a target namespace
```
grdryn: For exported functions & variables, the Go convention is to begin the comment with the exported name, e.g.:

```go
// CopyDefaultPullSecretToNameSpace copies the default pull secret to a target namespace
```

pkg/resources/pullSecret.go
```diff
@@ -16,40 +18,38 @@ const (
...
+func CopySecret(ctx context.Context, client k8sclient.Client, srcName, srcNamespace, destName, destNamespace string) error {
+	srcSecret := corev1.Secret{}
+	err := client.Get(ctx, types.NamespacedName{Name: srcName, Namespace: srcNamespace}, &srcSecret)
 	if err != nil {
 		return err
 	}
 
-	componentSecret := &corev1.Secret{
+	destSecret := &corev1.Secret{
 		Type: corev1.SecretTypeDockerConfigJson,
```
grdryn: I know it existed before your changes, but that's a strange secret type to specify :thinking: Is it necessary to specify a type here?

pkg/resources/pullSecret_test.go
```diff
@@ -20,63 +20,6 @@ func getBuildScheme() (*runtime.Scheme, error) {
 	return scheme, err
 }
 
-func TestGetDefaultPullSecret(t *testing.T) {
```
grdryn: Useless tests or broken tests?

pkg/products/ups/reconciler.go
```diff
@@ -25,17 +29,15 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/tools/record"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	defaultInstallationNamespace = "ups"
+	DefaultInstallationNamespace = "ups"
```
grdryn: Is this now being used outside this package? Also, I'm confused as to why that's the value...we never create a namespace with that name, do we?

matskiv:
Reviewed and verified
/approve
/lgtm

matskiv:
I didn't notice comments from @grdryn and jumped the gun with lgtm. 
Good comments, so I removed LGTM.

pkg/resources/pullSecret_test.go
philbrookes: Useless, they were testing a function  that I've deleted now.

pkg/products/ups/reconciler.go
philbrookes: oh, I was using it externally for a while,  but then went back on that decision. I've reverted this change now.

pkg/resources/pullSecret.go
philbrookes: yeah, this informs the cluster that this secret can be used for pulling in images to satisfy deployments / deploymentconfigs.

pkg/resources/pullSecret.go
philbrookes: Thanks, fixed this now.

PR #353, opened 2020-02-07
fix: generate smtp secret after addon installation

psturc:
This change is based on discussion with @philbrookes yesterday.
We would need to generate a dummy smtp secret after integreatly-operator's namespace is created, so the operator is able to start with installation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,6 @@ OPERATOR_SDK_VERSION=0.12.0
 AUTH_TOKEN=$(shell curl -sH "Content-Type: application/json" -XPOST https://quay.io/cnr/api/v1/users/login -d '{"user": {"username": "$(QUAY_USERNAME)", "password": "${QUAY_PASSWORD}"}}' | jq -r '.token')
 TEMPLATE_PATH="$(shell pwd)/templates/monitoring"
 INTEGREATLY_OPERATOR_IMAGE ?= $(REG)/$(ORG)/$(PROJECT):v$(TAG)
-KUBECONFIG="${HOME}/.kube/config"
 
 export SELF_SIGNED_CERTS   ?= true
 export INSTALLATION_TYPE   ?= managed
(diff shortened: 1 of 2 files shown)


mikenairn:
@philbrookes @psturc Should the addon not do this?

psturc:
@mikenairn as mentioned in the description, this is how it works now. It won't probably change in near future.

make/ocm.mk
```diff
@@ -69,6 +69,7 @@ ocm/install/rhmi-addon:
 	@$(call get_cluster_id)
 	@echo '{"addon":{"id":"rhmi"}}' | ${OCM} post /api/clusters_mgmt/v1/clusters/${OCM_CLUSTER_ID}/addons
 	$(call wait_command, oc --config=ocm/cluster.kubeconfig get installation -n redhat-rhmi-operator | grep -q integreatly, installation CR created, 10m, 30)
+	@make KUBECONFIG="ocm/cluster.kubeconfig" cluster/prepare/smtp
```
mikenairn: Should probably use the MAKE variable here $(MAKE)

Makefile
```diff
@@ -15,6 +15,7 @@ OPERATOR_SDK_VERSION=0.12.0
 AUTH_TOKEN=$(shell curl -sH "Content-Type: application/json" -XPOST https://quay.io/cnr/api/v1/users/login -d '{"user": {"username": "$(QUAY_USERNAME)", "password": "${QUAY_PASSWORD}"}}' | jq -r '.token')
 TEMPLATE_PATH="$(shell pwd)/templates/monitoring"
 INTEGREATLY_OPERATOR_IMAGE ?= $(REG)/$(ORG)/$(PROJECT):v$(TAG)
+KUBECONFIG="${HOME}/.kube/config"
```
mikenairn: I think the prow failures might be related to this, although I'm not 100% sure on that. It's like it's targeting the wrong cluster now.

psturc: Ok, good idea, I'll try to remove it from this Makefile and run the command inside the `ocm/install/rhmi-addon` target

PR #354, opened 2020-02-07
INTLY-2354 | Add AMQ-SLI PrometheusRule

HubertStefanski:
https://issues.redhat.com/browse/INTLY-2354
### New alert on alert page relating to AMQ-SLI-1.1
image
### Alert firing in alert manager
image

### DIFF
--- a/pkg/products/amqonline/reconciler.go
+++ b/pkg/products/amqonline/reconciler.go
@@ -362,6 +362,7 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation 
 
 func (r *Reconciler) reconcilePrometheusRule(ctx context.Context, installation *integreatlyv1alpha1.Installation, client k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
 	monitoringConfig := config.NewMonitoring(config.ProductConfig{})
+	keycloakServicePortCount := "2"
 	rule := &monitoringv1.PrometheusRule{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "rhmi-amq-online-sli",
@@ -371,21 +372,30 @@ func (r *Reconciler) reconcilePrometheusRule(ctx context.Context, installation *
 
 	rules := []monitoringv1.Rule{
 		{
-			Alert: fmt.Sprintf("AMQ-SLI-1.1: AMQ Online console is not available in namespace %s", r.Config.GetNamespace()),
+			Alert: fmt.Sprintf("AMQOnlineConsoleUnavailable"),
 			Annotations: map[string]string{
 				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
-				"message": "AMQ Online Console is not available",
+				"message": fmt.Sprintf("AMQ-SLI-1.1: AMQ Online console is not available in namespace \"%s\"", r.Config.GetNamespace()),
 			},
 			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"console\",namespace=\"%s\"}==1)", r.Config.GetNamespace())),
 			For:    "60s",
 			Labels: map[string]string{"severity": "critical"},
 		}, {
-			Alert: fmt.Sprintf("AMQ-SLI-1.3: Keycloak is not available in namespace %s", r.Config.GetNamespace()),
+			Alert: fmt.Sprintf("AMQOnlineKeycloakUnavailable"),
 			Annotations: map[string]string{
 				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
-				"message": "Keycloak is not available, addresses requiring user authentication will not receive messages",
+				"message": fmt.Sprintf("AMQ-SLI-1.3: Keycloak is not available in namespace %s", r.Config.GetNamespace()),
 			},
-			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"standard-authservice\",namespace=\"%s\"}==2)", r.Config.GetNamespace())),
+			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"standard-authservice\",namespace=\"%s\"}==\"%s\")", r.Config.GetNamespace(), keycloakServicePortCount)),
+			For:    "60s",
+			Labels: map[string]string{"severity": "critical"},
+		}, {
+			Alert: fmt.Sprintf("AMQOnlineAddressSpace"),
+			Annotations: map[string]string{
+				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
+				"message": fmt.Sprintf("AMQ-SLI-1.1: AMQ Online console is not available in namespace \"%s\"", r.Config.GetNamespace()),
+			},
+			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"console\",namespace=\"%s\"}==1)", r.Config.GetNamespace())),
 			For:    "60s",
 			Labels: map[string]string{"severity": "critical"},
 		},
@@ -404,7 +414,7 @@ func (r *Reconciler) reconcilePrometheusRule(ctx context.Context, installation *
 		return nil
 	})
 	if err != nil {
-		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("Error creating enmasse reconcilePrometheusRule: %w", err)
+		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("error creating enmasse reconcilePrometheusRule: %w", err)
 	}
 
 	return integreatlyv1alpha1.PhaseCompleted, nil
--- a/pkg/products/amqonline/reconciler.go
+++ b/pkg/products/amqonline/reconciler.go
@@ -369,17 +369,27 @@ func (r *Reconciler) reconcilePrometheusRule(ctx context.Context, installation *
 		},
 	}
 
-	rules := []monitoringv1.Rule{}
-	rules = append(rules, monitoringv1.Rule{
-		Alert: fmt.Sprintf("AMQ-SLI-1.1: AMQ Online console is not available in namespace %s", r.Config.GetNamespace()),
-		Annotations: map[string]string{
-			"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
-			"message": "AMQ Online Console is not available",
+	rules := []monitoringv1.Rule{
+		{
+			Alert: fmt.Sprintf("AMQ-SLI-1.1: AMQ Online console is not available in namespace %s", r.Config.GetNamespace()),
+			Annotations: map[string]string{
+				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
+				"message": "AMQ Online Console is not available",
+			},
+			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"console\",namespace=\"%s\"}==1)", r.Config.GetNamespace())),
+			For:    "60s",
+			Labels: map[string]string{"severity": "critical"},
+		}, {
+			Alert: fmt.Sprintf("AMQ-SLI-1.3: Keycloak is not available in namespace %s", r.Config.GetNamespace()),
+			Annotations: map[string]string{
+				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
+				"message": "Keycloak is not available, addresses requiring user authentication will not receive messages",
+			},
+			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"standard-authservice\",namespace=\"%s\"}==2)", r.Config.GetNamespace())),
+			For:    "60s",
+			Labels: map[string]string{"severity": "critical"},
 		},
-		Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"console\",namespace=\"%s\"}==1)", r.Config.GetNamespace())),
-		For:    "60s",
-		Labels: map[string]string{"severity": "critical"},
-	})
+	}
 
 	_, err := controllerutil.CreateOrUpdate(ctx, client, rule, func() error {
 		rule.ObjectMeta.Labels = map[string]string{"integreatly": "yes", monitoringConfig.GetLabelSelectorKey(): monitoringConfig.GetLabelSelector()}
--- a/pkg/products/amqonline/reconciler.go
+++ b/pkg/products/amqonline/reconciler.go
@@ -216,8 +216,8 @@ func (r *Reconciler) createResource(ctx context.Context, resourceName string, se
 	return resource, nil
 }
 
-func (r *Reconciler) reconcileTemplates(ctx context.Context, installation *integreatlyv1alpha1.Installation, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
-	// Iterate over template_list
+func (r *Reconciler) reconcileTemplates(ctx context.Context, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
+	// Interate over template_list
 	for _, template := range r.Config.GetTemplateList() {
 		// create it
 		_, err := r.createResource(ctx, template, serverClient)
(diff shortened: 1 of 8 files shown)


pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,43 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
...
+	}
+
+	rules := []monitoringv1.Rule{}
+	rules = append(rules, monitoringv1.Rule{
+		Alert: fmt.Sprintf("AMQ-SLI-1.1: AMQ Online console is not available in namespace %s", r.Config.GetNamespace()),
+		Annotations: map[string]string{
+			"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
+			"message": "AMQ Online Console is not available",
+		},
+		Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"console\",namespace=\"%s\"}==1)", r.Config.GetNamespace())),
```
HubertStefanski: Is this the query we want to use? or is there something better that we could replace it with? @david-martin @grdryn

pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,43 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
...
+	rule := &monitoringv1.PrometheusRule{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rhmi-amq-online-sli",
+			Namespace: r.Config.GetNamespace(),
+		},
+	}
+
+	rules := []monitoringv1.Rule{}
+	rules = append(rules, monitoringv1.Rule{
+		Alert: fmt.Sprintf("AMQ-SLI-1.1: AMQ Online console is not available in namespace %s", r.Config.GetNamespace()),
```
HubertStefanski: Is there a naming standard for the alerts/messages that would be better suited for these fields?
In Prometheus the `alert` seems to be a short string while `message` is more descriptive but we have it the other way around here

pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,43 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
...
+		Annotations: map[string]string{
+			"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
```
HubertStefanski: Is this the correct SOP and should we add a separate link for the SLI doc?

pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,43 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
...
+			"message": "AMQ Online Console is not available",
+		},
+		Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"console\",namespace=\"%s\"}==1)", r.Config.GetNamespace())),
+		For:    "60s",
+		Labels: map[string]string{"severity": "critical"},
```
HubertStefanski: Is this the right severity level? is the interval the right length?

HubertStefanski: Severity level seems to be correct, it adheres to the logic presented in other alerts whereby the level is _critical_ if a service/container is down or unavailable and warnings are reserved for situations that aren't an immediate impact, so we should probably keep it the same for the alerts being created here aswell. As for the interval It seems to be also fine, especially when comparing it to the UPS console where the interval is much higher 
image

pkg/products/amqonline/reconciler.go
```diff
@@ -369,17 +369,27 @@ func (r *Reconciler) reconcilePrometheusRule(ctx context.Context, installation *
...
+			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"console\",namespace=\"%s\"}==1)", r.Config.GetNamespace())),
+			For:    "60s",
+			Labels: map[string]string{"severity": "critical"},
+		}, {
+			Alert: fmt.Sprintf("AMQ-SLI-1.3: Keycloak is not available in namespace %s", r.Config.GetNamespace()),
+			Annotations: map[string]string{
+				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
+				"message": "Keycloak is not available, addresses requiring user authentication will not receive messages",
+			},
+			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"standard-authservice\",namespace=\"%s\"}==2)", r.Config.GetNamespace())),
```
HubertStefanski: Here the conditional has to be set as 2, as this is the default value being returned by the query, this is because there are indeed two containers as seen below on 
image
 @grdryn  Should I maybe implement a conditional to 1 or 2 containers? I'm not sure whether the init container gets deleted or if anything happens to it after it's finished, that may cause this alert to misfire?

grdryn: @HubertStefanski An "Init Container" is just run to completion at the start, before the regular container(s) run(s). It can be used for things like ensuring pre-requisite services are available.

I think the value here is 2 instead because there are two ports exposed from that one regular container:

image

HubertStefanski: @grdryn Guess that makes more sense, Would also explain why the query picks up on 2 endpoints, So it's completely fine to leave it at two?

grdryn: Yes, but maybe declare that as a named constant, so that we know why it's 2. Something like `KeycloakServicePortCount`.

https://en.wikipedia.org/wiki/Magic_number_%28programming%29#Unnamed_numerical_constants

pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,53 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
...
+	rule := &monitoringv1.PrometheusRule{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rhmi-amq-online-sli",
+			Namespace: r.Config.GetNamespace(),
+		},
+	}
+
+	rules := []monitoringv1.Rule{
+		{
+			Alert: fmt.Sprintf("AMQ-SLI-1.1: AMQ Online console is not available in namespace %s", r.Config.GetNamespace()),
```
grdryn: @david-martin I think this was asked already, but probably on an earlier iteration of this PR: is there a naming convention for these "Alert" fields? For some reason most alerts seem to have a short string here (a couple of words camelCased into a single identifier usually), then something more descriptive in `Annotations.message`. I kinda like the idea of this top-level thing being immediately descriptive, but I don't have a strong opinion either way.

HubertStefanski: @grdryn The template from Rob suggests we should probably switch it around to a more abstract alert and keep the more descriptive part for the message, I'll Refactor to adhere to this standard   https://github.com/EnMasseProject/enmasse/blob/master/controller-manager/src/main/resources/templates/PrometheusRules-enmasse.yaml

pkg/products/amqonline/reconciler.go
HubertStefanski: Leaving this as  `kube-endpoint-address-available`, as we discussed in person yesterday

pkg/products/amqonline/reconciler.go
grdryn: Sounds good! :+1:

david-martin: @grdryn @HubertStefanski I have an issue (downstream/private https://issues.redhat.com/browse/INTLY-4624) for reviewing Alert names for consistency. Including the SLI or SLO is something to consider for that.
Given the distributed nature of where alerts are defined for RHMI, it needs some thought on how best to make updates to all alerts.
Going with the current defacto pattern makes sense for this PR.

pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,63 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
...
+	rule := &monitoringv1.PrometheusRule{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rhmi-amq-online-sli",
+			Namespace: r.Config.GetNamespace(),
+		},
+	}
+
+	rules := []monitoringv1.Rule{
+		{
+			Alert: fmt.Sprintf("AMQOnlineConsoleUnavailable"),
```
david-martin: change to *Available

pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,63 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
...
+			Annotations: map[string]string{
+				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
+				"message": fmt.Sprintf("AMQ-SLI-1.1: AMQ Online console is not available in namespace \"%s\"", r.Config.GetNamespace()),
```
david-martin: These should be based on SLO's, not SLI's.

pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,63 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
...
+			},
+			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"console\",namespace=\"%s\"}==1)", r.Config.GetNamespace())),
+			For:    "60s",
+			Labels: map[string]string{"severity": "critical"},
+		}, {
+			Alert: fmt.Sprintf("AMQOnlineKeycloakUnavailable"),
```
david-martin: change to *Available

pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,63 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
...
+			Annotations: map[string]string{
+				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
+				"message": fmt.Sprintf("AMQ-SLI-1.3: Keycloak is not available in namespace %s", r.Config.GetNamespace()),
+			},
+			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"standard-authservice\",namespace=\"%s\"}==\"%s\")", r.Config.GetNamespace(), keycloakServicePortCount)),
+			For:    "60s",
+			Labels: map[string]string{"severity": "critical"},
+		}, {
+			Alert: fmt.Sprintf("AMQOnlineAddressSpace"),
```
david-martin: This should be about the ability to create AddressSpaces & Addresses, which is based on the enmasse-operator being available.
Maybe `AMQOnlineOperatorAvailable` is a better name

HubertStefanski:
Alert firing for SLO 1.5, the message has been updated to be the correct one for the alert after this screenshot has been taken

image

pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,64 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
...
+			},
+			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"standard-authservice\",namespace=\"%s\"}==%v)", r.Config.GetNamespace(), keycloakServicePortCount)),
+			For:    "300s",
+			Labels: map[string]string{"severity": "critical"},
+		},
+		{
+			Alert: fmt.Sprintf("AMQOnlineOperatorAvailable"),
+			Annotations: map[string]string{
+				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
+				"message": fmt.Sprintf("AMQ-SLO-1.5: amq-online(enmasse) operator in namespace %s", r.Config.GetNamespace()),
```
HubertStefanski: Should be AMQ-online (enmasse) operator is unavailable in namespace ...

pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,64 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
...
+			},
+			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint=\"console\",namespace=\"%s\"}==1)", r.Config.GetNamespace())),
+			For:    "300s",
+			Labels: map[string]string{"severity": "critical"},
+		},
+		{
+			Alert: fmt.Sprintf("AMQOnlineKeycloakAvailable"),
+			Annotations: map[string]string{
+				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
+				"message": fmt.Sprintf("AMQ-SLO-1.3: Keycloak is not available in namespace %s", r.Config.GetNamespace()),
```
grdryn:
```suggestion
				"message": fmt.Sprintf("AMQ-SLO-1.4: Keycloak is not available in namespace %s", r.Config.GetNamespace()),
```

pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,64 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
 
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
+
+func (r *Reconciler) reconcilePrometheusRule(ctx context.Context, client k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
+	monitoringConfig := config.NewMonitoring(config.ProductConfig{})
+	keycloakServicePortCount := 2
+	rule := &monitoringv1.PrometheusRule{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rhmi-amq-online-sli",
```
grdryn:
```suggestion
			Name:      "rhmi-amq-online-slo",
```

^maybe

pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,64 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
...
+			For:    "300s",
+			Labels: map[string]string{"severity": "critical"},
+		},
+		{
+			Alert: fmt.Sprintf("AMQOnlineOperatorAvailable"),
+			Annotations: map[string]string{
+				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
+				"message": fmt.Sprintf("AMQ-SLO-1.5: amq-online(enmasse) operator is not available in namespace %s", r.Config.GetNamespace()),
+			},
+			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_pod_status_ready{condition='true',namespace='%s',pod=~'enmasse-operator-.*'}==1)", r.Config.GetNamespace())),
```
grdryn: I see here that the query is nicely using single quotes, which means they don't need to be escaped. Would it be worth doing that for the two other queries being added above?

pkg/products/amqonline/reconciler.go
```diff
@@ -350,3 +359,64 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation
...
+				{
+					Name:  "amqonline.rules",
+					Rules: rules,
+				},
+			},
+		}
+		return nil
+	})
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("error creating enmasse reconcilePrometheusRule: %w", err)
```
grdryn:
```suggestion
		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("error creating enmasse PrometheusRule: %w", err)
```

pkg/products/amqonline/reconciler.go
```diff
@@ -207,8 +216,8 @@ func (r *Reconciler) createResource(ctx context.Context, resourceName string, se
 	return resource, nil
 }
 
-func (r *Reconciler) reconcileTemplates(ctx context.Context, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
-	// Interate over template_list
+func (r *Reconciler) reconcileTemplates(ctx context.Context, installation *integreatlyv1alpha1.Installation, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
```
grdryn: I'd put this back for now, I think it might be what's causing the failures.

HubertStefanski: hmm, that seemed like a typo, hence the change :thinking:

HubertStefanski: Fixed, it was the `installation` in the functions signature

HubertStefanski:
@grdryn @david-martin ready for approval if you'd like to merge it in?

david-martin:
/lgtm
@HubertStefanski I'm good to merge this if it's been tested & verified as well. Nice work.

@grdryn If you're happy, feel free to approve.

PR #356, opened 2020-02-10
bump cro v0.9.0

ciaranRoche:
## About
Bumps cro to v0.9.0, deploying cloud resources into the same VPC as openshift cluster. 

## Managed Install Prerequisite
Must be run on either:
- OSD BYOC cluster
- OCP cluster
Running a managed install on a regular OSD cluster will hang as the aws resources will be blocked. 

## Workshop Install Prerequisite
It is worth noting, that a workshop install can be installed on any cluster. As it deploys all CRO resources in cluster

## Verify 
- Run an RHMI install
- Verify the version of CRO is `v0.9.0`
- Verify successful install of `managed` type, RHMI. All AWS resources (s3, RDS and Elasticache) should be able to communicate with the cluster   

## Changes for 0.7.x to 0.9.x
- Metrics work being added in https://github.com/integr8ly/integreatly-operator/pull/336

## Note
`managed` installs will deploy resources in `aws` for development or in cluster deploys set the install `type` to `workshop`


david-martin:
@ciaranRoche theres 2 minor bumps from 0.7.1 to 0.9.0. Are there other changes besides vpc changes to be aware of?
If so, can you add details or link the related CRO PRs.

Also, based on all of above changes, are there any areas in the integreatly-operator that should be given extra attention while verifying this PR?

maleck13:
I am taking a look at this locally

ciaranRoche:
@david-martin updated the PR description.

maleck13:
verfied that install created RDS instances and could see connections to them. Unfortunately I seemed to be hitting issues with self signed certs in SSO but I think we can approve this PR based on the fact the RDS instances came up without any manual intervention and I could see connections to each instance

PR #358, opened 2020-02-10
INTLY-5184 Reconcile blackbox target config based on selfSignedCerts …

david-martin:
…field in Insdtallation CR


david-martin:
Closing this in favour of solving the updating of blackbox exporter config in the application-monitoring operator. 
A future PR will pass down the value of `selfSIgnedCerts` field in the Installation CR down to the Application Monitoring Operator via a new field of the same name in the Application Monitoring CR.

cc @pb82 @austinmartinh

PR #365, opened 2020-02-12
release 1.17.0

philbrookes:
I've removed all the older CSVs, they weren't really providing any value.


maleck13:
Sure trust you on that one /lgtm

PR #366, opened 2020-02-12
🐛 Use an in-cluster Postgresql DB rather than H2

grdryn:
JIRA: https://issues.redhat.com/browse/INTLY-5292

In the old Integreatly we used a Postgresql database for the standard
auth service, but when it was added here we just used the default
config which results in a H2 database being used instead.


grdryn:
@coveralls How could coverage have increased? I call shenanigans!

david-martin:
> How could coverage have increased? I call shenanigans

@grdryn I had trouble with @coveralls as well in the past where coverage went down when i just added new unit tests. Very discouraging.
Here's the PR if your interested https://github.com/integr8ly/integreatly-operator/pull/207

david-martin:
> Coverage increased (+0.06%) to 55.695% when pulling **[8e87545](https://github.com/integr8ly/integreatly-operator/commit/8e875455593364926dcb204efce658f5bb61c836) on grdryn:INTLY-5292** into **[620f04d](https://github.com/integr8ly/integreatly-operator/commit/620f04d3393be08298b080b8a51fcdb12a0e8f92) on integr8ly:master**.

@grdryn My guess is you added new lines of code to a place that was already being covered by tests.
So if you add enough lines of code to that area (and they all get covered by some existing test path) it will increase the overall coverage %.

pkg/products/amqonline/authservices.go
```diff
@@ -14,6 +16,18 @@ func GetDefaultAuthServices(ns string) []*v1beta1.AuthenticationService {
...
+						Type:     v1beta1.PostgresqlDatasource,
+						Host:     "postgresql",
+						Port:     5432,
+						Database: "amqonline",
+						CredentialsSecret: corev1.SecretReference{
+							Name:      "postgresql",
+							Namespace: ns,
+						},
+					},
+				},
```
david-martin: @grdryn I'm guessing there's more changes coming to create the `Postgres` CR for CRO as well?

grdryn: Yeah, sorry, I meant to mark the PR as WIP. All should be done now.

pkg/products/amqonline/reconciler.go
```diff
@@ -232,17 +242,127 @@ func (r *Reconciler) reconcileTemplates(ctx context.Context, serverClient k8scli
...
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: r.Config.GetNamespace(),
+			Name:      postgresqlName,
+		},
+	}
+	_, err = controllerutil.CreateOrUpdate(ctx, serverClient, keycloakSecret, func() error {
+		// I think it would be better to set the owner here to be the
+		// AuthService, but I can't immediately figure out how to get a
+		// reference to the scheme.
+		owner.AddIntegreatlyOwnerAnnotations(keycloakSecret, r.inst)
```
grdryn: @philbrookes @maleck13 half-baked idea: if the RHMI CRD was cluster-scoped, then I think we'd be able to use proper owner-references (even own namespaces?), rather than using annotations and finalizers to managed cleanup ourselves.

pkg/products/amqonline/reconciler.go
```diff
@@ -232,17 +242,127 @@ func (r *Reconciler) reconcileTemplates(ctx context.Context, serverClient k8scli
...
+		standardAuthSvc.Spec.Standard.Datasource.Database = string(croSecret.Data["database"])
+		standardAuthSvc.Spec.Standard.Datasource.Host = string(croSecret.Data["host"])
+		standardAuthSvc.Spec.Standard.Datasource.Port, err = strconv.Atoi(string(croSecret.Data["port"]))
+		return err // either nil or failed to convert port to an int
+	})
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed reconciling standard AuthenticationService: %w", err)
+	}
+
+	// TODO: wait for auth service pod to be ready?
```
austincunningham: Is this TODO required for this pr or is it a future wish ?

grdryn: Good catch! I'll remove that. According to @robshelly, in a future version, the phase in the AuthenticationService CR will be updated when everything is ready, so it will be easier to do that then.

austincunningham: low hanging fruit :cherries: :banana: :apple:

grdryn: We probably don't need this tbh: I don't think anything in future stages depends on the auth services, so there's no need to block on their readiness. If they don't get to ready then an alert will fire and someone will have to deal with it :slightly_smiling_face:

PR #367, opened 2020-02-12
fix generated files

briangallagher:
Command operator-sdk generate openapi seems to use the file path in generated files which resulted in github.com/briangallagher references.


briangallagher:
@grdryn Do you want to review?

PR #369, opened 2020-02-12
Allow dedicated-admins group to modify amq online configs

davidffrench:
https://issues.redhat.com/browse/INTLY-4805

### What
Giving the dedicated-admins group all permissions on AMQ Online configurations in the AMQ Online managed namespace. This will give them the ability to modify infra configs, authentication services, address space plans and address plans without the need to open up a JIRA with CS-SRE.

### DIFF
--- a/pkg/products/amqonline/reconciler.go
+++ b/pkg/products/amqonline/reconciler.go
@@ -25,6 +25,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -147,7 +148,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
-	phase, err = r.reconcileBrokerConfigs(ctx, serverClient, GetDefaultBrokeredInfraConfigs(ns), GetDefaultStandardInfraConfigs(ns))
+	phase, err = r.reconcileInfraConfigs(ctx, serverClient, GetDefaultBrokeredInfraConfigs(ns), GetDefaultStandardInfraConfigs(ns))
 	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
 		events.HandleError(r.recorder, installation, phase, "Failed to reconcile broker configs", err)
 		return phase, err
@@ -162,6 +163,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 	phase, err = r.reconcileAddressSpacePlans(ctx, serverClient, GetDefaultAddressSpacePlans(ns))
 	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
 		events.HandleError(r.recorder, installation, phase, "Failed to reconcile address space plans", err)
+		return phase, err
+	}
+
+	phase, err = r.reconcileServiceAdmin(ctx, serverClient, GetServiceAdminRole(ns))
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, "Failed to reconcile service admin role to dedicated admins group", err)
 		return phase, err
 	}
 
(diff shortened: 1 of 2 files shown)


grdryn:
> This will give them the ability to modify infra configs, authentication services [...]

Are they expected to be able to modify the existing standard AuthenticationService, or just create new ones? I'm making a change in #366 that does a CreateOrUpdate on the AuthenticationService CR, effectively making it read-only

davidffrench:
> > This will give them the ability to modify infra configs, authentication services [...]
> 
> Are they expected to be able to modify the existing standard AuthenticationService, or just create new ones? I'm making a change in #366 that does a CreateOrUpdate on the AuthenticationService CR, effectively making it read-only

The expectation is to always create new resources for their production use case and not use the example configs. I find this reasonable, we would need to be explicit in our documentation so they know why this is happening.

cc @pwright

davidffrench:
@grdryn Would you have the time to verify this PR?

grdryn:
:eyes:

pkg/products/amqonline/reconciler.go
```diff
@@ -412,6 +419,40 @@ func (r *Reconciler) reconcileAddressSpacePlans(ctx context.Context, serverClien
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
 
+func (r *Reconciler) reconcileServiceAdmin(ctx context.Context, serverClient k8sclient.Client, serviceAdminRole *rbacv1.Role) (integreatlyv1alpha1.StatusPhase, error) {
+	r.logger.Info("reconciling service admin role to the dedicated admins group")
+
+	err := serverClient.Create(ctx, serviceAdminRole)
```
grdryn: I generally prefer to see `controllerutil.CreateOrUpdate` where possible. Is there some reason that's not appropriate in this case?

pkg/products/amqonline/reconciler.go
```diff
@@ -412,6 +419,40 @@ func (r *Reconciler) reconcileAddressSpacePlans(ctx context.Context, serverClien
...
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Name: "dedicated-admins",
+				Kind: "Group",
+			},
+		},
+	}
+
+	err = serverClient.Create(ctx, serviceAdminRoleBinding)
```
grdryn: Same here, `controllerutil.CreateOrUpdate` should be preferred.

pkg/products/amqonline/reconciler.go
davidffrench: I had left it as create only because other resources in the amq online reconciler were create. you are correct, this should be update also. I will update the branch now.

grdryn:
For things that the operator creates, it should in general reconcile those to the expected state (modifications by users may not persist). Should users always be creating and editing _different_ resources than the ones the integreatly-operator creates (and hopefully maintains the state of)?

pkg/products/amqonline/reconciler.go
davidffrench: Resolved now, thanks @grdryn

pkg/products/amqonline/reconciler.go
davidffrench: Resolved now, thanks @grdryn

pkg/products/amqonline/reconciler.go
```diff
@@ -412,6 +419,51 @@ func (r *Reconciler) reconcileAddressSpacePlans(ctx context.Context, serverClien
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
 
+func (r *Reconciler) reconcileServiceAdmin(ctx context.Context, serverClient k8sclient.Client, serviceAdminRole *rbacv1.Role) (integreatlyv1alpha1.StatusPhase, error) {
+	r.logger.Info("reconciling service admin role to the dedicated admins group")
+
+	_, err := controllerutil.CreateOrUpdate(ctx, serverClient, serviceAdminRole, func() error {
+		owner.AddIntegreatlyOwnerAnnotations(serviceAdminRole, r.inst)
+
+		serviceAdminRole.Rules = GetServiceAdminRole("").Rules
```
grdryn:
```suggestion
		serviceAdminRole.Rules = GetServiceAdminRole(r.Config.GetNamespace()).Rules
```

^I presume it needs a valid namespace, not sure.

Also might be easier to just say the following at this point, and get rid of the `GetServiceAdminRole` function?

```suggestion
		serviceAdminRole.Rules = []rbacv1.PolicyRule{
			{
				APIGroups: []string{"admin.enmasse.io"},
				Resources: []string{"addressplans", "addressspaceplans", "brokeredinfraconfigs", "standardinfraconfigs", "authenticationservices"},
				Verbs:     []string{"create", "get", "update", "delete", "list", "watch", "patch"},
			},
		}
```

davidffrench: I think so too, I will refactor now.

davidffrench: refactored now with tests. Amended the last commit and pushed the branch

davidffrench:
> @davidffrench I've verified that this works as expected. All I did was (as customer-admin, in the dedicated-admins group) add an annotation to a resource of each kind mentioned in the Role.
> 
> For AuthenticationService, I instead actually modified a field that I know the integreatly-operator would reconcile back, and verified that it did get reset shortly after the manual modification. I wonder if that should be the expectation (and if so, should it be documented) for all of the instances of all of these types that that integreatly-operator reconciles (I'm not sure if the integreatly-operator does a CreateOrUpdate (or equivalent) for each, but maybe the expectation from users should be that it does)?

Thank you @grdryn . In my mind, the expectation should be that users will create and update their own AMQ Online configurations.

grdryn:
/lgtm

> Thank you @grdryn . In my mind, the expectation should be that users will create and update their own AMQ Online configurations.

@davidffrench maybe an integreatly-operator admission webhook to reject updates not from the operator SA? :sweat_smile:

davidffrench:
> /lgtm
> 
> > Thank you @grdryn . In my mind, the expectation should be that users will create and update their own AMQ Online configurations.
> 
> @davidffrench maybe an integreatly-operator admission webhook to reject updates not from the operator SA? 😅

That would actually be the perfect solution :-). Unfortunately not a high priority for now.

PR #371, opened 2020-02-12
refactor e2e tests

wei-lee:
https://issues.redhat.com/browse/INTLY-5165

Change the structure of the e2e tests so that we can easily add more reusable happy path tests


Dockerfile.osd.addon
```diff
@@ -0,0 +1,18 @@
...
+COPY vendor ./vendor
+COPY test ./test
+COPY Makefile ./
+RUN make test/compile/osd
+
+FROM registry.access.redhat.com/ubi7/ubi-minimal:latest
+
+COPY --from=builder /go/src/github.com/integr8ly/integreatly-operator/integreatly-operator-test-harness.test integreatly-operator-test-harness.test
+
+ENTRYPOINT [ "/integreatly-operator-test-harness.test" ]
```
mikenairn: We should add this as an image built by opesnhift CI on each PR, will create a jira about that.

Makefile
```diff
@@ -100,6 +102,18 @@ image/build/push: image/build image/push
...
+	go mod vendor
+	docker build . -f Dockerfile.osd.addon -t $(INTEGREATLY_OPERATOR_TEST_HARNESS_IMAGE)
+
+.PHONY: image/osd/push
+image/osd/push:
+	docker push $(INTEGREATLY_OPERATOR_TEST_HARNESS_IMAGE)
+
+.PHONY: image/osd/build/push
+image/osd/build/push: image/osd/build image/osd/push
+
```
mikenairn: Could we consider adding a new make file for these targets? https://github.com/integr8ly/integreatly-operator/tree/master/make. make/osd.mk ?

test/README.md
```diff
@@ -0,0 +1,53 @@
+# Tests
+
+## Structures
+
+There are 2 suites of tests here: [`e2e`](./e2e) and [`osd`](./osd). 
+
+* [`e2e`](./e2e)
+  
+  The tests in this directory are executed using the operator-sdk's test command against an OCP environment, and will be exectued everytime when a PR is opened.
```
mikenairn: The operator-sdk e2e tests can be run against any openshift cluster, including osd

test/README.md
```diff
@@ -0,0 +1,53 @@
...
+ 
+* [`osd`](./osd)
+
+  The tests in this directory are executed from inside of a container, as part of the [OSD Addon testing flow](https://github.com/openshift/osde2e/blob/master/docs/Addons.md), against an OSD environment. 
+  Currently these tests are executed periodically (every 4 hours).
```
mikenairn: The tests in this directory will be built into a container for testing via the osde2e add on flow, but can they not be executed outside of that?

Is there anything osd specific in these tests? Would this not be considered the functional test suite rather than anything sod specific.

test/README.md
```diff
@@ -0,0 +1,53 @@
...
+  
+Even though the test suites are executed against different environments, it should be possible to reuse some of the code. That's what the [`common`](./common) directory is for.
```
mikenairn: Maybe this answers my question, but i would expect the majority of tests to be common

test/e2e/integreatly_test.go
```diff
@@ -62,9 +64,42 @@ func TestIntegreatly(t *testing.T) {
...
 	// run subtests
 	t.Run("integreatly", func(t *testing.T) {
-		t.Run("Cluster", IntegreatlyCluster)
+		for _, test := range common.ALL_TESTS {
+			t.Run(test.Description, func(t *testing.T) {
+				test.Test(t, testingContext)
+			})
+		}
+
+		t.Run("Cluster", func(t *testing.T) {
```
philbrookes: This is passing off to the old not-yet-refactored tests, is that right?

Dockerfile.osd.addon
wei-lee: @mikenairn ok, once a jira is created, I can follow up on this.

Makefile
wei-lee: sure

test/e2e/integreatly_test.go
wei-lee: Yes, that's right.

wei-lee:
@mikenairn I think I have addressed your comments. You are right that the tests aren't really tied to OSD, so I renamed the folder to `functional`, and updated the README file and other commands accordingly.  Mind taking another look?

wei-lee:
@philbrookes tests are passed now. Can you approve the PR if you are happy with it?

make/osd.mk
```diff
@@ -0,0 +1,18 @@
...
+.PHONY: image/functional/push
+image/functional/push:
+	docker push $(INTEGREATLY_OPERATOR_TEST_HARNESS_IMAGE)
+
+.PHONY: image/functional/build/push
+image/functional/build/push: image/functional/build image/functional/push
+
+.PHONY: test/compile/functional
+test/compile/functional:
+	CGO_ENABLED=0 go test -v -c -o integreatly-operator-test-harness.test ./test/functional
```
mikenairn: Just to note that we should never really have to use these outside of creating dev images that people want to test themselves.

The following Jiras should cover the ability for our CI to build and push all images created out of this repo:

* https://issues.redhat.com/browse/INTLY-5332
* https://issues.redhat.com/browse/INTLY-5335 

All master artifacts will be synced automatically in that scenario, and also can be setup for any release branches we might be creating in the future.

mikenairn: Might also make sense to call the make file something other than osd since you renamed all the targets "functional"

PR #372, opened 2020-02-13
[INTLY-5109] Create AWS Resource Alerts

dimitraz:
### Description 
Jira: https://issues.redhat.com/browse/INTLY-5109

### Verification
 - Run an install of integreatly from this branch
- Ensure `useClusterStorage` is set to false 
- Wait for all RDS and Elasticache instances to come up 
- Navigate to the prometheus console, ensure all relevant alerts are green

PR #373, opened 2020-02-13
Intly 5213 - Add AWS connectivity alerts

ciaranRoche:
Jira - https://issues.redhat.com/browse/INTLY-5213

## Verify 
- Run an installation 
- Ensure CRO is version 0.11.0 
- Ensure the following alerts are created
Screenshot from 2020-02-22 18-59-08


pkg/resources/metrics.go
```diff
@@ -0,0 +1,60 @@
...
+// of a Redis cache
+func CreateRedisConnectivityAlert(ctx context.Context, client k8sclient.Client, cr *v1alpha1.Redis) (*prometheusv1.PrometheusRule, error) {
+	productName := cr.Labels["productName"]
+	alertName := productName + "RedisCacheConnectionFailed"
+	ruleName := fmt.Sprintf("connectivity-rule-%s", cr.Name)
+	alertExp := intstr.FromString(
+		fmt.Sprintf("absent(%s{exported_namespace='%s',resourceID='%s',productName='%s'} == 1)",
+			croResources.DefaultRedisConnectionMetricName, cr.Namespace, cr.Name, productName),
+	)
+	alertDescription := fmt.Sprintf("Unable to connect to Redis instance: '%s' (strategy: %s) for the product: %s", cr.Name, cr.Status.Strategy, productName)
```
david-martin: @ciaranRoche might be worth including the namespace & resourceID in the alert message to help debugging

pkg/products/codeready/reconciler.go
```diff
@@ -201,6 +201,11 @@ func (r *Reconciler) reconcileExternalDatasources(ctx context.Context, serverCli
 		return integreatlyv1alpha1.PhaseAwaitingComponents, nil
 	}
 
+	_, err = resources.CreatePostgresConnectivityAlert(ctx, serverClient, postgres)
```
david-martin: @aidenkeating This is what's needed for the sso reconcilers once using external db.

david-martin:
I'd like to see some positive and negative testing of the alerts before this is merged.
For example:

- the alerts are green after install
- an alert fails if security groups are changed to prevent connectivity
- an alert fails (or goes yellow temporarily) if the db/cache goes offline (e.g. by triggering a db engine change) and passes when it's back
- all alerts fails if cro isn't running (I'm unsure about this behaviour, but I think it's what we'd expect and want) and passes when cro is back

pkg/products/codeready/reconciler.go
aidenkeating: @david-martin 👍 To avoid stepping on anyone's toes, shall I hold off on any merging of SSO until this gets merged. Or is it ok for whichever PR gets merged last to include the changes?

david-martin: @aidenkeating I'd keep an eye on this PR (and #372) when you're finalising the SSO changes PR. Best to work it out between each other at merge time to avoid something being left out.

obrienrobert:
> 1. The alerts are green after install

Verfied.

> 2. An alert fails if security groups are changed to prevent connectivity

Verified. Changing the outbound SG rule caused all of the alerts to trigger. Reverting the rule ensured all alerts went back to green.

> 3. An alert fails (or goes yellow temporarily) if the db/cache goes offline (e.g. by triggering a db engine change) and passes when it's back

Tested by triggering a DB engine version and DB instance class change. The corresponding alerts did trigger, but only after the upgrade had taken affect - not during the upgrade (An DB outage occurs during both changes). Once the change has completed, the alerts **didn't** revert back to green.

> 4. All alerts fails if cro isn't running (I'm unsure about this behaviour, but I think it's what we'd expect and want) and passes when cro is back

Verified.

david-martin:
/lgtm
Needs rebase.

david-martin:
/hold

> Once the change has completed, the alerts **didn't** revert back to green.
> 

I misread this. This doesn't seem right

aidenkeating:
/lgtm

Tested the uninstallation changes on my OSD cluster. Worked as expected, CRO does not revert to it's defaults and the ConfigMap stays around until the end of uninstallation.

PR #379, opened 2020-02-14
upgrades operator-sdk

jjaferson:
upgrades operator-sdk to 0.15.1


manifests/integreatly-amq-online/amq-online-1.3.1/amq-online-addressplans.crd.yaml
```diff
@@ -48,9 +48,9 @@ spec:
               type: object
               properties:
                 router:
-                  type: number
+                  type: string
```
philbrookes: Do these changes need to go upstream too, for future releases? If so, can you make an issue in the technical debt epic to make sure we don't forget.

mikenairn:
This also needs updated when you are making changes to the operator SDK version https://github.com/integr8ly/integreatly-operator/blob/master/openshift-ci/Dockerfile.tools#L3

manifests/integreatly-amq-online/amq-online-1.3.1/amq-online-addressplans.crd.yaml
jjaferson: The recommendation is to use `resource.Quantity` when type is number, although I tried that and it didn't work so I went for type string instead. I'd say we better create a ticket to look into it further.

grdryn:
@jjaferson @philbrookes Running `make -B code/gen` now (after upgrading to 0.15.1 of the operator-sdk on my path), causes new CRD yaml files to be generated into `./deploy/crds/` for all of the types that we've got duplicated in `pkg/apis` from other projects (rather than depending on them as modules). Should these CRD yamls be ignored or committed?

grdryn:
Also this breaks `make code/run`. I've got a fix for that one in this PR: https://github.com/integr8ly/integreatly-operator/pull/387

jjaferson:
> @jjaferson @philbrookes Running `make -B code/gen` now (after upgrading to 0.15.1 of the operator-sdk on my path), causes new CRD yaml files to be generated into `./deploy/crds/` for all of the types that we've got duplicated in `pkg/apis` from other projects (rather than depending on them as modules). Should these CRD yamls be ignored or committed?

I think we should ignore as we don't use, I can add them to `.gitignore` file if we all agree

grdryn:
@jjaferson I've added it to the .gitignore in #390 feel free to review :wink: 

Long term, the solution should be to remove all that duplicate stuff from those other projects in `pkg/apis/`. :slightly_smiling_face:

maleck13:
@jjaferson @grdryn lets create jiras to track the need to do this work

grdryn:
@maleck13 I think the things being discussed here above are already done now. @jjaferson if you're doing something around getting the go.mod to actually reference 0.15.x of the operator-sdk library, maybe there should be an issue for that?

jjaferson:
> @maleck13 I think the things being discussed here above are already done now. @jjaferson if you're doing something around getting the go.mod to actually reference 0.15.x of the operator-sdk library, maybe there should be an issue for that?

it's been merged already in this PR https://github.com/integr8ly/integreatly-operator/pull/400 but I can create a jira if required

PR #381, opened 2020-02-14
INTLY-4958 - Rename public facing resources using with name integreatly to rhmi

KevFan:
## Description
Rename public facing areas and resources that use integreatly to RHMI

Jira:
* https://issues.redhat.com/browse/INTLY-4958

## Verification
* e2e test should pass - should be sufficient for install test
* Alternatively - checkout this branch and run install - and check resources are now named with RHMI prefixes


KevFan:
This PR includes a new release for changes for renaming from an the OLM installtion via OperatorHub

PR #385, opened 2020-02-17
INTLY-5166 - script for setting up testing IDP

matskiv:
JIRA: https://issues.redhat.com/browse/INTLY-5166

What:
I am adding a script for creation of reliable IDP for OSD clusters.
It adds a new realm into cluster SSO, adds a few testing users into this realm and sets it up as IDP of the cluster. Some users (customer-admin and customer-admin2) are added into dedicated-admins group.
[OCM API](https://api.openshift.com/#/default/post_api_clusters_mgmt_v1_clusters__cluster_id__identity_providers) is used to add IDP for a cluster and add users into dedicated-admins group.

Prerequisites:
OC session, with cluster admin permissions, in a target cluster
OCM session

Usage example:
`PASSWORD=Password1 ./scripts/setup-sso-idp.sh`

P.S: I also removed "deploy/integreatly-installation-cr.yml" file which I accidentally committed in my previous PR.


matskiv:
I noticed a bug in keycloak-operator.
We can't use a username that's prefix of another users username - test-user is prefix of test-user1. I am going to report it to Keycloak team and change usernames of additional users.

matskiv:
Keycloak bug filed here - d80846c79dee830fc32454571c68a2383652d242
I changed usernames "customer-admin2" -> "second-admin" and "test-user2" -> "second-user".

/unhold

jhellar:
LGTM. Verified.

maleck13:
@matskiv Should this be in the cluster sso? Would it be better to have in the user sso and allow dedicated-admins to manage those users then?

matskiv:
@maleck13 I picked cluster SSO because it is almost guaranteed to be installed. I imagined a development scenario where you install just subset of products without user sso.
I didn't think of using dedicated-admins this way. But using built in admin user in cluster SSO has few small advantages: you don't need to have integreatly operator running, you can test User SSO with developer user in the same browser window.
P.S: Sorry for late reply.

PR #390, opened 2020-02-17
🔥 Remove usage of github.com/pkg/errors (again!)

grdryn:


grdryn:
/cc @pb82 @briangallagher @davidkirwan @dimitraz @aidenkeating would one or more of you like to review this?

Note the new way of wrapping errors in Go 1.13+ means that we don't need to use this library any longer. Instead of:

```go
errors.Wrap(err, "failed to do a thing")
```

we now say:

```go
fmt.Errorf("failed to do a thing: %w", err)
```

PR #392, opened 2020-02-17
INTLY-5431 - Test CRO CR successful state

KevFan:
## Description
Adds tests to check that the CRO CR resources are in a successully provisioned state

Jira:
* https://issues.redhat.com/browse/INTLY-5431

### DIFF
--- a/test/common/cro_cr_success.go
+++ b/test/common/cro_cr_success.go
@@ -1,0 +1,146 @@
+package common
+
+import (
+	goctx "context"
+	"encoding/json"
+	"fmt"
+	crov1 "github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1"
+	croTypes "github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1/types"
+	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/constants"
+	"k8s.io/apimachinery/pkg/types"
+	"testing"
+)
+
+const (
+	requestNameSpace  = "redhat-rhmi-operator"
+	requestUrl        = "/apis/integreatly.org/v1alpha1"
+	openShiftProvider = "openshift"
+	externalProvider  = "aws"
+)
+
+var (
+	postgresToCheck = []string{
+		fmt.Sprintf("%s%s", constants.CodeReadyPostgresPrefix, InstallationName),
+		fmt.Sprintf("%s%s", constants.ThreeScalePostgresPrefix, InstallationName),
+		fmt.Sprintf("%s%s", constants.RHSSOPostgresPrefix, InstallationName),
+		fmt.Sprintf("%s%s", constants.RHSSOUserProstgresPrefix, InstallationName),
+		fmt.Sprintf("%s%s", constants.UPSPostgresPrefix, InstallationName),
+		// TODO - Add check for Fuse postgres here when task for supporting external resources is done - https://issues.redhat.com/browse/INTLY-3239
+		constants.AMQAuthServicePostgres,
+	}
+	redisToCheck = []string{
+		fmt.Sprintf("%s%s", constants.ThreeScaleBackendRedisPrefix, InstallationName),
+		fmt.Sprintf("%s%s", constants.ThreeScaleSystemRedisPrefix, InstallationName),
+	}
+	blobStorageToCheck = []string{
+		fmt.Sprintf("%s%s", constants.BackupsBlobStoragePrefix, InstallationName),
+		fmt.Sprintf("%s%s", constants.ThreeScaleBlobStoragePrefix, InstallationName),
+	}
+)
+
+func TestCROPostgresSuccessfulState(t *testing.T, ctx *TestingContext) {
+	var testErrors []string
+	originalStrategy := getResourceStrategy(t, ctx)
+
+	for _, postgresName := range postgresToCheck {
+		// AMQAuthService postgres is always in cluster
+		strategy := originalStrategy
+		if postgresName == constants.AMQAuthServicePostgres {
+			strategy = openShiftProvider
+		}
+
+		postgres := &crov1.Postgres{}
+		err := getResourceAndUnMarshalJsonToResource(ctx, "postgres", postgresName, postgres, testErrors)
+
+		if err == nil && postgres.Status.Phase != croTypes.PhaseComplete && postgres.Status.Strategy != strategy {
+			testErrors = append(testErrors, fmt.Sprintf("\n%s Postgres not ready with phase: %s, message: %s, provider, %s", postgresName, postgres.Status.Phase, postgres.Status.Message, postgres.Status.Provider))
+		}
+	}
+
+	if len(testErrors) != 0 {
+		t.Fatalf("Test CRO Postgress Succesful failed with the following errors: %s", testErrors)
+	}
+}
+
+func TestCRORedisSuccessfulState(t *testing.T, ctx *TestingContext) {
+	var testErrors []string
+	strategy := getResourceStrategy(t, ctx)
+
+	for _, redisName := range redisToCheck {
+		redis := &crov1.Redis{}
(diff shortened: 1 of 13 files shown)


test/common/cro_cr_success.go
```diff
@@ -0,0 +1,128 @@
...
+		fmt.Sprintf("%s%s", constants.ThreeScalePostgresPrefix, InstallationName),
+		fmt.Sprintf("%s%s", constants.RHSSOPostgresPrefix, InstallationName),
+		fmt.Sprintf("%s%s", constants.RHSSOUserProstgresPrefix, InstallationName),
+		fmt.Sprintf("%s%s", constants.UPSPostgresPrefix, InstallationName),
+		constants.AMQAuthServicePostgres,
+	}
+
+	for _, postGresName := range postGresToCheck {
+
+		strategy := getResourceStrategy(t, ctx)
```
b1zzu: I think the `getResourceStrategy` should be outside the loop, I think you already did this for the other tests

b1zzu: if it's because of the AMQ Postgres than I would suggest to do something like this
```go
originalStrategy := getResourceStrategy(t, ctx)
for _, postGresName := range postGresToCheck {
    strategy := originalStrategy
```

test/common/cro_cr_success.go
```diff
@@ -0,0 +1,128 @@
...
+
+		postGres := &crov1.Postgres{}
+		err := json.Unmarshal(requestBody, postGres)
+
+		if err != nil {
+			t.Fatalf("Failed to unmarshall json: %s", err)
+		}
+
+		if postGres.Status.Phase != croTypes.PhaseComplete && postGres.Status.Strategy != strategy {
+			t.Fatalf("%s Postgres not ready with phase: %s, message: %s, provider, %s", postGresName, postGres.Status.Phase, postGres.Status.Message, postGres.Status.Provider)
```
b1zzu: In my opinion, it shouldn't break on the first error, but verify all crd and break/fatal at the end if one or more are incorrect

test/common/cro_cr_success.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		requestBody, _ := getCustomResourceJson(t, ctx, "postgres", postGresName)
+
+		postGres := &crov1.Postgres{}
+		err := json.Unmarshal(requestBody, postGres)
+
+		if err != nil {
+			t.Fatalf("Failed to unmarshall json: %s", err)
+		}
```
b1zzu: Wdyt about creating a function with this code? The steps in all three tests are always `getCustomResourceJson`, `json.Unmarshal`, and `if err` if I'm not wrong

test/common/cro_cr_success.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+}
+
+// Function to get a custom resource json without needing to depend on operator-sdk
+func getCustomResourceJson(t *testing.T, ctx *TestingContext, resource string, resourceName string) ([]byte, error) {
+	request := ctx.ExtensionClient.RESTClient().Get().Resource(resource).Name(resourceName).Namespace(requestNameSpace).RequestURI(requestUrl).Do()
+	requestBody, err := request.Raw()
+
+	if err != nil {
+		t.Fatalf("failed to get %s resource %s: %s", resource, resourceName, err)
```
b1zzu: Sorry, me again bothering with the style :), wouldn't be better to return the error instead of passing `t`

test/common/cro_cr_success.go
```diff
@@ -0,0 +1,128 @@
...
+
+func TestCROPostGresSuccessfulState(t *testing.T, ctx *TestingContext) {
+	postGresToCheck := []string{
+		fmt.Sprintf("%s%s", constants.CodeReadyPostgresPrefix, InstallationName),
+		fmt.Sprintf("%s%s", constants.ThreeScalePostgresPrefix, InstallationName),
+		fmt.Sprintf("%s%s", constants.RHSSOPostgresPrefix, InstallationName),
+		fmt.Sprintf("%s%s", constants.RHSSOUserProstgresPrefix, InstallationName),
+		fmt.Sprintf("%s%s", constants.UPSPostgresPrefix, InstallationName),
+		constants.AMQAuthServicePostgres,
+	}
```
b1zzu: As part of the test case, we should also check for Fuse, which is not ready now so I thinking about which should be the best way to not forget this:

- add a TODO here, and create a follow-up task linked to the fuse task
- add a TODO here, and ensure that is part of the ac for the fuse task (my preferred)
- one of the previous without TODO

wdyt?

test/common/cro_cr_success.go
```diff
@@ -0,0 +1,128 @@
...
+
+const (
+	requestNameSpace  = "redhat-rhmi-operator"
+	requestUrl        = "/apis/integreatly.org/v1alpha1"
+	openShiftProvider = "openshift"
+	externalProvider  = "aws"
+)
+
+func TestCROPostGresSuccessfulState(t *testing.T, ctx *TestingContext) {
+	postGresToCheck := []string{
```
b1zzu: We are currently using three different styles to list resources to check or loop through in a test:

A function: https://github.com/integr8ly/integreatly-operator/blob/ca9144506bb8108423a1a800c1a3670d6d8207cc/test/common/deployment_types.go#L98

A constant: https://github.com/integr8ly/integreatly-operator/blob/ca9144506bb8108423a1a800c1a3670d6d8207cc/test/common/alerts_exist.go#L51

I would personally prefer the constant, but until the same style is used in all tests for me it's fine, wdyt?

We can also solve this in a follow-up task

test/common/cro_cr_success.go
KevFan: I was thinking something similar to this but I put it in the loop as I was thinking that for this to work, we always have to ensure the AMQ Postgres must be the last postgres to check as the strategy is only fetched once and would affect the other tests if this is not the case :thinking: 

I'm good with putting it outside the loop, if you think that is alright or isn't a concern ?

test/common/cro_cr_success.go
KevFan: Yup, this sounds good to me :+1:

test/common/cro_cr_success.go
KevFan: Yup, sounds good

test/common/cro_cr_success.go
KevFan: I'll add a TODO here but yeah should be best to add it in also as part of the fuse task

test/common/cro_cr_success.go
KevFan: Yeah, I'm good with using them as constants

test/common/cro_cr_success.go
KevFan: Yup, though the only difference is that we unmarshalling to difference CRO CR resources

test/common/cro_cr_success.go
b1zzu: I wouldn't give for granted that the AMQ is always the last, therefore my idea was to put the strategy in the `orignalStrategy` so that the original never gets overwritten. But putting the `getResourceStrategy` outside the loop is just a small improvement I would do because it would reduce the number of requests, but it doesn't change much else for the test so it's up to you

test/common/cro_cr_success.go
b1zzu: Yes, right the "Resource" or interface could be passed from outside like for the unmarshal function

test/common/cro_cr_success.go
KevFan: Oh I see, yeah I mis-read that. Sounds good to me so :+1:

test/common/cro_cr_success.go
b1zzu: Requirement "added" to the task https://issues.redhat.com/browse/INTLY-3239

KevFan: Great thanks !

b1zzu:
@KevFan what does CRO stand for? :sweat_smile:

KevFan:
> @KevFan what does CRO stand for?

@b1zzu Its an acronym for Cloud Resource Operator

b1zzu:
@KevFan is there something else we should do in order to get this PR merged? I seem ready to me

jjaferson:
@b1zzu I am only verifying this PR on an OSD cluster to approve it

PR #394, opened 2020-02-18
feat: ocm template - byoc support

psturc:
https://issues.redhat.com/browse/INTLY-5451

* `make ocm/aws/create_access_key` to generate access key via aws CLI
  * expects following env vars to be exported/passed as a parameter:
  * `AWS_ACCESS_KEY_ID`,      `AWS_ACCOUNT_ID`,         `AWS_SECRET_ACCESS_KEY`
* `make BYOC=true ocm/cluster.json` to generate a cluster config with aws credentials

I'll update the readme shortly

### DIFF
--- a/make/ocm.mk
+++ b/make/ocm.mk
@@ -9,6 +9,8 @@ OCM_CLUSTER_LIFESPAN=4
 RHMI_OPERATOR_NS=redhat-rhmi-operator
 # Path to the new cluster's kubeconfig
 CLUSTER_KUBECONFIG="ocm/cluster.kubeconfig"
+# Whether to use BYOC method (specify your own AWS credentials)
+BYOC=false
 
 define get_cluster_id
 	@$(eval OCM_CLUSTER_ID=$(shell mkdir -p ocm && touch ocm/cluster-details.json && jq -r .id < ocm/cluster-details.json ))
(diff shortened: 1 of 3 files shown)


make/ocm.mk
```diff
@@ -88,9 +100,24 @@ ocm/cluster/delete:
...
+	.expiration_timestamp = "$(OCM_CLUSTER_EXPIRATION_TIMESTAMP)" | \
+	.name = "$(OCM_CLUSTER_NAME)" | \
+	.region.id = "$(OCM_CLUSTER_REGION)"' < templates/ocm-cluster/cluster-template.json > ocm/cluster.json
+	@if [ $(BYOC) = true ]; then\
+	  jq '\
+	  .byoc = true | \
+	  .aws.access_key_id = "$(ACCESS_KEY)" | \
+	  .aws.secret_access_key = "$(SECRET_KEY)" | \
+	  .aws.account_id = "$(AWS_ACCOUNT_ID)"' < ocm/cluster.json > ocm/cluster.json.tmp \
+	  && mv ocm/cluster.json.tmp ocm/cluster.json;\
```
psturc: this is ugly workaround since `jq` doesn't allow to edit the file directly.

make/ocm.mk
```diff
@@ -88,9 +100,24 @@ ocm/cluster/delete:
...
+	fi
+	@cat ocm/cluster.json
+
+.PHONY: ocm/aws/create_access_key
+ocm/aws/create_access_key:
+	@mkdir -p ocm
+	@aws iam create-access-key --user-name osdCcsAdmin | jq -r .AccessKey | tee ocm/aws.json
```
mikenairn: Can we add the aws tooling to the opesshift ci Dockerfile or create a follow on Jira for it.

make/ocm.mk
```diff
@@ -23,12 +25,22 @@ define save_cluster_credentials
...
 ifeq ($(UNAME), Linux)
 	OCM_CLUSTER_EXPIRATION_TIMESTAMP=$(shell date --date="${OCM_CLUSTER_LIFESPAN} hour" "+%FT%TZ")
 else ifeq ($(UNAME), Darwin)
 	OCM_CLUSTER_EXPIRATION_TIMESTAMP=$(shell date -v+${OCM_CLUSTER_LIFESPAN}H "+%FT%TZ")
 endif
 
+ifeq ($(BYOC), true)
+	ACCESS_KEY=$(shell mkdir -p ocm && touch ocm/aws.json && jq -r .AccessKeyId < ocm/aws.json)
+	SECRET_KEY=$(shell mkdir -p ocm && touch ocm/aws.json && jq -r .SecretAccessKey < ocm/aws.json)
+endif
```
mikenairn: This looks exactly the same as what's in get_aws_credentials, no way we can reuse that here?

mikenairn:
Looks good, just a couple of minor comments.

make/ocm.mk
psturc: good catch, this one is actually being used, but `get_aws_credentials` should be removed. I'll fix that in next commit.

make/ocm.mk
psturc: sure, I think I can do it as a part of this PR

psturc: update: discussed and decided TBD as a part of [this jira ticket](https://issues.redhat.com/browse/INTLY-4892)

PR #404, opened 2020-02-20
Add OCP support into setup-sso-idp.sh

matskiv:
JIRA: https://issues.redhat.com/browse/INTLY-5166

I am updating the existing script to get it working on non-OSD clusters.
The script will edit "cluster" OAuth config if it can't detect cluster ID via OCM CLI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 scripts/htpasswd
 htpasswd
 .DS_Store
+router-ca.tmp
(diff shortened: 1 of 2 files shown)


matskiv:
On hold until I solve problems on a clusters with self-signed certs.

matskiv:
/unhold
Latest version will now work on OCP clusters with self-signed certificate as well.

PR #420, opened 2020-02-24
Intly 5277

philbrookes:
Addressing 2 tickets in one PR to reduce e2e time (hopefully...)
- INTLY-5277 - add solution explorer console link
- INTLY-5519 - Set default value of useClusterStorage to empty


pkg/products/solutionexplorer/reconciler.go
```diff
@@ -217,6 +238,31 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+			Name: "rhmi-solution-explorer",
+		},
+		Spec: consolev1.ConsoleLinkSpec{
+			ApplicationMenu: &consolev1.ApplicationMenuSpec{},
+		},
+	}
+
+	_, err := controllerutil.CreateOrUpdate(ctx, serverClient, cl, func() error {
+		cl.Spec.ApplicationMenu.ImageURL = "https://avatars1.githubusercontent.com/u/41683098"
+		cl.Spec.ApplicationMenu.Section = "Red Hat Managed Integrations"
```
tiffanynolan: @philbrookes Can the Solution Explorer be added to the current "Red Hat Applications" section instead of being included in a new one? This is the recommended location based on the UX guidelines for using this application menu as we only have one item and it should go with any other Red Hat Applications such as OCM and CRW.

philbrookes: Thanks @tiffanynolan I've made that change now

tiffanynolan: Great! Thanks, @philbrookes

pkg/products/solutionexplorer/reconciler.go
```diff
@@ -184,6 +199,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		r.ConfigManager.WriteConfig(r.Config)
 	}
 
+	phase, err = r.reconcileConsoleLink(ctx, serverClient)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, "Failed to reconcile oauth client", err)
```
JameelB: Think the log here should be `Failed to reconcile console link`?

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/operator-framework/operator-lifecycle-manager v0.0.0-20191115003340-16619cd27fa5
 	github.com/operator-framework/operator-marketplace v0.0.0-20191105191618-530c85d41ce7
 	github.com/operator-framework/operator-sdk v0.15.1
+	github.com/pkg/errors v0.8.1
```
KevFan: There's been an effort by Ger to remove the usage of `pkg.Errors` and to use instead the `fmt.Errors` package instead, so would be good to stick to one for consistency 
https://github.com/integr8ly/integreatly-operator/pull/390/files

pkg/products/solutionexplorer/reconciler.go
```diff
@@ -217,6 +238,31 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+	_, err := controllerutil.CreateOrUpdate(ctx, serverClient, cl, func() error {
+		cl.Spec.ApplicationMenu.ImageURL = "https://avatars1.githubusercontent.com/u/41683098"
+		cl.Spec.ApplicationMenu.Section = "Red Hat Applications"
+		cl.Spec.Href = r.Config.GetHost()
+		cl.Spec.Location = consolev1.ApplicationMenu
+		cl.Spec.Text = "Solution Explorer"
+		return nil
+	})
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, errors.Wrap(err, "error creating or updating solution explorer console link")
```
KevFan: Same, just would be nice to use `fmt.Errorf` instead of the errors package

go.mod
philbrookes: thanks Kevin, I'll remove this

pkg/products/solutionexplorer/reconciler.go
philbrookes: thanks, removed

KevFan:
Looks like recent commit merged to master added new alerts using the useClusterStorage boolean before it was changed to string breaking the prow tests
https://github.com/integr8ly/integreatly-operator/blob/4772f68049723e1a2db48a27dd3fb7decd17ee5b/pkg/resources/metrics.go#L49

https://github.com/integr8ly/integreatly-operator/blob/4772f68049723e1a2db48a27dd3fb7decd17ee5b/pkg/resources/metrics.go#L106

PR #423, opened 2020-02-24
INTLY-5500 - Get email from identity when creating keycloak user

KevFan:
## Description
Currently `@example.com` is appended to the username and used as the email as the user created in the openshift realm. Instead we should try to fetch the email from the associated `Identity` resource, from the Openshift user and use this value when creating the `KeyCloakUser`. 

If for some reason no email is found in the associated `Identity`, an `Update Profile` action is added to the created `KeyCloakUser` forcing them to add an email when first try to login to CodeReady or 3Scale Products

Jira:
* https://issues.redhat.com/browse/INTLY-5500
## Verification
* Install Integreatly onto cluster from this branch
* Run the IDP Script to create sample users 
```
```
* Login into RHSSO as admin (credentials can be found at `credential-rhsso` secret)
* Remove the email of the `customer-admin01` user from the`Testing-IDP realm` and save
* Open a private browser window
  * Log into the openshift console for `customer-admin01` using `testing-idp`
      * Wait for a minute to allow the reconcile to create the KeycloakUser in the `openshift` realm
      * Check RHSSO openshift realm, the user does not have an email
  * Using `customer-admin01`, try log into Codeready or 3Scale
    * Verify the user is asked to update profile with email, first name and last name
    * Update profile and check RHSSO openshift realm, the user details have been updated
* Close the private window and reopen a new one
  *  Log into the openshift console for `test-user01` using `testing-idp`
      * Wait for a minute to allow the reconcile to create the KeycloakUser in the `openshift` realm
      *  Check RHSSO openshift realm, the user has the expected email address 
  * Try logging into codeready or 3scale using `test-user01`
  * Verify user is logged in without being asked to update profile

### DIFF
--- a/pkg/products/rhsso/userHelper.go
+++ b/pkg/products/rhsso/userHelper.go
@@ -1,0 +1,43 @@
+package rhsso
+
+import (
+	"context"
+	"fmt"
+	keycloak "github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1"
+	usersv1 "github.com/openshift/api/user/v1"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Helper for User associated functions
+
+const (
+	updateProfileAction = "UPDATE_PROFILE"
+)
+
+func GetUserEmailFromIdentity(ctx context.Context, serverClient k8sclient.Client, user usersv1.User) (string, error) {
+	email := ""
+
+	// User can have multiple identities
+	for _, identityName := range user.Identities {
+		identity := &usersv1.Identity{}
+		err := serverClient.Get(ctx, k8sclient.ObjectKey{Name: identityName}, identity)
+
+		if err != nil {
+			return "", fmt.Errorf("failed to get identity provider: %w", err)
+		}
+
+		// Get first identity with email and break loop
+		if identity.Extra["email"] != "" {
+			email = identity.Extra["email"]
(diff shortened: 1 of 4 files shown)


pkg/products/rhsso/userHelper.go
```diff
@@ -0,0 +1,43 @@
...
+	updateProfileAction = "UPDATE_PROFILE"
+)
+
+func GetUserEmailFromIdentity(ctx context.Context, serverClient k8sclient.Client, user usersv1.User, ns string) (string, error) {
+	email := ""
+
+	// User can have multiple identities
+	for _, identityName := range user.Identities {
+		identity := &usersv1.Identity{}
+		err := serverClient.Get(ctx, k8sclient.ObjectKey{Name: identityName, Namespace: ns}, identity)
```
matskiv: Identities are not namespaced actually, so we don't need the `ns` var/param in this function. 
https://github.com/openshift/api/blob/d0b31d707c464221d1eb24846b0d0bbe57040102/user/v1/types.go#L47

matskiv:
Verified, works as expected.
Bonus points for tests :) 👍

pkg/products/rhsso/userHelper.go
KevFan: Thanks, good catch ! I've removed requiring passing / using the namespace name to the associated functions

PR #427, opened 2020-02-25
INTLY-2850: Create default `rhmi-developers` group in User SSO

sergioifg94:
Add phase in reconciliation that:

1. Creates a `rhmi-developers` group
2. Makes it a default group
3. Maps the `query-realms` role to the group

This allows the developer users to view the realm by default instead
of getting a `Forbidden` message on first login.

This operation is idempotent.

### DIFF
--- a/pkg/products/rhssouser/reconciler.go
+++ b/pkg/products/rhssouser/reconciler.go
@@ -47,8 +47,14 @@ var (
 )
 
 const (
-	masterRealmLabelKey   = "sso"
-	masterRealmLabelValue = "master"
+	masterRealmLabelKey         = "sso"
+	masterRealmLabelValue       = "master"
+	developersGroupName         = "rhmi-developers"
+	viewRealmRoleName           = "view-realm"
+	createRealmRoleName         = "create-realm"
+	masterRealmClientName       = "master-realm"
+	firstBrokerLoginFlowAlias   = "first broker login"
+	reviewProfileExecutionAlias = "review profile config"
 )
 
 type Reconciler struct {
@@ -1022,13 +1028,8 @@ func (r *Reconciler) reconcileBrowserAuthFlow(ctx context.Context, kc *keycloak.
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
 
-const (
-	developersGroupName     string = "rhmi-developers"
-	developersGroupRoleName        = "query-realms"
-	masterRealmClientName          = "master-realm"
-)
-
-// Create a default group called `rhmi-developers` with the "query-realms" role
+// Create a default group called `rhmi-developers` with the "view-realm" client role and
+// the "create-realm" realm role
 func (r *Reconciler) reconcileDevelopersGroup(kc *keycloak.Keycloak) (integreatlyv1alpha1.StatusPhase, error) {
 	// Get Keycloak client
 	kcClient, err := r.keycloakClientFactory.AuthenticatedClient(*kc)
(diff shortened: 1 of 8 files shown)


philbrookes:
looks like you need to run go mod tidy and go mod vendor locally and commit the changes.

pb82:
@sergioifg94 I think we said those users should also get `create-realm` permissions.

pkg/products/rhssouser/reconciler.go
```diff
@@ -1021,3 +1026,100 @@ func (r *Reconciler) reconcileBrowserAuthFlow(ctx context.Context, kc *keycloak.
...
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, err
+	}
+	// If the role is found, complete the phase and return
+	if existingRole != nil {
+		logrus.Infof("Role %s already mapped to group %s", developersGroupRoleName, developersGroupName)
+		return integreatlyv1alpha1.PhaseCompleted, nil
+	}
+
+	// Get the view-realm role from the list of available roles
```
matskiv: Comment says "view-realm", but we are actually using "query-realms".

matskiv:
@sergioifg94 Looks good overall. I added just two comments about minor things.
+1 to what @pb82 said :)

sergioifg94:
@matskiv @pb82 Added a new commit with the changes to map the `create-realm` role to the developers group

pb82:
thanks @sergioifg94 , i will get to this today

sergioifg94:
@pb82 Thanks. Talking to @matskiv we realized that instead of `query-realms` the correct role should be `view-realms`, so I changed it and also committed that optional part of the ticket to disable the "review profile" step when the user first logs in

matskiv:
I am running latest commit, and will test it in a minute.

pkg/products/rhssouser/reconciler.go
```diff
@@ -47,8 +47,14 @@ var (
 )
 
 const (
-	masterRealmLabelKey   = "sso"
-	masterRealmLabelValue = "master"
+	masterRealmLabelKey         = "sso"
+	masterRealmLabelValue       = "master"
+	developersGroupName         = "rhmi-developers"
+	viewRealmsRoleName          = "view-realms"
```
matskiv: Role name is actually `view-realm`
```
product installation errors : failed installation of rhssouser: failed to
    reconcile rhmi-developers group: view-realms role not found as available
```

sergioifg94: Oops I'll update it in a moment, thanks for pointing it out :grimacing:

pkg/products/rhssouser/reconciler.go
```diff
@@ -1016,3 +1032,188 @@ func (r *Reconciler) reconcileBrowserAuthFlow(ctx context.Context, kc *keycloak.
...
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, err
+	}
+
+	// If the execution is not found, nothing needs to be done
+	if authenticationExecution == nil {
+		return integreatlyv1alpha1.PhaseCompleted, nil
+	}
+	// If the execution is already disabled, nothing needs to be done
+	if strings.ToLower(authenticationExecution.Requirement) == "DISABLED" {
```
matskiv: @sergioifg94 we are changing string to lower case and comparing with upper case one, that will never match.

sergioifg94: Meant to call `ToUpper` there, I'll fix it now

PR #419, opened 2020-02-24
fix: jq compile error on linux OS

psturc:
For some reason, Linux slaves do not like line endings with slash in `jq` commands:
```
12:20:05  + make BYOC=true OCM_CLUSTER_LIFESPAN=5 ocm/cluster.json
12:20:05  jq: error: syntax error, unexpected INVALID_CHARACTER, expecting $end (Unix shell quoting issues?) at <top-level>, line 1:
12:20:05  \
12:20:05  jq: 1 compile error
12:20:05  make: *** [ocm/cluster.json] Error 3
```

Removing the slashes fixes it. Verified on Jenkins slave.

PR #432, opened 2020-02-25
added make ocm/cleanup

jhellar:
## Description

JIRA: https://issues.redhat.com/browse/INTLY-5518

Make command that checks if there are any soon to be retired BYOC clusters (2 hours) and if so, it deletes RHMI CR from them, so that AWS resources are released.

This command should be run periodically by Jenkins pipeline to make sure, there are no leftovers in AWS.

### DIFF
--- a/make/ocm.mk
+++ b/make/ocm.mk
@@ -125,6 +125,7 @@ ocm/aws/create_access_key:
 # Cleanup AWS resources for BYOC clusters that are to be retired soon
 .PHONY: ocm/cleanup
 ocm/cleanup:
+	@mkdir -p ocm
 	@$(call get_byoc_clusters_to_be_retired_soon)
 	@for cluster_id in $(CLUSTER_IDS); do \
 		echo "Removing RHMI CR from cluster: $$($(OCM) get /api/clusters_mgmt/v1/clusters/$$cluster_id | jq --raw-output '.name')" && \
--- a/make/ocm.mk
+++ b/make/ocm.mk
@@ -23,6 +23,19 @@ define save_cluster_credentials
 	@$(OCM) get /api/clusters_mgmt/v1/clusters/${OCM_CLUSTER_ID}/credentials | jq -r .admin | tee ocm/cluster-credentials.json
 endef
 
+# Get BYOC clusters that are to be retired in next 2 hours
+define get_byoc_clusters_to_be_retired_soon
+	@$(eval CLUSTER_IDS=$(shell $(OCM) get /api/clusters_mgmt/v1/clusters | \
+		jq --raw-output '\
+			.items[] | \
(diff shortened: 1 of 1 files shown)


make/ocm.mk
```diff
@@ -23,6 +23,19 @@ define save_cluster_credentials
 	@$(OCM) get /api/clusters_mgmt/v1/clusters/${OCM_CLUSTER_ID}/credentials | jq -r .admin | tee ocm/cluster-credentials.json
 endef
 
+# Get BYOC clusters that are to be retired in next 2 hours
+define get_byoc_clusters_to_be_retired_soon
+	@$(eval CLUSTER_IDS=$(shell $(OCM) get /api/clusters_mgmt/v1/clusters | \
```
psturc: @jhellar did you tested this only on Mac? I was hitting an issue on Jenkins slaves (Linux) when running the jq command line-separated with backslash line endings: https://github.com/integr8ly/integreatly-operator/pull/419#issue-378944831

jhellar: I've tested that mostly in container we use for our pipelines and hit no issues with this. By the Jenkins slaves did you meant the new container slaves in PSI?

psturc: Yep. I would maybe also check the PSI Openstack slave, since this is what master installation job uses. If it works, we should be fine.

psturc:
@jhellar looks good. Just couple of questions.
1. How frequently this job is going to be running? Should we talk about this job on some sync meeting with Eng so everyone is aware of this job?
2. There are nightly jenkins jobs for creating BYOC clusters with deletion timestamp for 5 hours. How could we ensure that this job is not going to interrupt any pipeline job?

make/ocm.mk
jhellar: Checked in both openstack [1] and container slaves [2].

[1] https://master-jenkins-csb-intly.cloud.paas.psi.redhat.com/job/Playground/job/jhellarByocCleanup/2/console
[2] https://master-jenkins-csb-intly.cloud.paas.psi.redhat.com/job/Playground/job/jhellarByocCleanup/1/console

jhellar:
@psturc I was thinking something like every 30min? In worst case there would be 1h30min to remove the aws resources. Do you know how long it usually takes to successfully delete RHMI CR (with aws resources)?

I think that nightly pipelines should be fine, can not think of situation where it would cause issues.

psturc:
The logic for cleanup of aws resources will be done in the scope of [this ticket](https://issues.redhat.com/browse/INTLY-5006) and we can use that tool in the pipeline that will run periodically instead. So I think we can close this PR, wdyt @jhellar

jhellar:
@psturc makes sense

PR #437, opened 2020-02-25
update release process in readme

philbrookes:


README.md
```diff
@@ -211,17 +211,16 @@ Run `make ocm/install/rhmi-addon` to trigger the installation. Once the installa
...
-
-Commit changes and open pull request. When the PR is accepted, create a new release tag:
-
-```sh
-git tag v<version> && git push upstream v<version>
-```
+- Create `release-<release-number>` branch
+- Update `tag` and `previoustag` in makefile
+- Run `make gen/csv`
+- - Perform any manual tidying up of the generated CSV as required.
```
maleck13: What manual tidying up is needed?

maleck13: What manual tidying up is needed?

philbrookes: well, it depends, I suppose this could be phrased as "sanity check the result" but there is quite often manual changes required, though they are different each time.

philbrookes: as above, it's not predictable, some examples are:
- set USE_CLUSTER_STORAGE to false (empty string on next release)
- set a ALERTING_EMAIL

maleck13: this would block automation. Do we have some examples? Should they be bugs in the operator sdk

KevFan: These 2 examples can be just set/added in `operator.yaml` I think and would be used in the generated csv
https://github.com/integr8ly/integreatly-operator/blob/153cb8bc3f0a8cfe82a346622bb974e8e85fb6c1/deploy/operator.yaml#L34

The csv generation from the last release was missing the `customresourcedefinitions.displayName` which was manually added (no longer brought over from the previous csv due to operator-sdk upgrade) but it looks like this can also be set using csv annotions
https://github.com/operator-framework/operator-sdk/blob/master/doc/user/olm-catalog/csv-annotations.md#examples

maleck13:
going to merge this so it can be seen

PR #438, opened 2020-02-25
 [INTLY-5214] Check all alerts as part of e2e & functional tests

dimitraz:
### Description 
Check all alerts exist as part of e2e and functional tests 
- Alerts we expected that weren't found 
- Alerts that were found and not expected 

### Verification
- Log in to an existing Openshift 4 cluster 
- Install integreatly from master
- Check out this branch and run `make test/functional`. The test is common to both e2e and functional tests, so running the functional ones is just quicker 
- See that the output of the test is as expected (e.g. try comment out some of the expected alerts, verify that the report output is correct)

### DIFF
--- a/test/common/alerts_exist.go
+++ b/test/common/alerts_exist.go
@@ -66,6 +66,12 @@ var expectedRules = []alertsTestRule{
 		},
 	},
 	{
+		File: "redhat-rhmi-codeready-workspaces-backupjobs-exist-alerts.yaml",
+		Rules: []string{
+			"CronJobExists_redhat-rhmi-codeready-workspaces_codeready-pv-backup",
+		},
+	},
+	{
 		File: "redhat-rhmi-rhsso-keycloak.yaml",
 		Rules: []string{
 			"KeycloakJavaHeapThresholdExceeded",
@@ -117,8 +123,7 @@ var expectedRules = []alertsTestRule{
 	{
 		File: "redhat-rhmi-apicurito-ksm-apicurito-alerts.yaml",
 		Rules: []string{
-			//"ApicuritoPodCount",
-			"test",
+			"ApicuritoPodCount",
 		},
 	},
 	{
@@ -142,6 +147,28 @@ var expectedRules = []alertsTestRule{
 		},
 	},
 	{
+		File: "redhat-rhmi-codeready-workspaces-ksm-codeready-alerts.yaml",
+		Rules: []string{
+			"CodeReadyPodCount",
+		},
+	},
+	{
+		File: "redhat-rhmi-3scale-ksm-3scale-alerts.yaml",
+		Rules: []string{
+			"ThreeScaleApicastStagingPod",
+			"ThreeScaleApicastProductionPod",
+			"ThreeScaleBackendWorkerPod",
+			"ThreeScaleBackendListenerPod",
+			"ThreeScaleSystemAppPod",
+			"ThreeScaleAdminUIBBT",
+			"ThreeScaleDeveloperUIBBT",
+			"ThreeScaleSystemAdminUIBBT",
+			"ThreeScalePodCount",
+			"ThreeScalePodHighMemory",
+			"ThreeScalePodHighCPU",
+		},
+	},
+	{
 		File: "redhat-rhmi-middleware-monitoring-operator-prometheus-application-monitoring-rules.yaml",
 		Rules: []string{
 			"DeadMansSwitch",
@@ -153,6 +180,12 @@ var expectedRules = []alertsTestRule{
 			"AMQOnlineConsoleAvailable",
 			"AMQOnlineKeycloakAvailable",
 			"AMQOnlineOperatorAvailable",
+		},
+	},
+	{
+		File: "redhat-rhmi-solution-explorer-ksm-solution-explorer-alerts.yaml",
+		Rules: []string{
+			"SolutionExplorerPodCount",
 		},
 	},
 	{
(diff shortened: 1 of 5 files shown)


test/common/alerts_exist.go
```diff
@@ -0,0 +1,281 @@
...
+		"prometheus", ctx)
+	if err != nil {
+		t.Fatal("failed to exec to pod:", err)
+	}
+
+	// get all found rules from the prometheus api
+	var promApiCallOutput apiResponse
+	err = json.Unmarshal([]byte(output), &promApiCallOutput)
+	if err != nil {
+		t.Logf("Failed to unmarshall json: %s", err)
```
aidenkeating: Will this allow the test to go past L183 if it cannot unmarshal the response, is this what we want?

test/common/alerts_exist.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	var rulesResult prometheusv1.RulesResult
+	err = json.Unmarshal([]byte(promApiCallOutput.Data), &rulesResult)
+	if err != nil {
+		t.Logf("Failed to unmarshall json: %s", err)
```
aidenkeating: Same question as above here

test/common/alerts_exist.go
```diff
@@ -0,0 +1,281 @@
...
+
+	var diff []string
+	actualRules := make(map[string][]string, 0)
+	for file, rules := range expectedRules {
+		// build a map of found rules for each file name. The difference
+		// between this actual map and the expected map will provide any missing rules
+		for _, group := range rulesResult.Groups {
+			if !strings.Contains(group.File, file) {
+				continue
+			}
```
aidenkeating: Is it expected that there'd only be one `file`-`group.File` match per group or could we get to L200 more than once per group here?

Also, I'm guessing we're using `strings#Contains` here because `group.File` isn't exactly the same as `file`?

test/common/types.go
```diff
@@ -19,3 +27,11 @@ type TestCase struct {
 	Description string
 	Test        func(t *testing.T, ctx *TestingContext)
 }
+
+type apiResponse struct {
```
aidenkeating: If this is going to be stored in the `common` package we should probably make the name of this clearer so it's clear what it's an API response from. wdyt?

test/common/alerts_exist.go
```diff
@@ -0,0 +1,281 @@
...
+	corev1 "k8s.io/api/core/v1"
+
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/remotecommand"
+)
+
+var expectedRules = map[string][]string{
+	"/redhat-rhmi-middleware-monitoring": {
```
aidenkeating: Can we sync up on the structure of these rules? I'd like to understand this structure some more, mainly out of curiosity.

test/common/alerts_exist.go
```diff
@@ -0,0 +1,281 @@
...
+	// is not being used
+	if !rhmi.Spec.UseClusterStorage {
+		for file, rules := range expectedExtRules {
+			expectedRules[file] = append(expectedRules[file], rules...)
+		}
+	}
+
+	// exec into the prometheus pod
+	output, err := execToPod("curl localhost:9090/api/v1/rules",
+		"prometheus-application-monitoring-0",
```
aidenkeating: Are we confident this Pod name should/will always be around?

test/common/types.go
dimitraz: Yes good point!

test/common/alerts_exist.go
dimitraz: Sure

test/common/alerts_exist.go
```diff
@@ -0,0 +1,281 @@
...
+		"AMQOnlinePodHighMemory",
+		"AMQOnlineConsoleAvailable",
+		"AMQOnlineKeycloakAvailable",
+		"AMQOnlineOperatorAvailable",
+		//"AddressHealth",
+		//"AddressSpaceHealth",
+		//"ComponentHealth",
+		//"PendingPods",
+		//"RestartingPods",
+		//"TerminatingPods",
```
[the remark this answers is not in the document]
dimitraz: Yep, will just need to remember to add them back in when the amq online version upgrade has been merged

dimitraz:
Hey @wei-lee, having troubles with the order of the e2e tests being run. `common` tests are currently executed first, but the test I've added relies on the cluster being set up properly. For now I've changed the order of execution but would be nice for the `crd exists` common test to fail sooner rather than at the end. 
What would your recommendation here be? 

cc @david-martin

wei-lee:
@dimitraz sorry for the delay, just saw your message. Yes, absolutely, we should change the order for now. I think we are doing too much in `IntegreatlyCluster` for now, ideally many of the tests should be moved out into the `common` directory so that they should become test cases, but it is outside the scope of your PR.

The `crd exists` is really just an example, we don't really need to keep it if it's not required. If we want it to fail sooner, one thing we can look at is to have different stages for the tests and run them at different points. E.g. we could have tests that can run before install, after operator install, after products install etc.  If we do this way, we can split `ALL_TESTS` into a few arrays and invoke them at different point.

Wdyt?

test/common/alerts_exist.go
```diff
@@ -0,0 +1,530 @@
...
+	// add external database alerts to list of expected rules if
+	// cluster storage is not being used
+	if rhmi.Spec.UseClusterStorage != "true" {
+		for _, rule := range expectedAWSRules {
+			expectedRules = append(expectedRules, rule)
+		}
+	}
+
+	// exec into the prometheus pod
+	output, err := execToPod("curl localhost:9090/api/v1/rules",
```
b1zzu: Wouldn't it be better to use `port-forward` to create a local proxy to the pod and then hit it using the golang `http` instant of executing curl inside the pod?

test/common/alerts_exist.go
```diff
@@ -0,0 +1,530 @@
...
+}
+
+func TestIntegreatlyAlertsExist(t *testing.T, ctx *TestingContext) {
+	// get the RHMI custom resource to check what storage type is being used
+	rhmi := &v1alpha1.RHMI{}
+	ns := fmt.Sprintf("%soperator", namespacePrefix)
+	err := ctx.Client.Get(goctx.TODO(), types.NamespacedName{Name: installationName, Namespace: ns}, rhmi)
+	if err != nil {
+		t.Fatal("error getting RHMI CR:", err)
+	}
```
b1zzu: This information should be a test parameter or flag (don't know what can be used in golang), and this would be a problem of many other tests so it should be solved globally

ping @wei-lee

test/common/alerts_exist.go
```diff
@@ -0,0 +1,530 @@
...
+		if v.Status == fileMissing || len(v.MissingRules) > 0 {
+			missingCount++
+		}
+		if v.Status == fileAdditional || len(v.AdditionalRules) > 0 {
+			extraCount++
+		}
+	}
+
+	if missingCount > 0 {
+		fmt.Println("\nMissing alerts were found from Prometheus. If the removal of these Alert rules was intentional, please update this test to remove them from the check. If the removal of these Alert rules was not intendended or you are not sure, please create a Jira & discuss with the monitoring team on how best to proceed")
```
b1zzu: It would be good to print which alerts are missing or are extra

test/common/alerts_exist.go
wei-lee: @b1zzu can be elaborate on what exactly do you mean here?

b1zzu: The code here is checking if the cluster is using or not AWS resources, but when we run the test pipeline we already know if the cluster we have provisioned should use or not the aws resources, therefore we should pass this information to the tests instant of the tests finding it out

test/common/alerts_exist.go
dimitraz: Yep that happens here: https://github.com/integr8ly/integreatly-operator/pull/438/files#diff-fb2239bb04ca074b4768a268fef37e9aR421-R424

test/common/alerts_exist.go
dimitraz: I can look into it, why would this be a better way of doing it rather than doing an exec?

david-martin: I would question the benefit in changing this implementation without any benefit. Using port-forward sounds more cumbersome and prone to error also

test/common/alerts_exist.go
david-martin: I think checking the CR is valid. It's the system of record for what type of storage is being used.
No need to double up on this. The test would instruct what to set to the CR to, not the other way around.

test/common/alerts_exist.go
b1zzu: Ok thanks, didn't notice it

test/common/alerts_exist.go
b1zzu: I don't have strong opinion for this, but I think that if we could reach the api directly it would simplify the code, as we would not have to deal with streams, stdout and stderr and the `execToPod` method wouldn't be necessary anymore but there is no easy way to use `port-forwarding` too

test/common/alerts_exist.go
```diff
@@ -0,0 +1,536 @@
...
+	for _, x := range a {
+		if _, found := mb[x]; !found {
+			diff = append(diff, x)
+		}
+	}
+	return diff
+}
+
+// difference returns the elements in list of alertsTestRule `a` that aren't in list `b`.
+func ruleDifference(a, b []alertsTestRule) []alertsTestRule {
```
aidenkeating: Not important as the comment covers what it does. But I think the var names here could be more descriptive.

For example `mb`, I think means `map representation of  b`.

```
//ruleDifference one-way diff that return rules in diffSource that are not in diffTarget
ruleDifference(diffTarget, diffSource []alertsTestRule) []alertsTestRule {
  //lookup map to check diffSource contains element simple
  diffSourceLookupMap := make(map[string]struct{}, len(diffSource))
  for _, rule := range diffSource {
    diffSourceLookupMap[rule.File] = struct{}{}
  }
  //store files in diffSource that are not in diffTarget
  var diff []alertsTestRule
  for _, rule := range diffTarget {
    if _, found := diffSourceLookupMap[rule.File]; !found {
      diff = append(diff, rule)
    }
  }
  return diff
}
```

I'd then also be tempted to swap source and target.

Again, not very important, just thought I'd mention it.

dimitraz: Thanks Aiden, I agree, that is **a lot** more readable! Will update now

jhellar:
Haven't yet had a chance to verify, but code LGTM. Just two things that maybe are not part of this piece of work: tests are currently not executed with `make test/functional`, also there is no test report in console.

dimitraz:
fixed, thanks @jhellar

test/common/alerts_exist.go
b1zzu: Would like to mark this ad resolved but somehow I can't, same for the other two comments

PR #440, opened 2020-02-26
Collect the functional test suite output while printing it to the console

b1zzu:
## Why

Capturing and the output of the tests and print it all only at the end make it difficult to debug a test

## How

While capturing the output into a byte Buffer we are also writing it back to the original stdout and stderr files

### DIFF
--- a/test/functional/main_test.go
+++ b/test/functional/main_test.go
@@ -3,15 +3,17 @@ package functional
 import (
 	"bytes"
 	"fmt"
-	"github.com/integr8ly/integreatly-operator/test/metadata"
-	"github.com/jstemmer/go-junit-report/formatter"
-	"github.com/jstemmer/go-junit-report/parser"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
+
+	"github.com/integr8ly/integreatly-operator/test/metadata"
+	"github.com/jstemmer/go-junit-report/formatter"
+	"github.com/jstemmer/go-junit-report/parser"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


b1zzu:
@wei-lee I would like to simplify the functional test suite and make it easier to run locally, wdyt?

test/functional/main_test.go
```diff
@@ -1,104 +1,29 @@
...
-		if err != nil {
-			fmt.Printf("error while writing the junit report file: %v", err)
-			os.Exit(1)
-		}
-
-		err = metadata.Instance.WriteToJSON(filepath.Join(testResultsDirectory, addonMetadataName))
-		if err != nil {
-			fmt.Printf("error while writing metadata: %v", err)
-			os.Exit(1)
-		}
```
wei-lee: @b1zzu This is required because we want to build the functional tests as a container and run it as part of the OSD E2E testing flow. See the `Required output directory` section of [their doc](https://docs.google.com/document/d/1sqpJ0ChJeya3QdsnIOiLDyOqCMF48OaOQkPoyDxjO48/edit#heading=h.hsz39z8b9re1).

So to support both, maybe adding another flag to this test suite so that by default it will use the current implement, but if the flag is set, it will use the new implementation?

b1zzu: I will try to read the osd testing flow doc, but having a flag could be a good solution

b1zzu:
@wei-lee I saw that some of the things that I was concern have been already fixed, like checking if the `test-run-results` directory, but I still think that printing the logs in real-time will simplify the life of who will have to write the tests, so I've found an alternative solution, wdyt?

test/functional/main_test.go
```diff
@@ -21,31 +23,83 @@ const (
...
+				break
+			}
+		}
+
+		wg.Done()
+	}
+
+	wg.Add(2)
+
+	go t(stdoutReader, originalStderr)
```
wei-lee: should the second parameter here be `originalStdout`?

b1zzu: Yes right, thanks for catching it

b1zzu:
@wei-lee it should be ready now

b1zzu:
ping @wei-lee

damienomurchu:
/lgtm

You'll need an approver from someone in https://github.com/integr8ly/integreatly-operator/blob/master/OWNERS though as you changed the Makefile though

PR #443, opened 2020-02-26
add manifest/check command to Makefile

StevenTobin:
Add manifest/check command to Makefile so we can check if any image_mirror_mapping files exist and block them from getting back to master

To verify:

1. Checkout this branch
1. run `make manifest/check`
2. run `touch manifests/integreatly-3scale/image_mirror_mapping`
3. make manifest/check
4. rm manifests/integreatly-3scale/image_mirror_mapping


laurafitzgerald:
Verified the change works as described.

Makefile
```diff
@@ -269,3 +270,14 @@ vendor/check: vendor/fix
...
+.PHONY: manifest/check/image_mirror_mapping
+manifest/check/image_mirror_mapping:
+ifneq ( ,$(findstring image_mirror_mapping,$(IMAGE_MAPPINGS)))
+		$(error found image_mirror_mapping in $(IMAGE_MAPPINGS))
+else
+		@echo "No image_mirror_mapping files found in manifests directory"
+endif
+
+.PHONY: manifest/check
+manifest/check: manifest/check/image_mirror_mapping
```
mikenairn: I was thinking it might be good to keep all related make targets in separate make files see https://github.com/integr8ly/integreatly-operator/tree/master/make. Should we consider moving this to a manifest.mk file?

This will also allow us to add an OWNERS file in there so that we can add a few more people that can approve these changes.

StevenTobin: Sure sounds good to me to keep things separated. I've pushed the change

PR #447, opened 2020-02-27
INTLY-5468 - Fix invalid characters used in keycloak user meta name

KevFan:
## Description
Jira
* https://issues.redhat.com/browse/INTLY-5468

## Verification
* Checkout this branch and install integreatly
* Run the IDP sccript
```
```
* Create a keycloak user with special characters to the testing-ipd realm
```
apiVersion: keycloak.org/v1alpha1
kind: KeycloakUser
metadata:
  name: test-user-irregular-username
  namespace: redhat-rhmi-rhsso
spec:
  realmSelector:
    matchLabels:
      sso: testing-idp
  user:
    lastName: Tester01
    enabled: true
    emailVerified: true
    clientRoles:
      account:
        - manage-account
        - view-profile
      broker:
        - read-token
    username: tester@example.com!!
    firstName: Test
    credentials:
      - type: password
        value: Password1
    email: tester@example.com!!!
```
* In a private window, sign into openshift console with this user
* Check that the generated keycloak user for the keycloal realm doesn't not contain the special characters in it's resource name
  * Should be named `generated-tester-example-com-<userId>`
### Github IDP Test (Optional)
* If you have a github username with a uppercase letter, try setting up Github IDP on cluster and sign into cluster using your github user name
* Generated keycloak user in openshift realm should be all lower case

### DIFF
--- a/pkg/products/rhsso/reconciler.go
+++ b/pkg/products/rhsso/reconciler.go
@@ -3,6 +3,7 @@ package rhsso
 import (
 	"context"
 	"fmt"
+	userHelper "github.com/integr8ly/integreatly-operator/pkg/resources/user"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -783,7 +784,7 @@ func syncronizeWithOpenshiftUsers(ctx context.Context, keycloakUsers []keycloak.
 	}
 
 	for _, osUser := range added {
-		email, err := GetUserEmailFromIdentity(ctx, serverClient, osUser)
+		email, err := userHelper.GetUserEmailFromIdentity(ctx, serverClient, osUser)
 
 		if err != nil {
 			return nil, err
@@ -802,7 +803,7 @@ func syncronizeWithOpenshiftUsers(ctx context.Context, keycloakUsers []keycloak.
 				},
 			},
 		}
-		AppendUpdateProfileActionForUserWithoutEmail(&newKeycloakUser)
+		userHelper.AppendUpdateProfileActionForUserWithoutEmail(&newKeycloakUser)
 
 		keycloakUsers = append(keycloakUsers, newKeycloakUser)
 	}
@@ -833,7 +834,7 @@ func deleteKeycloakUsers(allKcUsers []keycloak.KeycloakAPIUser, deletedUsers []k
 		// Delete the CR
 		kcUser := &keycloak.KeycloakUser{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      fmt.Sprintf("generated-%v", delUser.UserName),
+				Name:      userHelper.GetValidGeneratedUserName(delUser),
 				Namespace: ns,
 			},
 		}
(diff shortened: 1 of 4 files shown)


matskiv:
@KevFan So what will happen if there is a user with `abCd` username and another one with `ab-d` username? Or `abC` and `ab`.

KevFan:
> @KevFan So what will happen if there is a user with `abCd` username and another one with `ab-d` username? Or `abC` and `ab`.

@matskiv The generated resource name for the created `KeyCloakUser` should be:
*  `abCd` => `generated-abcd` (will be lowercased after tranformation)
*  `ab-d` => `generated-ab-d`(would remain the same after going through tranformation)
*  `abC` => `generated-abcd` (will be lowercased after tranformation)
*  `ab` => `generated-ab-d`(would remain the same after going through tranformation)

I can add these these to the unit tests if you want ?

matskiv:
@KevFan oops, I didn't notice that .toLower call and gave bad examples.
Give me one more shot - `test.a` and `test-a`?

KevFan:
> @KevFan oops, I didn't notice that .toLower call and gave bad examples.
> Give me one more shot - `test.a` and `test-a`?

Oh in that case, they would both end up being `generated-test-a` which would cause problems :thinking: I can add `.-_` to the regex to not replace these as they are valid but theres will still be a chance there can be a resource already exists with the same name :confused:

matskiv:
@KevFan yes, as you point out, it can still have conflicts.
Another idea is to use User resource UID in KeycloakUser resource name. That one should be unique.

KevFan:
> @KevFan yes, as you point out, it can still have conflicts.
> Another idea is to use User resource UID in KeycloakUser resource name. That one should be unique.

Yeah sure, that would be easiest :+1: 
For the `KeyCloakUser` resource name do which do you think is best ?:
* `<user-uid>`
* `generated-<user-uid>` 
* `generated-<processed-username>-<user-uid>`

matskiv:
@KevFan I was thinking about `generated-<user-uid>`, but it's not a strong preference.
`generated-<processed-username>-<user-uid>` would be convenient, but also more code.
It's up to you :)

Boomatang:
@matskiv Would this have fixed the issue I had the other day where it was a member of the SRE team that had logged into the cluster before I had set up the operator?

KevFan:
> @KevFan I was thinking about `generated-<user-uid>`, but it's not a strong preference.
> `generated-<processed-username>-<user-uid>` would be convenient, but also more code.
> It's up to you :)

Cool, I went for `generated-<processed-username>-<user-uid` if you want to have a look again when you get the chance :smiley:

matskiv:
@Boomatang yes, it would.

matskiv:
Code looks ok, but some usernames in tests are too complex and unrealistic 😀 
I'll test this in the morning, I don't have a cluster now.

KevFan:
> Code looks ok, but some usernames in tests are too complex and unrealistic 
> I'll test this in the morning, I don't have a cluster now.

True, was just me testing that most special characters would be replaced :laughing: 
I updated the tests to be more normal cases anyhow :+1:

matskiv:
I tested this and it worked as expected.
Unfortunately KeycloakUser CR with upper case username didn't reconcile completely. I fixed one issue that was causing that and reported another one here - https://issues.redhat.com/browse/KEYCLOAK-13158, but it should not block us.

/approve
/lgtm

PR #449, opened 2020-02-27
INTLY-4360  add a postgres backup cronjob to amq online

austincunningham:
# Why
issue: https://issues.redhat.com/browse/INTLY-4360

# What
Add in a backup cron job for postgres to amq online 

# Verification
## See if cronjob created
- checkout this branch
- deploy rhmi
- check prometheus alerts route for the following alert rule for amq online postgres
image
to see if the cron job exists
and check all projects/cron jobs to confirm enmasse-postgres-backup present. 
image
## Test the cronjob
- Setup the s3 bucket and access as outline in the [readme](https://github.com/integr8ly/integreatly-operator#local-setup)
- Create the job
```sh
oc create job enmasse-postgres-backup-$(date +"%Y-%m-%d-%H%M%S") --from=cronjob/enmasse-postgres-backup -n redhat-rhmi-amq-online
```
This should create a job in the form `enmasse-postgres-backup-2020-03-03-`
- Check the jobs pod logs to confirm that the backup was successful
e.g.
image

### DIFF
--- a/pkg/config/amqonline.go
+++ b/pkg/config/amqonline.go
@@ -112,6 +112,10 @@ func (a *AMQOnline) GetBackupsSecretName() string {
 	return "backups-s3-credentials"
 }
 
+func (c *AMQOnline) GetPostgresBackupSecretName() string {
+	return "enmasse-postgres-secret"
+}
+
 func (a *AMQOnline) GetBackupSchedule() string {
 	return "30 2 * * *"
 }
--- a/pkg/products/amqonline/reconciler.go
+++ b/pkg/products/amqonline/reconciler.go
@@ -8,6 +8,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	monitoringv1alpha1 "github.com/integr8ly/application-monitoring-operator/pkg/apis/applicationmonitoring/v1alpha1"
+	cro1types "github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1/types"
 	enmasseadminv1beta1 "github.com/integr8ly/integreatly-operator/pkg/apis/enmasse/admin/v1beta1"
 	enmassev1beta1 "github.com/integr8ly/integreatly-operator/pkg/apis/enmasse/v1beta1"
 	enmassev1beta2 "github.com/integr8ly/integreatly-operator/pkg/apis/enmasse/v1beta2"
@@ -514,6 +515,10 @@ func (r *Reconciler) reconcileBackup(ctx context.Context, serverClient k8sclient
 	if err != nil {
 		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to reconcile postgres: %w", err)
 	}
+	if postgres.Status.Phase != cro1types.PhaseComplete {
+		return integreatlyv1alpha1.PhaseAwaitingComponents, nil
+	}
+
 	// get the secret created by the cloud resources operator
 	croSec := &corev1.Secret{}
 	err = serverClient.Get(ctx, k8sclient.ObjectKey{Name: postgres.Status.SecretRef.Name, Namespace: postgres.Status.SecretRef.Namespace}, croSec)
--- a/pkg/products/amqonline/reconciler.go
+++ b/pkg/products/amqonline/reconciler.go
@@ -514,6 +514,12 @@ func (r *Reconciler) reconcileBackup(ctx context.Context, serverClient k8sclient
 		},
 		Components: []resources.BackupComponent{
 			{
+				Name:     "enmasse-postgres-backup",
+				Type:     "postgres",
+				Secret:   resources.BackupSecretLocation{Name: r.Config.GetPostgresBackupSecretName(), Namespace: r.Config.GetNamespace()},
+				Schedule: r.Config.GetBackupSchedule(),
+			},
+			{
 				Name:     "enmasse-pv-backup",
 				Type:     "enmasse_pv",
 				Schedule: r.Config.GetBackupSchedule(),
--- a/pkg/products/amqonline/reconciler.go
+++ b/pkg/products/amqonline/reconciler.go
@@ -295,6 +295,7 @@ func (r *Reconciler) reconcileStandardAuthenticationService(ctx context.Context,
 			return nil
 		},
 	)
+
 	if postgres.Status.Phase != cro1types.PhaseComplete {
 		return integreatlyv1alpha1.PhaseAwaitingComponents, nil
 	}
@@ -343,13 +344,6 @@ func (r *Reconciler) reconcileStandardAuthenticationService(ctx context.Context,
 		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed reconciling enmasse standard auth service keycloak secret: %w", err)
 	}
 
-	// get the secret created by the cloud resources operator
-	croSec := &corev1.Secret{}
-	err = serverClient.Get(ctx, k8sclient.ObjectKey{Name: postgres.Status.SecretRef.Name, Namespace: postgres.Status.SecretRef.Namespace}, croSec)
-	if err != nil {
-		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to get postgres credential secret: %w", err)
-	}
-
 	// create backup secret
 	logrus.Info("Reconciling amq-online postgres backup secret")
 	amqOnlneBackUpSecret := &corev1.Secret{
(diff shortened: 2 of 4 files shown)


pkg/config/amqonline.go
```diff
@@ -112,6 +112,10 @@ func (a *AMQOnline) GetBackupsSecretName() string {
 	return "backups-s3-credentials"
 }
 
+func (c *AMQOnline) GetPostgresBackupSecretName() string {
+	return "enmasse-postgres-secret"
```
KevFan: Just wondering is the meant to backup the `standard-auth-service` postgres?
Cause this secret doesn't exist or created by CRO? :thinking: 

https://github.com/integr8ly/integreatly-operator/blob/c3ce78c1abf241cb3de792c1b3b7acf7fefe6022/pkg/products/amqonline/reconciler.go#L275
https://github.com/integr8ly/integreatly-operator/blob/c3ce78c1abf241cb3de792c1b3b7acf7fefe6022/pkg/products/amqonline/reconciler.go#L304

So running the backup job would fail :confused: 
```
oc create job enmasse-postgres-backup-$(date +"%Y-%m-%d-%H%M%S") --from=cronjob/enmasse-postgres-backup -n redhat-rhmi-amq-online
```

austincunningham: :eyes:

austincunningham: /hold

austincunningham: good catch , the secret needs to be created like
https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/codeready/reconciler.go#L219L241

KevFan: Hmm yeah, I'm not too sure also, maybe just changing `GetPostgresBackupSecretName()` to return `standard-authservice-postgresql` would do :confused: 

@grdryn Might be able to better comment on this since he added the change for the in cluster postgres of amq online

austincunningham: `standard-authservice-postgresql` doesn't connect to psql
```sh
psql: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
==> Component data dump completed
ls: cannot access /tmp/intly/archives/*: No such file or directory```

Comment (grdryn) [reply pkg/config/amqonline.go]: If you're going to change this to return `standard-authservice-postgresql`, then change https://github.com/integr8ly/integreatly-operator/blob/c3ce78c1abf241cb3de792c1b3b7acf7fefe6022/pkg/products/amqonline/reconciler.go#L275 to get the value from this function rather than having two different places where it's defined. 

Also maybe change the function name since it wouldn't be specific to backups then.

Comment (grdryn) [reply pkg/config/amqonline.go]: @austincunningham is the secret formatted in the expected way for the backup script?

Comment (austincunningham) [reply pkg/config/amqonline.go]: @grdryn I suspect you are correct the format is wrong , not quite sure where to find the correct format, but the previous example in 
codeready-postgres-secret  which has
```yaml
data:
  POSTGRES_DATABASE: 
  POSTGRES_HOST: 
  POSTGRES_PASSWORD: 
  POSTGRES_PORT: 
  POSTGRES_USERNAME: 
  POSTGRES_VERSION: 
``` 


where as standard-authservice-postgresql has 
```yaml
data:
  database-password: 
  database-user: 
```

grdryn: Yeah, that looks correct: https://github.com/integr8ly/backup-container-image/blob/0d0fb63707f3485fa22c1fc2e109bb185da01e2c/image/tools/lib/component/postgres.sh#L3..L25

I think you'll need to either create a new secret, or maybe check the secret that the cloud-resource-operator creates in the redhat-rhmi-operator repo...maybe that's in the right format?

pkg/products/amqonline/reconciler.go
```diff
@@ -505,6 +506,45 @@ func (r *Reconciler) reconcileConfig(ctx context.Context, serverClient k8sclient
...
+		amqOnlneBackUpSecret.Data["POSTGRES_USERNAME"] = croSec.Data["username"]
+		amqOnlneBackUpSecret.Data["POSTGRES_PASSWORD"] = croSec.Data["password"]
+		amqOnlneBackUpSecret.Data["POSTGRES_DATABASE"] = croSec.Data["database"]
+		amqOnlneBackUpSecret.Data["POSTGRES_PORT"] = croSec.Data["port"]
+		amqOnlneBackUpSecret.Data["POSTGRES_VERSION"] = []byte("10")
+		return nil
+	})
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to create or update %s connection secret: %w", r.Config.GetPostgresBackupSecretName(), err)
+	}
```
KevFan: What do you think of moving this to `reconcileStandardAuthenticationService()` where the postgres is created and making the necessary changes? By doing this there will be a single place for the creation/reconcile of the postGres and fetch / creation of the secrets for the backup seret :thinking: 

Also I think it would be good to check that the postgres is available before fetching the croSecret as otherwise it would throw a panic if it's not present like it's done in codeready

https://github.com/integr8ly/integreatly-operator/blob/54e3cc6b2e6903b2c0763e2e296c424facdd4a6e/pkg/products/codeready/reconciler.go#L199

austincunningham: :+1:

pkg/products/amqonline/reconciler.go
```diff
@@ -337,6 +343,37 @@ func (r *Reconciler) reconcileStandardAuthenticationService(ctx context.Context,
 		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed reconciling enmasse standard auth service keycloak secret: %w", err)
 	}
 
+	// get the secret created by the cloud resources operator
+	croSec := &corev1.Secret{}
+	err = serverClient.Get(ctx, k8sclient.ObjectKey{Name: postgres.Status.SecretRef.Name, Namespace: postgres.Status.SecretRef.Namespace}, croSec)
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to get postgres credential secret: %w", err)
+	}
```
KevFan: I think this fetch can replace the initial `croSecret` fetch in the code block above it
https://github.com/integr8ly/integreatly-operator/blob/5ab142cc591cdcbf4540e4b46bc26abfdf9ae000/pkg/products/amqonline/reconciler.go#L301-L307

pkg/products/amqonline/reconciler_test.go
```diff
@@ -184,6 +186,21 @@ func setupRecorder() record.EventRecorder {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-postgres-",
+			Namespace: "test-namespace",
+		},
+		Spec: crov1.PostgresSpec{},
+		Status: crov1.PostgresStatus{
+			Phase: crotypes.PhaseComplete,
+			SecretRef: &crotypes.SecretRef{
+				Name:      "test-postgres-",
+				Namespace: "test-postgres-namespace",
```
KevFan:
```suggestion
		ObjectMeta: metav1.ObjectMeta{
			Name:      "standard-authservice-postgresql",
			Namespace: "test-namespace",
		},
		Spec: crov1.PostgresSpec{},
		Status: crov1.PostgresStatus{
			Phase: crotypes.PhaseComplete,
			SecretRef: &crotypes.SecretRef{
				Name:      "standard-authservice-postgresql",
				Namespace: "test-namespace",
```
I think unit tests are failing with these changes :thinking:

pkg/products/amqonline/reconciler.go
davidkirwan: Is the backup secret the same as the croSecret? (I've no idea lol)

davidkirwan:
Could you take another look please @KevFan and see if the code I pushed is similar to what you were thinking?

pkg/products/amqonline/reconciler.go
KevFan: Yup, the backup secret is the same as the `croSec`, it's just that the backup secret is in the product namespace under a different name for the backup container to use

On the above code suggestion, I just think these 2 fetches are the same, so would be good to remove of them ( `croSecret` for `croSec` fetch ) :thinking:

KevFan:
/lgtm

Verified cronjob alert in prometheus and running the backup cronjob

david-martin:
/approve.
Thanks @austincunningham & @davidkirwan

davidkirwan:
@david-martin Looks like prow does not like the fullstop in /approve

PR #452, opened 2020-02-27
dont unassign 3scale admin role in workshop mode

pb82:
Check for workshop mode also when assigning `member` role to 3scale users: we want to make sure that users who got admin permissions but are not Openshift admins get to keep the permissions in workshop mode. This is required to be able to complete the Walkthroughs.


matskiv:
Amazing! I didn't know that in Go, "else if" gets executed even when first "if" was true.
/approve
/lgtm

pb82:
@matskiv It doesn't, but it will go into this branch in the next reconcile because now the user is already an admin, not a member and not an Openshift Admin.

matskiv:
@pb82 That makes much more sense, thanks for the explanation. :)

PR #453, opened 2020-02-28
DEL-19 change manifest directory name from fuse to fuse-online

StevenTobin:
* Change manifest directory name to integreatly-fuse-online from integreatly-fuse
* Update fuse reconciler manifestPackage var to reflect this

The product name is fuse-online so the manifest should reflect that. In delorean we expect the product name to be fuse-online not fuse.


laurafitzgerald:
Looks good to me. Will wait for someone from openshift 4 team to take a look before lgtm-ing

PR #454, opened 2020-02-28
Fix(env type): Chnage USE_CLUSTER_STORAGE to Boolean true

Boomatang:
# Jira
https://issues.redhat.com/browse/INTLY-5809

# Reason
Preflight checks was failing when using `make code/run`

# Fix
Change `USE_CLUSTER_STORAGE` to boolean version over string

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ export SELF_SIGNED_CERTS   ?= true
 export INSTALLATION_TYPE   ?= managed
 export INSTALLATION_NAME   ?= rhmi
 export INSTALLATION_PREFIX ?= redhat-rhmi
-export USE_CLUSTER_STORAGE ?= "true"
+export USE_CLUSTER_STORAGE ?= true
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -20,7 +20,7 @@ export SELF_SIGNED_CERTS   ?= true
 export INSTALLATION_TYPE   ?= managed
 export INSTALLATION_NAME   ?= rhmi
 export INSTALLATION_PREFIX ?= redhat-rhmi
-export USE_CLUSTER_STORAGE ?= "true"
+export USE_CLUSTER_STORAGE ?= true
```
matskiv: @philbrookes you were the one that added double quotes, was that intentional?

philbrookes: yes, this is confusing, when I submitted the PR to change this value to a string, the e2e failed until I changed this to a string, as it couldn't create the CR. It seems to be passing now, so maybe I addressed this issue in 2 places and they have overlapped.

PR #455, opened 2020-02-28
INTLY-5760 - Fix config maps are not deleted on uninstall

KevFan:
## Description
The `foreground` finalizer gets removed immediately when a resource is marked for deletion. This prevents the operator from removing the created config maps after the product uninstalls. In addition, since this finalizer is used in the "cloud-resource-config", it can be deleted during a namespace deletion before the operator has completed the product uninstalls which may cause problems when used with aws resources 

Jira:
* https://issues.redhat.com/browse/INTLY-5760

## Verification
* Install integreatly and ensure the `RHMI` CR and `cloud-resources-config` config map now have a `kubernetes` finalizer
* Delete the `RHMI CR`, ensure the `cloud-resource-config` config map is not deleted until the CR has been deleted 
* Ensure the logs contain "uninstall completed"
* Repeat install but instead of deleting the CR, delete the `redhat-rhmi-operator` namespace
* Ensure the `cloud-resource-config` config map is not deleted until the CR has been deleted 
* Ensure the uninstall completes successfully

### DIFF
--- a/pkg/controller/installation/installation_controller.go
+++ b/pkg/controller/installation/installation_controller.go
@@ -3,11 +3,10 @@ package installation
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/types"
 	"os"
 	"strings"
 	"time"
-
-	"k8s.io/apimachinery/pkg/types"
 
 	usersv1 "github.com/openshift/api/user/v1"
 
@@ -40,7 +39,7 @@ import (
 )
 
 const (
-	deletionFinalizer                = "foregroundDeletion"
+	deletionFinalizer                = "finalizer/configmaps"
 	DefaultInstallationName          = "rhmi"
 	DefaultInstallationConfigMapName = "installation-config"
 	DefaultInstallationPrefix        = "redhat-rhmi-"
@@ -286,12 +285,12 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 				return retryRequeue, merr
 			}
 
-			if err = r.handleCROConfigDeletion(*installation); err != nil {
+			if err = r.handleCROConfigDeletion(*installation); err != nil && !k8serr.IsNotFound(err) {
 				merr.Add(fmt.Errorf("failed to remove Cloud Resource ConfigMap: %w", err))
 				return retryRequeue, merr
 			}
 
-			err = resources.RemoveFinalizer(r.context, installation, r.client, deletionFinalizer)
+			err = resources.RemoveFinalizer(context.TODO(), installation, r.client, deletionFinalizer)
 			if err != nil {
 				merr.Add(fmt.Errorf("Failed to remove finalizer: %w", err))
 				return retryRequeue, merr
@@ -576,8 +575,8 @@ func (r *ReconcileInstallation) handleCROConfigDeletion(rhmi integreatlyv1alpha1
 	// get cloud resource config map
 	croConf := &corev1.ConfigMap{}
 	err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: rhmi.Namespace, Name: DefaultCloudResourceConfigName}, croConf)
-	if err != nil && !k8serr.IsNotFound(err) {
-		return fmt.Errorf("error occurred trying to get cro config map %w", err)
+	if err != nil {
+		return err
 	}
 
 	// remove cloud resource config deletion finalizer if it exists


pkg/controller/installation/installation_controller.go
```diff
@@ -40,7 +39,7 @@ import (
 )
 
 const (
-	deletionFinalizer                = "foregroundDeletion"
+	deletionFinalizer                = "kubernetes"
```
matskiv: `kubernetes` finalizer name is used by k8s itself, so IMHO we should use a custom finalizer name, which we fully control

KevFan: Yup, sure, I've just pushed a commit to use `finalizer/configmaps` instead.
Its using `/` instead of  `.`  like the other product finalizers as it would otherwise give an invalid finalizer error.

Looks to be due to this so could be different validation for finalizer between different resources :thinking: :
https://github.com/kubernetes/kubernetes/blob/86141c0cce2496508c2f9c1f1af3e6cbfb69f50c/pkg/apis/core/validation/validation.go#L5448

matskiv: @KevFan And have you tried adding it with our suffix, e.g. `finalizer.configmaps.integreatly.org` ?

KevFan: @matskiv Yup, got this error 
```
    Bootstrap stage reconcile failed: ConfigMap "cloud-resource-config" is
    invalid: metadata.finalizers[0]: Invalid value:
    "finalizer.configmaps.integreatly.org": name is neither a standard finalizer
    name nor is it fully qualified
```
I'm guessing it's due to missing a `/` in it's name from: 
https://github.com/kubernetes/kubernetes/blob/86141c0cce2496508c2f9c1f1af3e6cbfb69f50c/pkg/apis/core/validation/validation.go#L5446

Boomatang:
I have test both ways as said in the verification steps.
- Deleting the `RHMI` CR will tidy up the resources as expected.
- Deleting the `redhat-rhmi-operator` namespace **does not** tidy the resources up fully.

KevFan:
> I have test both ways as said in the verification steps.
> 
> * Deleting the `RHMI` CR will tidy up the resources as expected.
> * Deleting the `redhat-rhmi-operator` namespace **does not** tidy the resources up fully.

Yeah I have also seen that. Looks like cloud-resource no longer can clean up the resource when the 
`redhat-rhmi-operator` namespace is triggered for deletion :man_shrugging: 
```
{"level":"error","ts":1583165303.4293768,"logger":"controller-runtime.controller","msg":"Reconciler error","controller":"postgres-controller","request":"redhat-rhmi-operator/codeready-postgres-rhmi",
"error":"failed to update resource status phase and message: postgres.integreatly.org \"codeready-postgres-rhmi\" is forbidden: User \"system:serviceaccount:redhat-rhmi-cloud-resources-operator:cloud-resource-operator\" cannot update resource 
\"postgres/status\" in API group \"integreatly.org\" in the namespace \"redhat-rhmi-operator\": RBAC: role.rbac.authorization.k8s.io \"cloud-resources.v0.13.0-kbnr5\" not found","errorVerbose":"postgres.integreatly.org \"codeready-postgres-rhmi\" 
is forbidden: User \"system:serviceaccount:redhat-rhmi-cloud-resources-operator:cloud-resource-operator\" cannot update resource \"postgres/status\" in API group \"integreatly.org\" in the namespace \"redhat-rhmi-operator\": RBAC: role.rbac.authorization.k8s.io 
\"cloud-resources.v0.13.0-kbnr5\" not found\nfailed to update resource status phase and message\
```
Not too sure is this ever going to be a scenario for uninstall via deletion the `redhat-rhmi-operator` namespace :confused: The `cloud-resource-config` is still kept in this scenario as the  cloud resource finalizer isnt removed from the RHMI CR. 

@matskiv What do you think on this? :thinking:

matskiv:
@KevFan sounds like role and rolebinding for the CRO is removed because ns is marked for deletion. Do you still have a cluster with this? Can you check roles and rolebindings in redhat-rhmi-operator ns?

KevFan:
> @KevFan sounds like role and rolebinding for the CRO is removed because ns is marked for deletion. Do you still have a cluster with this? Can you check roles and rolebindings in redhat-rhmi-operator ns?

Just checked. Yeah looks like the the cloud resource role and rolebinding in the `redhat-rhmi-operator` namespace was removed when the namespace was marked for deletion :confused:

matskiv:
In that case, I don't think we can easily prevent problems when the whole `redhat-rhmi-operator` ns is deleted.

KevFan:
@matskiv Oh right, yeah I'm not too sure either. Unless maybe we provision the CRO resource in it's own namespace? :confused: 

But do you think it's a likely case for cleanup by deleting the `redhat-rhmi-operator` namespace? Cause we know that it should be uninstalled via deleting the CR but not sure how others would approach it :man_shrugging: 

Would you think this work would be part of this PR? As it does cleanup in the normal flow of deleting the CR and the `cloud-resource-config` is prevented to be deleted if the resource is deleted before the CR so should prevent any issues during in cluster dev work where this config is deployed :thinking:

matskiv:
> maybe we provision the CRO resource in it's own namespace?
That would probably work, but I don't think we have time to investigate this approach now

I don't think we need to cover the case of `redhat-rhmi-operator` ns getting delete, in this PR.

PR #459, opened 2020-03-02
bumping amo from 1.1.2 to 1.1.4

R-Lawton:
Bumping amo version in the integreatly operator

Verification:
Run integreatly operator from this PR branch


dimitraz:
Do we want to bump the manifest and the image as well?

R-Lawton:
No not unless we are doing a release . This is not a vital thing it will go out in the next release

manifests/integreatly-monitoring/monitoring-1.1.4/application-monitoring-operator.v1.1.4.clusterserviceversion.yaml
```diff
@@ -0,0 +1,360 @@
...
+  - Grafana
+  - Monitoring
+  - AlertManager
+  maintainers:
+  - email: placeholder@email.com
+    name: Integreatly
+  maturity: alpha
+  provider:
+    name: Red Hat
+  replaces: application-monitoring-operator.v1.1.2
```
ciaranRoche: I think this needs to be removed, we do not have this field in any other CSV

dimitraz: We actually do keep it in the last manifest: https://github.com/integr8ly/integreatly-operator/blob/master/manifests/integreatly-monitoring/monitoring-1.1.2/application-monitoring-operator.v1.1.2.clusterserviceversion.yaml#L347

I think we can do this now because we aren't deploying through operator hub

R-Lawton: Ok ill remove that. Why is it created by the make gen/csv if its not needed?

R-Lawton: @ciaranRoche Just checked with Oleg we do need it

PR #465, opened 2020-03-02
Del 19 fuse online release pipeline

StevenTobin:
Add fuse specific parts to the  manifest processing script.

Successful dryrun of pipeline here https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Delorean%202.0/job/delorean-2-ews/job/fuse-online/job/discovery/12/

### DIFF
--- a/scripts/process-image-manifests
+++ b/scripts/process-image-manifests
@@ -52,7 +52,7 @@ extract_manifests_from_image() {
     ls -la ${MANIFEST_TMP_DIR}
 }
 
-set_package_name_for_current_csv() {
+get_current_csv_filename() {
   # In the case of 3scale the package name needs to be customised
   case $PRODUCT in
   "3scale")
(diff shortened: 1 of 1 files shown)


scripts/process-image-manifests
```diff
@@ -52,21 +52,55 @@ extract_manifests_from_image() {
...
+  # In the case of 3scale the package name needs to be customised
+  case $PRODUCT in
+  "3scale")
+    CURRENT_CSV_FILE_NAME=${PRODUCT}-operator.package.yaml
+    ;;
+  *)
+    CURRENT_CSV_FILE_NAME=${PRODUCT}.package.yaml
+  esac
+}
+
```
mikenairn: Any reason you are calling this set* and all the rest are get* even though they pretty much all set values?

Would just calling it "get_current_csv_filename" not be better to match the var it's going to set?

StevenTobin: No real reason. Set made sense to me and I didn't notice the rest were already gets, i'll make that change.

StevenTobin: Actually this was the commit that @laurafitzgerald added so she should speak to whether there was a reason for it

StevenTobin:
Changed the name of that function as per comment @mikenairn

PR #467, opened 2020-03-03
INTLY-4107 - filter listed KeycloakUsers by our prefix

matskiv:
JIRA: https://issues.redhat.com/browse/INTLY-4107

User SSO reconciler should only check KeycloakUser resources that it creates - the ones with `generated-` prefix.
This will allow SRE's to create a KeycloakUser resource to gain access into SSO admin console.

Verification:
Run the operator with current PR (you can use a cluster installed from a different branch)
Create a testing IDP
Follow new SOP for getting admin access - https://github.com/RHCloudServices/integreatly-help/pull/410 (use a user from testing IDP)
Verify you can login into User SSO console as admin.
Verify that "generated-..." KeycloakUser for your test user was not created in user sso ns.


philbrookes:
@matskiv can you rerun the e2e test locally and paste the output here, so we can override prow and get these merged

matskiv:
@philbrookes ok, will do

PR #471, opened 2020-03-03
update keycloak operator image

pb82:
Instead of pointing to a fork and image in a private account, this PR updates the keycloak operator image to one based on a fork in the integreatly org (https://github.com/integr8ly/keycloak-operator-rhmi/tree/all-branches) hosted on the integreatly quay repo.


davidffrench:
/lgtm
/approve

What is the reason for the prow failures?

matskiv:
@pb82 btw: there is also some image reference on line 8:
https://github.com/integr8ly/integreatly-operator/blob/ff4550ab4fcabf4737b6cf029eac38c33427c457/manifests/integreatly-rhsso/rhsso-8.0.1/keycloakoperator.8.0.1.clusterserviceversion.yaml#L8

Not sure how it's used by OLM, but would be good to fix just in case.

pb82:
Good point @matskiv, updated now

PR #466, opened 2020-03-02
INTLY-5589: Create dedicated-admins and realm-managers groups

sergioifg94:
Implement functionallity in RHSSO User operator to create a dedicated-admins group
with a realm-managers child group.
The dedicated-admins group has the manage-users and view-realm client roles,
and the realm-managers group has a manage-realm role for each role that is
created.
Implement functionallity to add the dedicated admin generated users from
Openshift to the dedicated-admins and realm-managers groups.

Refactor the logic to reconcile Keycloak groups based on a `keycloakGroupSpec`
struct that contains the desired status of the group.

## Verification

Running on a cluster, set up the SSO IdP
```sh
PASSWORD=Password1 ./scripts/setup-sso-idp.sh
```
Log in the User SSO as the `customer-admin01` user and check that (_Note: it might take a couple of minutes for the admin permissions to propagate after first login_):

- [ ] A group in the User SSO instance called `dedicated-admins` is created
  - [ ] The group has the `manage-users` role mapped to the `master-realm` client
  - [ ] The group has the `view-realm` role mapped to the `master-realm` client
  - [ ] A child group of the `dedicated-admins` group called `realm-managers` is created
    - [ ] The group has the `manage-realm` role mapped to the `master-realm` client
    - [ ] Create a new realm and, after reconciliation, the group has the all roles except `impersonation` mapped to the `<realm name>-realm` client
- [ ] The dedicated admin users from the Openshift IdP are members of the `dedicated-admins` and `realm-managers` groups in SSO

PR #472, opened 2020-03-03
INTLY-5840 - setup default roles in User SSO only once

matskiv:
JIRA: https://issues.redhat.com/browse/INTLY-5840

Verify that:
User SSO reconciler adds roles to "rhmi-developers" Keycloak group only once, during the installation.
If a role is removed, it should not be re-added by the operator.
Verify that redhat-rhmi-installation-config contains:
`DEFAULT_ROLES_CONFIGURED: "true"` in "rhssouser" section.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,6 @@ require (
 	github.com/integr8ly/keycloak-client v0.0.0-20200302145927-c6e3f0174498
 	github.com/jstemmer/go-junit-report v0.0.0-20190106144839-af01ea7f8024
 	github.com/keycloak/keycloak-operator v0.0.0-20200207072807-b527c8b26465
-	github.com/matryer/moq v0.0.0-20200125112110-7615cbe60268 // indirect
 	github.com/openshift/api v3.9.1-0.20191031084152-11eee842dafd+incompatible
 	github.com/openshift/client-go v3.9.0+incompatible
 	github.com/openshift/cluster-samples-operator v0.0.0-20191113195805-9e879e661d71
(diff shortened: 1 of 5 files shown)


matskiv:
@sergioifg94 You know context of this change, can you review and verify if you have time?

sergioifg94:
@matskiv  Sure I'll give it a look

sergioifg94:
@matskiv It looks all right to me. The only thing that concerns me is that it will conflict with the changes introduced in #466, as I refactored the group creation and role mapping into a more generic function. I'm wondering if the `if !rolesConfigured` block could instead wrap the whole `r.reconcileDevelopersGroup(kc)` call. The only difference is that the `rhmi-developers` group creation wouldn't be reconciled once the flag is set to true.

If not then I can still figure out a way to keep the generic function. Perhaps adding a predicate parameter to the [reconcileGroup](https://github.com/integr8ly/integreatly-operator/blob/46d02475b17f628a9c15cf53e6ae123555f0cb3b/pkg/products/rhssouser/reconciler.go#L1458) function:
```golang
r.reconcileGroup(kcClient, groupSpec, func() bool { return !rolesConfigured })
```

matskiv:
@sergioifg94 oh, I see.
Yeah, I can probably wrap `r.reconcileDevelopersGroup(kc)` instead. I'll wait for your PR to land, and then ping you again, once I rebase this.

maleck13:
@matskiv The creation of roles only happening once. This seems like something we would want to happen via the KC operator user CR. If so lets open an issue for that feature.

matskiv:
@maleck13 So this is about roles assigned to Keycloak groups. Keycloak operator can't create/manage groups, so this would be a feature request, not a bug.
And I was actually thinking about compiling various ideas or feature request that we have for keycloak operator now. With the user flow and permission changes are previous requirements became outdated. I just didn't have time to work on this yet. I create a task to get back to this in the future - https://issues.redhat.com/browse/INTLY-5957

sergioifg94:
:+1: 
/lgtm

PR #483, opened 2020-03-05
Update OWNERS for test folder

david-martin:
@b1zzu @pawelpaszki Who from QE do you think should be added as an owner of the test folder?

### DIFF
--- a/test/OWNERS
+++ b/test/OWNERS
@@ -3,7 +3,12 @@
 approvers:
 - mikenairn
 - pmccarthy
-- tremes
 - maleck13
 - philbrookes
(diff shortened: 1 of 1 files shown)


pawelpaszki:
@david-martin - so tremes should be removed. About adding new people - I think it is tricky, because you shouldn't really have too many members there. Maybe add @psturc and any of (@jhellar me and @b1zzu )

grdryn:
/lgtm

@philbrookes I'll leave it to you to approve

damienomurchu:
@david-martin can you add me to the owners in this PR too?

PR #489, opened 2020-03-06
INTLY-5844: Remove logic that demotes manually promoted users on 3scale

sergioifg94:
> [JIRA ticket](https://issues.redhat.com/browse/INTLY-5844)

## Why

The operator currently promotes or demotes 3scale users based on their membership to the
admins group in Openshift. However there might be the case that an admin **manually** promotes
a non-admin user. If this happens, the operator will demote that user as they don't belong to the Openshift admin group.

## How

Remove the logic from `reconcileOpenshiftUsers` function that demotes users that have an admin
role in 3scale but are not admins in Openshift

# Verification steps

In a cluster:

- [ ] Create 2 users admin and developer
- [ ] Login to cluster with both users
- [ ] Login to 3scale with admin user and promote developer to admin
- [ ] Run a reconcile of the integreatly-operator
- [ ] See that the developer user is still an admin in 3scale

### DIFF
--- a/pkg/products/threescale/reconciler.go
+++ b/pkg/products/threescale/reconciler.go
@@ -793,28 +793,34 @@ func (r *Reconciler) reconcileOpenshiftUsers(ctx context.Context, installation *
 	if err != nil {
 		return integreatlyv1alpha1.PhaseInProgress, err
 	}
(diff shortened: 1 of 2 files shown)


david-martin:
/approve

This change makes sense.

PR #495, opened 2020-03-09
INTLY-5944 Increase prometheus pvc storage & retention

david-martin:
https://issues.redhat.com/browse/INTLY-5638

The new values match what's used for RHMI 1.x by SRE.

After install, to verify:

```
oc get pvc -n redhat-rhmi-middleware-monitoring-operator
```

Verify the pvc is 50GB.

The retention size can be verified via the Prometheus UI > Status > Configuration.

Also, verify the PVCStorageWillFillIn4Days and PVCStorageWillFillIn4Hours alerts are not triggering for the prometheus pvc

### DIFF
--- a/pkg/config/monitoring.go
+++ b/pkg/config/monitoring.go
@@ -108,11 +108,11 @@ func (m *Monitoring) GetAdditionalScrapeConfigSecretKey() string {
 }
 
 func (m *Monitoring) GetPrometheusRetention() string {
-	return "15d"
+	return "45d"
 }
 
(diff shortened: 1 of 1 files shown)


davidffrench:
@david-martin Are there any changes required to pre-sales/sales documentation around cluster capacity for customers with this change?

davidffrench:
/approve

This is a bug that should be resolved for 2.0.0. So this should be merged.

Talked to @david-martin offline, he will follow up on the resource capacity changes separately to this PR.

PR #497, opened 2020-03-09
copy and link pull secret

StevenTobin:
https://issues.redhat.com/browse/DEL-43

When installing branches created by the delorean ews pipelines, the serviceAccount pulling the image must have the requisite imagePullSecret to successfully pull the image. This PR alters the namespace reconciliation so that:

- If there is a secret defined in the installation spec that secret is copied to each namespace created by the operator
- For each SA in that namespace the secret is added as an imagePullSecret 

*Verification*

1. make cluster/prepare
2. make cluster/prepare/crd
3. create secret in rhmi-redhat-operator namespace
4. add this line to installation cr `pullSecret: {"name": "<secret-name>","namespace":"rhi-redhat-operator"}`
5. make code/run/service_account
6. verify that:

1. The secret specified in the cr has been copied to each namespace created by the operator
2. The secret is in the imagePullSecrets array in every serviceAccount in each operator created namespace
3. A syndesis-pull-secret exists in the fuse namespace
4. A `samples-registry-credentials` secret exists in the 3scale namespace
5. An `apicurito-pull-secret` exists in the apicurito and apicurito-operator namespaces


pkg/resources/pullSecret.go
```diff
@@ -54,3 +54,43 @@ func CopySecret(ctx context.Context, client k8sclient.Client, srcName, srcNamesp
...
+		currentSa := &corev1.ServiceAccount{}
+		err = client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: sa.Name}, currentSa)
+		if err != nil {
+			return err
+		}
+
+		pullSecretFound := false
+		for _, ips := range currentSa.ImagePullSecrets {
+			if ips.Name == secretName {
+				pullSecretFound = true
```
wei-lee: I think we can stop the loop once the pull secret is found

pkg/resources/pullSecret.go
```diff
@@ -54,3 +54,43 @@ func CopySecret(ctx context.Context, client k8sclient.Client, srcName, srcNamesp
...
+			}
+
+		}
+
+		if !pullSecretFound {
+			newPullSecret := corev1.LocalObjectReference{Name: secretName}
+			imagePullSecret := append(currentSa.ImagePullSecrets, newPullSecret)
+			_, err = controllerutil.CreateOrUpdate(ctx, client, currentSa, func() error {
+				currentSa.ImagePullSecrets = imagePullSecret
```
wei-lee: what's this line for?

pkg/resources/reconciler.go
```diff
@@ -130,6 +130,18 @@ func (r *Reconciler) ReconcileNamespace(ctx context.Context, namespace string, i
 		return integreatlyv1alpha1.PhaseCompleted, nil
 	}
 
+	if inst.Spec.PullSecret.Name != "" {
+		_, err := r.ReconcilePullSecret(ctx, namespace, inst.Spec.PullSecret.Name, inst, client)
```
wei-lee: IIRC, there are some products are calling this function atm.  If we are reconciling the pull secret here, and this function is called for all products, then I think we can remove the existing references to this function.

pkg/resources/pullSecret.go
StevenTobin: Good point, i'll change that

pkg/resources/pullSecret.go
StevenTobin: You are correct it's unneccessary, the append does what I want it to, changed that.

pkg/resources/reconciler.go
StevenTobin: We could, but when this function is called with the product reconcilers it's called with a reference to a specific secret like here https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/apicurito/reconciler.go#L130 whereas how I've used it we are checking for a secret in the installation spec and using that.

I have no objection to removing the function from the other products as you've said it doesn't seem to be used for anything atm but as I'm doing something different here than how it's currently used, I left in what was there.

KevFan: I agree, I'm not sure would it be alright to remove this call in the individual product reconcilers as originally it was added as part of https://issues.redhat.com/browse/INTLY-3511 to fix starting walkthrough 1 correctly as they might expect pull secrets with a specifc name in the namespace or would it be fine with the service accounts linked to the pull secrets now :thinking:
@JameelB might be able to better comment on this

Would be good to make sure the walkthroughs that use these products still work correctly if we do remove these references.

However, if we don't remove these references and the `PullSecret` is not intially set in the RHMI CR, I believe it will be set to the default pull secret from any one of the products that call `ReconcilePullSecret()`. Then this part of the code will execute on next namespace reconcile, copying and linking this pull secret to the namespace and service accounts. Just wondering is this what you are expecting? :thinking:  
https://github.com/integr8ly/integreatly-operator/blob/6fecb88f02d8bc16d33270a73f8be26ad511ab65/pkg/resources/pullSecret.go#L22-L31

JameelB: @KevFan is correct. As far as I remember, Fuse was the only one that expected the secret to be there. It has to be named a certain way as well. If that secret is missing, I think the fuse operator reconcile will fail if it can't find that secret even if the pull secret is already attached to the service account. There is a new version of Fuse coming up. I'm not exactly sure if this is still the case when we upgrade.

I don't think we have this issue for the other products though so we can probably remove it there. As @KevFan best to check the walkthroughs and make sure that they work.

wei-lee: @KevFan

> However, if we don't remove these references and the PullSecret is not intially set in the RHMI CR, I believe it will be set to the default pull secret from any one of the products that call ReconcilePullSecret(). Then this part of the code will execute on next namespace reconcile, copying and linking this pull secret to the namespace and service accounts. Just wondering is this what you are expecting?

From what I can see, it does set some default values for the fields in the CR, however, they are not saved as part of this function, so I don't know if they will have any impact on the next reconciliation. Also I don't know why we set these default values: just coping the default secrets to a namespace does nothing. I feel like we should just review the logic to set the default values.

KevFan: > From what I can see, it does set some default values for the fields in the CR, however, they are not saved as part of this function, so I don't know if they will have any impact on the next reconciliation. Also I don't know why we set these default values: just coping the default secrets to a namespace does nothing. I feel like we should just review the logic to set the default values.

@wei-lee Yeah, it's not saved as part of this function, but I'm pretty sure it does get saved into the CR further down the call chain somewhere. Maybe by the controller here:
https://github.com/integr8ly/integreatly-operator/blob/03ae73c39d3b917bf2bf10f0c8cd54f97eb8410f/pkg/controller/installation/installation_controller.go#L353
Also had a quick run of this PR yesterday, and it looks like the case. But I do agree on reviewing this logic for setting the default values

pkg/products/threescale/reconciler.go
```diff
@@ -166,7 +166,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return integreatlyv1alpha1.PhaseFailed, err
 	}
 
-	phase, err = r.ReconcilePullSecret(ctx, r.Config.GetNamespace(), "", installation, serverClient)
+	phase, err = r.ReconcilePullSecret(ctx, r.Config.GetNamespace(), "3scale-pull-secret", installation, serverClient)
```
wei-lee: normally string values are defined in constants.

StevenTobin: Made that change

pkg/resources/reconciler.go
```diff
@@ -179,15 +191,23 @@ func (r *Reconciler) ReconcileFinalizer(ctx context.Context, client k8sclient.Cl
...
+	if inst.Spec.PullSecret.Name == "" {
+		srcSecretName = DefaultOriginPullSecretName
+	} else {
+		srcSecretName = inst.Spec.PullSecret.Name
+	}
+	if inst.Spec.PullSecret.Namespace == "" {
+		srcSecretNamespace = DefaultOriginPullSecretNamespace
+	} else {
+		srcSecretNamespace = inst.Spec.PullSecret.Namespace
 	}
```
mikenairn: Is this logic correct? If we are explicitly copying the default pull secret in the reconcilers that require it, do we still need to do this here?

StevenTobin: So it shouldn't get to there because we check if the name is empty but I didn't think it hurts to leave in a default. 

I suppose it will error out if it doesn't find the secret it expects which is a better way to go. I'll remove it

mikenairn:
I verified this in the following scenarios:

* CR with no pullSecret added  (Same as the current e2e test) - All worked as expected
* CR with pullSecret added but nothing actually needing it - All worked as expected
* CR with pullSecret added and this branch merged with amq-online-next - All worked as expected, but did hit this issue https://issues.redhat.com/browse/DEL-210 (Will be addressed separately)

@philbrookes @maleck13 Is the code freeze for master finsihed. If it is can we look at getting this merged back?

pkg/resources/pullSecret.go
```diff
@@ -85,3 +72,43 @@ func ReconcileSecretToRHMIOperatorNamespace(ctx context.Context, client k8sclien
...
+				currentSa.ImagePullSecrets = append(currentSa.ImagePullSecrets, newPullSecret)
+				return nil
+			})
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
```
mikenairn, line 114: @StevenTobin @wei-lee Did we not decide to remove the SA linking?

StevenTobin: Yeah I removed where this function was called but not the function itself, just an oversight, I'll remove it now.

mikenairn:
/lgtm

@wei-lee Would you mind giving this a once over, if you are happy i will approve.

PR #498, opened 2020-03-09
add imagePullSecret to csv

StevenTobin:
The amq-online csv does not include an imagePullSecrets var but the deployment of the operator defaults to the enmasse-operator dockercfg, this blocks image pulls for the mirrored images.

This pr adds a step to the image manifest processing script which will add an imagePullSecrets var for the secret we are specifying in the installation cr spec (and which will be shared to all namespaces with this pr https://github.com/integr8ly/integreatly-operator/pull/497).

Defaulting to integreatly-delorean-pull-secret which is the name of the secret when copied from quay.

### DIFF
--- a/scripts/delorean/process-csv-images
+++ b/scripts/delorean/process-csv-images
@@ -28,6 +28,7 @@ RH_REGISTRY_PROD=registry.redhat.io
 RH_REGISTRY_STAGE=registry.stage.redhat.io
 RH_REGISTRY_OSBS=registry-proxy.engineering.redhat.com/rh-osbs
 DELOREAN_REGISTRY=quay.io/integreatly/delorean
+IMAGE_PULL_SECRET=integreatly-delorean-pull-secret
 
 collect_csv_images() {
     csv_images=(`grep -o ''${RH_REGISTRY_PROD}'.*\|'${RH_REGISTRY_STAGE}'.*' $1 | sed 's/"//' || true`)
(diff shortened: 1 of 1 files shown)


scripts/delorean/process-image-manifests
```diff
@@ -21,10 +21,11 @@
 set -e
 
 WORK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-MANIFESTS_DIR=${WORK_DIR}/../manifests
+MANIFESTS_DIR=${WORK_DIR}/../../manifests
```
mikenairn: bah :man_facepalming: I will check for more of these

scripts/delorean/process-image-manifests
```diff
@@ -197,6 +208,7 @@ process_image() {
    get_file_name_for_new_csv
    update_new_csv
    update_package_yaml
+   update_image_pull_secret
```
mikenairn: Hmm, i wonder if this should only happen if the image mirror mapping file has anything in it. We only need to do this if we intend to pull images from the delorean quay.io registry right?

StevenTobin: Ah true, we only need if we have images to mirror so we should check that. I'll make that change

StevenTobin: So to do this I thought to just move the update logic to the csv script, as the image processing script runs before the csv script, so it's possible that the image_mirror_mapping might not exist before it runs.

But that relies on a bunch of variables that only exist in the image-manifest script. So i'll need to think about this a bit,.

StevenTobin: The get_current_csv_file function gets me what I need.

mikenairn:
/close

@StevenTobin Can you make these changes against the scripts in this repo https://github.com/integr8ly/delorean

PR #499, opened 2020-03-09
Fix manifest directory paths in delorean scripts

mikenairn:


mikenairn:
Verified this actually works this time

/approve

PR #501, opened 2020-03-10
fix: check array length before indexing

pb82:
Fixes a crash when users are added to OpenShift but have not yet logged in (and thus the federated identities array is empty).

Verification steps:

1) Install Integreatly
2) Run the `setup-sso-idp.sh` script to create an IDP and some users
3) leave the operator running: it will crash after a while

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,6 @@ require (
 	github.com/integr8ly/keycloak-client v0.0.0-20200302145927-c6e3f0174498
 	github.com/jstemmer/go-junit-report v0.0.0-20190106144839-af01ea7f8024
 	github.com/keycloak/keycloak-operator v0.0.0-20200207072807-b527c8b26465
-	github.com/matryer/moq v0.0.0-20200125112110-7615cbe60268 // indirect
 	github.com/openshift/api v3.9.1-0.20191031084152-11eee842dafd+incompatible
 	github.com/openshift/client-go v3.9.0+incompatible
 	github.com/openshift/cluster-samples-operator v0.0.0-20191113195805-9e879e661d71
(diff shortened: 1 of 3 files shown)


matskiv:
@pb82 what is adding these KeycloakUser resources into OpenShift though? 🤔 
If they are the ones added by the operator, they should have a FederatedIdentity.

I am not against this change, just trying to understand what was causing this.

pb82:
@matskiv good question. I'd assume its the script to setup the IDP:

```
➜  ~ oc get keycloakuser --all-namespaces
NAMESPACE              NAME                                                              AGE
redhat-rhmi-rhsso      generated-customer-admin01-d394502d-62b9-11ea-b172-0a580a800012   133m
redhat-rhmi-rhsso      generated-opentlc-mgr-652409ac-62ab-11ea-83cd-0a580a820012        3h51m
redhat-rhmi-rhsso      generated-test-user01-cec8b456-62c7-11ea-83cd-0a580a820012        29m
redhat-rhmi-rhsso      testing-idp-customer-admin01                                      139m
redhat-rhmi-rhsso      testing-idp-customer-admin02                                      139m
redhat-rhmi-rhsso      testing-idp-customer-admin03                                      139m
redhat-rhmi-rhsso      testing-idp-test-user01                                           139m
redhat-rhmi-rhsso      testing-idp-test-user02                                           139m
redhat-rhmi-rhsso      testing-idp-test-user03                                           139m
redhat-rhmi-rhsso      testing-idp-test-user04                                           139m
redhat-rhmi-rhsso      testing-idp-test-user05                                           139m
redhat-rhmi-rhsso      testing-idp-test-user06                                           139m
redhat-rhmi-rhsso      testing-idp-test-user07                                           139m
redhat-rhmi-rhsso      testing-idp-test-user08                                           139m
redhat-rhmi-rhsso      testing-idp-test-user09                                           139m
redhat-rhmi-rhsso      testing-idp-test-user10                                           139m
redhat-rhmi-user-sso   generated-customer-admin01-d394502d-62b9-11ea-b172-0a580a800012   132m
```

matskiv:
@pb82 but those "testing-idp" users are in cluster SSO, and your change is in user sso recnciler

maleck13:
@matskiv do you still have concerns here?

matskiv:
@maleck13 yes
Code makes sense as a safeguard for preventing crashing the operator. 
I am just worried that it might be hiding a real cause of the missing FederatedIdentity.
@pb82 you are assuming that this is caused by KeycloakUsers set up by the testing IDP script. But those users land in cluster sso, while your changes are to UserSSO reconciler.

pb82:
@matskiv Hmm right. customer sso should not reconcile those users. I'll try to reproduce the crash and see which users causes it.

pb82:
@maleck13 @matskiv The problem was that we were using a user array with a preallocated length and appending to it: so the first entry was an empty users with all references set to nil. This caused the exception. The fix is to use an empty array, but I also left the size checks just in case.

PR #502, opened 2020-03-10
INTLY-5683 Check the available pvc storage threshold before triggerin…

david-martin:
…g prediction alerts

This change is to prevent alerts triggering soon after an install where the initial working set of storage is written to the volume in a short space of time, resulting in a false positive alert.
Now, the alerts should only trigger if the storage is predicted to fill in the next 4 days/4 hours *and* the storage usage has fallen below a certain percentage (0.5 / 50%).

To verify, either modify the threshold percentage in the alert so it detects some pvcs that meed the threshold AND are predicted to run out of storage soon (typically syndesis-db & syndesis-prometheus),
or fill up a mounted volume in a pv until it reaches the threshold.
The relevant alert should trigger in either case.

cc @alanmoran @pat-cremin @sclarkso @brizrobbo  I'd be interested in your thoughts on this as it's related to recent triggering alerts in RHMI 1

also cc @obrienrobert for review as it's something we discussed recently.

### DIFF
--- a/templates/monitoring/kube_state_metrics_alerts.yaml
+++ b/templates/monitoring/kube_state_metrics_alerts.yaml
@@ -92,7 +92,12 @@ spec:
         annotations:
           message: The {{ "{{" }} $labels.persistentvolumeclaim {{ "}}" }} PVC will run of disk space in the next 4 days.
         expr: |
-          ((sum by(persistentvolumeclaim, namespace) (predict_linear(kubelet_volume_stats_available_bytes{job="kubelet"}[6h], 4 * 24 * 3600)) * on(namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key="middleware"}) * on (persistentvolumeclaim, namespace) kube_persistentvolumeclaim_info)  <= 0
+          (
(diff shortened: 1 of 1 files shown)


alanmoran:
@david-martin This LGTM. I noticed you said 0.5 / 50% in the comment but the change is for 0.25 / 25%

obrienrobert:
Verified by completing a fresh install and following the verification steps. LGTM.

david-martin:
> I noticed you said 0.5 / 50% in the comment but the change is for 0.25 / 25%

Good catch :)
I could say I did that on purpose to promote discussion, but I'd be lying.

If there's no objection to 25% being the threshold, we can stick with that.

brizrobbo:
@david-martin It could be related, but I am thinking that it may simply be something is filling up quick and then emptying regularly, as the alert typically resolves itself in some short time later (maybe 1-2 hrs). We have pulled together a graph to see if we can see this pattern (filling up quick and then emptying), but it hasn't occured for awhile.

david-martin:
Thanks.
@alanmoran @brizrobbo @obrienrobert Could I get an lgtm and approve?

PR #504, opened 2020-03-10
Initial Best Practices for e2e and functional tests

b1zzu:
This document is an extract of the major points initially discussed in this document: https://docs.google.com/document/d/1gZQ0jFo_onaT643E0eDrzG47VNs2iSAnkI7jzdhxvyA

The goal of this document is to help writing better tests and review new tests.

### DIFF
--- a/test/BEST_PRACTICES.md
+++ b/test/BEST_PRACTICES.md
@@ -39,11 +39,7 @@ A test must always be independent of other tests, which means:
 
 - it should clean up after itself so it is possible to run the test concurrently or multiple times
 
-- it should not rely on a static order when querying data because multiple tests may run in parallel
-
 - it should not delete or modify resources used by other tests so tests may run in parallel
-
-- it should not destruct any resource because other tests may rely on them
 
 ## Secrets
 
--- a/test/BEST_PRACTICES.md
+++ b/test/BEST_PRACTICES.md
@@ -33,15 +33,15 @@ instead you should create a loop to check if the resource is available or the op
 
 ## Independent
 
-A test must always be independent of other tests, which mean:
+A test must always be independent of other tests, which means:
 
-- it should not rely on other tests changes so that it would be possible to run it independently
+- it should not rely on the successful execution of other tests so the test can be run independently
 
-- it should clean up after himself so that it would be possible to run it multiple time
+- it should clean up after itself so it is possible to run the test concurrently or multiple times
 
 - it should not rely on a static order when querying data because multiple tests may run in parallel
 
-- it should not delete or modify other tests data because multiple tests may run in parallel
+- it should not delete or modify resources used by other tests so tests may run in parallel
 
 - it should not destruct any resource because other tests may rely on them
 
--- a/test/BEST_PRACTICES.md
+++ b/test/BEST_PRACTICES.md
@@ -1,16 +1,15 @@
 # Test Best Practices
 
-This is a collection of best practices to write high quality tests. The best practices
-can be referenced in review comments to speed up the review process.
+This is a collection of best practices to write high quality tests. These best practices should be referenced in the comments that you provide when you perform code reviews of tests.
 
-- [Code Style](#code-style)
+- [Coding Style](#coding-style)
 - [Test Case Traceability](#test-case-traceability)
 - [Don't Sleep](#dont-sleep)
 - [Independent](#independent)
 - [Secrets](#secrets)
 - [Logging](#logging)
 
-## Code Style
+## Coding Style
 
 The code style should be consistent with the style used in the integreatly-operator codebase [TODO: link to golang practices used by engineering]
 
@@ -21,15 +20,11 @@ Also, in case of doubt, the Golang standard best practices can be used:
 
 ## Test Case Traceability
 
-In order to compute test coverage and test automation progress across multiple test suites all
-automated tests must have an ID and must be traced back to the [integreatly-test-cases](https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases)
-repo following the [How to automate a test case and link it back](https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases#how-to-automate-a-test-case-and-link-it-back)
-tutorial.
+In order to compute test coverage and test automation progress across multiple test suites all automated tests must have an ID and must be traced back to the [integreatly-test-cases](https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases) repo following the [How to automate a test case and link it back](https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases#how-to-automate-a-test-case-and-link-it-back) tutorial.
 
 ## Don't Sleep
 
-When waiting for a resource to become available or an operation to complete do not use fixed wait times and/or sleep statements,
-instead you should create a loop to check if the resource is available or the operation has completed
+When waiting for a resource to become available or an operation to complete do not use fixed wait times and/or sleep statements, instead you should create a loop to check if the resource is available or the operation has completed
 
 ## Independent
 
@@ -43,11 +38,10 @@ A test must always be independent of other tests, which means:
 
 ## Secrets
 
-Ensure secret or sensitive data is not included in commits or log output. 
+Ensure secret or sensitive data is not included in commits or log output.
 
 ## Logging
 
-Do not be afraid of logging, if the test pass we will ignore the logs but if the test fail logs will help
-understand why the test has failed, especially if the tests fail during an unexpected step.
+Do not be afraid of logging, if the test pass we will ignore the logs but if the test fail logs will help understand why the test has failed, especially if the tests fail during an unexpected step.
 
 Use the [`t.Log`](https://golang.org/pkg/testing/#B.Log) to print to the console instead of `fmt.Print`.
--- a/test/BEST_PRACTICES.md
+++ b/test/BEST_PRACTICES.md
@@ -43,7 +43,7 @@ A test must always be independent of other tests, which means:
 
 ## Secrets
 
-Do not commit or log any sensitive data or secrets, and always double check for it because it can happen unconsciously
+Ensure secret or sensitive data is not included in commits or log output. 
 
 ## Logging
 
--- a/test/BEST_PRACTICES.md
+++ b/test/BEST_PRACTICES.md
@@ -28,7 +28,7 @@ tutorial.
 
 ## Don't Sleep
 
-When waiting for a resource to become available or an operation to complete do not use fixed wait time,
+When waiting for a resource to become available or an operation to complete do not use fixed wait times and/or sleep statements,
 instead you should create a loop to check if the resource is available or the operation has completed
 
 ## Independent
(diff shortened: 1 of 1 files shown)


b1zzu:
Ping @ldimaggi @damienomurchu @wei-lee

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,54 @@
...
+- [Go Style](#go-style)
+- [Test Case Traceability](#test-case-traceability)
+- [Don't Sleep](#dont-sleep)
+- [Independent](#independent)
+- [Secrets](#secrets)
+- [Logging](#logging)
+
+## Go Style
+
+We don't want to redefine best practices for go, therefore we should reuse the existing ones
```
ldimaggi: The reason here is that we want these best practices to focus on test design and implementation. The same coding standards used throughout the product team should also be followed for test code.

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,54 @@
...
+## Test Case Traceability
+
+In order to compute test coverage and test automation progress across multiple test suites all
+automated tests must have an ID and must be traced back to the [integreatly-test-cases](https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases)
+repo following the [How to automate a test case and link it back](https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases#how-to-automate-a-test-case-and-link-it-back)
+tutorial.
+
+## Don't Sleep
+
+When waiting for a resource to become available or an operation to complete do not use fixed wait time,
```
ldimaggi: '...do not use fixed wait times  and/or sleep statements...'

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,54 @@
...
+instead you should create a loop and keep pooling until the operation doesn't complete.
+
+## Independent
+
+A test must always be independent of other tests, which mean:
+
+- it should not rely on other tests changes so that it would be possible to run it independently
```
ldimaggi: '...It should not rely on the successful execution of other tests...'

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,54 @@
...
+- [Don't Sleep](#dont-sleep)
+- [Independent](#independent)
+- [Secrets](#secrets)
+- [Logging](#logging)
+
+## Go Style
+
+We don't want to redefine best practices for go, therefore we should reuse the existing ones
+described in the [Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments)
+and in the[Effective Go](https://golang.org/doc/effective_go.html) documents.
```
damienomurchu:
```suggestion
and in the [Effective Go](https://golang.org/doc/effective_go.html) documents.
```

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,54 @@
...
+
+In order to compute test coverage and test automation progress across multiple test suites all
+automated tests must have an ID and must be traced back to the [integreatly-test-cases](https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases)
+repo following the [How to automate a test case and link it back](https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases#how-to-automate-a-test-case-and-link-it-back)
+tutorial.
+
+## Don't Sleep
+
+When waiting for a resource to become available or an operation to complete do not use fixed wait time,
+instead you should create a loop and keep pooling until the operation doesn't complete.
```
damienomurchu:
```suggestion
instead you should create a loop and check within that loop if the resource is available or operation has completed.
```

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Independent
+
+A test must always be independent of other tests, which mean:
```
damienomurchu:
```suggestion
A test must always be independent of other tests, which means:
```

test/BEST_PRACTICES.md
damienomurchu:
```suggestion
- it should not rely on the successful execution of other tests so the test can be run independently
```

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,54 @@
...
+
+- it should not rely on other tests changes so that it would be possible to run it independently
+
+- it should clean up after himself so that it would be possible to run it multiple time
```
damienomurchu:
```suggestion
- it should clean up after itself so it is possible to run the test concurrently or multiple times
```

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,54 @@
...
+
+- it should not rely on a static order when querying data because multiple tests may run in parallel
```
damienomurchu: I think we should reword this somehow so its a bit clearer

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,54 @@
...
+
+- it should not delete or modify other tests data because multiple tests may run in parallel
```
damienomurchu:
```suggestion
- it should not delete or modify resources used by other tests so tests may run in parallel
```

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,54 @@
...
+
+- it should not destruct any resource because other tests may rely on them
```
damienomurchu: I think this point is redundant as its covered by the previous point

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Secrets
+
+Do not commit or log any sensitive data or secrets, and always double check for it because it can happen unconsciously
```
damienomurchu:
```suggestion
Ensure secret or sensitive data is not included in commits or log output. 
```

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Logging
+
+Do not be afraid of logging, if the test pass we will ignore the logs but if the test fail logs will help
+understand why the test has failed, especially if the tests fail during an unexpected step.
```
damienomurchu:
```
Log statements should be used to add meaningful test execution information at regular points in the testcase. This is important so log output is useful and unnecessary additional, lengthy test runs to debug failures can be avoided. 
```

damienomurchu: @b1zzu I think we should mention the use of logging levels here too

test/BEST_PRACTICES.md
b1zzu: The problem is that the project is completely missing a Contribution or Best Practices guide, so I don't know which shared practices eng are using

test/BEST_PRACTICES.md
b1zzu: I tweaked it a little bit agai

test/BEST_PRACTICES.md
b1zzu: @ldimaggi tried to reword it wdyt?

test/BEST_PRACTICES.md
b1zzu: @damienomurchu I don't get why removing the `so that`, I would like to emphasize that there is a reason why we do it and a consequence if you don't do it and I think that using `so that` will strength the sentence. I would like just o understand why removing it?

test/BEST_PRACTICES.md
b1zzu: Using the standard log testing library for go is not possible to specify a log level, but instead of introducing a new library for the moment using the standard testing library is already an improvement over the fmt.Print because it will print also the timestamp and the line number

test/BEST_PRACTICES.md
b1zzu: This is actually more an example of what not to do, maybe we can just remove this

test/BEST_PRACTICES.md
damienomurchu:
```suggestion
Test code style to be consistent with that of integreatly-operator codebase [TODO: link to golang practices used by engineering]
```
perhaps something like the above?
tbh if there are no shared practices to link to, I would question the value of this section in the doc though.

b1zzu: Try to reworded again, wdty?

b1zzu: ping @damienomurchu

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,53 @@
+# Test Best Practices
+
+This is a collection of best practices to write high quality tests. The best practices
+can be referenced in review comments to speed up the review process.
```
ldimaggi: Suggested wording:
'...These best practices should be referenced in the comments that you provide when you perform code reviews of tests...'

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,53 @@
...
+
+- [Code Style](#code-style)
+- [Test Case Traceability](#test-case-traceability)
+- [Don't Sleep](#dont-sleep)
+- [Independent](#independent)
+- [Secrets](#secrets)
+- [Logging](#logging)
+
+## Code Style
```
ldimaggi: Suggested - change this to "Coding Style"

test/BEST_PRACTICES.md
```diff
@@ -0,0 +1,53 @@
...
+## Test Case Traceability
+
+In order to compute test coverage and test automation progress across multiple test suites all
+automated tests must have an ID and must be traced back to the [integreatly-test-cases](https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases)
+repo following the [How to automate a test case and link it back](https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases#how-to-automate-a-test-case-and-link-it-back)
+tutorial.
+
+## Don't Sleep
+
+When waiting for a resource to become available or an operation to complete do not use fixed wait times and/or sleep statements,
```
ldimaggi: Very samll point - remove the NEWLINE.   ;-)

test/BEST_PRACTICES.md
b1zzu: @ldimaggi if you want to make the suggestion even more effectively you can use the github suggestions so that I can accept them with one click, less work for me :)

test/BEST_PRACTICES.md
b1zzu: I removed the newlines from all paragraphs, is that right

test/BEST_PRACTICES.md
ldimaggi: Looks good. Were you able to locate the Eng Golang practices doc?

damienomurchu: I wasn't when I looked, but lets not let this block this PR. 
@b1zzu make the above change with the TODO, and we can update this doc later after its merged?

damienomurchu:
@b1zzu looks good to me - just make the change re `Go Style` and I will approve?

ldimaggi:
lgtm too !

damienomurchu:
Approved now, and should auto-merge once tests pass (unfortunately we need to wait for e2e tests to pass even though there are no code changes)

PR #505, opened 2020-03-11
remove sre users from rhmi group

briangallagher:
### What

The RHMI developer group is a catchall for developers on the cluster, it used to ensure developers have the right cluster perms and the right access to products. At the moment it is also adding SRE users, these users don't need to be set up in 3scale etc.

### How to test
Create a user in the following groups:
user1: dedicated-admins
user2: osd-sre-admins
Create a third user in no group. 
user3:

Run the RHMI-Operator and **log in with all 3 users**

#### Verify

1. Verify that user1 and user3 get added to the rhmi-developers group
2. Verify that user2 does not get added to the rhmi-developers group
3. Verify that user1 and user3 get added to the RHSSO OpenShift realm
4. Verify that user2 does not get added to the cluster RHSSO OpenShift realm
5. Verify that user1 and user3 get added as 3Scale users
6. Verify that user2 does not get added as a 3Scale users

### DIFF
--- a/pkg/controller/user/user_controller.go
+++ b/pkg/controller/user/user_controller.go
@@ -2,6 +2,7 @@ package user
 
 import (
 	"context"
+	userHelper "github.com/integr8ly/integreatly-operator/pkg/resources/user"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	usersv1 "github.com/openshift/api/user/v1"
@@ -19,10 +20,6 @@ import (
 var (
 	log = logf.Log.WithName("controller_user")
 	// A set of pre configured groups used to exclude a user from rhmi specific groups
-	exclusionGroups = []string{
-		"layered-cs-sre-admins",
-		"osd-sre-admins",
-	}
 )
 
 // Add creates a new User Controller and adds it to the Manager. The Manager will set fields on the Controller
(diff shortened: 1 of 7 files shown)


pkg/controller/user/user_controller.go
```diff
@@ -17,7 +16,14 @@ import (
...
 )
 
-var log = logf.Log.WithName("controller_user")
+var (
+	log = logf.Log.WithName("controller_user")
+	// A set of pre configured groups used to exclude a user from rhmi specific groups
+	exclusionGroups = []string{
+		"layered-cs-sre-admins",
+		"osd-sre-admins",
+	}
```
KevFan: Looks to be the `exclusionGroups` in common in `user_controller ` and `rhsso` reconciler, what do you think on moving this to `userHelper.go`?

pkg/controller/user/user_controller.go
```diff
@@ -83,11 +96,32 @@ func (r *ReconcileUser) Reconcile(request reconcile.Request) (reconcile.Result,
...
+				for _, groupUser := range group.Users {
+					if groupUser == user.Name {
+						return true
+					}
+				}
+			}
+		}
+	}
+	return false
+}
```
KevFan: Looks to be comon function in here and rhsso reconciler, what do you think on moving this to `userHelper.go`?

pkg/products/rhsso/reconciler.go
```diff
@@ -45,6 +45,10 @@ var (
 	idpAlias                = "openshift-v4"
 	githubIdpAlias          = "github"
 	manifestPackage         = "integreatly-rhsso"
+	exclusionGroups         = []string{
+		"layered-cs-sre-admins",
+		"osd-sre-admins",
+	}
```
KevFan: Same as above

pkg/products/rhsso/reconciler.go
```diff
@@ -767,13 +771,38 @@ func getUserDiff(keycloakUsers []keycloak.KeycloakAPIUser, openshiftUsers []user
...
+				for _, groupUser := range group.Users {
+					if groupUser == user.Name {
+						return true
+					}
+				}
+			}
+		}
+	}
+	return false
+}
```
KevFan: Same as above

briangallagher: Updated as suggested

pkg/controller/user/user_controller.go
```diff
@@ -17,7 +17,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-var log = logf.Log.WithName("controller_user")
+var (
+	log = logf.Log.WithName("controller_user")
+	// A set of pre configured groups used to exclude a user from rhmi specific groups
```
KevFan: Just leftover comment from moving the constants to `userHelper.go`. Otherwise looks good ! :+1:

PR #506, opened 2020-03-11
creation of grafana dashboards test

obrienrobert:
## Description

https://issues.redhat.com/browse/INTLY-5729

These changes automate the following test: https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases/blob/master/tests/dashboards/e02-verify-that-all-dashboards-are-installed-and-all-the-graphs.md

## Verification

- Login to an OS4 cluster with RHMI installed
- Using this branch, run the functional tests: `make test/functional`, the test should initially pass, as the expected dashboards should match the list of dashboards in Grafana.
- Modify the list of  `expectedDashboards` in the `dashboards_exits` file by changing the names of the some of the existing dashboards in the list, and also by adding new dashboard to the list. Ensure that the tests fails accordingly, and that the failure message is clear and concise.
- Login to Grafana as the admin user (login credentails are stored as a secret in the `middleware-monitoring` namespace) and add a new dashboard. Ensure the test detects the new dashboard, and that the test fails accordingly. Additionally, adding the title of the newly created dashboard to the list should allow the test to pass.

### DIFF
--- a/test/common/dashboards_exist.go
+++ b/test/common/dashboards_exist.go
@@ -1,0 +1,173 @@
+package common
+
+import (
+	"bytes"
+	goctx "context"
+	"encoding/json"
+	"fmt"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/remotecommand"
+)
+
+type dashboardsTestRule struct {
+	Title string `json:"title"`
+}
+
+var expectedDashboards = []dashboardsTestRule{
+	{
+		Title: "Endpoints Detailed",
+	},
+	{
+		Title: "Endpoints Report",
+	},
+	{
+		Title: "Endpoints Summary",
+	},
+	{
+		Title: "Keycloak",
+	},
+	{
+		Title: "Resource Usage By Namespace",
+	},
+	{
+		Title: "Resource Usage By Pod",
+	},
+	{
+		Title: "Resource Usage for Cluster",
+	},
+	{
+		Title: "Syndesis - Infra - API",
+	},
+	{
+		Title: "Syndesis - Infra - DB",
+	},
+	{
+		Title: "Syndesis - Infra - Home",
+	},
+	{
+		Title: "Syndesis - Infra - JVM",
+	},
+	{
+		Title: "Syndesis - Integrations - Camel",
+	},
+	{
+		Title: "Syndesis - Integrations - Home",
+	},
+	{
+		Title: "Syndesis - Integrations - JVM",
+	},
+	{
+		Title: "UnifiedPush Operator",
+	},
+	{
+		Title: "UnifiedPush Server",
+	},
+}
+
+func TestIntegreatlyDashboardsExist(t *testing.T, ctx *TestingContext) {
+	pods := &corev1.PodList{}
+	opts := []k8sclient.ListOption{
+		k8sclient.InNamespace(namespacePrefix + "middleware-monitoring-operator"),
+		k8sclient.MatchingLabels{"app": "grafana"},
+	}
+
+	err := ctx.Client.List(goctx.TODO(), pods, opts...)
+	if err != nil {
+		t.Fatal("failed to list pods", err)
+	}
+
+	if len(pods.Items) != 1 {
+		t.Fatal("grafana pod not found")
+	}
+
+	type Output []dashboardsTestRule
+
+	output, err := execToGrafanaPod("curl localhost:3000/api/search",
+		pods.Items[0].ObjectMeta.Name,
+		namespacePrefix+"middleware-monitoring-operator",
+		"grafana", ctx)
+	if err != nil {
+		t.Fatal("failed to exec to pod:", err)
+	}
+
+	var grafanaApiCallOutput Output
+	err = json.Unmarshal([]byte(output), &grafanaApiCallOutput)
+	if err != nil {
+		t.Logf("Failed to unmarshall json: %s", err)
+	}
+
+	if len(grafanaApiCallOutput) == 0 {
+		t.Fatal("grafana dashboard not found : %w", grafanaApiCallOutput)
+	}
+
+	ruleDiff := dashboardDifference(grafanaApiCallOutput, expectedDashboards)
+	if len(ruleDiff) > 0 {
+		t.Fatalf("Unexpected dashboards were found. If these dashboards were intentionally added, please update this test to include them. If these dashboards were not added intentionally or you are not sure, please create a JIRA and discuss with the monitoring team on how best to proceed. Additional dashboards: %s", strings.Join(ruleDiff, ", "))
+	}
+
+	ruleDiff = dashboardDifference(expectedDashboards, grafanaApiCallOutput)
+	if len(ruleDiff) > 0 {
+		t.Fatalf("Missing dashboards were found. If the removal of these dashboards was intentional, please update this test to remove them from the check. If the removal of these dashboards was not intended or you are not sure, please create a JIRA and discuss with the monitoring team on how best to proceed. Missing dashboards: %s", strings.Join(ruleDiff, ", "))
+	}
+}
+
+func execToGrafanaPod(command string, podName string, namespace string, container string, ctx *TestingContext) (string, error) {
+	req := ctx.KubeClient.CoreV1().RESTClient().Post().
+		Resource("pods").
+		Name(podName).
+		Namespace(namespace).
+		SubResource("exec").
+		Param("container", container)
+	scheme := runtime.NewScheme()
+	if err := corev1.AddToScheme(scheme); err != nil {
+		return "", fmt.Errorf("error adding to scheme: %v", err)
+	}
+	parameterCodec := runtime.NewParameterCodec(scheme)
+	req.VersionedParams(&corev1.PodExecOptions{
+		Container: container,
+		Command:   strings.Fields(command),
+		Stdin:     false,
+		Stdout:    true,
+		Stderr:    true,
+		TTY:       false,
+	}, parameterCodec)
+
+	exec, err := remotecommand.NewSPDYExecutor(ctx.KubeConfig, "POST", req.URL())
+	if err != nil {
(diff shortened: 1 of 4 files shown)


test/common/dashboards_exist.go
```diff
@@ -0,0 +1,173 @@
...
+		t.Fatalf("Unexpected dashboards were found. If these dashboards were intentionally added, please update this test to include them. If these dashboards were not added intentionally or you are not sure, please create a JIRA and discuss with the monitoring team on how best to proceed. Additional dashboards: %s", strings.Join(ruleDiff, ", "))
+	}
+
+	ruleDiff = dashboardDifference(expectedDashboards, grafanaApiCallOutput)
+	if len(ruleDiff) > 0 {
+		t.Fatalf("Missing dashboards were found. If the removal of these dashboards was intentional, please update this test to remove them from the check. If the removal of these dashboards was not intended or you are not sure, please create a JIRA and discuss with the monitoring team on how best to proceed. Missing dashboards: %s", strings.Join(ruleDiff, ", "))
+	}
+}
+
+func execToGrafanaPod(command string, podName string, namespace string, container string, ctx *TestingContext) (string, error) {
```
dimitraz: Is this function any different to the `execToPod` function from the other tests? If not, we can probably move this out to a common file

dimitraz: Is this function any different to the `execToPod` function from the other tests? If not, we can probably move this out to a common file

test/common/dashboards_exist.go
```diff
@@ -0,0 +1,173 @@
...
+		Tty:    false,
+	})
+	if err != nil {
+		return "", fmt.Errorf("error in Stream: %v", err)
+	}
+	return stdout.String(), nil
+}
+
+// dashboardDifference returns the elements in list of dashboardsTestRule `a` that aren't in list `b`.
+func dashboardDifference(a, b []dashboardsTestRule) []string {
```
dimitraz: Aiden pointed out that the variable names `a`, `b`, `x` etc might be a little too abstract and that we could probably use more meaningful names

test/common/dashboards_exist.go
```diff
@@ -0,0 +1,173 @@
...
+	err = json.Unmarshal([]byte(output), &grafanaApiCallOutput)
+	if err != nil {
+		t.Logf("Failed to unmarshall json: %s", err)
+	}
+
+	if len(grafanaApiCallOutput) == 0 {
+		t.Fatal("grafana dashboard not found : %w", grafanaApiCallOutput)
+	}
+
+	ruleDiff := dashboardDifference(grafanaApiCallOutput, expectedDashboards)
```
dimitraz: Probably could rename this to `dashboardDiff` or something like that

test/common/dashboards_exist.go
```diff
@@ -0,0 +1,173 @@
...
+	err := ctx.Client.List(goctx.TODO(), pods, opts...)
+	if err != nil {
+		t.Fatal("failed to list pods", err)
+	}
+
+	if len(pods.Items) != 1 {
+		t.Fatal("grafana pod not found")
+	}
+
+	type Output []dashboardsTestRule
```
dimitraz: Do you think having this `Output` type makes it a bit more clear to read? I'm not entirely sure about it, but obviously that's just a personal opinion

test/common/dashboards_exist.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	var grafanaApiCallOutput Output
+	err = json.Unmarshal([]byte(output), &grafanaApiCallOutput)
+	if err != nil {
+		t.Logf("Failed to unmarshall json: %s", err)
+	}
+
+	if len(grafanaApiCallOutput) == 0 {
+		t.Fatal("grafana dashboard not found : %w", grafanaApiCallOutput)
```
dimitraz: Could probably have a slightly more clear failure message here, something like `No grafana dashboards were found`?

test/common/dashboards_exist.go
```diff
@@ -0,0 +1,173 @@
...
+
+	ruleDiff := dashboardDifference(grafanaApiCallOutput, expectedDashboards)
+	if len(ruleDiff) > 0 {
+		t.Fatalf("Unexpected dashboards were found. If these dashboards were intentionally added, please update this test to include them. If these dashboards were not added intentionally or you are not sure, please create a JIRA and discuss with the monitoring team on how best to proceed. Additional dashboards: %s", strings.Join(ruleDiff, ", "))
+	}
+
+	ruleDiff = dashboardDifference(expectedDashboards, grafanaApiCallOutput)
+	if len(ruleDiff) > 0 {
+		t.Fatalf("Missing dashboards were found. If the removal of these dashboards was intentional, please update this test to remove them from the check. If the removal of these dashboards was not intended or you are not sure, please create a JIRA and discuss with the monitoring team on how best to proceed. Missing dashboards: %s", strings.Join(ruleDiff, ", "))
+	}
```
dimitraz: If there are any unexpected dashboards, the test fails before it gets to this point. 

I think it would be useful to output both lists before failing, so that the e2e test doesn't have to fail twice before all of the problems are addressed. I think that would give a better overview of what is failing and why

test/common/dashboards_exist.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if len(dashboardDiffUnexpected) > 0 || len(dashboardDiffMissing) > 0 {
+		t.Logf("unexpected dashboards found: %s", strings.Join(dashboardDiffUnexpected, ", "))
+		t.Logf("missing dashboards found: %s", strings.Join(dashboardDiffMissing, ", "))
+		t.Fail()
+	}
+}
+
+// dashboardDifference returns the elements in list of dashboardsTestRule `diffSource` that aren't in list `diffTarget`.
+func dashboardDifference(diffSource, diffTarget []dashboardsTestRule) []string {
```
jhellar: This function is almost identical to these other 2:

- https://github.com/integr8ly/integreatly-operator/blob/master/test/common/alerts_exist.go#L493
- https://github.com/integr8ly/integreatly-operator/blob/master/test/common/alerts_exist.go#L511

Can we make it more general and move it to `shared_functions` as well?

jhellar:
LGTM, just one comment

test/common/dashboards_exist.go
b1zzu: +1

b1zzu:
@obrienrobert the manual test is also verifying if the list of graphs in each dashboard and if there are data points. Thanks to this step when managed to catch 4 bugs during the manual testing:
https://issues.redhat.com/browse/INTLY-5953
https://issues.redhat.com/browse/INTLY-5952
https://issues.redhat.com/browse/INTLY-5499
https://issues.redhat.com/browse/INTLY-5501
therefore I think that they are important steps, any chance to implement them in the same PR? otherwise, we will need to create a new task

test/common/dashboards_exist.go
```diff
@@ -0,0 +1,119 @@
...
+		actualDashboardTitles = append(actualDashboardTitles, dashboard.Title)
+	}
+
+	dashboardDiffUnexpected := difference(actualDashboardTitles, expectedDashboardTitles)
+	dashboardDiffMissing := difference(expectedDashboardTitles, actualDashboardTitles)
+
+	if len(dashboardDiffUnexpected) > 0 || len(dashboardDiffMissing) > 0 {
+		t.Logf("unexpected dashboards found: %s", strings.Join(dashboardDiffUnexpected, ", "))
+		t.Logf("missing dashboards found: %s", strings.Join(dashboardDiffMissing, ", "))
+		t.Fail()
```
dimitraz: I know it's not as nice, but do you think we could do something like this here:
https://github.com/integr8ly/integreatly-operator/blob/60e36bee0f4e091e1ce4d1aa0b0d513aa05505d0/test/common/alerts_exist.go#L435-L443

I think printing out the statements regardless, even if one of those diff lists are empty, could be slightly misleading

obrienrobert: Good call. Makes sense.

obrienrobert:
> therefore I think that they are important steps, any chance to implement them in the same PR? otherwise, we will need to create a new task

@b1zzu @jhellar I discussed this with the team. We decided it would be better if this work was completed as part of a follow-on task, which resulted in https://issues.redhat.com/browse/INTLY-6234 being created. Please feel free to add any comments/suggestions you may have to the this JIRA.

b1zzu:
Thanks @obrienrobert

test/common/dashboards_exist.go
```diff
@@ -0,0 +1,125 @@
...
+	var actualDashboardTitles []string
+	for _, dashboard := range grafanaApiCallOutput {
+		actualDashboardTitles = append(actualDashboardTitles, dashboard.Title)
+	}
+
+	dashboardDiffUnexpected := difference(actualDashboardTitles, expectedDashboardTitles)
+	dashboardDiffMissing := difference(expectedDashboardTitles, actualDashboardTitles)
+
+	if len(dashboardDiffUnexpected) > 0 {
+		t.Logf("unexpected dashboards found: %s", strings.Join(dashboardDiffUnexpected, ", "))
```
b1zzu: Tip: using Errorf the test will be marked as failed but the execution will proceed so that it will also log the missing dashboards and there is no need for the last Fatal
```suggestion
		t.Errorf("unexpected dashboards found: %s", strings.Join(dashboardDiffUnexpected, ", "))
```

b1zzu: Even better you could loop over the unexpected dashboard and Errorf each one

test/common/dashboards_exist.go
```diff
@@ -0,0 +1,125 @@
...
+		namespacePrefix+"middleware-monitoring-operator",
+		"grafana", ctx)
+	if err != nil {
+		t.Fatal("failed to exec to pod:", err)
+	}
+
+	var grafanaApiCallOutput []dashboardsTestRule
+	err = json.Unmarshal([]byte(output), &grafanaApiCallOutput)
+	if err != nil {
+		t.Logf("failed to unmarshall json: %s", err)
```
b1zzu: I believe we should completely fail if this happened
```suggestion
		t.Fatalf("failed to unmarshall json: %s", err)
```

b1zzu:
Didn't know Prow would have merged it straight away :sweat_smile:

PR #507, opened 2020-03-11
[INTLY-5947] Update product and operator versions

dimitraz:
### Description
Some of the product and operator versions in `rhmi_types.go` contained inconsistent/incorrect product versions. These are the updated versions after talking to several engineering teams (syndesis, enmasse, strimzi). 

The operator versions seem to refer to the version in the CSV and in general not the actual operator release image, so I've left that as is.

### DIFF
--- a/pkg/apis/integreatly/v1alpha1/rhmi_types.go
+++ b/pkg/apis/integreatly/v1alpha1/rhmi_types.go
@@ -53,10 +53,10 @@ var (
 	// Could not find a way to determine these versions dynamically, so they are hard-coded
 	// It is preferable to determine the version of a product dynamically (from a CR, or configmap, etc)
 	// Follow up Jira: https://issues.redhat.com/browse/INTLY-5946
-	VersionAMQOnline           ProductVersion = "1.3.1"
+	VersionAMQOnline           ProductVersion = "1.3"
 	VersionApicurito           ProductVersion = "7.5"
 	VersionAMQStreams          ProductVersion = "1.1.0"
-	VersionCodeReadyWorkspaces ProductVersion = "2.0.0.GA"
+	VersionCodeReadyWorkspaces ProductVersion = "2.0.0"
 	VersionFuseOnOpenshift     ProductVersion = "master"
 	VersionMonitoring          ProductVersion = "1.1.4"
 	Version3Scale              ProductVersion = "2.7"
@@ -71,18 +71,19 @@ var (
 	PreflightSuccess    PreflightStatus = "successful"
 	PreflightFail       PreflightStatus = "failed"
 
+	// Operator image tags
 	OperatorVersionAMQStreams          OperatorVersion = "1.1.0"
-	OperatorVersionAMQOnline           OperatorVersion = "1.3.1"
+	OperatorVersionAMQOnline           OperatorVersion = "1.3"
 	OperatorVersionMonitoring          OperatorVersion = "1.1.4"
 	OperatorVersionSolutionExplorer    OperatorVersion = "0.0.45"
 	OperatorVersionRHSSO               OperatorVersion = "8.0.1"
 	OperatorVersionRHSSOUser           OperatorVersion = "8.0.1"
-	OperatorVersionCodeReadyWorkspaces OperatorVersion = "2.0.0"
-	OperatorVersionFuse                OperatorVersion = "7.5.0"
-	OperatorVersion3Scale              OperatorVersion = "0.4.0"
+	OperatorVersionCodeReadyWorkspaces OperatorVersion = "2.0"
+	OperatorVersionFuse                OperatorVersion = "1.5.0"
+	OperatorVersion3Scale              OperatorVersion = "1.10-4"
 	OperatorVersionCloudResources      OperatorVersion = "0.13.1"
 	OperatorVersionUPS                 OperatorVersion = "0.5.0"
-	OperatorVersionApicurito           OperatorVersion = "7.5.0"
+	OperatorVersionApicurito           OperatorVersion = "1.5.0"
 
 	// Event reasons to be used when emitting events
 	EventProcessingError       string = "ProcessingError"
--- a/pkg/apis/integreatly/v1alpha1/rhmi_types.go
+++ b/pkg/apis/integreatly/v1alpha1/rhmi_types.go
@@ -80,7 +80,7 @@ var (
 	OperatorVersionRHSSOUser           OperatorVersion = "8.0.1"
 	OperatorVersionCodeReadyWorkspaces OperatorVersion = "2.0"
 	OperatorVersionFuse                OperatorVersion = "1.5.0"
-	OperatorVersion3Scale              OperatorVersion = "1.10-4"
+	OperatorVersion3Scale              OperatorVersion = "0.4.0"
 	OperatorVersionCloudResources      OperatorVersion = "0.13.1"
 	OperatorVersionUPS                 OperatorVersion = "0.5.0"
 	OperatorVersionApicurito           OperatorVersion = "1.5.0"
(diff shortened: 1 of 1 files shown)


dimitraz:
cc @JameelB @philbrookes @KevFan @pb82

pkg/apis/integreatly/v1alpha1/rhmi_types.go
```diff
@@ -50,22 +50,22 @@ var (
...
+	// Follow up Jira: https://issues.redhat.com/browse/INTLY-5946
 	VersionAMQOnline           ProductVersion = "1.3.1"
-	VersionApicurito           ProductVersion = "1.0.1"
+	VersionApicurito           ProductVersion = "7.5"
 	VersionAMQStreams          ProductVersion = "1.1.0"
 	VersionCodeReadyWorkspaces ProductVersion = "2.0.0.GA"
 	VersionFuseOnOpenshift     ProductVersion = "master"
 	VersionMonitoring          ProductVersion = "1.1.4"
 	Version3Scale              ProductVersion = "2.7"
 	VersionUps                 ProductVersion = "2.3.2"
```
JameelB: Is this the correct product version of UPS that we install? I looked at the UPS console but I couldn't find info of what version it is from there. The manifest file we use for this only has has reference to `0.5.0` but not `2.3.2`.

maybe @secondsun @grdryn can help with this?

pkg/apis/integreatly/v1alpha1/rhmi_types.go
```diff
@@ -78,11 +78,11 @@ var (
...
 	OperatorVersionRHSSOUser           OperatorVersion = "8.0.1"
 	OperatorVersionCodeReadyWorkspaces OperatorVersion = "2.0.0"
-	OperatorVersionFuse                OperatorVersion = "1.5.0"
+	OperatorVersionFuse                OperatorVersion = "7.5.0"
 	OperatorVersion3Scale              OperatorVersion = "0.4.0"
-	OperatorVersionCloudResources      OperatorVersion = "0.13.0"
+	OperatorVersionCloudResources      OperatorVersion = "0.13.1"
 	OperatorVersionUPS                 OperatorVersion = "0.5.0"
-	OperatorVersionApicurito           OperatorVersion = "1.5.0"
+	OperatorVersionApicurito           OperatorVersion = "7.5.0"
```
JameelB: Same with Fuse. The product version is `7.5.0` but the actual image tag we use for the operator is `1.5.0`

pkg/apis/integreatly/v1alpha1/rhmi_types.go
grdryn: The version of the operator is 0.5.0, and the version of UPS itself is \<checks notes> 2.3.2: https://github.com/aerogear/unifiedpush-operator/blob/master/pkg/constants/constants.go#L4

grdryn: I presume that the integreatly-operator needs this to be able to display it somewhere or something?

JameelB: Yeah, it's displayed in the solution explorer :smiley: . Thanks for the confirmation @grdryn! :+1:

dimitraz: I think it's taken from here: https://github.com/aerogear/unifiedpush-operator/blob/f8b3254fed3ad63a8a2e8a75b5b2b45c09d5c8ab/pkg/constants/constants.go#L4 

Let's see if they can confirm if this is correct!

pkg/apis/integreatly/v1alpha1/rhmi_types.go
```diff
@@ -78,11 +78,11 @@ var (
 	OperatorVersionRHSSO               OperatorVersion = "8.0.1"
 	OperatorVersionRHSSOUser           OperatorVersion = "8.0.1"
 	OperatorVersionCodeReadyWorkspaces OperatorVersion = "2.0.0"
-	OperatorVersionFuse                OperatorVersion = "1.5.0"
+	OperatorVersionFuse                OperatorVersion = "7.5.0"
```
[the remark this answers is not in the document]
dimitraz: Yeah, I'm a bit confused by this. I changed it because all of the other operator versions seem to make reference to the CSV version and not the actual operator image. 

I'm not sure which one is supposed to be used? 

cc @philbrookes

pkg/apis/integreatly/v1alpha1/rhmi_types.go
```diff
@@ -50,22 +50,22 @@ var (
...
 
-	// PBrookes 08/08/2019:
 	// Could not find a way to determine these versions dynamically, so they are hard-coded
 	// It is preferable to determine the version of a product dynamically (from a CR, or configmap, etc)
+	// Follow up Jira: https://issues.redhat.com/browse/INTLY-5946
 	VersionAMQOnline           ProductVersion = "1.3.1"
-	VersionApicurito           ProductVersion = "1.0.1"
+	VersionApicurito           ProductVersion = "7.5"
 	VersionAMQStreams          ProductVersion = "1.1.0"
 	VersionCodeReadyWorkspaces ProductVersion = "2.0.0.GA"
```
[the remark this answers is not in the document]
dimitraz: Sure, I'll just double check this

pkg/apis/integreatly/v1alpha1/rhmi_types.go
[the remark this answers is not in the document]
JameelB: It makes sense for me that this version should be the actual operator version and the product version to be defined in `Version<product>`. 

I don't think we have consistency across all products yet whether this csv version refers to the operator/product version. At the moment some of them refer to the product version and some the operator version.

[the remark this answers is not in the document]
dimitraz: Ok, that makes sense, thanks Jameel

I'll change these now

pkg/apis/integreatly/v1alpha1/rhmi_types.go
```diff
@@ -71,18 +71,19 @@ var (
...
 	OperatorVersionMonitoring          OperatorVersion = "1.1.4"
 	OperatorVersionSolutionExplorer    OperatorVersion = "0.0.45"
 	OperatorVersionRHSSO               OperatorVersion = "8.0.1"
 	OperatorVersionRHSSOUser           OperatorVersion = "8.0.1"
-	OperatorVersionCodeReadyWorkspaces OperatorVersion = "2.0.0"
-	OperatorVersionFuse                OperatorVersion = "7.5.0"
-	OperatorVersion3Scale              OperatorVersion = "0.4.0"
+	OperatorVersionCodeReadyWorkspaces OperatorVersion = "2.0"
+	OperatorVersionFuse                OperatorVersion = "1.5.0"
+	OperatorVersion3Scale              OperatorVersion = "1.10-4"
```
JameelB: Seems like the actual 3Scale operator version is `0.4.0` based on their version [here](https://github.com/3scale/3scale-operator/blob/v0.4.0/version/version.go#L4) which doesn't match their image tag. I would be inclined to change this back to `0.4.0` since that is what the actual operator version is within their operator.

wdyt @dimitraz @philbrookes? pinging @laurafitzgerald for visibility

philbrookes: yes, let's use the actual operator version, rather than the image tag. Do we know why they are using 2 different versions like this?

dimitraz: Sounds good, thanks guys. I'll get in touch with the 3scale team to ask about this

dimitraz: Apparently there is no relation between the operator version and the operator image tag. I've updated the 3scale version now

PR #508, opened 2020-03-11
INTLY-5942 -  Backup & restore rhsso admin credentials

KevFan:
## Description
If RHSSO / User SSO namespace is deleted manially, the reconcilation fails due to different admin credentials are regenerated while the existing credentials in the DBs remain the same. To prevent this,  a generated password is bootstrapped and copied over to the RHSSO / User SSO for reconcilation  

Jira:
* https://issues.redhat.com/browse/INTLY-5942

## Verfication
* Install RHMI on this branch
* RHSSO:
  * Check the `redhat-rhmi-rhsso` and `redhat-rhmi-operator` namespace for the `credential-rhsso` secret and ensure they are the name
  * Manually delete the  `redhat-rhmi-rhsso`  namespace
    * Would need to manually finalizer in:
      * openshift KeycloakRealm CR
      * che-client KeyCloakClient CR
      * threescale-client KeyCloakClient CR
  * Verify the `redhat-rhmi-rhsso` namespace successfully reconciles
  * Verify the `credential-rhsso` secret in the  `redhat-rhmi-rhsso`  namespace is the same as thecredential-rhsso` secret in the `redhat-rhmi-operator` namespace
* User SSO:
  * Check the `credential-rhssouser` secret in the  `redhat-rhmi-user-sso` namespace and the `redhat-rhmi-operator` namespace and ensure they are the same
  * Manually delete the `redhat-rhmi-user-sso` namespace
    * Would need to manually finalizer in:
      * master KeycloakRealm CR
  * Verify the `redhat-rhmi-user-sso` namespace successfully reconciles
  * Verify the `credential-rhssouser` secret in the `redhat-rhmi-user-sso` namespace is the same as the `credential-rhssouser` secret in the `redhat-rhmi-operator` namespace


philbrookes:
@KevFan can you rebase this on master please?

KevFan:
> @KevFan can you rebase this on master please?

/hold

@philbrookes Will do. Could you hold off reviewing this? I'm going to change the implementation to match what @pb82  & @JameelB  is doing for 3scale.

They are copying the secrets from the 3scale namespace to the RHMI operator namespace when it finishes the initial install and this secret (if it exists in the RHMI operator namespace) will be copied over to the product namespace again after product namespace re-creation . This will remove the need of bootstrapping the secrets ourselves

KevFan:
/unhold

@philbrookes @davidffrench Feel free to have a look now whenever you get the chance

pkg/resources/pullSecret.go
```diff
@@ -54,3 +57,25 @@ func CopySecret(ctx context.Context, client k8sclient.Client, srcName, srcNamesp
...
+
+// Copies a secret from the RHMI operator namespace to a target namespace
+func ReconcileSecretToProductNamespace(ctx context.Context, client k8sclient.Client, configManager config.ConfigReadWriter, secretName string, namespace string) (integreatlyv1alpha1.StatusPhase, error) {
+	err := CopySecret(ctx, client, secretName, configManager.GetOperatorNamespace(), secretName, namespace)
+
+	if err != nil && !k8serr.IsNotFound(err) {
+		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to seed copy secret to product namespace: %w", err)
+	}
+
+	return integreatlyv1alpha1.PhaseCompleted, nil
```
davidffrench: @KevFan This will return complete even if the secret doesn't exist which it won't on the first try. I don't think it needs to be changed for this PR since this needs to get merged quickly. I would prefer to see a non-blocking failure  or even a log here in this case as it might be hard to debug an issue in the future if there is an issue caused because of this case.

pkg/products/rhsso/reconciler.go
```diff
@@ -48,9 +48,10 @@ var (
 )
 
 const (
-	SSOLabelKey   = "sso"
-	SSOLabelValue = "integreatly"
-	RHSSOProfile  = "RHSSO"
+	SSOLabelKey               = "sso"
+	SSOLabelValue             = "integreatly"
+	RHSSOProfile              = "RHSSO"
+	adminCredentialSecretName = "credential-rhsso"
```
davidffrench: This needs to be dynamic, based on `credential-` + keycloakName var from above

pkg/products/rhssouser/reconciler.go
```diff
@@ -61,6 +61,7 @@ const (
 	masterRealmClientName       = "master-realm"
 	firstBrokerLoginFlowAlias   = "first broker login"
 	reviewProfileExecutionAlias = "review profile config"
+	adminCredentialSecretName   = "credential-rhssouser"
```
davidffrench: This needs to be dynamic based on `credential-` + keycloakName var from above

pkg/products/rhsso/reconciler.go
KevFan: Good catch ! Will make it combination based on the `keyCloakName` var

pkg/products/rhssouser/reconciler.go
KevFan: Good catch ! Will make it combination based on the `keyCloakName` var

pkg/resources/pullSecret.go
KevFan: Yup, sure sounds good to me :+1:

KevFan:
@davidffrench I've addressed your comments now if you want to have a look again

PR #510, opened 2020-03-12
Intly 5437 - User Permission Functional Test

ciaranRoche:
## JIRA
https://issues.redhat.com/browse/INTLY-5437

## About
This change only tests the first two steps of functional test B02 for both `rhmi developer` user and a `dedicated admin` user

## Verification
Note you must be signed into and OSD cluster with a full RHMI install
- Run `make test/functional`
- Ensure tests are successful


test/common/user_permissions.go
```diff
@@ -0,0 +1,227 @@
...
+			// verify an rhmi developer can access the pods logs
+			if err := verifySuccessfullHTTPRequest(fmt.Sprintf("%s/%s/%s/log?%s", masterURL, pathFusePods, p.Name, lv.Encode()), rhmiDevUserToken); err != nil {
+				t.Fatalf("test-failed - rhmi devolper unable to access fuse logs at %s, error : %v", p.Name, err)
+			}
+		}
+	}
+}
+
+// verifies that there is at least 1 project with a prefix `openshift` , `redhat` and `kube`
+func verifyFoundProjects(projects []projectv1.Project) bool {
```
aidenkeating: These private functions are still going to be package visible in `common`, which may cause confusion are more tests are added. IMO we could either move these inline or make the function name more specific. e.g.

```
verifyDedicatedAdminProjectPermissions(projects []projectv1.Project) bool {
...
}
```

wdyt?

test/common/user_permissions.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+		if strings.HasPrefix(ns.Name, "kube") {
+			hasKubePrefix = true
+		}
+	}
+	return hasKubePrefix && hasRedhatPrefix && hasOpenshiftPrefix
+}
+
+// makes call to url, and checks if the resp is successful, returns error if request is not successful
+func verifySuccessfullHTTPRequest(url, token string) error {
```
aidenkeating: Typo in the function name here `Successful`. Again I think we may need to make this function name more descriptive. As it's for an OpenShift request (or at least sets a cookie for OpenShift based on the provided token) and only performs a GET. e.g.

```
func doOpenshiftGetRequest(url, token string) error {
...
}
```

We could probably also consider using `ocRequest` inside this function, right? Or maybe removing this altogether and doing the 200 check inline as it's only used in one place currently [1]?

[1] https://github.com/integr8ly/integreatly-operator/pull/510/files#diff-036b3912c851d760c7f05430eee9e3d8R127

test/common/user_permissions.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	req.Header.Set("Cookie", fmt.Sprintf("%s=%s", resources.OpenshiftTokenName, token))
+	// handle request
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("error occured while making http request : %w", err)
+	}
+	// check response
+	if resp.StatusCode != 200 {
+		return errors.New("user unable to connect to fuse logs")
```
aidenkeating: Is this log message correct?

test/common/user_permissions.go
```diff
@@ -0,0 +1,227 @@
...
+
+	// check if six pods are running
+	runningCount := 0
+	for _, p := range podlist.Items {
+		if p.Status.Phase == "Running" {
+			runningCount++
+		}
+	}
+	if runningCount != expectedFusePodCount {
+		t.Fatalf("test-failed - %d found fuse pod count not match expected count of %d", runningCount, expectedFusePodCount)
```
obrienrobert: Could this error message be slightly clearer? Something like: `expected fuse pod count: <count>, found fuse pod count: <count>`

test/common/user_permissions.go
```diff
@@ -0,0 +1,227 @@
...
+	"net/http"
+	"testing"
+)
+
+const (
+	openshiftAuthenticationNamespace = "openshift-authentication"
+	openshiftOAuthRouteName          = "oauth-openshift"
+
+	pathProjects = "api/kubernetes/apis/project.openshift.io/v1/projects"
+	pathFusePods = "api/kubernetes/api/v1/namespaces/redhat-rhmi-fuse/pods"
```
obrienrobert: Does this hard-coded path need to be updated to use the `namespacePrefix` value from the installation CR?

obrienrobert:
When I run `make test/functional`, it fails with the following message: 
```
user_permissions.go:114: test-failed - 0 found fuse pod count not match expected count of 6
```
I've verified that there are 6 running pods in the `redhat-rhmi-fuse` namespace. I'm running the test's against an OSD DEV cluster, with RHMI installed from master.

test/resources/user_auth.go
```diff
@@ -0,0 +1,196 @@
...
+	jar, err := cookiejar.New(&cookiejar.Options{PublicSuffixList: publicsuffix.List})
+	if err != nil {
+		return "", fmt.Errorf("error occured creating a new cookie jar : %w", err)
+	}
+
+	// declare http client
+	client := &http.Client{
+		Transport: tr,
+		Jar:       jar,
+	}
```
JameelB: Since more than one test may need an HTTP client, maybe the main test can initialise this instead and be passed to the common tests by context? This would make it easier to reuse. wdyt @ciaranRoche?

test/common/user_permissions.go
```diff
@@ -0,0 +1,227 @@
...
+	TailLines string `url:"tailLines"`
+}
+
+func TestIntegreatlyUserPermissions(t *testing.T, ctx *TestingContext) {
+	// get the RHMI custom resource to check what storage type is being used
+	rhmi := &v1alpha1.RHMI{}
+	ns := fmt.Sprintf("%soperator", namespacePrefix)
+	if err := ctx.Client.Get(goctx.TODO(), types.NamespacedName{Name: installationName, Namespace: ns}, rhmi); err != nil {
+		t.Fatal("error getting RHMI CR:", err)
+	}
```
b1zzu: There is a shared function to retrieve this info:
https://github.com/integr8ly/integreatly-operator/blob/edcbe34d3b2b94b93573d5236b85e43e69670697/test/common/shared_functions.go#L75

test/common/user_permissions.go
```diff
@@ -0,0 +1,227 @@
...
+		if projectCount >= expectedRhmiDeveloperProjectCount {
+			t.Fatal(fmt.Sprintf("test failed - project count for rhmi-developer exceeded expected 1"))
+		}
+		if p.Name != expectedRhmiDeveloperNamespace {
+			t.Fatal(fmt.Sprintf("test failed - found project for rhmi-developer does not match expected : %s", expectedRhmiDeveloperNamespace))
+		}
+	}
+	t.Log("test-passed - found projects for rhmi-developer are as expected")
+
+	// get dedicated admin token
```
b1zzu: Wouldn't it be better to split the developer and dedicated admin test permissions in two different tests? They are one manual tests just to simplify the manual execution, but I would prefer to have two automated tests because they are not dependent from each other

test/resources/openshift_api.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"io/ioutil"
+	"net/http"
+)
+
+// returns all pods in a namesapce
+func DoOpenshiftGetNamespacePods(masterUrl, path, token string) (*corev1.PodList, error) {
+	resp, err := DoOpenshiftGetRequest(masterUrl, path, token)
+	if err != nil {
+		return nil, fmt.Errorf("error occured durning oc request : %w", err)
```
dimitraz:
```suggestion
		return nil, fmt.Errorf("error occured performing oc get request : %w", err)
```

test/resources/openshift_api.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	return foundPods, nil
+}
+
+// returns all projects
+func DoOpenshiftGetProjects(masterUrl, token string) (*projectv1.ProjectList, error) {
+	resp, err := DoOpenshiftGetRequest(masterUrl, PathProjects, token)
+	if err != nil {
+		return nil, fmt.Errorf("error occured durning oc request : %w", err)
```
dimitraz: Same as above

test/resources/openshift_api.go
```diff
@@ -0,0 +1,63 @@
...
+
+// makes a get request, expects master url, a path and a token
+func DoOpenshiftGetRequest(masterUrl string, path string, token string) (*http.Response, error) {
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+	}
+	client := &http.Client{Transport: tr}
+	req, err := http.NewRequest("GET", fmt.Sprintf("https://%s%s", masterUrl, path), nil)
+	if err != nil {
+		return nil, fmt.Errorf("error occured while creating new requests : %w", err)
```
dimitraz:
```suggestion
		return nil, fmt.Errorf("error occurred while creating new http request: %w", err)
```

test/resources/openshift_auth.go
```diff
@@ -0,0 +1,201 @@
...
+	callbackUrl := "https%3A%2F%2F" + masterUrl + "%2Fauth%2Fcallback&" + cv.Encode()
+	// create request url, not this url needed to be created in this format to avoid any unwanted escape character
+	reqUrl := "https://" + oauthUrl + "/oauth/authorize?" + lv.Encode() + "&redirect_uri=" + callbackUrl
+	req, err := http.NewRequest("GET", reqUrl, nil)
+	if err != nil {
+		return "", fmt.Errorf("error occured while creating a new request, %w", err)
+	}
+
+	// perform request, handling redirects
+	resp, err := client.Do(req)
```
dimitraz: Should you be closing `resp.Body` at some point?

test/resources/openshift_api.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// returns all projects
+func DoOpenshiftGetProjects(masterUrl, token string) (*projectv1.ProjectList, error) {
+	resp, err := DoOpenshiftGetRequest(masterUrl, PathProjects, token)
+	if err != nil {
+		return nil, fmt.Errorf("error occured durning oc request : %w", err)
+	}
+
+	respBody, err := ioutil.ReadAll(resp.Body)
```
dimitraz: Need to close `resp.Body`

test/resources/openshift_api.go
```diff
@@ -0,0 +1,63 @@
...
+	"net/http"
+)
+
+// returns all pods in a namesapce
+func DoOpenshiftGetNamespacePods(masterUrl, path, token string) (*corev1.PodList, error) {
+	resp, err := DoOpenshiftGetRequest(masterUrl, path, token)
+	if err != nil {
+		return nil, fmt.Errorf("error occured durning oc request : %w", err)
+	}
+	respBody, err := ioutil.ReadAll(resp.Body)
```
dimitraz: Need to close `resp.Body`

test/common/user_rhmi_developer_permissions.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// get projects for rhmi developer
+	rhmiDevfoundProjects, err := resources.DoOpenshiftGetProjects(masterURL, rhmiDevUserToken)
+	if err != nil {
+		t.Fatalf("error occured while getting user projects : %v", err)
+	}
+
+	// check if projects are as expected for rhmi developer
+	for projectCount, p := range rhmiDevfoundProjects.Items {
+		t.Log(fmt.Sprintf("found rhmi-developer project - %s", p.Name))
```
dimitraz: Do we need this log statement? If so, could use `t.Logf()`

test/common/user_rhmi_developer_permissions.go
```diff
@@ -0,0 +1,102 @@
...
+		if p.Status.Phase == "Running" {
+			logOpt := LogOptions{p.Spec.Containers[0].Name, "false", "10"}
+			lv, err := query.Values(logOpt)
+			if err != nil {
+				t.Fatal(err)
+			}
+			// verify an rhmi developer can access the pods logs
+			resp, err := resources.DoOpenshiftGetRequest(fmt.Sprintf("%s/%s/%s/log?%s", masterURL, resources.PathFusePods, p.Name, lv.Encode()), "", rhmiDevUserToken)
+			if err != nil {
+				t.Fatalf("error occured while making Openshift request: %v", err)
```
dimitraz:
```suggestion
				t.Fatalf("error occurred making oc get request: %v", err)
```

test/common/user_rhmi_developer_permissions.go
```diff
@@ -0,0 +1,102 @@
...
+			}
+			if resp.StatusCode != 200 {
+				t.Fatalf("test-failed - rhmi devolper unable to access fuse logs at %s, error : %v", p.Name, err)
```
dimitraz:
```suggestion
				t.Fatalf("RHMI developer unable to access fuse logs in pod %s: %v", p.Name, err)
```

dimitraz: Could be worth logging the exact status code as well maybe?

test/resources/openshift_api.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+	projectv1 "github.com/openshift/api/project/v1"
+	corev1 "k8s.io/api/core/v1"
+
+	"io/ioutil"
+	"net/http"
+)
+
+// returns all pods in a namesapce
+func DoOpenshiftGetNamespacePods(masterUrl, path, token string) (*corev1.PodList, error) {
```
dimitraz: Not huge but `DoOpenshiftGetPodsForNamespace` seems a bit more clear to me

test/resources/openshift_api.go
```diff
@@ -0,0 +1,63 @@
...
+	resp, err := DoOpenshiftGetRequest(masterUrl, path, token)
```
dimitraz: We could probably build the path endpoint dynamically, which means we'd only have to pass in the namespace to this function:
```go
path := fmt.Sprintf("/api/kubernetes/api/v1/namespaces/%s/pods", namespace)
```
Might be slightly more readable, instead of having to somewhat duplicate the api endpoint every time?

test/resources/openshift_auth.go
```diff
@@ -0,0 +1,201 @@
...
+func DoAuthOpenshiftUser(oauthUrl string, masterURL string, idp string, username string, password string) (string, error) {
+	// declare transport
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+	}
+
+	// declare new cookie jar om nom nom
+	jar, err := cookiejar.New(&cookiejar.Options{PublicSuffixList: publicsuffix.List})
+	if err != nil {
+		return "", fmt.Errorf("error occured creating a new cookie jar : %w", err)
```
dimitraz:
```suggestion
		return "", fmt.Errorf("error occurred creating a new cookie jar: %w", err)
```

test/common/user_rhmi_developer_permissions.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// check if projects are as expected for rhmi developer
+	for projectCount, p := range rhmiDevfoundProjects.Items {
+		if projectCount >= expectedRhmiDeveloperProjectCount {
+			t.Fatalf("test failed - found rhmi developer project count : %s expected rhmi-developer project count : %s", projectCount, expectedRhmiDeveloperProjectCount)
+		}
+		if p.Name != expectedRhmiDeveloperNamespace {
+			t.Fatalf("test failed - found rhmi developer project: %s expected rhmi developer project : %s", p.Name, expectedRhmiDeveloperNamespace)
+		}
+	}
```
jhellar: Could this be simplified - just two asserts (no `for` loop):

- projectCount is as expected
- name is as expected

test/resources/openshift_auth.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	// perform request, handling redirects
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("error occured at client do: %w", err)
+	}
+
+	// parse response body
+	doc, err := html.Parse(resp.Body)
```
b1zzu: Instead of using the html parser we can introduce the **goquery**, it will simplify the parsing of the html, I'm already using it successfully to authenticate AMQ 
https://github.com/integr8ly/integreatly-operator/blob/1abb8e717b6e19a69e2616d9b6cd025dc362379c/test/common/amq_address_and_address_space.go#L53

test/common/tests.go
```diff
@@ -7,7 +7,15 @@ var (
 		{"Verify CRD Exists", TestIntegreatlyCRDExists},
 	}
 
+	AFTER_INSTALL_PREREQUISITES = []TestCase{
+		// Prerequisite checks, to be executed before tests
+		{"Setup Testing IDP", SetupTestingIDP},
```
b1zzu: Not sure about this approach, doing so if the setup of the testing idp fails, then none of the other tests will be executed even if not all tests depend on the testing idp, this is breaking the principle of having [Independent](https://github.com/integr8ly/integreatly-operator/blob/master/test/BEST_PRACTICES.md#independent) tests

test/common/testing_idp.go
```diff
@@ -0,0 +1,583 @@
...
+	tlsCrt := routerSecret.Data["tls.crt"]
+
+	idpConfigMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("idp-ca-%s", testingIDPRealm),
+			Namespace: "openshift-config",
+		},
+	}
+	if _, err := controllerutil.CreateOrUpdate(context.TODO(), ctx.Client, idpConfigMap, func() error {
+		if idpConfigMap == nil {
```
JameelB: this should be `idpConfigMap.Data == nil`

test/common/tests.go
ciaranRoche: makes sense, good catch :+1:

aidenkeating:
/lgtm
/approve

The e2e tests have been run as part of the Prow tests and passed.
Functional test suite was run locally against an OSD cluster.

```
$ make test/functional
go clean -testcache && go test ./test/functional
ok  	github.com/integr8ly/integreatly-operator/test/functional	102.119s
```

I can see the IDP has been set up in the `cluster` `OAuth` resource, I can also see the correct users have been given the correct groups.

```
$ oc get groups
NAME                         USERS
dedicated-admins             customer-admin-0, customer-admin-1
rhmi-developers              customer-admin-1, test-user-1 # only customer-admin-1 and test-user-1 login in the existing functional tests, so only these two are expected here
```

PR #511, opened 2020-03-12
backup and restore 3scale secrets for namespace restoration

pb82:
Backup the 3scale system seed and master apicast secrets into the rhmi-operator namespace and restore them from there (in case the namespace gets removed).

### DIFF
--- a/pkg/products/threescale/reconciler.go
+++ b/pkg/products/threescale/reconciler.go
@@ -58,6 +58,9 @@ const (
 	externalPostgresSecretName     = "system-database"
 
 	numberOfReplicas int64 = 2
+
+	systemSeedSecretName          = "system-seed"
+	systemMasterApiCastSecretName = "system-master-apicast"
 )
 
 func NewReconciler(configManager config.ConfigReadWriter, installation *integreatlyv1alpha1.RHMI, appsv1Client appsv1Client.AppsV1Interface, oauthv1Client oauthClient.OauthV1Interface, tsClient ThreeScaleInterface, mpm marketplace.MarketplaceInterface, recorder record.EventRecorder) (*Reconciler, error) {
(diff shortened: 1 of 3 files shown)


KevFan:
@pb82 Phil has made this `CopySecret` function should remove a lot of the logic if you want to have a look it
https://github.com/integr8ly/integreatly-operator/blob/534db38b1aaa06f462266f76cdf142d8ca16c7db/pkg/resources/pullSecret.go#L34

pkg/products/threescale/reconciler.go
```diff
@@ -261,7 +276,33 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+
+	return integreatlyv1alpha1.PhaseCompleted, nil
+}
+
+// Copies the seed and master api cast secrets for later restoration
+func (r *Reconciler) backupSystemSecrets(ctx context.Context, serverClient k8sclient.Client, installation *integreatlyv1alpha1.RHMI) (integreatlyv1alpha1.StatusPhase, error) {
+	for _, secretName := range []string{systemSeedSecretName, systemMasterApiCastSecretName} {
+		err := resources.CopySecret(ctx, serverClient, secretName, r.Config.GetNamespace(), secretName, installation.Namespace)
+		if err != nil {
+			return integreatlyv1alpha1.PhaseFailed, err
```
philbrookes: is failed the best state here? The secret could already exist and the error be really quite trivial, right?

pkg/products/threescale/reconciler.go
```diff
@@ -261,7 +276,33 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
 }
 
-func (r *Reconciler) reconcileTemplates(ctx context.Context, installation *integreatlyv1alpha1.RHMI, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
+// restores seed and master api cast secrets if available
+func (r *Reconciler) restoreSystemSecrets(ctx context.Context, serverClient k8sclient.Client, installation *integreatlyv1alpha1.RHMI) (integreatlyv1alpha1.StatusPhase, error) {
+	for _, secretName := range []string{systemSeedSecretName, systemMasterApiCastSecretName} {
+		err := resources.CopySecret(ctx, serverClient, secretName, installation.Namespace, secretName, r.Config.GetNamespace())
+		if err != nil {
+			if !k8serr.IsNotFound(err) {
+				return integreatlyv1alpha1.PhaseFailed, err
```
philbrookes: is failed the best state here? The secret could already exist and the error be really quite trivial, right?

pb82: Good question. If the secret already exists in 3scale then it would get updated. `PhaseFailed` here means that the integreatly operator can't reconcile (restore) the 3scale secrets. But we should probably also ignore conflicts because the 3scale operator might also update those secrets.

JameelB: I think it's still best to be able to see those conflict errors rather than ignore them because then we won't know that the restore has actually failed. 

If we get any errors back from `CopySecret`, the secret wouldn't have been updated with the original credentials right? I think we should capture those errors as an event.

philbrookes: ok, I'm happy for this to merge.

PR #514, opened 2020-03-13
Add cloud-resources-0.13.2 manifest

grdryn:
JIRA: https://issues.redhat.com/browse/INTLY-5819

----

#


philbrookes:
as we have no releases yet, we don't need to duplicate the manifests and set a replaces field, instead just delete the old one and add the new one without a replaces field.

PR #515, opened 2020-03-13
INTLY-6011 Improve expected replicas tests

b1zzu:
Resolves: https://issues.redhat.com/browse/INTLY-6011

**What**
* Verify that the expected numbers of replicas are also available (up and running)
* Add TestStatefulSetsExpectedReplicas to the test list
* Move resources lists to global variables
* Report all errors before failing

### DIFF
--- a/test/common/deployment_types.go
+++ b/test/common/deployment_types.go
@@ -20,160 +20,8 @@ type Product struct {
 	ExpectedReplicas int32
 }
 
-func TestDeploymentExpectedReplicas(t *testing.T, ctx *TestingContext) {
-	deployment := getDeployment()
-
-	isClusterStorage, err := isClusterStorage(ctx)
-	if err != nil {
-		t.Fatal("error getting isClusterStorage:", err)
-	}
-	// If the cluster is using in cluster storage instead of AWS resources
-	// These deployments will also need to be checked
-	if isClusterStorage {
-		clusterStorageDeployments := getClusterStorageDeployments()
-		for _, clusterStorageDeployment := range clusterStorageDeployments {
-			deployment = append(deployment, clusterStorageDeployment)
-		}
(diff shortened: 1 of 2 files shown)


test/common/deployment_types.go
```diff
@@ -264,22 +112,202 @@ func getDeployment() []Namespace {
...
-				Product{Name: "rhsso-postgres-rhmi", ExpectedReplicas: 1},
-				Product{Name: "rhssouser-postgres-rhmi", ExpectedReplicas: 1},
+				Product{Name: "codeready-postgres-" + InstallationName, ExpectedReplicas: 1},
+				Product{Name: "threescale-backend-redis-" + InstallationName, ExpectedReplicas: 1},
+				Product{Name: "threescale-postgres-" + InstallationName, ExpectedReplicas: 1},
+				Product{Name: "threescale-redis-" + InstallationName, ExpectedReplicas: 1},
+				Product{Name: "ups-postgres-" + InstallationName, ExpectedReplicas: 1},
+				Product{Name: "rhsso-postgres-" + InstallationName, ExpectedReplicas: 1},
+				Product{Name: "rhssouser-postgres-" + InstallationName, ExpectedReplicas: 1},
 				Product{Name: "standard-authservice-postgresql", ExpectedReplicas: 1},
```
KevFan: These looks to be the same as the CRO prefix contants that I externalised for the CRO tests, not sure if you want to use them instead :thinking: 
https://github.com/integr8ly/integreatly-operator/blob/cf24ffb9ea80c1f201e0965beed10e4ee2e6e9a2/pkg/resources/constants/croConstants.go#L4-L11

b1zzu: Thanks @KevFan, yes we can use the constants

PR #521, opened 2020-03-16
bump cro version to 0.13.3

aidenkeating:


aidenkeating:
Can confirm that the monitoring cluster came up successfully with the correct image version for CRO.

Can also confirm the changes included in v0.13.3 of CRO are available.

- In RDS, including tags in final snapshots

Screenshot 2020-03-16 at 14 08 32
Screenshot 2020-03-16 at 14 00 41

- In Elasticache, including tags in final snapshots

Screenshot 2020-03-16 at 14 24 25

Cluster deletion also worked successfully.

PR #525, opened 2020-03-17
fix: no empty strings in address space resources

pb82:
AMQ Online does not allow empty strings in an address plan resouce. Router or Broker can be either left out be defined. The spec however does not include `omitempty` which means even empty strings will be serialized to yaml. Instead of modifying the spec, this PR sets the previously undefined strings to `"0.0"`. This is the equivalent of not having them in the yaml.

Verification steps:

1. Create an address space in the AMQ Online console
2. Create one address of each type, standard and brokered
3. Make sure that all addresses are created successfully

### DIFF
--- a/pkg/products/amqonline/addressplans.go
+++ b/pkg/products/amqonline/addressplans.go
@@ -20,6 +20,7 @@ func GetDefaultAddressPlans(ns string) []*v1beta2.AddressPlan {
 				AddressType:      "topic",
 				Resources: v1beta2.AddressPlanResources{
 					Broker: "0.0",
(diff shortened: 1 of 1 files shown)


davidffrench:
@pb82 The code changes look good to me. Do you have a cluster I can look at the created address plans on?

davidffrench:
Verified I can create all modified addresses from this PR on Peters cluster
- Small, medium, large anycast addresses in standard addressspace
- Small, medium, large multicast addresses in standard addressspace
- Queue and topic addresses in brokered addressspace


/lgtm
/approve

PR #526, opened 2020-03-18
INTLY-5436: Tests to verify that all routes have been created

sergioifg94:
Create and add test to the `AFTER_INSTALL_TESTS` list that enumerates the expected routes
that must have been created and checks that they exist in the cluster.

Remove test logic that becomes redundant after adding this test.

### DIFF
--- a/test/common/routes_exist.go
+++ b/test/common/routes_exist.go
@@ -1,0 +1,228 @@
+package common
+
+import (
+	"fmt"
+	"testing"
+
+	goctx "context"
+
+	routev1 "github.com/openshift/api/route/v1"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var expectedRoutes = map[string][]ExpectedRoute{
+	"3scale": []ExpectedRoute{
+		ExpectedRoute{
+			Name:  "backend",
+			isTLS: true,
+		},
+		ExpectedRoute{
+			Name:            "zync-3scale-api-",
+			IsGeneratedName: true,
+			isTLS:           true,
+			ServiceName:     "apicast-staging",
+		},
+		ExpectedRoute{
+			Name:            "zync-3scale-api-",
+			IsGeneratedName: true,
+			isTLS:           true,
+			ServiceName:     "apicast-production",
+		},
+		ExpectedRoute{
+			Name:            "zync-3scale-master-",
+			IsGeneratedName: true,
+			isTLS:           true,
+			ServiceName:     "system-master",
+		},
+		ExpectedRoute{
+			Name:            "zync-3scale-provider-",
+			IsGeneratedName: true,
+			isTLS:           true,
+			ServiceName:     "system-developer",
+		},
+		ExpectedRoute{
+			Name:            "zync-3scale-provider-",
+			IsGeneratedName: true,
+			isTLS:           true,
+			ServiceName:     "system-provider",
+		},
+	},
+
+	"amq-online": []ExpectedRoute{
+		ExpectedRoute{
+			Name:  "console",
+			isTLS: true,
+		},
+		ExpectedRoute{
+			Name:  "standard-authservice",
+			isTLS: true,
+		},
+	},
+
+	"codeready-workspaces": []ExpectedRoute{
+		ExpectedRoute{
+			Name:  "codeready",
+			isTLS: true,
+		},
+		ExpectedRoute{
+			Name:  "devfile-registry",
+			isTLS: true,
+		},
+		ExpectedRoute{
+			Name:  "plugin-registry",
+			isTLS: true,
+		},
+	},
+
+	"fuse": []ExpectedRoute{
+		ExpectedRoute{
+			Name:  "syndesis",
+			isTLS: true,
+		},
+	},
+
+	"middleware-monitoring-operator": []ExpectedRoute{
+		ExpectedRoute{
+			Name:  "alertmanager-route",
+			isTLS: true,
+		},
+		ExpectedRoute{
+			Name:  "grafana-route",
+			isTLS: true,
+		},
+		ExpectedRoute{
+			Name:  "prometheus-route",
+			isTLS: true,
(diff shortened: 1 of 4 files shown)


test/common/routes_exist.go
```diff
@@ -0,0 +1,224 @@
...
+// TestIntegreatlyRoutesExist tests that the routes for all the products are created
+func TestIntegreatlyRoutesExist(t *testing.T, ctx *TestingContext) {
+	var err error
+
+	for product, routes := range expectedRoutes {
+		for _, expectedRoute := range routes {
+			err = checkRoute(t, ctx, product, expectedRoute)
+
+			if err != nil {
+				t.Fatalf("Failed checking route %v for product %v: %v",
```
b1zzu: I will suggest using `t.Errorf`, the error will be logged, the test marked as failed but the execution of the test will proceed and we could catch other problems 
```suggestion
				t.Errorf("Failed checking route %v for product %v: %v",
```

sergioifg94: Makes sense, I'll change that in a moment. Thanks @b1zzu !

b1zzu:
:eyes: I'm trying it against BYOC

b1zzu:
Verified also against BYOC

/lgtm

test/common/routes_exist.go
```diff
@@ -0,0 +1,224 @@
...
+	for _, route := range routes.Items {
+		// Skip routes that don't point to services
+		if route.Spec.To.Kind != "Service" {
+			continue
+		}
+
+		generatedName := route.GetObjectMeta().GetGenerateName()
+		to := route.Spec.To.Name
+
+		if expectedRoute.Name == generatedName && expectedRoute.ServiceName == to {
```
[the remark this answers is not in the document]
sergioifg94: I added the logic to check for the service in order to avoid false positives:
If `expectedRoute.Name` is "keycloak" and there's a route called "keycloak-edge" the `strings.HasPrefix` call would be evaluated to true.
Also there's the case in 3scale:
```go
expectedRoute{
	Name:            "zync-3scale-api-",
	IsGeneratedName: true,
	isTLS:           true,
	ServiceName:     "apicast-staging",
},
expectedRoute{
	Name:            "zync-3scale-api-",
       IsGeneratedName: true,
	isTLS:           true,
	ServiceName:     "apicast-production",
},
```
Where they have the same generated name so the test would pass even if there's only one of these routes available

sergioifg94:
@dimitraz Thanks a lot for the suggestions! I added a fixup to the commit with the changes

test/common/routes_exist.go
```diff
@@ -0,0 +1,220 @@
...
+	routev1 "github.com/openshift/api/route/v1"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type expectedRoute struct {
+	Name            string
+	isTLS           bool
+	ServiceName     string
+	IsGeneratedName bool
+}
```
KevFan: Looks like people are starting to put all the types in `types.go` so just would be good for consistency :thinking: 
https://github.com/integr8ly/integreatly-operator/blob/master/test/common/types.go

test/common/routes_exist.go
```diff
@@ -0,0 +1,220 @@
...
+		expectedRoute{
+			Name:  "keycloak-edge",
+			isTLS: true,
+		},
+	},
+}
+
+const (
+	intlyNamespacePrefix = "redhat-rhmi-"
+)
```
KevFan: This constant looks to be already defined in `types.go` so just would be good to prevent duplicates :thinking: 
https://github.com/integr8ly/integreatly-operator/blob/8ff8ac219da0428dfbde872de8adbe104c39059d/test/common/types.go#L15

sergioifg94:
@KevFan Thanks for the suggestions! I hadn't noticed that `types.go` file. I updated the PR with these changes

test/common/routes_exist.go
```diff
@@ -0,0 +1,209 @@
...
+
+// TestIntegreatlyRoutesExist tests that the routes for all the products are created
+func TestIntegreatlyRoutesExist(t *testing.T, ctx *TestingContext) {
+	for product, routes := range expectedRoutes {
+		for _, expectedRoute := range routes {
+			foundRoute, err := getRoute(t, ctx, product, expectedRoute)
+
+			if err != nil {
+				t.Errorf("Failed checking route %v for product %v: %v",
+					expectedRoute.Name, product, err)
```
b1zzu:
```suggestion
				t.Errorf("Failed checking route %v for product %v: %v",
					expectedRoute.Name, product, err)
				continue
```

test/common/routes_exist.go
```diff
@@ -0,0 +1,209 @@
...
+		},
+	},
+}
+
+// TestIntegreatlyRoutesExist tests that the routes for all the products are created
+func TestIntegreatlyRoutesExist(t *testing.T, ctx *TestingContext) {
+	for product, routes := range expectedRoutes {
+		for _, expectedRoute := range routes {
+			foundRoute, err := getRoute(t, ctx, product, expectedRoute)
+
```
b1zzu:
```suggestion
			foundRoute, err := getRoute(t, ctx, product, expectedRoute)
```

test/common/routes_exist.go
b1zzu: `t.Errorf` doesn't break the execution so we need to continue the loop otherwise it will proceed and fail in the `isTLS` function

sergioifg94: Oops missed that one, thanks!

test/common/routes_exist.go
b1zzu: Just the new line :sweat_smile: , usually the if err is attached to the function call

sergioifg94: Oops, the diff didn't show it. I just pushed that change!

sergioifg94:
/hold
As some routes are missing in the test

PR #527, opened 2020-03-18
Add common test to check PDBs exist

davidffrench:
# Description
https://issues.redhat.com/browse/INTLY-5735

Adding common test to check for the existence of PodDisruptionBudgets

## Type of change

## Checklist

### DIFF
--- a/test/common/pdbs_exist.go
+++ b/test/common/pdbs_exist.go
@@ -1,0 +1,49 @@
+package common
+
+import (
+	"testing"
+
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	namespaces = []Namespace{
+		{
+			Name: "redhat-rhmi-rhsso",
+			PodDisruptionBudgetNames: []string{
+				"keycloak",
+			},
+		},
+		{
+			Name: "redhat-rhmi-user-sso",
+			PodDisruptionBudgetNames: []string{
+				"keycloak",
+			},
+		},
+		// {
+		// 	Name: "redhat-rhmi-3scale",
+		// 	PodDisruptionBudgetNames: []string{
+		// 		"apicast-production",
+		// 		"apicast-staging",
+		// 		"backend-cron",
+		// 		"backend-listener",
+		// 		"backend-worker",
+		// 		"system-app",
+		// 		"system-sidekiq",
+		// 		"zync",
+		// 		"zync-que",
+		// 	},
+		// },
+	}
+)
+
+func TestIntegreatlyPodDisruptionBudgetsExist(t *testing.T, ctx *TestingContext) {
+	for _, namespace := range namespaces {
+		for _, podDisruptionBudgetName := range namespace.PodDisruptionBudgetNames {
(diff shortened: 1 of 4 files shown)


test/common/pdbs_exist.go
```diff
@@ -0,0 +1,49 @@
...
+				"keycloak",
+			},
+		},
+		{
+			Name: "redhat-rhmi-user-sso",
+			PodDisruptionBudgetNames: []string{
+				"keycloak",
+			},
+		},
+		// {
```
davidffrench: Adding these in but commented out for now until the new version of 3scale is included at the beginning of April

davidffrench:
Unable to add in the check for AMQ Online PodDisruptionBudgets for two reasons:
1. This requires the non-released AMQ Online 1.4 release.
2. The AMQ Online PDBs are only created dynamically once address spaces are created. @wei-lee @b1zzu Once the https://issues.redhat.com/browse/INTLY-5434 ticket is resolved, should be able to then check for PDBs?

test/common/pdbs_exist.go
```diff
@@ -0,0 +1,49 @@
...
+		// },
+	}
+)
+
+func TestIntegreatlyPodDisruptionBudgetsExist(t *testing.T, ctx *TestingContext) {
+	for _, namespace := range namespaces {
+		for _, podDisruptionBudgetName := range namespace.PodDisruptionBudgetNames {
+			_, err := ctx.KubeClient.PolicyV1beta1().PodDisruptionBudgets(namespace.Name).Get(podDisruptionBudgetName, v1.GetOptions{})
+			if err != nil {
+				t.Fatalf("PodDisruptionBudget %s not found in namespace %s - Error: %s", podDisruptionBudgetName, namespace.Name, err)
```
b1zzu: I suggest using `t.Errorf` so that the log will be printed and the test marked as failed but the execution can continue in order to check the remaining PDBs
```suggestion
				t.Errorf("PodDisruptionBudget %s not found in namespace %s - Error: %s", podDisruptionBudgetName, namespace.Name, err)
```

test/common/types.go
```diff
@@ -36,3 +36,14 @@ type prometheusAPIResponse struct {
...
+type Namespace struct {
+	Name                     string
+	Products                 []Product
+	PodDisruptionBudgetNames []string
+}
+
+type Product struct {
+	Name             string
+	ExpectedReplicas int32
+}
```
b1zzu: In my opinion, is better to keep these types specialized per test, the code will be duplicated but it would be cleaner for me

davidffrench: The problem is with all the tests being in a single package (which is right in my opinion), we will have to come up with different names for each file which isn't ideal in my opinion. This is why I think it is better to have a common type that makes sense for the tests that need it. If it starts to get unruly, maybe decide to split it back out again then but I don't think this will happen.

test/common/pdbs_exist.go
davidffrench: excellent, thanks @b1zzu . Will make this change now

test/common/types.go
b1zzu: I see I forget about this golang constrain, in this case there is nothing else we can do

davidffrench:
@b1zzu rebased from master with your merged PR now. Will need another `lgtm` when you get a chance

PR #528, opened 2020-03-18
Clean test cache before running functional tests

davidffrench:
# Description
I was noticing a cached response when running functional tests even when I switched the target cluster which should have failed the tests.
```
ok      github.com/integr8ly/integreatly-operator/test/functional       (cached)
```

Cleaning the cachetest before each run is the change here as per https://golang.org/cmd/go/#hdr-Build_and_test_caching

## Type of change

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,7 +131,7 @@ test/e2e: cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/prepare/c
 .PHONY: test/functional
 test/functional:
 	# Run the functional tests against an existing cluster. Make sure you have logged in to the cluster.
-	go test ./test/functional
+	go clean -testcache && go test ./test/functional
 
 .PHONY: install/olm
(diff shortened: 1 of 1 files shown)


davidffrench:
@wei-lee If you have a second to look over this minor change, that would be great.

Makefile
```diff
@@ -131,7 +131,7 @@ test/e2e: cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/prepare/c
 .PHONY: test/functional
 test/functional:
 	# Run the functional tests against an existing cluster. Make sure you have logged in to the cluster.
-	go test ./test/functional
+	go clean -testcache && go test ./test/functional
```
b1zzu: Using the `-v` flag is enough:

Form the go doucumentation
```
When 'go test' runs in package list mode, 'go test' caches successful
package test results to avoid unnecessary repeated running of tests. To
disable test caching, use any test flag or argument other than the
cacheable flags. The idiomatic way to disable test caching explicitly
is to use -count=1.
```

davidffrench: Just tested, unfortunately `-v` still uses cached results. I had seen the alternative solution of using `-count=1` but this seems more like a workaround to me vs using `go clean -testcache`.

PR #531, opened 2020-03-19
INTLY-6303-fix-amq-memory-usage-alert

austinmartinh:
# Description
When memory usage in `api-server` container in AMQ namespace is increased above 90% of capacity, `AMQOnlinePodHighMemory` alert is not being triggered. The expression defining the trigger for the alert needed to be modified as it referred to a label that did not exist.

Also, an unnecessary alert RHSSOPodHighMemory has been removed

JIRA: https://issues.redhat.com/browse/INTLY-6303
          https://issues.redhat.com/browse/INTLY-5491
 
# Verification
- Open terminal on `api-server` container and use a bash script to get memory utilisation to 90%+
- Wait 15 minutes 
- Verify Alert fires
- Ensure RHSSOHighPodMemory Alert is no longer showing
## Type of change

## Checklist


david-martin:
/approve

@austinmartinh Would you mind checking all alert & dashboard files for use of 'container_name' in case they need to be updated too?
Note there's another PR where some have been updated https://github.com/integr8ly/integreatly-operator/pull/532

grdryn:
I don't know enough about alerts to be able to just eyeball this and know that it's correct. I'll verify it on a cluster & if all of the modified alerts work as expected, then I'll lgtm. :+1:

templates/monitoring/kube_state_metrics_rhsso_alerts.yaml
```diff
@@ -13,7 +13,7 @@ spec:
             message: The {{ "{{" }} $labels.container {{ "}}" }} pod has been using {{ "{{" }} printf "%.0f" $value {{ "}}" }}% of available memory for longer than 15 minutes.
             scaling_plan: https://github.com/integr8ly/middleware-load-testing/blob/master/sops/sso-scaling.md
           expr: |
-            sum by(container) (label_replace(container_memory_usage_bytes{container_name!="",namespace="{{ index .Params "Namespace" }}"}, "container", "$1", "container_name", "(.*)")) / sum by(container) (kube_pod_container_resource_limits_memory_bytes{namespace="{{ index .Params "Namespace" }}"}) * 100 > 90
+            sum by(container) (container_memory_usage_bytes{container!="",namespace="{{ index .Params "Namespace" }}"}) / sum by(container) (kube_pod_container_resource_limits_memory_bytes{namespace="{{ index .Params "Namespace" }}"}) * 100 > 90
```
grdryn: @austinmartinh were you able to successfully trigger this alert for RHSSO? It looks like there aren't any resource requests or limits on the Keycloak pods/containers.

@david-martin FYI. If I'm right, should there be resource limits, or should we remove the alert?

templates/monitoring/kube_state_metrics_3scale_alerts.yaml
```diff
@@ -128,7 +128,7 @@ spec:
           message: The {{ "{{" }} $labels.container {{ "}}" }} pod has been using {{ "{{" }} printf "%.0f" $value {{ "}}" }}% of available CPU for longer than 15 minutes.
           scaling_plan: https://github.com/integr8ly/middleware-load-testing/blob/master/sops/3scale-scaling.md
         expr: |
-          sum(label_replace(namespace_pod_name_container_name:container_cpu_usage_seconds_total:sum_rate{namespace="{{ index .Params "Namespace" }}"}, 'container', '$1', 'container_name', '(.*)')) by (container) / sum(kube_pod_container_resource_limits_cpu_cores{namespace="{{ index .Params "Namespace" }}"}) by (container) * 100 > 90
+          sum(namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{namespace="{{ index .Params "Namespace" }}"}) by (container) / sum(kube_pod_container_resource_limits_cpu_cores{namespace="{{ index .Params "Namespace" }}"}) by (container) * 100 > 90
```
grdryn:
```suggestion
          sum(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{namespace="{{ index .Params "Namespace" }}"}) by (container) / sum(kube_pod_container_resource_limits_cpu_cores{namespace="{{ index .Params "Namespace" }}"}) by (container) * 100 > 90
```

I'm not sure if this is more correct or not, but I was at least able to get a value using it. With `namespace_pod_container` (as in your change) I wasn't able to trigger the alert or get a value when querying.

templates/monitoring/kube_state_metrics_rhsso_alerts.yaml
david-martin: SSO used to have resource limits when installed with the older integreaty/keycloak-operator.
So this alert should be removed as it's of no use in it's current state.
Also, it's already been marked for removal for a different reason (downstream issue https://issues.redhat.com/browse/INTLY-5491), so @austinmartinh let's remove it.

templates/monitoring/kube_state_metrics_3scale_alerts.yaml
austinmartinh: Ah good catch, that was my intended change.

R-Lawton:
:eyes:

grdryn:
I've just deployed this on my cluster & have modified the thresholds to something I think should cause alerts, but I haven't seen anything yet. I'll leave it run while I'll be on a call, and will try harder after if there's still nothing there

R-Lawton:
finding it quite difficult to hit the threshold i have also modified it to be lower

grdryn:
Verified :100: 

/lgtm

PR #533, opened 2020-03-19
INTLY-5734-verify-aws-resources-are-created

HubertStefanski:
# Description
https://issues.redhat.com/browse/INTLY-5734
This change adds in new functional tests to check if aws resources (postgres and redis) have been created by the CRO during an RHMI install

# Verify
**Note** this test is only a functional test and will require an RHMI cluster backed by AWS resources 
- Run `make test/functional` 
- Ensure tests pass successfully


test/functional/aws_test.go
```diff
@@ -0,0 +1,85 @@
...
+		t.Log("checking tags database cluster")
+		tagListInput := &rds.ListTagsForResourceInput{
+			ResourceName: dbInstance.DBInstanceArn,
+		}
+		tagListOutput, err := c.rdsClient.ListTagsForResource(tagListInput)
+		if err != nil {
+			return nil, fmt.Errorf("failed to list tags for database cluster", err)
+		}
+		t.Logf("checking for cluster tag match (%s=%s) on database", tagKeyClusterId, clusterIDString)
+		if common.FindTag(tagKeyClusterId, clusterIDString, tagListOutput.TagList) == nil {
```
dimitraz: Do we need this cluster tag check? All of the database instances we're listing should be from the same cluster regardless

test/functional/aws_test.go
```diff
@@ -0,0 +1,85 @@
...
+			t.Logf("database did not contain cluster tag match (%s=%s)", tagKeyClusterId, clusterIDString)
+			continue
+		}
+		extraTagsMatch := true
+		for extraTagKey, extraTagVal := range tags {
```
dimitraz: Do we need these extra tag checks as well? I would just be in favour of returning the entire list of found DB instances and then checking if their names/tags matched the ones we expected in the main function :thinking:

test/functional/aws_test.go
HubertStefanski: Valid point, will refactor

test/functional/aws_test.go
```diff
@@ -0,0 +1,162 @@
...
+		ClusterSSOResourceName,
+		ApplicationSSOResourceName,
+	}
+)
+
+func TestAWS(t *testing.T) {
+	//TODO
+}
+
+func testExpectedResourcesExist(t testing.T, f *framework.Framework, instances []*rds.DBInstance, cacheClusters []*elasticache.CacheCluster, buckets []*s3.Bucket, expectedResources []string) (map[string]bool, error) {
```
dimitraz: Does this function have to be wrapped? Could probably just have this as the main function

test/functional/aws_test.go
```diff
@@ -0,0 +1,162 @@
...
+	//check RDS resources
+
+	for _, instance := range instances {
+		for _, name := range expectedResources {
+			if instance.DBName != aws.String(name) {
+				continue
+			}
+			if instance.MultiAZ != aws.Bool(true) {
+				t.Logf("instance: %v , has multi-AZ disabled, instance ARN: %v ", instance.DBName, instance.DBInstanceArn)
+				return nil, errors.New("error: Multi-AZ is disabled")
```
dimitraz: Over here then we wouldn't need to return the error, just a simple `t.Errorf()` statement

test/functional/aws_test.go
```diff
@@ -0,0 +1,162 @@
...
+	return &awsClients{
+		rdsClient:         rds.New(session),
+		elasticacheClient: elasticache.New(session),
+		s3Client:          s3.New(session),
+	}
+}
+
+func setAWSCredEnvar(f *framework.Framework) error {
+	secret := &corev1.Secret{}
+	err := f.Client.Get(goctx.TODO(), types.NamespacedName{Name: awsCredsSecretName, Namespace: awsCredsNamespace}, secret)
```
ciaranRoche: Why are we using the `kube-system` secret. Should we be using the secret created by CRO?

test/functional/aws_test.go
```diff
@@ -0,0 +1,162 @@
...
+	for _, cacheCluster := range cacheClusters {
+		if cacheCluster.CacheClusterStatus != aws.String(resourceStatusAvailable) {
+			return false
+		}
+	}
+	return true
+}
+
+func getRDSResources(session *session.Session) ([]*rds.DBInstance, error) {
+	client := newAWSClients(session)
```
dimitraz: Does `newAWSClients` have to be called every time? Can we just declare it once and then pass it into all of these functions instead of `session`?

test/functional/aws_test.go
```diff
@@ -0,0 +1,162 @@
...
+	clusterDescribeInput := &rds.DescribeDBInstancesInput{}
+	clusterDescribeOutput, err := client.rdsClient.DescribeDBInstances(clusterDescribeInput)
```
ciaranRoche: Can we pass the `&rds.DescribeDBInstance{}` directly in the `DescribeDBInstances()` function, instead of declaring it as a variable?

test/functional/aws_test.go
```diff
@@ -0,0 +1,162 @@
...
+	return RDSDBInstances, nil
+}
+func getElasticacheResources(session *session.Session) ([]*elasticache.CacheCluster, error) {
+	client := newAWSClients(session)
+	clusterDescribeInput := &elasticache.DescribeCacheClustersInput{}
+	clusterDescribeOutput, err := client.elasticacheClient.DescribeCacheClusters(clusterDescribeInput)
+	if err != nil {
+		return nil, fmt.Errorf("failed to describe elasticache clusters : %w", err)
+	}
+	var ElasticacheInstances []*elasticache.CacheCluster
```
dimitraz: Does this var need to be exported?

test/functional/aws_test.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func getRDSResources(session *session.Session) ([]*rds.DBInstance, error) {
+	client := newAWSClients(session)
+	clusterDescribeInput := &rds.DescribeDBInstancesInput{}
+	clusterDescribeOutput, err := client.rdsClient.DescribeDBInstances(clusterDescribeInput)
+	if err != nil {
+		return nil, fmt.Errorf("failed to describe database clusters : %w", err)
+	}
+	var RDSDBInstances []*rds.DBInstance
```
dimitraz: Does this var need to be exported?

test/functional/aws_test.go
```diff
@@ -0,0 +1,162 @@
...
+	for _, dbInstance := range clusterDescribeOutput.DBInstances {
+		RDSDBInstances = append(RDSDBInstances, dbInstance)
+	}
+	return RDSDBInstances, nil
```
ciaranRoche: I think we can just return `clusterDescribeOutput.DBInstances` as this is an `[]*rds.DBInstance` we do not need to build it ourselves

test/functional/aws_test.go
```diff
@@ -0,0 +1,162 @@
...
+	clusterDescribeOutput, err := client.elasticacheClient.DescribeCacheClusters(clusterDescribeInput)
+	if err != nil {
+		return nil, fmt.Errorf("failed to describe elasticache clusters : %w", err)
+	}
+	var ElasticacheInstances []*elasticache.CacheCluster
+	for _, cacheCluster := range clusterDescribeOutput.CacheClusters {
+
+		ElasticacheInstances = append(ElasticacheInstances, cacheCluster)
+	}
+	return ElasticacheInstances, nil
```
ciaranRoche: Same as above we can just return `clusterDescribeOutput.CacheClusters`

test/functional/aws_test.go
```diff
@@ -0,0 +1,162 @@
...
+	listBucketsRequestOutput, err := client.s3Client.ListBuckets(listBucketsRequestInput)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list buckets : %w", err)
+	}
+	var s3Buckets []*s3.Bucket
+	for _, bucket := range listBucketsRequestOutput.Buckets {
+		s3Buckets = append(s3Buckets, bucket)
+
+	}
+	return s3Buckets, nil
```
ciaranRoche: Same here, just return `listBucketsRequestOutput.Buckets`

test/functional/aws_test.go
```diff
@@ -0,0 +1,162 @@
...
+	if string(secret.Data["aws_access_key_id"]) != "" && string(secret.Data["aws_secret_access_key"]) != "" {
+		os.Setenv("AWS_ACCESS_KEY_ID", string(secret.Data["aws_access_key_id"]))
+		os.Setenv("AWS_SECRET_ACCESS_KEY", string(secret.Data["aws_secret_access_key"]))
+		return nil
+	}
+	return fmt.Errorf("error setting %v envar, secret data may not exist", awsCredsSecretName)
+}
+
+// Wait for status of all resources to become "available", return true if all available
+func awsResourcesAvailable(dbinstances []*rds.DBInstance, cacheClusters []*elasticache.CacheCluster) bool {
```
dimitraz: Is this function needed? We probably have other tests (e.g. the alerts test) that will pick up on whether or not the resources are available, so we shouldn't need to wait for that (because that might also never happen)

test/functional/aws_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+func setAWSCredEnvar(f *framework.Framework) error {
+	secret := &corev1.Secret{}
+	err := f.Client.Get(goctx.TODO(), types.NamespacedName{Name: awsCredsSecretName, Namespace: awsCredsNamespace}, secret)
+	if err != nil {
+		return fmt.Errorf("error getting secret: %v from cluster: %w ", awsCredsSecretName, err)
+	}
+	if string(secret.Data["aws_access_key_id"]) != "" && string(secret.Data["aws_secret_access_key"]) != "" {
+		os.Setenv("AWS_ACCESS_KEY_ID", string(secret.Data["aws_access_key_id"]))
+		os.Setenv("AWS_SECRET_ACCESS_KEY", string(secret.Data["aws_secret_access_key"]))
```
ciaranRoche: Do we need to set these `env-vars`? Can we pass the secret values directly when we create the AWS session?

dimitraz: Are these env vars required to create the session?

test/functional/aws_test.go
HubertStefanski: Yes, noticed this yesterday, going to change this in the next commit

test/functional/aws_test.go
HubertStefanski: We can, will do

test/functional/aws_test.go
HubertStefanski: They are required for the session, although I'm looking at a way to set them when calling a new session

test/functional/aws_test.go
HubertStefanski: I wasn't entirely sure, so just created this just in case, but I've no objection if this is to be deleted

test/functional/aws_test.go
HubertStefanski: Changed to be used as the main function :+1:

test/functional/aws_test.go
HubertStefanski: removed as per Ciarans suggestion to return the value without wrapping it into a seperate slice

test/functional/aws_test.go
HubertStefanski: Will refactor to incorporate this

test/functional/aws_test.go
HubertStefanski: Initially this was the easier approach, but I am looking at passing these directly

test/functional/aws_test.go
```diff
@@ -0,0 +1,242 @@
...
+			if expectedInstance.DBName != returnedInstance.DBName {
+				continue
+			}
+			if !common.EqualResourceBool(
+				[]bool{aws.BoolValue(expectedInstance.DeletionProtection), aws.BoolValue(expectedInstance.MultiAZ), aws.BoolValue(expectedInstance.StorageEncrypted)},
+				[]bool{aws.BoolValue(returnedInstance.DeletionProtection), aws.BoolValue(returnedInstance.MultiAZ), aws.BoolValue(returnedInstance.StorageEncrypted)}) {
+				t.Logf("RDS resources are not in the expected state")
+				resourceMatchesExpectedState[aws.StringValue(expectedInstance.DBName)] = false
+			}
+			resourceMatchesExpectedState[aws.StringValue(expectedInstance.DBName)] = true
```
jhellar, line 106: Would not it be marked as true every time? Should this be in `else` branch?

HubertStefanski: Good catch! completely missed that when I put it in :+1:

test/functional/aws_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+		}
+	}
+	//Check Elasticache
+	for _, expectedReplicationGroup := range expectedElasticacheResources {
+		for arn, replicationgroups := range elasticacheResourcesMap {
+			if strings.Contains(arn, ThreeScaleNameElement) {
```
jhellar, line 113: If I understand it correctly, this will make sure that if there is 3scale resource, it has correct configuration, but it won't fail if the resource is not there. Should we also check that the resources are there? (Same for other resource types)

test/functional/aws_test.go
```diff
@@ -0,0 +1,242 @@
...
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"os"
+	"strings"
+	"testing"
+)
+
+var (
+	expectedRDSResources = []*rds.DBInstance{
```
jhellar, line 23: Manual test case checks also automatic backup, can this be added?

test/functional/aws_test.go
```diff
@@ -0,0 +1,242 @@
...
+			StorageEncrypted:   aws.Bool(true),
+		},
+		{
+			DBName:             aws.String(ApplicationSSOResourceName),
+			DeletionProtection: aws.Bool(true),
+			MultiAZ:            aws.Bool(true),
+			StorageEncrypted:   aws.Bool(true),
+		},
+	}
+	expectedElasticacheResources = []*elasticache.ReplicationGroup{
```
jhellar, line 62: Can we add check for multi-az and automatic backup?

HubertStefanski: ReplicationGroups by themselves don't have multi-AZ fields, this means that a slightly more extensive test is needed to check if the cache clusters for the respective groups are in different AZ's, but these should also be implemented, Noted to be added

test/functional/aws_test.go
HubertStefanski: Left this out while investigating ways of checking for that, the best solution @aidenkeating suggested was just to check if the `PreferredBackupWindow` field is not empty, That's why theres no field for it in the `DBInstance{}` slice, as we can't predict what values are going to be present, so the approach for this is to do a nil check or something similar on the `returnedInstance.PrefferedBackupWindow`, this change will come in the next few commits :+1:

test/functional/aws_test.go
HubertStefanski: The code here is a bit outdated, but essentially there will be a final check that will ensure `resourceMatchesExpectedState` returns all true, else it will return what fields are missing or misconfigured, the results for Elasticache and s3 will be added into this map aswell :+1:

test/functional/aws_elasticache_resources_exist.go
```diff
@@ -0,0 +1,104 @@
...
+		foundElasticacheReplicationGroups, err := elasticacheapi.DescribeReplicationGroups(&elasticache.DescribeReplicationGroupsInput{
+			ReplicationGroupId: aws.String(resourceID),
+		})
+		if err != nil {
+			testErrors = append(testErrors, fmt.Sprintf("failed to get %s elasticache replicationgroups with error : %v", resourceID, err))
+			continue
+		}
+
+		replicationGroup := foundElasticacheReplicationGroups.ReplicationGroups[0]
+		nodeGroup := replicationGroup.NodeGroups[0]
```
dimitraz, line 75: Should we check if the length of the `NodeGroups` is 2 before we continue?

test/functional/aws_elasticache_resources_exist.go
```diff
@@ -0,0 +1,104 @@
...
+		t.Fatalf("test elasticache exists failed with the following errors : %s", testErrors)
+	}
+
+}
+
+// helper method for verifying nodes are in different availability zones
+func verifyMultiAZ(member []*elasticache.NodeGroupMember) error {
+	if len(member) < 1 {
+		return errors.New("insufficient number of nodes in elasticache group")
+	}
```
dimitraz, line 99: I think we can remove this if we perform the node group length check above

test/functional/aws_elasticache_resources_exist.go
```diff
@@ -0,0 +1,104 @@
...
+	if member[0].PreferredAvailabilityZone == member[1].PreferredAvailabilityZone {
+		return errors.New("replication group is not multiAZ")
+	}
+	return nil
+}
```
dimitraz, line 104: Not a huge improvement, but may be slightly cleaner to change the function signature to return a `bool` instead: 

```suggestion
func verifyMultiAZ(members []*elasticache.NodeGroupMember) bool {
	return members[0].PreferredAvailabilityZone == members[1].PreferredAvailabilityZone
}
```
Then we wouldn't need the error check when calling the function either.

test/functional/aws_elasticache_resources_exist.go
```diff
@@ -0,0 +1,104 @@
...
+		if !aws.BoolValue(replicationGroup.AtRestEncryptionEnabled) {
+			testErrors = append(testErrors, fmt.Sprintf("elasticache resource %s does not have encryption enabled", resourceID))
+		}
+		if replicationGroup.SnapshotWindow == nil {
+			testErrors = append(testErrors, fmt.Sprintf("elasticache resource %s does not have automatic snapshotting enabled", resourceID))
+		}
+	}
+
+	if len(testErrors) != 0 {
+		t.Fatalf("test elasticache exists failed with the following errors : %s", testErrors)
```
dimitraz, line 90:
```suggestion
		t.Fatalf("test elasticache instances exist failed with the following errors : %s", testErrors)
```

test/functional/aws_rds_resources_exist.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// check ever expected resource
+	rdsapi := rds.New(sess)
+	for _, resourceIdentifier := range rdsResourceIDs {
+		// get the rds instance
+		foundRDSInstances, err := rdsapi.DescribeDBInstances(&rds.DescribeDBInstancesInput{
+			DBInstanceIdentifier: aws.String(resourceIdentifier),
+		})
+		if err != nil {
+			testErrors = append(testErrors, fmt.Sprintf("failed to get %s rds instances with error : %v", resourceIdentifier, err))
```
dimitraz, line 69:
```suggestion
			testErrors = append(testErrors, fmt.Sprintf("failed to get rds instance: %s, with error : %v", resourceIdentifier, err))
```

test/functional/integreatly_test.go
```diff
@@ -19,14 +19,15 @@ func TestIntegreatly(t *testing.T) {
 	t.Run("Integreatly Happy Path Tests", func(t *testing.T) {
 		for _, test := range common.ALL_TESTS {
 			t.Run(test.Description, func(t *testing.T) {
-				testingContext, err = common.NewTestingContext(config)
```
dimitraz, line 22: Why are we deleting this?

test/functional/aws_rds_resources_exist.go
```diff
@@ -0,0 +1,87 @@
...
+			continue
+		}
+		// verify the rds instance is as expected
+		if !verifyRDSInstanceConfig(*foundRDSInstances.DBInstances[0]) {
+			testErrors = append(testErrors, fmt.Sprintf("failed as rds %s resources is not as expected", resourceIdentifier))
+		}
+	}
+
+	if len(testErrors) != 0 {
+		t.Fatalf("test cro postgres exists failed with the following errors : %s", testErrors)
```
dimitraz, line 79:
```suggestion
		t.Fatalf("test rds instances exist failed with the following errors : %s", testErrors)
```

test/functional/aws_elasticache_resources_exist.go
HubertStefanski: I Think leaving it to return an error is better, due to the fact that when It does fail, at least we've a specific description of why it failed as opposed to just getting a "Fail" and being left in the dark as to why

dimitraz: You could always return the same error above when you call the function

test/functional/integreatly_test.go
HubertStefanski: We're passing it from outside of the `for` loop from what I can see, we can probably do the same with the `functional loop` to avoid code repetition

ciaranRoche: I think we should add this back in. Each test that we run should have its own testing context. This will ensure each test does not modify or change the context on another test. It will also ensure that every test run will have the latest/most up to date schemes/apis. 

So yeah, I think for `ALL_TEST` , `AFTER_INSTALL_TESTS` and `FUNCTIONAL_TESTS` we should add this back in.

test/functional/integreatly_test.go
```diff
@@ -19,16 +19,19 @@ func TestIntegreatly(t *testing.T) {
 	t.Run("Integreatly Happy Path Tests", func(t *testing.T) {
 		for _, test := range common.ALL_TESTS {
 			t.Run(test.Description, func(t *testing.T) {
+				testingContext, err = common.NewTestingContext(config)
```
ciaranRoche, line 22: i think we should be handling the `err` in these. 

Also I think we can remove from line 15 - 18,  we do not need to create the testing context here as we do it inline

david-martin:
/approve

@jhellar do any manual tests need to be removed from test cases?

HubertStefanski:
@david-martin  last commit before this is merged, assigned seperate test case ID's for test cases after discussing this with @b1zzu

david-martin:
Test failure looks like the already logged test flake https://issues.redhat.com/browse/INTLY-6679
```
        --- FAIL: TestIntegreatly/integreatly/B05_-_Verify_Codeready_CRUDL_permissions (43.63s)
            codeready_crudl.go:97: Test codeready workspace creation
            codeready_crudl.go:125: failed to login to codeready: failed to get access token: {"error":"invalid_grant","error_description":"Code not valid"}
```

/lgtm
/test e2e

PR #535, opened 2020-03-19
replaced parameters to quicken scrape duration

R-Lawton:
[CSSRE-830](https://issues.redhat.com/browse/CSSRE-830.) for the observed issue in RHMI 1.x which caused intermittent scrape timeouts and hence intermittent alerts.The following fixes this in RHMI2.x https://issues.redhat.com/browse/INTLY-6406

Before: (high 3 -4)
no
After: (low 1-2)
Screenshot from 2020-03-23 11-33-36
### Prerequisite:
OSD cluster
### Verification steps:
1. Install integreatly using this branch
2. when everythngs up run the following expression in Prometheus : `scrape_duration_seconds{job="openshift-monitoring-federation"}`
3. The graph should look like the after graph above.
4. Check that the results of the following expressions are using the rhmi namespace:
`kubelet_volume_stats_used_bytes`
`kubelet_volume_stats_available_bytes`


templates/monitoring/jobs/openshift_monitoring_federation.yaml
```diff
@@ -9,7 +9,8 @@
   metrics_path: /federate
   params:
     match[]:
-    - '{{ "{" }} endpoint="https-metrics" {{ "}" }}'
+    - 'kubelet_volume_stats_used_bytes{endpoint="https-metrics",namespace=~"openshift-.*"}'
```
david-martin: @R-Lawton the namespace regex is too restrictive for on OSD4. It's fine for OSD3 where RHMI namespaces are prefixed with `openshift-`. However on OSD4 they are prefixed with `redhat-rhmi-`.
Based on that, I'd suggest this regex be changed to `redhat-rhmi-`.

@philbrookes Does that seem correct to you for any volumes we'd be interested in on OSD4?

R-Lawton: Cool ill change that and see does it change the graph

maleck13: yes the prefix in OSD is ```redhat-rhmi-*```

templates/monitoring/jobs/openshift_monitoring_federation.yaml
```diff
@@ -9,7 +9,8 @@
   metrics_path: /federate
   params:
     match[]:
-    - '{{ "{" }} endpoint="https-metrics" {{ "}" }}'
+    - 'kubelet_volume_stats_used_bytes{endpoint="https-metrics",namespace=~"redhat-rhmi-.*"}'
+    - 'kubelet_volume_stats_available_bytes{endpoint="https-metrics",namespace=~"redhat-rhmi-.*"}'
     - '{{ "{" }} service="kube-state-metrics" {{ "}" }}'
     - '{{ "{" }} service="node-exporter" {{ "}" }}'
     - '{{ "{" }} __name__=~"namespace_pod_name_container_name:.*" {{ "}" }}'
```
dimitraz: I don't think these metrics need to be federated, since as far as I know this is replaced by `node_namespace_pod_container` in newer versions of Kubernetes/Prometheus. 

If that were the case, then this line in the 3scale dashboard would also need to be updated:
https://github.com/integr8ly/integreatly-operator/blob/ace4b90437c62ad2653b793ef300d0a26f7139a2/templates/monitoring/kube_state_metrics_3scale_alerts.yaml#L131

R-Lawton: removed the unused metric

david-martin:
@alanmoran can you review for an lgtm?

alanmoran:
Assuming we've tested that `namespace_pod_name_container_name:.*` is not used. 

Otherwise /lgtm

R-Lawton:
@alanmoran namespace_pod_name_container_name:.* dosent exist anymore in osd4 its been replaced by node_namespace_pod_container:.*

PR #538, opened 2020-03-20
Fixing incorrect field type in AMQ Online CRD and types

davidffrench:
# Description
https://issues.redhat.com/browse/INTLY-6411

A change was introduced during the operator-sdk update of the integreatly-operator that modified the field type from float32 to string for AMQ Online (https://github.com/integr8ly/integreatly-operator/pull/379/files#diff-abd13f258f80f94a2c1f152e5f91c80aL53-L57). The reason this was changed was that the `operator-sdk generate crds` command using the newer controller-tools does not support floats any longer.

Changing to strings causes issues when we will go to upgrade AMQ Online to the newest version (1.4). We should not modify the CRD or types that we copy in from other operators.

The solution here since alternatives didn't work is to move the apis from dependant operators outside of pkg/api so the `operator-sdk generate crds` command will not try to generate a crd that we don't use anyway.

## Type of change

## Checklist


maleck13:
as the automated tests passed I think we are good to merge

pkg/products/amqonline/addressspaceplans.go
```diff
@@ -118,9 +118,9 @@ func GetDefaultAddressSpacePlans(ns string) []*v1beta2.AddressSpacePlan {
 				LongDescription:  "Messaging infrastructure based on Apache Qpid Dispatch Router and Apache ActiveMQ Artemis and MQTT support. This plan allows an unlimited number of routers (minimum 2) and brokers, and is suitable for applications where you do not want to impose any restrictions.",
 				AddressSpaceType: "standard",
 				ResourceLimits: v1beta2.AddressSpacePlanResourceLimits{
-					Broker:    "10000.0",
-					Router:    "10000.0",
-					Aggregate: "10000.0",
+					Broker:    10000.0,
+					Router:    1000.0,
```
maleck13: I notice this changed from 10000 to 1000 any concern with that? I assume we just copied this in

davidffrench: Good spotting @maleck13 . Will create a PR to fix this now

davidffrench: PR up now - https://github.com/integr8ly/integreatly-operator/pull/539

PR #534, opened 2020-03-19
[WIP] INTLY-5434 - CRUDL an Address/Address Space as a user in the rhmi-developer group

b1zzu:
# Description

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist
- [ ] This change requires a documentation update 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Verified independently on a cluster by reviewer

PR #540, opened 2020-03-23
Intly 5439 - Fuse e2e tests

pb82:
Tests that a rhmi-developer user can log into fuse and interact with the API (create an integration). This requires login through the OAuth proxy.

Tests that as dedicated-admin user can log into 3scale and interact with the API (create a product). This requires to login to OpenShift and 3Scale separately.

This PR is based on the work from @ciaranRoche  (https://github.com/integr8ly/integreatly-operator/pull/510) and @b1zzu (https://github.com/integr8ly/integreatly-operator/pull/534)


test/resources/oauth_proxy_auth.go
```diff
@@ -0,0 +1,250 @@
+package resources
```
pb82: @b1zzu I've put your code for oauth proxy authentication into this file, it should work with all product host URLs. I've added a step for permissions approval on first logins.

jjaferson:
/lgtm  

I have tested the PR on my cluster setting up testing IDP manually. 

The reason e2e test is falling is because it's missing the creation of testing IDP which is addressed [here INTLY-6418](https://issues.redhat.com/browse/INTLY-6418).

test/resources/oauth_proxy_auth.go
```diff
@@ -0,0 +1,250 @@
...
+		return nil, errorWithResponseDump(response, fmt.Errorf("the request to %s failed with code %d", u, response.StatusCode))
+	}
+
+	document, err = parseResponse(response)
+	if err != nil {
+		return nil, err
+	}
+
+	// find the form for the approval
+	form = document.Find("form")
```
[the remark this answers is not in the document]
pb82: I removed this because the URL is different when you get the approval form. But I can add that check after the approval (or if no approval is needed).

test/common/fuse_crudl.go
```diff
@@ -0,0 +1,94 @@
...
+	// Get the fuse host url from the rhmi status
+	fuseHost := rhmi.Status.Stages[v1alpha1.ProductsStage].Products[v1alpha1.ProductFuse].Host
+
+	err = loginOpenshift(ctx, masterURL, fuseLoginUser, fuseLoginPassword, rhmi.Spec.NamespacePrefix)
+	if err != nil {
+		t.Fatalf("error logging into openshift: %v", err)
+	}
+
+	// Get a client that authenticated to fuse via oauth
+	authenticatedFuseClient, err := resources.ProxyOAuth(ctx.HttpClient, fuseHost, fuseLoginUser, DefaultPassword)
```
ciaranRoche, line 53: Do we need to do this full ProxyOAuth flow, if we have already logged the user in via Openshift, if we use the httpClient we can go directly to the syndesis endpoint and select `testing-idp`, we can skip going to `/oauth/start`

pb82: It looks like I still need to select the testing-idp, fetch the form and submit username and password. Hitting the syndesis url directly instead of oauth/start doesn't seem to work.

PR #543, opened 2020-03-23
fix: update test image release process

psturc:
# Description
Update readme - add test image release process


mikenairn:
@psturc This seems out of date, can it be closed?

psturc:
@mikenairn ah yes, definitely, thanks

PR #549, opened 2020-03-24
Fix replicas not reporting on test failure

davidffrench:
# Description
https://issues.redhat.com/browse/INTLY-6434

The common tests are not reporting the replica count of workloads on failure. Changing to use the correct type conversion

## Type of change

## Checklist

### DIFF
--- a/test/common/deployment_types.go
+++ b/test/common/deployment_types.go
@@ -205,22 +205,22 @@ func TestDeploymentExpectedReplicas(t *testing.T, ctx *TestingContext) {
 			}
 
 			if deployment.Status.Replicas != product.ExpectedReplicas {
-				t.Errorf("Deployment %s in namespace %s doens't match the number of expected replicas %s/%s",
+				t.Errorf("Deployment %s in namespace %s doesn't match the number of expected replicas. Replicas: %v / Expected Replicas: %v",
 					product.Name,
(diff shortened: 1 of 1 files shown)


test/common/deployment_types.go
```diff
@@ -205,22 +206,22 @@ func TestDeploymentExpectedReplicas(t *testing.T, ctx *TestingContext) {
...
 
 			// Verify that the expected replicas are also available, means they are up and running and consumable by users
 			if deployment.Status.AvailableReplicas != product.ExpectedReplicas {
-				t.Errorf("Deployment %s in namespace %s doens't match the number of expected available replicas %s/%s",
+				t.Errorf("Deployment %s in namespace %s doesn't match the number of expected available replicas. Available Replicas: %s / Expected Replicas: %s",
 					product.Name,
 					namespace.Name,
-					string(deployment.Status.AvailableReplicas),
-					string(product.ExpectedReplicas),
+					strconv.Itoa(int(deployment.Status.AvailableReplicas)),
```
pb82: @davidffrench using `%v` instead of `%s` and then not casting the avaible and expected replicas to strings also works

davidffrench: Thanks @pb82 . Much cleaner. Updated now

PR #550, opened 2020-03-24
[WIP] Refactor common.TestCase type for better encapsulation

sergioifg94:
Simple demostration on refactoring the `TestCase` struct with
an interface instead of a function field to improve encapsulation
of the different tests within the `common` package.

Refactor `pdbs_exist.go` as an example of implementing the `TestExecutor` interface.

**⚠️ Only one test is included and all the other tests were removed for the sake of simplicity ⚠️**

## Further improvements
* A `Cleanup` function in the `TestExecutor` interface to clean up any possible temporal resources that might have been created just for the test

### DIFF
--- a/test/common/pdbs_exist.go
+++ b/test/common/pdbs_exist.go
@@ -6,8 +6,12 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var (
-	namespaces = []Namespace{
+type podDisruptionBudgetTestExecutor struct {
+	namespaces []Namespace
+}
+
+var PodDisruptionBudgetTest = podDisruptionBudgetTestExecutor{
+	namespaces: []Namespace{
 		{
 			Name: "redhat-rhmi-rhsso",
 			PodDisruptionBudgetNames: []string{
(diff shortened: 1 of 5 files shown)


sergioifg94:
In the case of this test, the `namespaces` variable becomes a field in the `podDisruptionBudgetTestExecutor` struct instead of being in the global scope of the package. Other tests have more variables and functions that can be accessed from other tests in the package despite being unrelated. That's the main reason for this change, keep all members related to each test in their own struct and thus the scope of the `common` package is cleaner.

It's not really a problem, but as the test list continues to grow I can see it can get messy with all the unrelated functions and variables mixed in the same scope.

wei-lee:
@sergioifg94 ah, yes, I get it now. In that case, should we just make `TestCase` an interface instead? It should just have the `Description()` and `Execute()` method, and then each test can implement that interface and encapsulate its own fields.

sergioifg94:
Yes. Another option would be making the list of test cases a `map[string]TestCase` where the key would be the description, but I don't have anything against a `Description()` method either

wei-lee:
@sergioifg94 cool, sounds good then. Are you going to make the changes in this PR, or we can close this one and you will open another PR?

sergioifg94:
I could make them in this PR. Should I go ahead and do the whole refactor or just update the test that I included here?

wei-lee:
@sergioifg94 I say do one of them for now, and then ask a few other people for feedbacks, especially @philbrookes  and @b1zzu, before refactoring all other tests.

sergioifg94:
@wei-lee Makes sense, will do. Thanks!

philbrookes:
this looks like a good change to me, I really like it.

b1zzu:
@sergioifg94 I also like it +1

sergioifg94:
Cool, I'll create a JIRA for this and close the PR if that's all right with you :+1:

sergioifg94:
With a bit of delay, here's the JIRA for the refactor:
https://issues.redhat.com/browse/INTLY-6513

PR #551, opened 2020-03-24
Intly-4776

jjaferson:
https://issues.redhat.com/browse/INTLY-4776

Test network policy allows cross namespace access to services, so basically a project is created with an endpoint by a user in the dedicated-admin group and a request is made from the 3scale apicast pod to the created endpoint.  

Authentication to the Openshift API is required to create a project as user in the dedicated-admin group, which is a part of this PR https://github.com/integr8ly/integreatly-operator/pull/510

@ciaranRoche


test/common/network_policy_access_ns_to_svc.go
```diff
@@ -0,0 +1,278 @@
...
+
+	// check if project exist
+	err := ctx.Client.Get(goctx.TODO(), k8sclient.ObjectKey{Name: projectName}, &projectv1.Project{})
+	if err != nil && !k8serr.IsNotFound(err) {
+		return nil, fmt.Errorf("error occured while retrieving a project: %v", err)
+	} else if k8serr.IsNotFound(err) {
+		if err := resources.DoOpenshiftCreateProject(openshiftAPIURL, token, projectCR); err != nil {
+			return nil, fmt.Errorf("error occured while making request to create a project: %v", err)
+		}
+	}
```
Boomatang: I wonder would this be better write as below. The `k8serr.IsNotFound` the one you care most about and I think you should check for that first.
```
	if k8serr.IsNotFound(err) {
		if err := resources.DoOpenshiftCreateProject(openshiftAPIURL, token, projectCR); err != nil {
			return nil, fmt.Errorf("error occured while making request to create a project: %v", err)
		}
	} else if err != nil {
		return nil, fmt.Errorf("error occured while retrieving a project: %v", err)
	}
```

test/common/network_policy_access_ns_to_svc.go
```diff
@@ -0,0 +1,278 @@
...
+	key := k8sclient.ObjectKey{Name: serviceCR.GetName(), Namespace: serviceCR.GetNamespace()}
+	err := ctx.Client.Get(goctx.TODO(), key, serviceCR)
+	if err != nil && !k8serr.IsNotFound(err) {
+		return nil, fmt.Errorf("error occured while retrieving service %s : %v", serviceCR.GetName(), err)
+	} else if k8serr.IsNotFound(err) {
+		err = resources.DoOpenshiftCreateServiceInANamespace(openshiftAPIURL, token, projectName, serviceCR)
+		if err != nil {
+			return nil, fmt.Errorf("error occured while making request to create service %s : %v", serviceCR.GetName(), err)
+		}
+	}
```
Boomatang: This has the same comment as above.
```
	if k8serr.IsNotFound(err) {
		err = resources.DoOpenshiftCreateServiceInANamespace(openshiftAPIURL, token, projectName, serviceCR)
		if err != nil {
			return nil, fmt.Errorf("error occured while making request to create service %s : %v", serviceCR.GetName(), err)
		}
	} else if err != nill {
		return nil, fmt.Errorf("error occured while retrieving service %s : %v", serviceCR.GetName(), err)
	}
```

test/common/network_policy_access_ns_to_svc.go
```diff
@@ -0,0 +1,278 @@
...
+	err := ctx.Client.Get(goctx.TODO(), key, podCR)
+	if err != nil && !k8serr.IsNotFound(err) {
+		return nil, fmt.Errorf("error occured while retrieving pod %s : %v", podCR.GetName(), err)
+	} else if k8serr.IsNotFound(err) {
+		err = resources.DoOpenshiftCreatePodInANamespace(openshiftAPIURL, token, projectName, podCR)
+		if err != nil {
+			return nil, fmt.Errorf("error occured while making request to create pod %s : %v", podCR.GetName(), err)
+			// t.Fatalf("error occured while creating pod: %v", err)
+		}
+	}
```
Boomatang: As commented before:
```
	if k8serr.IsNotFound(err) {
		err = resources.DoOpenshiftCreatePodInANamespace(openshiftAPIURL, token, projectName, podCR)
		if err != nil {
			return nil, fmt.Errorf("error occured while making request to create pod %s : %v", podCR.GetName(), err)
			// t.Fatalf("error occured while creating pod: %v", err)
		}
	} else if err != nill {
		return nil, fmt.Errorf("error occured while retrieving pod %s : %v", podCR.GetName(), err)
	}
```

test/common/network_policy_access_ns_to_svc.go
```diff
@@ -0,0 +1,278 @@
...
+			return false, fmt.Errorf("error getting pod: %v", err)
+		}
+		for _, cnd := range podCR.Status.Conditions {
+			if cnd.Type == corev1.ContainersReady {
+				if cnd.Status != corev1.ConditionStatus("True") {
+					checkPodStatus = false
+				} else {
+					checkPodStatus = true
+					break
+				}
```
Boomatang: Here is a similar thing. Would it not be better to check for the passing statement first and then handle a failing statement.
```
 if cnd.Status == corev1.ConditionStatus("True") {
					checkPodStatus = true
					break
				} else {
					checkPodStatus = false
				}
```

Boomatang:
I was unable to test the e2e tests locally. The error I got was `error setting up testing idp: error occurred executing testing idp script: exit status 127` and was raised by user_rhmi_developer_permissions.go:30, user_dedicated_admin_permissions.go:44 & network_policy_access_ns_to_svc.go:35

test/common/network_policy_access_ns_to_svc.go
jjaferson: Makes sense but after having a better look at the code I think we can get rid of the `else` statement as it'll only progress if the resources was not found, wdyt?

```
	err := ctx.Client.Get(goctx.TODO(), key, serviceCR)
	if err != nil && !k8serr.IsNotFound(err) {
		return nil, fmt.Errorf("error occured while retrieving service %s : %v", serviceCR.GetName(), err)
	}

	err = resources.DoOpenshiftCreateServiceInANamespace(openshiftAPIURL, token, projectName, serviceCR)
	if err != nil {
		return nil, fmt.Errorf("error occured while making request to create service %s : %v", serviceCR.GetName(), err)
	}
```

jjaferson: I was thinking here after your suggestion and I am not sure if it really matter much as `!k8serr.IsNotFound(err)` will be only checked if `err` is not nil, but for the sake of readability I always prefer to handle the errors after an operation.

jjaferson:
> I was unable to test the e2e tests locally. The error I got was `error setting up testing idp: error occurred executing testing idp script: exit status 127` and was raised by user_rhmi_developer_permissions.go:30, user_dedicated_admin_permissions.go:44 & network_policy_access_ns_to_svc.go:35

The changes made by @ciaranRoche to set up testing idp https://github.com/integr8ly/integreatly-operator/pull/510 need to be added to this PR. I'll add them on and ping you @Boomatang 

thanks

test/common/network_policy_access_ns_to_svc.go
```diff
@@ -0,0 +1,264 @@
...
+		ObjectMeta: v1.ObjectMeta{
+			Name: projectName,
+		},
+	}
+
+	// check if project exist
+	err := ctx.Client.Get(goctx.TODO(), k8sclient.ObjectKey{Name: projectName}, &projectv1.Project{})
+	if err != nil && !k8serr.IsNotFound(err) {
+		return nil, fmt.Errorf("error occured while retrieving a project: %v", err)
+	} else if k8serr.IsNotFound(err) {
```
ciaranRoche: Probably don't need to explicitly declare the `else` statement. if there is an error and it is not an `isNotFound` error this if statement will be called anyway

test/common/network_policy_access_ns_to_svc.go
```diff
@@ -0,0 +1,264 @@
...
+					checkPodStatus = false
+				} else {
+					checkPodStatus = true
+					break
+				}
+			}
+		}
+	}
+	return checkPodStatus, nil
+}
```
ciaranRoche: just a suggestion to shorten this, might be worth wrapping this in a poll, to handle other pod status, like a `crashBackLoop`, an example of a [poll](https://github.com/integr8ly/integreatly-operator/pull/510/files#diff-630e5cb2ca42a372b23875a0a962ee94R120)
```suggestion
	key := k8sclient.ObjectKey{Name: podCR.GetName(), Namespace: podCR.GetNamespace()}
	err := ctx.Client.Get(goctx.TODO(), key, podCR)
	if err != nil {
		return false, fmt.Errorf("error getting pod: %v", err)
	}
	for _, cnd := range podCR.Status.Conditions {
		if cnd.Type == corev1.ContainersReady  && cnd.Status == corev1.ConditionStatus("True") {
			return true, nil
		}
	}
	return false, nil
}
```

test/common/network_policy_access_ns_to_svc.go
jjaferson: That makes sense but only if there is an error,  if there isn't any error `err` will be nil and then it'll attempt to create the CR even though it already exist. 
The reason for the check is to avoid return an error when the CR is present.

PR #553, opened 2020-03-24
Remove pod count as it isn't needed to check pods are ready

CathalOConnorRH:
# Description

removed the pod count as it didn't match the number of pods actually deployed.
It is not needed since the checkPodStatus ensures pods are ready.
To verify, change install type to workshop and install as normal. Solution explorer should deploy along with AMQ Streams. 

## Type of change

## Checklist

### DIFF
--- a/pkg/products/amqstreams/reconciler.go
+++ b/pkg/products/amqstreams/reconciler.go
@@ -214,12 +214,7 @@ func (r *Reconciler) handleProgressPhase(ctx context.Context, client k8sclient.C
 		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to check amq streams installation: %w", err)
 	}
 
-	//expecting 8 pods in total
-	if len(pods.Items) < 8 {
-		return integreatlyv1alpha1.PhaseInProgress, nil
-	}
-
-	//and they should all be ready
+	//checking pods are all ready
 checkPodStatus:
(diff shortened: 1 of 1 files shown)


pkg/products/amqstreams/reconciler.go
```diff
@@ -215,11 +215,11 @@ func (r *Reconciler) handleProgressPhase(ctx context.Context, client k8sclient.C
 	}
 
 	//expecting 8 pods in total
-	if len(pods.Items) < 8 {
+	if len(pods.Items) == 8 {
```
Boomatang: This is a small thing. Can the value of 8 be turned into a variable instead? A well named variable would make the comment above not required.

KevFan: Should this if block be removed if there is no need to check the pod count? :thinking: 

Depending on the decision, some unit tests would need to be updated as the expected phase returned is now different in these tests as they create pods in the ready or unready state:
* https://github.com/integr8ly/integreatly-operator/blob/ee67540e2bf7b580366731804f8650f8dcce2442/pkg/products/amqstreams/reconciler_test.go#L365

* https://github.com/integr8ly/integreatly-operator/blob/ee67540e2bf7b580366731804f8650f8dcce2442/pkg/products/amqstreams/reconciler_test.go#L236

PR #560, opened 2020-03-27
Resolve AMQ Streams install

CathalOConnorRH:
Updated amq streams to version 1.4.0, modified number of replicas from 3 to 2 and changed monitoring to use 48Gi instead of 50Gi due to PV limits.

Change installation type to Workshop and build. 
Confirm there are 4 pods in AMQ Streams project redhat-rhmi-amq-streams and that solution explorer installs. 

# Description

## Type of change

## Checklist

### DIFF
--- a/pkg/products/amqstreams/reconciler.go
+++ b/pkg/products/amqstreams/reconciler.go
@@ -32,9 +32,9 @@ var (
 	defaultInstallationNamespace         = "amq-streams"
 	clusterName                          = "rhmi-cluster"
 	manifestPackage                      = "integreatly-amq-streams"
-	replicas                             = 3
-	offsetsTopicReplicationFactor        = "3"
-	transactionStateLogReplicationFactor = "3"
+	replicas                             = 2
+	offsetsTopicReplicationFactor        = "2"
+	transactionStateLogReplicationFactor = "2"
 )
 
 type Reconciler struct {
@@ -167,7 +167,7 @@ func (r *Reconciler) handleCreatingComponents(ctx context.Context, client k8scli
 		kafka.Name = clusterName
 		kafka.Namespace = r.Config.GetNamespace()
 
-		kafka.Spec.Kafka.Version = "2.1.1"
+		kafka.Spec.Kafka.Version = "2.4.0"
 		if kafka.Spec.Kafka.Replicas < replicas {
 			kafka.Spec.Kafka.Replicas = replicas
 		}
@@ -180,14 +180,14 @@ func (r *Reconciler) handleCreatingComponents(ctx context.Context, client k8scli
 		kafka.Spec.Kafka.Config.TransactionStateLogMinIsr = "2"
 		kafka.Spec.Kafka.Config.LogMessageFormatVersion = "2.1"
 		kafka.Spec.Kafka.Storage.Type = "persistent-claim"
-		kafka.Spec.Kafka.Storage.Size = "10Gi"
+		kafka.Spec.Kafka.Storage.Size = "5Gi"
 		kafka.Spec.Kafka.Storage.DeleteClaim = false
 
 		if kafka.Spec.Zookeeper.Replicas < replicas {
 			kafka.Spec.Zookeeper.Replicas = replicas
 		}
 		kafka.Spec.Zookeeper.Storage.Type = "persistent-claim"
-		kafka.Spec.Zookeeper.Storage.Size = "10Gi"
+		kafka.Spec.Zookeeper.Storage.Size = "5Gi"
 		kafka.Spec.Zookeeper.Storage.DeleteClaim = false
 
 		kafka.Spec.EntityOperator.TopicOperator = kafkav1alpha1.KafkaTopicOperator{}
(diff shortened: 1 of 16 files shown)


pkg/products/amqstreams/reconciler.go
```diff
@@ -161,21 +162,21 @@ func (r *Reconciler) handleCreatingComponents(ctx context.Context, client k8scli
 		kafka.Name = clusterName
 		kafka.Namespace = r.Config.GetNamespace()
 
-		kafka.Spec.Kafka.Version = "2.1.1"
-		kafka.Spec.Kafka.Replicas = 3
+		kafka.Spec.Kafka.Version = "2.4.0"
```
davidffrench: @CathalOConnorRH What is this Kafka version for? is `2.4.0` correct here?

pkg/products/amqstreams/reconciler.go
```diff
@@ -161,21 +162,21 @@ func (r *Reconciler) handleCreatingComponents(ctx context.Context, client k8scli
...
+		kafka.Spec.Kafka.Replicas = 2
 		kafka.Spec.Kafka.Listeners = map[string]kafkav1.KafkaListener{
 			"plain": {},
 			"tls":   {},
 		}
-		kafka.Spec.Kafka.Config.OffsetsTopicReplicationFactor = "3"
-		kafka.Spec.Kafka.Config.TransactionStateLogReplicationFactor = "3"
+		kafka.Spec.Kafka.Config.OffsetsTopicReplicationFactor = "2"
+		kafka.Spec.Kafka.Config.TransactionStateLogReplicationFactor = "2"
```
davidffrench: Do you know what these properties are used for?

pkg/config/monitoring.go
```diff
@@ -112,7 +112,7 @@ func (m *Monitoring) GetPrometheusRetention() string {
 }
 
 func (m *Monitoring) GetPrometheusStorageRequest() string {
-	return "50Gi"
+	return "48Gi"
```
davidffrench: @david-martin Are there any implications for reducing this?

pkg/products/amqstreams/reconciler.go
CathalOConnorRH: If we leave it at 2.1.1, we get errors that the version must be 2.3.0 or 2.4.0. I decided to use 2.4.0 as it would be the newest version to use. 

- 2.3.0=registry.redhat.io/amq7/amq-streams-kafka-23-rhel7:1.4.0
- 2.4.0=registry.redhat.io/amq7/amq-streams-kafka-24-rhel7:1.4.0

pkg/products/amqstreams/reconciler.go
CathalOConnorRH: It wouldn't deploy with these set to the original values so I matched them based on the docs to the number of replicas in use. 

From the docs : 

- offsets.topic.replication.factor: The replication factor for the offsets topic (set higher to ensure availability). Internal topic creation will fail until the cluster size meets this replication factor requirement.

Type: short
Default: 3
Valid Values: [1,...]
Importance: high
Update Mode: read-only

- transaction.state.log.replication.factor: The replication factor for the transaction topic (set higher to ensure availability). Internal topic creation will fail until the cluster size meets this replication factor requirement.

Type: short
Default: 3
Valid Values: [1,...]
Importance: high
Update Mode: read-only

pkg/products/amqstreams/reconciler.go
davidffrench: Sounds good, thanks @CathalOConnorRH

pkg/products/amqstreams/reconciler.go
davidffrench: Perfect, thanks @CathalOConnorRH

pkg/config/monitoring.go
davidffrench: ping @david-martin

david-martin: The amount is brought over from RHMI 1 (ping @alanmoran).
I'm unsure exactly how calculated this number is based on the stable state of metrics on a running 45d window.

However, I'd have concerns about a product that isn't installed by default affecting other products/services that are installed by default. It seems an odd thing to do.
Also, this solution seems to be ignoring the capacity impact of amq streams and should be discussed at the appropriate forum so the impact is more fully understood and planned for.

david-martin:
/hold
putting on hold as there's an impact on the prometheus volume size.
If the volume size is to be changed, there are upgrade considerations that I'm not sure are covered in this PR

matskiv:
@CathalOConnorRH is there a JIRA associated with this work?
This has been open for a long time, we need to make sure this is still something that we plan to do. I know IPT team is looking at installing new AMQ Streams as well...

CathalOConnorRH:
@matskiv The Jira is https://issues.redhat.com/browse/INTLY-5140  
Phil has said to keep it open to track but I think we could close the PR and just keep the jira open ?

matskiv:
@CathalOConnorRH Thanks.
I added comment on JIRA, and for now I'll close this PR.

PR #561, opened 2020-03-30
only cleanup resources when there are no test failures

wei-lee:
# Description
https://issues.redhat.com/browse/DEL-78

## Type of change

- [] Bug fix (non-breaking change which fixes an issue)
- [x] Test improvements

## Checklist

### DIFF
--- a/test/e2e/integreatly_test.go
+++ b/test/e2e/integreatly_test.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"path"
 	"strings"
 	"testing"
 	"time"
@@ -49,6 +50,7 @@ const (
 	authenticationStageTimeout   = time.Minute * 30
 	productsStageTimout          = time.Minute * 30
 	solutionExplorerStageTimeout = time.Minute * 10
+	artifactsDirEnv              = "ARTIFACT_DIR"
 )
 
 func TestIntegreatly(t *testing.T) {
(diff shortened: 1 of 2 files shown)


test/e2e/integreatly_test.go
```diff
@@ -114,6 +113,11 @@ func TestIntegreatly(t *testing.T) {
 		}
 	})
 
+	t.Fatal("Failed test")
```
mikenairn: I assume this is here for testing purposes?

wei-lee: yes, that's right

wei-lee:
When the test is failed, you can see all logs for rhmi installed products are kept here: https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/integr8ly_integreatly-operator/561/pull-ci-integr8ly-integreatly-operator-master-e2e/2114/artifacts/e2e/pods/, and also the rhmi cr is kept here: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/integr8ly_integreatly-operator/561/pull-ci-integr8ly-integreatly-operator-master-e2e/2114/artifacts/e2e/rhmi.yaml.

wei-lee:
@mikenairn I just realised that we didn't actually delete RHMI during our tests. This is because the RHMI CR is never created during the tests: it is created by the rhmi-operator itself. This means we have always kept the log files from all pods in Prow. See this build that was passed: https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/integr8ly_integreatly-operator/561/pull-ci-integr8ly-integreatly-operator-master-e2e/2107/artifacts/e2e/pods/.

mikenairn:
@wei-lee So the tests used to create the CR but i guess now they don't, it's done upfront before running the tests, and so the logs are still there during archiving. The only change here is we are including the CR in the artifacts?

@StevenTobin You mentioned this as being something that was still an issue? Is capturing the CR all that was missing?

wei-lee:
@mikenairn yes, that's right. Then `cleanup` is called, we do remove some of the resources that are created during the tests. But that's about it. 

I can change the way how RHMI CR is created, but that means the tests will wait for the uninstall to complete, and may take even longer to finish. Probably something we can consider in a separate PR.

Is there anything else we need to do for this PR?

mikenairn:
> Is there anything else we need to do for this PR?

No, if we have the pod logs and now the cr i think that shoudl be all we want from this.

/lgtm

@philbrookes You OK to approve this Testing changes only?

PR #563, opened 2020-03-30
Refactored threescale external resource metric creation

obrienrobert:
## Description
https://issues.redhat.com/browse/INTLY-6490

This PR addresses 2 issues:

1. When installing RHMI using external AWS resources resources, the Threescale Postgres/Redis alerts would trigger initially, and then resolve when the resources were in an available state. This PR refactors the existing Threescale alert creation flow to ensure that the alerts are created only when the resources are in an available state.

threescale_alerts

2. AWS Postgres and Redis alerts were noticed to be firing continuously, on what looked to be every hour. Any AWS resource related Prometheus alerts are now created with `for = '5m'`.

## Type of change

## Verification
- Using this branch, run an install using external Postgres/Redis AWS resources
- Ensure the Postgres/Redis Threescale alerts are created only once the accompanying AWS resources are in an available state, and that they are **not** firing.
- Verify that the alerts are not firing every hour.

### DIFF
--- a/pkg/products/threescale/reconciler.go
+++ b/pkg/products/threescale/reconciler.go
@@ -562,15 +562,6 @@ func (r *Reconciler) reconcileExternalDatasources(ctx context.Context, serverCli
 		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to reconcile backend redis request: %w", err)
 	}
 
-	// create the prometheus availability rule
-	if _, err = resources.CreateRedisAvailabilityAlert(ctx, serverClient, r.installation, backendRedis); err != nil {
-		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to create backend redis prometheus alert for threescale: %w", err)
-	}
-	// create backend connectivity alert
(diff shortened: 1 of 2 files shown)


david-martin:
/approve
This change makes sense.
Needs an lgtm from someone familiar with the alerts. @ciaranRoche can you help with that?

ciaranRoche:
@david-martin @obrienrobert Just looking at these changes, should we be alerting during installation?  I am just thinking in the case a resource provision fails, the RHMI CR will just read `awaiting components` and we will have no alert to capture this failure from the aws resource provisioning side. What do you think?

david-martin:
> we will have no alert to capture this failure from the aws resource provisioning side.

Agreed that there is a gap here.
I believe this gap can be filled separately by a dual approach:
- alerting on the RHMI CR if it isn't 'Ready' within a certain timeframe
- exposing metrics from CRO related to calls to AWS (e.g. http status & response times) and alerting if there are failures

I'm planning on writing up some enhancements related to this soon

PR #568, opened 2020-03-31
[WIP] INTLY-5764 - Test PR (Do not merge)

KevFan:
# Description

**Just testing some stuff - Do not merge**
## Type of change

## Checklist


davidffrench:
Excellent investigation @KevFan . I don't believe there is any way to avoid this based on looking at the links you have sent on. 

Look at https://golang.org/src/flag/flag.go?s=24628:24694#L749 and look at the Var function here - https://golang.org/pkg/flag/?m=src

Unless we somehow switched how we are running the e2e tests which I don't recommend, this will require upstream changes in keycloak operator for now. But this could happen in another dependency in the future so it is no guarantee of success. I would recommend we mitigate this and future issues by not importing our product packages in the tests.

KevFan:
Thanks @davidffrench. Yeah I agree, so I will drop the commits for moving the constants to the product packages in favour of using the constants package instead in the parent PR :+1:

davidffrench:
> Thanks @davidffrench. Yeah I agree, so I will drop the commits for moving the constants to the product packages in favour of using the constants package instead in the parent PR 👍

Perfect, thanks @KevFan

PR #573, opened 2020-04-02
change the alert name to include what redis is being alerted on using the cr name

austincunningham:
# Description
JIRA: https://issues.redhat.com/browse/INTLY-6613

## Type of change
There are Duplicate Alert names for system and backend redis in 3scale , This pr is to generate unique names 

## Verification

- Confirm that the alert rules are updated as follow

image

### DIFF
--- a/pkg/resources/metrics.go
+++ b/pkg/resources/metrics.go
@@ -86,8 +86,7 @@ func CreateRedisAvailabilityAlert(ctx context.Context, client k8sclient.Client, 
 		return nil, nil
 	}
 	productName := cr.Labels["productName"]
-	// remove hyphen as prometheus rules have difficulty with them
-	redisCRName := strings.Replace(cr.Name, "-", "", -1)
+	redisCRName := strings.Title(strings.Replace(cr.Name, "redis-example-rhmi", "", -1))
 	alertName := redisCRName + "RedisCacheUnavailable"
 	ruleName := fmt.Sprintf("availability-rule-%s", cr.Name)
 	alertExp := intstr.FromString(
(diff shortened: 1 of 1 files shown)


austincunningham:
/assign @david-martin 
Haven't verified this yet think I will need a byoc cluster

david-martin:
/approve 
Approving the change, will hold off on lgtm until verified.

austincunningham:
@david-martin verified now, and change added for updating rules on reconcile.

pkg/resources/metrics.go
```diff
@@ -170,6 +171,24 @@ func reconcilePrometheusRule(ctx context.Context, client k8sclient.Client, ruleN
 
 	// create or update the resource
 	_, err := controllerutil.CreateOrUpdate(ctx, client, rule, func() error {
+		rule.Name = ruleName
+		rule.Namespace = ns
+		rule.Spec.Groups = []prometheusv1.RuleGroup{
```
david-martin, line 176: is it possible there are multiple groups, and if so, would it make it easier to overwrite the entirety of `spec.groups` instead of overriding individual fields in 1 group?

david-martin:
image

@austincunningham  The alerts are distinguishable now.
1 minor thing is the naming is different for these 3scale alerts vs. others (Threescale vs. 3Scale)

austincunningham:
@david-martin , the name difference is caused by a use of a different format in the redis CR and the productname, can change it with some string substitution if that is better

pkg/resources/metrics.go
austincunningham: :eyes: Think I tried something like that earlier but didn't appear to update.

austincunningham: just testing 
```
		rule.Spec = prometheusv1.PrometheusRuleSpec{groups}
```

austincunningham:
Test flake, no codeready changes in this pr
```
        --- FAIL: TestIntegreatly/integreatly/Verify_Codeready_CRUDL_permissions (45.83s)
            codeready_crudl.go:97: Test codeready workspace creation
            codeready_crudl.go:125: failed to login to codeready: failed to get access token: {"error":"invalid_grant","error_description":"Code not valid"}
```
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/integr8ly_integreatly-operator/573/pull-ci-integr8ly-integreatly-operator-master-e2e/2245
/test e2e

austincunningham:
@david-martin having had another, postgres is the only one that is using 3scale all other alerts are Threescale will apply this change to postgres as well.

pkg/resources/metrics.go
austincunningham: From what I can see debugging the `groups` object doesn't update inside the controllerutil.CreateOrUpdate block but does outside of that block, not sure why, will revert to original method for updating

austincunningham:
image
duplicate alert names for postgres in rhsso also resolved with latest change.

PR #575, opened 2020-04-02
🐛 Add missing federated metric

grdryn:
# Description

JIRA: https://issues.redhat.com/browse/INTLY-6660

There was  a metric  missing, so the  `PVCStorageWillFillIn4Days`, and
`PVCStorageWillFillIn4Hours`  alerts could  have unexpected  behaviour
(potentially never firing).

I added those metrics to the `PVCStorageMetricsAvailable` alert (if
they're missing, that alert should fire), and I also parameterized
where the namespace prefix was previously hardcoded.

## Type of change

## Checklist

(partially -- by adding those metrics to the other alert)


david-martin:
/approve

Getting 13 results for `kubelet_volume_stats_capacity_bytes` now.

Modified the `PVCStorageMetricsAvailable` alert query to verify an absent metric results in a value.

PR #582, opened 2020-04-06
Update Apicurito to version 7.6.0

briangallagher:
# Description
Update Apicurito to version 7.6.0 from 7.5.0
https://issues.redhat.com/browse/INTLY-5732

# How to test
- Install RHMI
- Open Apicurito via route
- Verify you can create a new API
- Verify you can generate and download a Camel Project (zip file) from the newly created API. Link is in the top right of web app


manifests/integreatly-apicurito/7.5.0/apicuritooperator.v7.5.0.clusterserviceversion.yaml
```diff
@@ -1,173 +0,0 @@
-apiVersion: operators.coreos.com/v1alpha1
```
KevFan, line 1: @philbrookes Are we doing upgrades for Apicurito? If we are, then we'd need to keep this csv and add a `replaces` field to the new one :thinking:

pkg/products/apicurito/reconciler.go
```diff
@@ -185,8 +185,10 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
 
 	or, err := controllerutil.CreateOrUpdate(ctx, serverClient, apicuritoCR, func() error {
 		// Specify 2 pods to provide HA
-		apicuritoCR.Spec.Size = 2
-		apicuritoCR.Spec.Image = "registry.redhat.io/fuse7/fuse-apicurito:1.5"
+		apicuritoCR.Spec.Size = 3
+		//apicuritoCR.Spec.Image = "registry.redhat.io/fuse7/fuse-apicurito:1.5"
```
KevFan, line 189: Is there a reason for keeping this commented out line? :thinking:

briangallagher: @KevFan no, also, I think we can drop Size back to 2. Will update

manifests/integreatly-apicurito/7.5.0/apicuritooperator.v7.5.0.clusterserviceversion.yaml
philbrookes: yes, this should also work for an upgrade. I overlooked this in the jira as we thought it was going into 2.0.0 initially.

briangallagher: Will add back in, thanks

KevFan:
@briangallagher Verified upgrading but has an issue with product version not being updated to the current version of the RHMI CR.

The operator version and product version is set here if the config have an empty namespace when creating the reconciler but this would not be the case when upgrading I think :thinking: 

https://github.com/integr8ly/integreatly-operator/blob/9c064b936916d570ae0a93f17931925f6f3f4192/pkg/products/apicurito/reconciler.go#L61-L65

Operator version looks to be updated becuase it is always returning the hardcoded version number:
https://github.com/integr8ly/integreatly-operator/blob/9c064b936916d570ae0a93f17931925f6f3f4192/pkg/config/apicurito.go#L56-L58

But for the product version, it is always returning the version set in the config
https://github.com/integr8ly/integreatly-operator/blob/9c064b936916d570ae0a93f17931925f6f3f4192/pkg/config/apicurito.go#L52-L54

Screenshot from 2020-04-07 15-02-46

PR #591, opened 2020-04-07
Intly 6588

CathalOConnorRH:
# Description

## Type of change

## Checklist
- [ X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/pkg/products/threescale/reconciler.go
+++ b/pkg/products/threescale/reconciler.go
@@ -418,40 +418,72 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, serverClient k8scl
 			Name:      apiManagerName,
 			Namespace: r.Config.GetNamespace(),
 		},
-	}
-
-	status, err := controllerutil.CreateOrUpdate(ctx, serverClient, apim, func() error {
-
-		apim.Spec = threescalev1.APIManagerSpec{
-			HighAvailability: &threescalev1.HighAvailabilitySpec{
-				Enabled: true,
-			},
+		Spec: threescalev1.APIManagerSpec{
+			HighAvailability: &threescalev1.HighAvailabilitySpec{},
 			APIManagerCommonSpec: threescalev1.APIManagerCommonSpec{
-				WildcardDomain:              r.installation.Spec.RoutingSubdomain,
 				ResourceRequirementsEnabled: &resourceRequirements,
 			},
 			System: &threescalev1.SystemSpec{
 				DatabaseSpec: &threescalev1.SystemDatabaseSpec{
 					PostgreSQL: &threescalev1.SystemPostgreSQLSpec{},
 				},
-				FileStorageSpec: fss,
-				AppSpec:         &threescalev1.SystemAppSpec{Replicas: &[]int64{numberOfReplicas}[0]},
-				SidekiqSpec:     &threescalev1.SystemSidekiqSpec{Replicas: &[]int64{numberOfReplicas}[0]},
+				FileStorageSpec: &threescalev1.SystemFileStorageSpec{
+					S3: &threescalev1.SystemS3Spec{},
+				},
+				AppSpec:     &threescalev1.SystemAppSpec{Replicas: &[]int64{0}[0]},
+				SidekiqSpec: &threescalev1.SystemSidekiqSpec{Replicas: &[]int64{0}[0]},
 			},
 			Apicast: &threescalev1.ApicastSpec{
-				ProductionSpec: &threescalev1.ApicastProductionSpec{Replicas: &[]int64{numberOfReplicas}[0]},
-				StagingSpec:    &threescalev1.ApicastStagingSpec{Replicas: &[]int64{numberOfReplicas}[0]},
+				ProductionSpec: &threescalev1.ApicastProductionSpec{Replicas: &[]int64{0}[0]},
+				StagingSpec:    &threescalev1.ApicastStagingSpec{Replicas: &[]int64{0}[0]},
 			},
 			Backend: &threescalev1.BackendSpec{
-				ListenerSpec: &threescalev1.BackendListenerSpec{Replicas: &[]int64{numberOfReplicas}[0]},
-				WorkerSpec:   &threescalev1.BackendWorkerSpec{Replicas: &[]int64{numberOfReplicas}[0]},
-				CronSpec:     &threescalev1.BackendCronSpec{Replicas: &[]int64{numberOfReplicas}[0]},
+				ListenerSpec: &threescalev1.BackendListenerSpec{Replicas: &[]int64{0}[0]},
+				WorkerSpec:   &threescalev1.BackendWorkerSpec{Replicas: &[]int64{0}[0]},
+				CronSpec:     &threescalev1.BackendCronSpec{Replicas: &[]int64{0}[0]},
 			},
 			Zync: &threescalev1.ZyncSpec{
-				AppSpec: &threescalev1.ZyncAppSpec{Replicas: &[]int64{numberOfReplicas}[0]},
-				QueSpec: &threescalev1.ZyncQueSpec{Replicas: &[]int64{numberOfReplicas}[0]},
+				AppSpec: &threescalev1.ZyncAppSpec{Replicas: &[]int64{0}[0]},
+				QueSpec: &threescalev1.ZyncQueSpec{Replicas: &[]int64{0}[0]},
 			},
+		},
+	}
+
+	status, err := controllerutil.CreateOrUpdate(ctx, serverClient, apim, func() error {
+
+		apim.Spec.HighAvailability = &threescalev1.HighAvailabilitySpec{Enabled: true}
+		apim.Spec.APIManagerCommonSpec.ResourceRequirementsEnabled = &resourceRequirements
+		apim.Spec.APIManagerCommonSpec.WildcardDomain = r.installation.Spec.RoutingSubdomain
+		apim.Spec.System.FileStorageSpec = fss
+
+		if *apim.Spec.System.AppSpec.Replicas < numberOfReplicas {
+			*apim.Spec.System.AppSpec.Replicas = numberOfReplicas
 		}
+		if *apim.Spec.System.SidekiqSpec.Replicas < numberOfReplicas {
+			*apim.Spec.System.SidekiqSpec.Replicas = numberOfReplicas
+		}
+		if *apim.Spec.Apicast.ProductionSpec.Replicas < numberOfReplicas {
+			*apim.Spec.Apicast.ProductionSpec.Replicas = numberOfReplicas
+		}
+		if *apim.Spec.Apicast.StagingSpec.Replicas < numberOfReplicas {
+			*apim.Spec.Apicast.StagingSpec.Replicas = numberOfReplicas
+		}
+		if *apim.Spec.Backend.ListenerSpec.Replicas < numberOfReplicas {
+			*apim.Spec.Backend.ListenerSpec.Replicas = numberOfReplicas
+		}
+		if *apim.Spec.Backend.WorkerSpec.Replicas < numberOfReplicas {
+			*apim.Spec.Backend.WorkerSpec.Replicas = numberOfReplicas
+		}
+		if *apim.Spec.Backend.CronSpec.Replicas < numberOfReplicas {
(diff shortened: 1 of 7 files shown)


davidffrench:
👀

test/common/scale_up_replicas_threescale.go
```diff
@@ -0,0 +1,152 @@
...
+	numberOfReplicas int64 = 2
+	scaleUpReplicas  int64 = 3
+	name                   = "3scale"
+	namespace              = "redhat-rhmi-3scale"
+)
+
+func TestReplicasInThreescale(t *testing.T, ctx *TestingContext) {
+
+	createWorkspaceRetryInterval := time.Second * 20
+	createWorkspaceTimeout := time.Minute * 7
```
davidffrench, line 26:
```suggestion
	retryInterval := time.Second * 20
	timeout := time.Minute * 7
```
Just a name change here. A workspace is a CodeReady workspace concept

philbrookes:
Hey Cathal, these changes look good. Is 3scale the only product with a value like this? Are there other places where we can implement the same logic?

philbrookes:
this needs a rebase also.

test/common/scale_up_replicas_threescale.go
```diff
@@ -0,0 +1,152 @@
...
+	if err := checkNumberOfReplicasAgainstValue(apim, numberOfReplicas, createWorkspaceRetryInterval, createWorkspaceTimeout, t); err != nil {
+		t.Fatalf("Incorrect number of replicas : %v", err)
+	}
+}
+
+func getAPIManager(dynClient k8sclient.Client, t *testing.T) (threescalev1.APIManager, error) {
+	apim := &threescalev1.APIManager{}
+
+	if err := dynClient.Get(goctx.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, apim); err != nil {
+		t.Fatalf("Error : %v ", err)
```
davidffrench, line 63:
```suggestion
		return nil, err
```
Would be good to return the error here. You are error handling anyway in the main function. You can avoid passing in `t` into the function.

CathalOConnorRH:
@philbrookes I'm not sure, I'll check it out and if there is I'll create a ticket for those be updated in the same way.

test/common/scale_up_replicas_threescale.go
davidffrench, line 26: You should be able to define these as consts as well above.

test/common/scale_up_replicas_threescale.go
```diff
@@ -0,0 +1,152 @@
...
+	createWorkspaceRetryInterval := time.Second * 20
+	createWorkspaceTimeout := time.Minute * 7
+
+	apim, err := getAPIManager(ctx.Client, t)
+	if err != nil {
+		t.Fatalf("failed to get APIManager : %v", err)
+	}
+
+	t.Log("Checking correct number of replicas are set")
+	if err := checkNumberOfReplicasAgainstValue(apim, numberOfReplicas, createWorkspaceRetryInterval, createWorkspaceTimeout, t); err != nil {
```
davidffrench, line 34: You shouldn't need to pass in the numberOfReplicas as it is defined as a const.

test/common/scale_up_replicas_threescale.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		t.Fatalf("Unable to update : %v", err)
+	}
+
+	t.Log("Checking correct number of updated replicas are set")
+	if err := checkNumberOfReplicasAgainstValue(apim, scaleUpReplicas, createWorkspaceRetryInterval, createWorkspaceTimeout, t); err != nil {
+		t.Fatalf("Incorrect number of replicas : %v", err)
+	}
+
+	apim, err = updateAPIManager(ctx.Client, t, numberOfReplicas)
```
davidffrench, line 48: This should be passing in a lower numberOfReplicas such as 1. To ensure the operator changes it back to 2.

test/common/scale_up_replicas_threescale.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	return *apim, nil
+}
+
+func updateAPIManager(dynClient k8sclient.Client, t *testing.T, replicas int64) (threescalev1.APIManager, error) {
+
+	apim, err := getAPIManager(dynClient, t)
+	if err != nil {
+		t.Fatalf("Error : %v ", err)
```
davidffrench, line 73: Same as above, I would just return the error here.

test/common/scale_up_replicas_threescale.go
```diff
@@ -0,0 +1,152 @@
...
+			},
+			Zync: &threescalev1.ZyncSpec{
+				AppSpec: &threescalev1.ZyncAppSpec{Replicas: &[]int64{replicas}[0]},
+				QueSpec: &threescalev1.ZyncQueSpec{Replicas: &[]int64{replicas}[0]},
+			},
+		},
+	}
+
+	if err := dynClient.Update(goctx.TODO(), newApim, &k8sclient.UpdateOptions{}); err != nil {
+		t.Fatalf("Error : %v ", err)
```
davidffrench, line 114: Same as above, I would just return the error here.

test/common/scale_up_replicas_threescale.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func updateAPIManager(dynClient k8sclient.Client, t *testing.T, replicas int64) (threescalev1.APIManager, error) {
+
+	apim, err := getAPIManager(dynClient, t)
+	if err != nil {
+		t.Fatalf("Error : %v ", err)
+	}
+	resourceRequirements := true
+	newApim := &threescalev1.APIManager{
```
davidffrench, line 76: Thankfully you don't need to define the entire object again. You are retrieving the APIManager CR just above. You can use `apim` here and just modify the replica values.

test/common/scale_up_replicas_threescale.go
```diff
@@ -0,0 +1,152 @@
...
+		t.Fatalf("Error : %v ", err)
+	}
+
+	return *newApim, nil
+}
+
+func checkNumberOfReplicasAgainstValue(apim threescalev1.APIManager, numberOfRequiredReplicas int64, retryInterval, timeout time.Duration, t *testing.T) error {
+	return wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		if *apim.Spec.System.AppSpec.Replicas != numberOfRequiredReplicas {
+			return false, fmt.Errorf("Number of replicas not correct")
```
davidffrench, line 123: In all of these errors, it would be good to know the current cluster replicas vs what was expected. If the tests do fail in the future, it will be much easier to investigate what has happened.

test/common/scale_up_replicas_threescale.go
```diff
@@ -0,0 +1,152 @@
...
+
+	if err := dynClient.Update(goctx.TODO(), newApim, &k8sclient.UpdateOptions{}); err != nil {
+		t.Fatalf("Error : %v ", err)
+	}
+
+	return *newApim, nil
+}
+
+func checkNumberOfReplicasAgainstValue(apim threescalev1.APIManager, numberOfRequiredReplicas int64, retryInterval, timeout time.Duration, t *testing.T) error {
+	return wait.Poll(retryInterval, timeout, func() (done bool, err error) {
```
davidffrench, line 121: The APIManager CR will need to be retrieved again from the cluster when polling to get the updated version.

test/common/tests.go
```diff
@@ -29,6 +29,6 @@ var (
 		{"C01 - Verify Alerts are not firing apart from DeadMansSwitch", TestIntegreatlyAlertsFiring},
 		{"B05 - Verify Codeready CRUDL permissions", TestCodereadyCrudlPermisssions},
 		{"A09 - Verify Subscription Install Plan Strategy", TestSubscriptionInstallPlanType},
-		{"B06 - Verify users with no email get default email", TestDefaultUserEmail},
+		{"Verify users with no email get default email", TestDefaultUserEmail},
```
davidffrench, line 32: Would be good to follow Davides steps and see if this is replacing a current manual test, the links are at the top of this file

philbrookes:
Make sure to squash these commits into a single commit too

test/common/scale_up_replicas_threescale.go
```diff
@@ -0,0 +1,166 @@
...
+
+	t.Log("Checking correct number of updated replicas are set")
+	if err := checkNumberOfReplicasAgainstValue(apim, ctx, scaleDownReplicas, retryInterval, timeout, t); err != nil {
+		t.Fatalf("Incorrect number of replicas : %v", err)
+	}
+
+	apim, err = updateAPIManager(ctx.Client, numberOfReplicas)
+	if err != nil {
+		t.Fatalf("Unable to update : %v", err)
+	}
```
davidffrench, line 61: I don't think we need to check the scale down replias were set. It will be enough to check they are reset to the minimum below

test/common/scale_up_replicas_threescale.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func checkNumberOfReplicasAgainstValue(apim threescalev1.APIManager, ctx *TestingContext, numberOfRequiredReplicas int64, retryInterval, timeout time.Duration, t *testing.T) error {
+	return wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		apim, err = getAPIManager(ctx.Client)
+		if err != nil {
+			t.Fatalf("failed to get APIManager : %v", err)
+		}
+		if *apim.Spec.System.AppSpec.Replicas != numberOfRequiredReplicas {
+			return false, fmt.Errorf("Number of replicas for apim.Spec.System.AppSpec is not correct : %v", *apim.Spec.System.AppSpec.Replicas)
```
davidffrench, line 137:
```suggestion
			return false, fmt.Errorf("Number of replicas for apim.Spec.System.AppSpec is not correct : Replicas - %v, Expected - %v", *apim.Spec.System.AppSpec.Replicas, numberOfRequiredReplicas)
```
Would be good to also list the expected replicas in this output.

test/common/tests.go
```diff
@@ -30,5 +30,8 @@ var (
 		{"B05 - Verify Codeready CRUDL permissions", TestCodereadyCrudlPermisssions},
 		{"A09 - Verify Subscription Install Plan Strategy", TestSubscriptionInstallPlanType},
 		{"B06 - Verify users with no email get default email", TestDefaultUserEmail},
+		{"F05 - Verify Replicas Scale correctly in Threescale", TestReplicasInThreescale},
+		{"F06 - Verify Replicas Scale correctly in Apicurito", TestReplicasInApicurito},
+		{"F07 - Verify Replicas Scale correctly in Amq Streams", TestReplicasInAmqstreams},
```
davidffrench, line 35: @CathalOConnorRH You are correct by the looks of it. The E2E tests are only checking the managed type. This is a larger issue to resolve and should be outside of this PR @philbrookes in my opinion.

For now, I would remove this line here so the AMQ Streams test does not get run and open a new JIRA please for adding the ability to support different installation types in the e2e tests.

CathalOConnorRH: I've updated that now and created a ticket 
https://issues.redhat.com/browse/INTLY-6767

davidffrench: Thanks @CathalOConnorRH . e2e tests should pass now. 

@philbrookes Are you ok with this path forward?

pkg/products/threescale/reconciler.go
```diff
@@ -418,40 +418,72 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, serverClient k8scl
...
 		},
-	}
-
-	status, err := controllerutil.CreateOrUpdate(ctx, serverClient, apim, func() error {
-
-		apim.Spec = threescalev1.APIManagerSpec{
-			HighAvailability: &threescalev1.HighAvailabilitySpec{
-				Enabled: true,
-			},
+		Spec: threescalev1.APIManagerSpec{
```
philbrookes, line 421: Why do we populate all the spec here, and then do it again in the createOrUpdate method? Also, are the values read into the object in the mutator function, or will they always come through as zero because of this block?

CathalOConnorRH: If we don't have this here the createOrUpdate fails because the objects are nil on initial install, When it gets to the createOrUpdate the values are updated so we can compare required vs current values

PR #593, opened 2020-04-07
add delorean pullsecrets script

StevenTobin:
# Description
As part of the delorean setup to ensure pods can pull from the private delorean repo we can edit the pull-secret secret in the openshift-config namespace to include our secret. The operator will then deploy this to each node and they can then by default pull from the private repo.

## Type of change

## Checklist

## Verification

1. Download token from `docker configuration` tab here https://quay.io/organization/integreatly?tab=robots and place it in the integreatly-operator root directory
2. Run `make cluster/prepare/delorean/pullsecret`
3. Run `make code/run/delorean`
4. Verify that the delorean pull secret is in the pull-secret secret in the openshift-config namespace
5. Verify that the installation completes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,6 +222,10 @@ ifneq ( ,$(findstring image_mirror_mapping,$(IMAGE_MAPPINGS)))
 	@ oc apply -f integreatly-delorean-secret.yml --namespace=$(NAMESPACE)
 endif
 
+.PHONY: cluster/prepare/delorean/pullsecret
+cluster/prepare/delorean/pullsecret:
+	@./scripts/setup-delorean-pullsecret.sh
+
 .PHONY: cluster/cleanup
 cluster/cleanup:
 	@-oc delete -f deploy/integreatly-rhmi-cr.yml --timeout=240s --wait
--- a/scripts/setup-delorean-pullsecret.sh
+++ b/scripts/setup-delorean-pullsecret.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+set -e
+set -o pipefail
+
+CONFIG_PULLSECRET="$(pwd)/config-pullsecret"
+DECODED_PULLSECRET="$(pwd)/config-pullsecret-decoded"
+DELOREAN_PULLSECRET="$(pwd)/integreatly-delorean-auth.json"
+STAGED_PULLSECRET="$(pwd)/staged-pullsecret"
+COMBINED_PULLSECRET="$(pwd)/combined-pullsecret"
+
+if [ ! -f ${DELOREAN_PULLSECRET} ]; then
+  echo "Error: integreatly-delorean-auth.json not found at ${DELOREAN_PULLSECRET}"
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -222,6 +222,10 @@ ifneq ( ,$(findstring image_mirror_mapping,$(IMAGE_MAPPINGS)))
 	@ oc apply -f integreatly-delorean-secret.yml --namespace=$(NAMESPACE)
 endif
 
+.PHONY: cluster/prepare/delorean/pullsecret
```
laurafitzgerald, line 225: @StevenTobin wdy think about putting this into `cluster/prepare/delorean` make target above. It has the check for whether it's a delorean pr included?

StevenTobin: I don't know if we want to do both, since the makefile change you introduced has less impact on the cluster and the one I'm introducing here will cause every node to drain and reboot so for products where we have a pullsecret solution it's much less impact to just use cluster/prepare/delorean.

scripts/setup-delorean-pullsecret.sh
```diff
@@ -0,0 +1,23 @@
...
+  exit 1
+fi
+
+oc get secret pull-secret -n openshift-config -o yaml > "$CONFIG_PULLSECRET"
+sed -i 's+quay.io+quay.io/integreatly/delorean+g' $DELOREAN_PULLSECRET
+yq r $CONFIG_PULLSECRET 'data' | awk '{print $2}' | base64 -d > $DECODED_PULLSECRET
+jq -s -c '{auths: map(.auths) | add}' $DECODED_PULLSECRET $DELOREAN_PULLSECRET | base64 > $STAGED_PULLSECRET
+awk '{ printf "%s", $0 }' $STAGED_PULLSECRET  > $COMBINED_PULLSECRET
+oc patch secret pull-secret -n openshift-config -p='{"data": {".dockerconfigjson": "'$(cat ${COMBINED_PULLSECRET})'"}}'
+rm $CONFIG_PULLSECRET $STAGED_PULLSECRET $COMBINED_PULLSECRET $DECODED_PULLSECRET
```
mikenairn, line 23: What does this do exactly? Can we add a description. Does this wait for something to happen or does it just update and then openshift does what it needs to do behind the scenes?

Makefile
mikenairn: We do not want this to happen all the time, but ideally if it needs to be done it should just happen as part of the `cluster/prepare/delorean` target. I guess for that we would need some way of knowing if a manifest existed in the manifests dir that required this to be done :thinking:

scripts/setup-delorean-pullsecret.sh
StevenTobin: It updates the pull-secret secret which will cause a drain and restart of each node. It doesn't wait because Openshift will handle it all and you can install RHMI while that's happening.

There'll be notes to add to any documentation we put around this that the cluster will be unstable while that is happening.

PR #595, opened 2020-04-08
Add build description to weekly jobs report

matskiv:
# Description
@psturc @pawelpaszki Wdyt?

I decided to improve weekly report generated based on the results of nightly jobs.
Now the "Build information" text is added as a "Note" in the weekly report.
This way we can communicate additional information about the particular build.

Output of the script now looks like this: https://gist.github.com/matskiv/1d2e16d9300ac01fe210e9bc43571b1d

I started adding "Build information" into the results of failed builds this week. It can be seen in Jenkins as well:
Screenshot 2020-04-08 at 18 17 50

## Type of change
Script for pipeline. No impact on product.

### DIFF
--- a/scripts/jenkins/nightly-report.sh
+++ b/scripts/jenkins/nightly-report.sh
@@ -64,7 +64,12 @@ do
       buildUrl=$(cat build.json | jq -r .url)
       getDateString
       buildResult=$(cat build.json | jq -r .result)
-      echo -e "\t#$buildNumber\t$buildDate\tResult: $buildResult\tUrl: $buildUrl\n" >> report.txt
+      echo -e "\n\t#$buildNumber\t$buildDate\tResult: $buildResult\tUrl: $buildUrl" >> report.txt
+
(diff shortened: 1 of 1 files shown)


pawelpaszki:
That looks good Oleg! Just one suggestion (you can ignore if you like :) ), if there is no description, there is no new line added after the build info. Also - I think it would be worth moving this script to rhmi-utils repo, wdyt?

matskiv:
@pawelpaszki new line is added before build info and not in the "Note" output.
Yes, I agree that this repo isn't the best place to host this script. But that requires an update to the pipeline, and I don't have time to do that right now.

matskiv:
E2E test are irrelevant here.
/override ci/prow/e2e

psturc:
👍 Great, thanks for adding this Oleg.
I created a [follow-up jira](https://issues.redhat.com/browse/INTLY-6735) for moving the script to rhmi-utils repo.

PR #596, opened 2020-04-08
Fixing alerts names in e2e tests on AWS

matskiv:
# Description
Fix for test failures seen on the last nightly build of master:
https://master-jenkins-csb-intly.cloud.paas.psi.redhat.com/job/Nightly/job/rhmi-install-master/40/console

Test failures were caused by a change in #573 and went unnoticed in the PR because the tests are only executed when cloud resources are used, which is not the case with PROW. 

Ping @pawelpaszki @psturc @jjaferson 

## Type of change


matskiv:
Changes made to tests here are not executed by PROW, so skipping the e2e tests.

/override ci/prow/e2e

PR #601, opened 2020-04-09
fix: parameter in subscription template

psturc:
# Description
* Tiny fix for a subscription template used for OLM installation


davidffrench:
@psturc The subscription should be created in the rhmi-operator namespace. Are you seeing an issue with that currently?

psturc:
@davidffrench namespace is configured correctly, but `sourceNamespace` should point to `openshift-marketplace` (where rhmi operator source is located)

PR #603, opened 2020-04-09
add docs for creating a new reconciler

philbrookes:
# Description
Adding a guide on how to create a new product reconciler

## Type of change
docs

@matskiv @JameelB @pb82 I would appreciate your thoughts on this.

### DIFF
--- a/guides/ADDING_A_NEW_PRODUCT.md
+++ b/guides/ADDING_A_NEW_PRODUCT.md
@@ -1,0 +1,221 @@
+# Adding a new product's Operator
+The goal of this guide is to show how to add a new product's operator and CR to the range of operator's managed by the
+integreatly-operator. This will touch on a few different areas of the code-base, and explain the purpose of each area.
+
+## Terminology
+This guide assumes familiarity with Openshift and Kubernetes terminology, and terminology specific to the Integreatly-operator
+is covered [here](./TERMINOLOGY.md).
+
+## Areas of code-base to modify
+- Add manifests files for the new operator to `manifests/` directory.
+- The product variables to the `pkg/apis/integreatly/v1alpha1/rhmi_types.go` file.
+- Add product to applicable installation types.
+- A new reconciler for the product in the `pkg/products` directory.
+- Update reconciler factory.
+- A new config for the product in the `pkg/config` directory.
+- Update config manager.
+
+## Add Manifest Files
+Every product has an operator, and every operator is installed and maintained via OLM. To enable a particular version of
+the integreatly-operator to always install a specific version of each product, it maintains it's own set of manifests 
+for each product. To do this:
+- Create a new directory in the `manifests` directory, named: `integreatly-<product-name>`
+- Create a `<product-name>.package.yaml` file, the channel defined in this file must be `rhmi`.
+- Create a directory for the release of the operator to be deployed `<product>-<version>`.
+- Copy all the manifest files for the operator into this directory.
+
+**Note:** The operator installs the operator and product in separate namespaces and leverages the operatorgroup
+functionality of OLM to do this. In order to be compatible with how OLM does this the new CSV must define the namespace
+watched by the operator like so:
+```asciidoc
+- name: <WATCH_NAMESPACE_ENV_VAR>
+  valueFrom:
+    fieldRef:
+      fieldPath: metadata.annotations['olm.targetNamespaces']
+``` 
+
+## Define Product Variables 
+Every product has a series of variables defined in `pkg/apis/integreatly/v1alpha1/rhmi_types.go`, which are used through
+out the code-base, you will also need to define these for the new product.
+- ProductName this must be DNS-valid (i.e. all lower-case, and only alphanumeric and dashes)
+- ProductVersion
+- OperatorVersion
+
+## Add Product to Applicable Installation Types
+For a product to be installed as part of an installation type, the configuration of that installation type needs to be
+updated to include the new product. These installation types are defined in `pkg/controller/installation/types.go`, at
+the time of writing, there are 2 installation types defined in variables in this file:
+- managed defined in `allManagedStages`
+- workshop defined in `allWorkshopStages`
+
+### Deciding when to create a new stage
+New stages are not desirable, as the operator will not progress to the following phase, until everything in the current
+phase has reported that it has completed, so adding new stages can slow down installation time.
+
+The idea of a stage is to allow the operator to complete some action, that will be required in the following stage, if 
+product A is required by product B, product A should probably go in an earlier stage than product B. For example, RHSSO
+requires cloud resource, so the cloud resources operator had to go into it's own stage, prior to the stage that RHSSO is 
+in.
+
+In general, if the new product is another tool for developers to use, it probably belongs in the products stage. Where 
+you add it inside a stage is immaterial; as these are not processed in any particular order, and nothing in a stage 
+should depend on something else in the same stage.
+
+## New Product Reconciler
+The reconciler must implement the `Products.Interface` interface, in order to work with  the installation controller. 
+The methods are defined in more detail below:
+
+### Reconcile
+This is the primary entry point of your reconciler, on each reconcile loop of the integreatly-operator,
+all the logic in here should be written with the assumption that resources may or may not exist, and can
+be created or updated based on their current state. 
+
+This is called every time the stage containing this reconciler is processed, the stages are processed any time a 
+watched CRD is modified, or alternatively, every 10 minutes - and all preceding stages are complete.
+
+#### Useful Notes on Reconciler design
+A few hints and tips that may come in useful when putting a new reconciler together.
+ 
+##### Helper Library: Resources
+There are quite a lot of helpful functions in the resource package for handling the more common tasks, such as setting 
+up the CSV, catalog source, subscription and namespaces. Many examples of this are available in the other reconcilers 
+(such as [Code-Ready](../pkg/products/codeready/reconciler.go), [RHSSO](../pkg/products/rhsso/reconciler.go), etc). 
+
+There is also a helper function to add a finalizer to the RHMI CR, this also requires a function allowing the organisation 
+of the tear-down of the product at uninstall time, again examples of this can are available in the existing reconcilers.
+
+##### Authentication
+Our approach to authentication for products consoles has been to authenticate against Openshift where possible. If the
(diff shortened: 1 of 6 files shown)


guides/ADDING_A_NEW_PRODUCT.md
```diff
@@ -0,0 +1,160 @@
...
+in the installation type. It follows a fairly simple pattern for most products.
+
+## Create a Config Object for the Product
+Each product has a config object, this is used for 2 purposes:
+1. It is persisted in a configmap, so should the operator crash and restart, the values are still available.
+2. The config of one product can be read from the reconciler of another product (e.g. getting realm and namespace of the 
+cluster SSO).
+
+The config object exists in the `pkg/config` directory and must satisfy the `Config.ConfigReadable` interface. The 
+methods of this interface are expanded on below:
```
matskiv, line 132: IMHO: method explanations would be more useful as comments on the ConfigReadable interface source code.

guides/TERMINOLOGY.md
```diff
@@ -0,0 +1,7 @@
+#Terminology
+- Product - A piece of software controlled by an operator (e.g. RHSSO, 3scale, etc)
+- Resource - Any CRD that is managed by the product's operator.
```
matskiv, line 3: IMHO: resource isn't necessarily a CRD, it can reference k8s/OpenShift native objects/resources :)

guides/TERMINOLOGY.md
```diff
@@ -0,0 +1,7 @@
+#Terminology
+- Product - A piece of software controlled by an operator (e.g. RHSSO, 3scale, etc)
+- Resource - Any CRD that is managed by the product's operator. 
+- Reconciler - this refers to the package in the integreatly-operator that is concerned with reconciling a single 
+products operator, resources and configuration.
+- Stage - A set of products that can be installed at the same time as each other.
+- Installation Type - Defines a set of stages to process.
```
matskiv, line 7: From this definition it might not be obvious that installation type also defines set of products that will be installed.

guides/ADDING_A_NEW_PRODUCT.md
```diff
@@ -0,0 +1,160 @@
...
+Before the integreatly-operator will begin an installation, it will initially check the cluster has no existing installs
+of the same products; this is to avoid potential issues of 2 operators trying to act on one resource. 
+
+This function informs the operator of what object it should look for, to check if the product is already installed. The 
+namespace argument is the namespace currently being scanned for existing installations.
+
+For example, codeready looks for a deployment in the scanned namespace with the name "codeready", if found this 
+installation will stall until that product is removed.
+
+## Update Reconciler Factory
```
davidffrench, line 121: Would be good to put a link to this and exactly what needs to be updated.

guides/ADDING_A_NEW_PRODUCT.md
```diff
@@ -0,0 +1,160 @@
...
+
+In general, if the new product is another tool for developers to use, it probably belongs in the products stage. Where 
+you add it inside a stage is immaterial; as these are not processed in any particular order, and nothing in a stage 
+should depend on something else in the same stage.
+
+## New Product Reconciler
+The reconciler must implement the `Products.Interface` interface, in order to work with  the installation controller. 
+The methods are defined in more detail below:
+
+### Reconcile
```
matskiv, line 68: It would be good to explain when, or how often, is the reconcile called.

guides/ADDING_A_NEW_PRODUCT.md
```diff
@@ -0,0 +1,160 @@
...
+For example, codeready looks for a deployment in the scanned namespace with the name "codeready", if found this 
+installation will stall until that product is removed.
+
+## Update Reconciler Factory
+The reconciler factory is used by the installation_controller to build your reconciler when it comes across your product
+in the installation type. It follows a fairly simple pattern for most products.
+
+## Create a Config Object for the Product
+Each product has a config object, this is used for 2 purposes:
+1. It is persisted in a configmap, so should the operator crash and restart, the values are still available.
```
davidffrench, line 127: > It

What exactly is persisted in the configmap?

guides/TERMINOLOGY.md
```diff
@@ -0,0 +1,7 @@
+#Terminology
```
davidffrench, line 1:
```suggestion
# Terminology
```

matskiv:
Few bits that would be "nice to have":
- tests - unit and functional e2e
- custom types vendoring if product operator is in Go
- auth recomendations (use OpenShift for auth if possible, or cluster SSO if OpenShift is not supported. How we use groups, etc.)

guides/ADDING_A_NEW_PRODUCT.md
```diff
@@ -0,0 +1,160 @@
...
+``` 
+
+## Define Product Variables 
+Every product has a series of variables defined in `pkg/apis/integreatly/v1alpha1/rhmi_types.go`, which are used through
+out the code-base, you will also need to define these for the new product.
+- ProductName this must be DNS-valid (i.e. all lower-case, and only alphanumeric and dashes)
+- ProductVersion
+- OperatorVersion
+
+## Add Product to Applicable Installation Types.
```
JameelB, line 44: Remove `.` from title

guides/TERMINOLOGY.md
davidffrench, line 3: I don't think this should be here in my opinion, Resource is a common k8s term.

matskiv:
This would probably be a separate doc, but we should also explain how a new product can be added into Solution Explorer.

guides/ADDING_A_NEW_PRODUCT.md
```diff
@@ -0,0 +1,160 @@
...
+This is a pointer to the this reconciler's product in the status block of the CR, it can be used to set values
+such as version, host and operator version.
+
+##### serverClient
+This is the client to the cluster, and is used for getting, creating, updating and deleting resources in the cluster.
+
+#### Return Values
+The return values from this method are `state` and `err`:
+##### State
+This is communicated back to the user via the status block of the RHMI CR, this is usually either in progress
```
JameelB, line 101: Could also mention here that the available options for this state are these 3 rhmi status phase types: https://github.com/integr8ly/integreatly-operator/blob/master/pkg/apis/integreatly/v1alpha1/rhmi_types.go#L24-L26

philbrookes:
thanks @matskiv I'll update this doc

guides/ADDING_A_NEW_PRODUCT.md
```diff
@@ -0,0 +1,160 @@
...
+the namespace prefix. 
+
+##### product 
+This is a pointer to the this reconciler's product in the status block of the CR, it can be used to set values
+such as version, host and operator version.
+
+##### serverClient
+This is the client to the cluster, and is used for getting, creating, updating and deleting resources in the cluster.
+
+#### Return Values
```
JameelB, line 98: We should also mention creating events using the events handler functions in [resources/events](https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/events/events.go). If a product reconciles successfully, it should call the event product complete handler and if an error was returned, it should call out the event error handler.

guides/TERMINOLOGY.md
philbrookes: removed this line as my resources definition essentially matches a kubernetes resources definition now.

guides/TERMINOLOGY.md
philbrookes: updated this.

guides/ADDING_A_NEW_PRODUCT.md
philbrookes: added a paragraph on this.

guides/TERMINOLOGY.md
philbrookes: cool, removed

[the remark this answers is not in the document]
philbrookes: I've removed this line now.

pkg/config/manager.go
```diff
@@ -60,12 +60,27 @@ type ConfigReadWriter interface {
...
 	GetProductVersion() integreatlyv1alpha1.ProductVersion
+
+	//GetOperatorVersion returns the value of the globally defined OperatorVersion
 	GetOperatorVersion() integreatlyv1alpha1.OperatorVersion
+
+	//GetHost returns a URL that can be used to access the product, either an API, or console, or blank if not applicable.
 	GetHost() string
+
+	//GetWatchableCRDs should return an array of CRDs that should be watched by the integreatly-operator, if a change of one of these CRDs
+	//in any namespace is detected, it will trigger a full reconcile of the integreatly-operator. This usually returns all of
```
matskiv, line 79: I just checked the implementation, and it only watches for changes in the operand namespace:
https://github.com/integr8ly/integreatly-operator/blob/f53be4cea890ff2d61f798530b8b1e1b7a63f3c3/pkg/controller/installation/installation_controller.go#L542-L549

philbrookes: thanks, updated!

guides/ADDING_A_NEW_PRODUCT.md
philbrookes: good thinking, added!

guides/ADDING_A_NEW_PRODUCT.md
philbrookes: done, thanks!

guides/ADDING_A_NEW_PRODUCT.md
philbrookes: > This would probably be a separate doc, but we should also explain how a new product can be added into Solution Explorer.

https://issues.redhat.com/browse/INTLY-6774

philbrookes: updated to be more specific

philbrookes:
> This would probably be a separate doc, but we should also explain how a new product can be added into Solution Explorer.

https://issues.redhat.com/browse/INTLY-6774

guides/ADDING_A_NEW_PRODUCT.md
philbrookes: Good thinking, I've added quite a few links to stuff now.

philbrookes:
@matskiv @davidffrench @JameelB any further thoughts on this PR, or OK to merge?

matskiv:
LGTM 👍

PR #605, opened 2020-04-14
Allow integreatly operator to approve its own install plan

davidffrench:
# Description
https://issues.redhat.com/browse/INTLY-6170

Adding ability for RHMI operator to approve its own manual approval install plan when the release is non-service affecting. Currently, the detection for service affecting or not is based on an annotation for the new integreatly-operator CSV.

## Verification
Verification is a little complicated since this is dealing with OLM and upgrades.

1. Run `PREVIOUS_TAG=2.0.0 TAG=2.0.1 make gen/csv`
2. Manually modify new CSV
-- Change containerImage and operator deployment image field to my image `quay.io/davidffrench/integreatly-operator:vINTLY-6170`
-- Add `displayName: RHMI Installation` to spec.customresourcedefinitions.owned
3. Push the CSV to your application registry. `QUAY_USERNAME=<your username> QUAY_PASSWORD=<your password> REPO=<your repo name, I used my username> make push/csv`
4. Setup OperatorSource on cluster
```
kind: OperatorSource
metadata:
  name: RHMI
spec:
  authorizationToken: {}
  displayName: RHMI Operators
  registryNamespace: <YOUR QUAY REPO>
  type: appregistry
```
5. Run `make cluster/prepare/project && make cluster/prepare/smtp && make cluster/prepare/dms && make cluster/prepare/pagerduty`
6. Go through cluster OLM UI to install operator in `redhat-rhmi-operator` namespace with manual approval.
7. Manually approve Install Plan on cluster
8. Rerun steps 1 & 2 but with `PREVIOUS_TAG=2.0.1 TAG=2.0.2 ` set
9. Add  `serviceAffecting: "false"` to CSV metadata.annotations
10. Rerun step 3
11. On cluster, in the openshift-marketplace namespace, delete the rhmi-<> pod.
-- Note: This will allow the new CSV to be picked up from your application registry immediately
12. Verify the integreatly-operator automatically approves its own new install plan.

## Type of change

## Checklist

### DIFF
--- a/pkg/controller/installation/installation_controller.go
+++ b/pkg/controller/installation/installation_controller.go
@@ -24,13 +24,14 @@ import (
 	"github.com/integr8ly/integreatly-operator/pkg/resources/marketplace"
 
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/record"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


pkg/controller/installation/installation_controller.go
```diff
@@ -202,17 +211,48 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
 	}
 
+	rhmiSubscription, err := r.getIntegreatlyOperatorSubscription(request.NamespacedName.Namespace)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	rhmiUpgradeAvailable := r.isRHMIUpgradeAvailable(rhmiSubscription)
+
+	if rhmiUpgradeAvailable {
```
maleck13, line 227: could simplify with ``` if r.isRHMIUpgradeAvailable(rhmiSubscription) ```

pkg/controller/installation/installation_controller.go
```diff
@@ -202,17 +211,48 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+
+		latestRHMICSV, err := r.getIntegreatlyOperatorCSV(latestRHMIInstallPlan)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		serviceAffectingUpgrade := r.isRHMIUpgradeServiceAffecting(latestRHMICSV)
+
+		if !serviceAffectingUpgrade {
+			err = r.approveRHMIUpgrade(latestRHMIInstallPlan)
```
maleck13, line 245: same as above could inline

pkg/controller/installation/installation_controller.go
davidffrench: Thanks @maleck13 . Done

pkg/controller/installation/installation_controller.go
davidffrench: Thanks @maleck13 . Done

pkg/controller/installation/installation_controller.go
```diff
@@ -202,17 +211,44 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+			return reconcile.Result{}, err
+		}
+
+		latestRHMICSV, err := r.getIntegreatlyOperatorCSV(latestRHMIInstallPlan)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		if !r.isRHMIUpgradeServiceAffecting(latestRHMICSV) {
+			err = r.approveRHMIUpgrade(latestRHMIInstallPlan)
```
matskiv, line 241: A case where a previous upgrade is still in progress is not covered here.

What if we have a non service affecting release out, before scheduled upgrade of a service affecting release happened? 
Once the scheduled time comes, we will approve first InstallPlan, then second InstallPlan(for non affecting release) will be created shortly after first one is applied. The second installPlan might be applied before operator finished all of it's intended upgrade actions(for first upgrade).

Also, ignoring the possibility of two upgrades in a row, we should consider checking that RHMI status.phase is PhaseCompleted, as it's probably not a good idea to roll out upgrade if previous reconcile didn't finish completely.

davidffrench: Good spot @matskiv . I will check the phaseCompleted here also. 

In the case of handling multiple upgrades in a row. this will be covered in this spike JIRA - https://issues.redhat.com/browse/INTLY-7273

matskiv: @davidffrench I was concerned that, when a second upgrade is available, operator might be restarted before it finishes first upgrade, because it will approve non service affecting InstallPlan at the beginning of reconcile loop, and new deployment will be rolled out, killing the operator mid-reconcile.
I don't think INTLY-7273 addressed this specific concern.

davidffrench: Based on our research today, OLM will only ever create one pending approval install plan for a subscription and moving the logic to the after the products reconcilers should ensure a second upgrade does not take place until the first is complete.

davidffrench:
/unhold

This is ready for review again. Would really like your input @philbrookes and @matskiv

PR #607, opened 2020-04-14
federate alert metrics to cluster monitoring

pb82:
# Description

Federate metrics about alerts to the cluster monitoring stack

## Type of change

Adds a new service monitor in a new namespace to be scraped by the cluster monitoring stack (exposing metrics about rhmi alerts).

## Verification steps

1. Install RHMI using this branch
2. Make sure a namespace `redhat-rhmi-middleware-monitoring-federate` is created
3. Make sure the namespace contains a service monitor named `rhmi-alerts-federate`
4. Check the config of the prometheus instance in `openshift-monitoring`: a job to scrape the `rhmi-alerts-federate` service monitor should be present

The cluster prometheus should now have a target for the federated endpoint:

federation

### DIFF
--- a/pkg/products/monitoring/reconciler.go
+++ b/pkg/products/monitoring/reconciler.go
@@ -174,7 +174,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 			}
 			return integreatlyv1alpha1.PhaseInProgress, nil
 		}
+
 		phase, err := resources.RemoveNamespace(ctx, installation, serverClient, r.Config.GetOperatorNamespace())
+		if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+			return phase, err
+		}
+
+		phase, err = resources.RemoveNamespace(ctx, installation, serverClient, r.Config.GetFederationNamespace())
 		if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
 			return phase, err
 		}
--- a/pkg/products/monitoring/reconciler.go
+++ b/pkg/products/monitoring/reconciler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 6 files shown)


david-martin:
Trying this out & reviewing now

briangallagher:
@pb82 Alan confirmed {__name__="ALERTS",alertstate="firing" } is required.

philbrookes:
I think we'll need a change to `deploy/role.yaml` here?

pb82:
@philbrookes the operator already has permissions to create rolebindings: https://github.com/integr8ly/integreatly-operator/blob/master/deploy/role.yaml#L82

Or is there another change required?

philbrookes:
> @philbrookes the operator already has permissions to create rolebindings: https://github.com/integr8ly/integreatly-operator/blob/master/deploy/role.yaml#L82
> 
> Or is there another change required?

oh ok, that's awesome!

pkg/products/monitoring/reconciler.go
```diff
@@ -180,6 +194,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
+	phase, err = r.ReconcileNamespace(ctx, r.Config.GetFederationNamespace(), installation, serverClient)
```
david-martin, line 197: @pb82 Leaving a comment for what was discussed on a call.
This ReconcileNamespace fn will set a label of `monitoring-key:middleware` on the namespace, which causes any prometheus operator CRDs in the namespace to be reconciled by the rhmi prometheus operator instance.
This manifests as 1 new ServiceMonitor being reconciled by both the rhmi & cluster prometheus operators.

Recommendation is to have specific logic in this file for creating the namespace so that generic rhmi stuff doesn't get applied to it.

philbrookes:
Has this been verified by someone?

philbrookes:
oh yes, verified by @david-martin 
/lgtm
/approve

david-martin:
late to comment, but /lgtm
Verified on cluster

PR #612, opened 2020-04-15
INTLY-6766 - fix auth issues affecting nightly tests

matskiv:
# Description
This is a fix for the e2e tests failure we've seen on nightly pipeline
JIRA: https://issues.redhat.com/browse/INTLY-6766

This issue has been caused by the tests code that is adding/updating "testing-idp" related resources. Main problem was with the KeycloakClient CR, which was overridden with a different secret, but OAuth CR was not being updated to use this secret.
Minor bug fixes around error handling and KeycloakUser CR updates.
I also reverted my previous "fix".  

## Type of change

## Checklist


matskiv:
Adding override to avoid blocking merging in case of retest fail
/override ci/prow/e2e

PR #613, opened 2020-04-15
Certain 3scale alerts not firing when pods are down

R-Lawton:
# Description
[JIRA](https://issues.redhat.com/browse/INTLY-6769) the following PR is to fix certain alerts that are counting deploy pods which have already completed and are not running which stops the alerts from triggering.

#Verifcation 
1. install integreatly from this branch
2. Run the following command to bring the relevant pods down 
```
while true; do
  if oc get deployment -n redhat-rhmi-3scale-operator -o json | jq '.items[0].spec.replicas' | grep 1; then
    oc scale deployment 3scale-operator --replicas=0 -n redhat-rhmi-3scale-operator
  fi
  if oc get dc apicast-production -n redhat-rhmi-3scale -o json | jq '.spec.replicas' | grep 2; then
    oc scale dc apicast-production --replicas=0 -n redhat-rhmi-3scale
  fi
  sleep 5
done
```
Substituting apicast-production with a deployment config from below:

- apicast-production
- apicast-staging
- backend-listener
- backend-worker
- system-app
3.  Ensure the following rules fire when the relevant pods are down
- ThreeScaleApicastProductionPod
- ThreeScaleApicastStagingPod
- ThreeScaleBackendListenerPod
- ThreeScaleBackendWorkerPod
- ThreeScaleSystemAppPod
Screenshot from 2020-04-16 10-46-01

### DIFF
--- a/templates/monitoring/kube_state_metrics_3scale_alerts.yaml
+++ b/templates/monitoring/kube_state_metrics_3scale_alerts.yaml
@@ -13,7 +13,7 @@ spec:
           message: >-
             3Scale apicast-staging has no pods in a ready state.
         expr: |
-          absent(kube_pod_status_ready{namespace="{{ index .Params "Namespace" }}", condition="true", pod=~"apicast-staging.*"})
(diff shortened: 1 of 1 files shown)


obrienrobert:
The `namespace="redhat-rhmi-3scale"` is hard-coded in each of the alerts. Should we be using `{{ index .Params "Namespace" }}` instead?

R-Lawton:
A lot of the alerts have hardcoded namespaces so im not sure if this Pr would be the best for it @david-martin what do you think?

david-martin:
@R-Lawton Given the changes in this PR are currently removing the `{{ index .Params "Namespace" }}` templating, I think they should be added back.

If there are other alerts besides these that have hardcoded namespaces, please create a jira for them.

R-Lawton:
Perfect will do

obrienrobert:
Changes verified on an OSD cluster.

PR #626, opened 2020-04-17
Functional test:  Code ready login - Add more logging information

Rajagopalan-Ranganathan:
# Description
Jira link: https://issues.redhat.com/browse/INTLY-6679
Added more logging information for the code ready login flow flake. Theoratical understanding is
the code received from oauth is not valid or expired. One possible reason could be username in IDP
is not yet synced correctly (fixed in: https://issues.redhat.com/browse/INTLY-6766).

## Type of change

## Checklist

### DIFF
--- a/test/common/codeready_crudl.go
+++ b/test/common/codeready_crudl.go
@@ -120,7 +120,7 @@ func TestCodereadyCrudlPermisssions(t *testing.T, ctx *TestingContext) {
 	}
 
 	// login to codeready
-	token, err := loginClient.CodereadyLogin(keycloakHost, redirectUrl)
+	token, err := loginClient.CodereadyLogin(keycloakHost, redirectUrl, t)
 	if err != nil {
 		t.Fatalf("failed to login to codeready: %v", err)
 	}
--- a/test/resources/codeready_login.go
+++ b/test/resources/codeready_login.go
@@ -10,9 +10,9 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"strings"
+	"testing"
 
 	"github.com/PuerkitoBio/goquery"
-	"github.com/sirupsen/logrus"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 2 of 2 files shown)


matskiv:
It would be nicer to use .Log or .Logf funciton from ["testing" package](https://golang.org/pkg/testing/). In case we start running the tests in parallel, it will organize the output of each test correctly.

Rajagopalan-Ranganathan:
> It would be nicer to use .Log or .Logf funciton from ["testing" package](https://golang.org/pkg/testing/). In case we start running the tests in parallel, it will organize the output of each test correctly.

The issue with testing  log or logf would be the information is only printed when the test fails or the test flag of verbose "v" added i think. This would mean that, when the test case fails we get the information, but we would not have logs from a successful test case to compare against. If that is ok, I can modify the logging to use testing package.

matskiv:
@Rajagopalan-Ranganathan I don't think we need the output on each successful test, as it would be fairly similar, right?

Rajagopalan-Ranganathan:
> @Rajagopalan-Ranganathan I don't think we need the output on each successful test, as it would be fairly similar, right?
My point was, if we wanted to reproduce it locally, the issue being a flake, would again require code changes to get the logs for successful case. I will make the changes to use the testing log/logf methods.

matskiv:
Locally, I would run this with `make test/functional`, and [that runs tests in a verbose mode](https://github.com/integr8ly/integreatly-operator/blob/c581a290aad237cf8d271a3144668a94e394a847/Makefile#L146-L148), so we should get the output.
@Rajagopalan-Ranganathan Thanks in advance for the change. 👍

Rajagopalan-Ranganathan:
> Locally, I would run this with `make test/functional`, and [that runs tests in a verbose mode](https://github.com/integr8ly/integreatly-operator/blob/c581a290aad237cf8d271a3144668a94e394a847/Makefile#L146-L148), so we should get the output.
> @Rajagopalan-Ranganathan Thanks in advance for the change. +1

Changes are done.

Rajagopalan-Ranganathan:
Adding override to avoid blocking merging in case of retest fail
/override ci/prow/e2e

PR #629, opened 2020-04-20
add release preparation script

StevenTobin:
# Description
https://issues.redhat.com/browse/DEL-219

Add a script to generate a new csv and edit appropriate files with the new version. 

The script takes in a SEMVER variable which it will use to define the version and tag ( if it exists ) and edit/create associated files.

## Type of change

## Checklist

## Verification

1. Run `make release/prepare` and verify that nothing changes
2. Edit SEMVER in the Makefile to change the rc number
3. Run `make release/prepare` and verify that the images in the csv have changed
4. Edit SEMVER in the Makefile to change the version number
5. Run `make release/prepare` and verify that the new csv has been created and the package, makefile and version.go are all referencing the new version


mikenairn:
@StevenTobin Does this script do the same as this make target? https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L280

StevenTobin:
@mikenairn actually yes, looks like it does, that's all the changes in the ticket and all the references I could find to the version string, maybe this is not required? but what is the intention of the ticket then.

wei-lee:
@StevenTobin sorry, my bad. I thought there's something additional here that needs to be done. But if the everything is covered by the make target, let's just use that and close the ticket.

cc @mikenairn

StevenTobin:
@mikenairn @wei-lee I've actually just made some changes so that this script now supports updating the current csv with a new release ( like going from rc1 to rc2 ) which the current makefile doesn't and may be something we need in the automation.

I also think that managing this behavior in a script rather than a make target is just a better choice, especially since that makefile is already complex and if we need to modify release behavior in future separating the logic out like this will make it easier.

scripts/prepare-release.sh
```diff
@@ -0,0 +1,77 @@
...
+  sed -i "s/PREVIOUS_TAG ?=.*/PREVIOUS_TAG ?=$PREVIOUS_VERSION/g" Makefile
+  sed -i "s/image:.*/image: quay\.io\/integreatly\/integreatly-operator:v$VERSION-$TAG/g" deploy/operator.yaml
+  sed -i "s/containerImage:.*/containerImage: quay\.io\/integreatly\/integreatly-operator:v$VERSION/g" deploy/olm-catalog/integreatly-operator/${VERSION}/integreatly-operator.v${VERSION}.clusterserviceversion.yaml
+  sed -i "s/image:.*/image: quay\.io\/integreatly\/integreatly-operator:v$VERSION-$TAG/g" deploy/olm-catalog/integreatly-operator/${VERSION}/integreatly-operator.v${VERSION}.clusterserviceversion.yaml
+  sed -i "s/version: $VERSION.*/version: $VERSION-$TAG/g" deploy/olm-catalog/integreatly-operator/${VERSION}/integreatly-operator.v${VERSION}.clusterserviceversion.yaml
+}
+
+clean_up() {
+  rm -rf deploy/olm-catalog/integreatly-operator/${PREVIOUS_VERSION}
+}
```
mikenairn, line 40: After talking to @philbrookes i don't think we should do this any more. 2.1.0+ will be setting the replaces and will need the previous versions kept.

scripts/prepare-release.sh
```diff
@@ -0,0 +1,77 @@
...
+set_new_tag() {
+  sed -i "s/image:.*/image: quay\.io\/integreatly\/integreatly-operator:v$PREVIOUS_VERSION-$TAG/g" deploy/operator.yaml
+  sed -i "s/image:.*/image: quay\.io\/integreatly\/integreatly-operator:v$PREVIOUS_VERSION-$TAG/g" deploy/olm-catalog/integreatly-operator/${PREVIOUS_VERSION}/integreatly-operator.v${PREVIOUS_VERSION}.clusterserviceversion.yaml
+  sed -i "s/version: ${PREVIOUS_VERSION}.*/version: ${PREVIOUS_VERSION}-${TAG}/g" deploy/olm-catalog/integreatly-operator/${PREVIOUS_VERSION}/integreatly-operator.v${PREVIOUS_VERSION}.clusterserviceversion.yaml
+}
+
+# Creating a new release with a tag
+set_new_version_with_tag() {
+  sed -i "s/$PREVIOUS_VERSION/$VERSION/g" Makefile
+  sed -i "s/PREVIOUS_TAG ?=.*/PREVIOUS_TAG ?=$PREVIOUS_VERSION/g" Makefile
```
mikenairn, line 31: I don't think "PREVIOUS_TAG" needs to be in the Makefile any more and can just be removed.

scripts/prepare-release.sh
```diff
@@ -0,0 +1,77 @@
...
+  sed -i "s/image:.*/image: quay\.io\/integreatly\/integreatly-operator:v$VERSION-$TAG/g" deploy/operator.yaml
```
mikenairn, line 32: I'm wondering if it makes sense to actually change the image tag in deploy/operator.yaml before generating the csv. Would it not make more sense to just leave this set to master but modify the deployment in the generated csv after?

@philbrookes What do you think about this? Would it not e better to leave master pointing to the master image on quay.io?

scripts/prepare-release.sh
```diff
@@ -0,0 +1,77 @@
...
+  sed -i "s/containerImage:.*/containerImage: quay\.io\/integreatly\/integreatly-operator:v$VERSION/g" deploy/olm-catalog/integreatly-operator/${VERSION}/integreatly-operator.v${VERSION}.clusterserviceversion.yaml
+  sed -i "s/image:.*/image: quay\.io\/integreatly\/integreatly-operator:v$VERSION-$TAG/g" deploy/olm-catalog/integreatly-operator/${VERSION}/integreatly-operator.v${VERSION}.clusterserviceversion.yaml
+  sed -i "s/version: $VERSION.*/version: $VERSION-$TAG/g" deploy/olm-catalog/integreatly-operator/${VERSION}/integreatly-operator.v${VERSION}.clusterserviceversion.yaml
+}
```
mikenairn, line 36: Quite a lot of duplication generally here, i think we should probably try and get it down to one place where each change is being made.

mikenairn:
The make file target doing the same stuff as this should be replaced to call this script instead https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L279-L280

scripts/prepare-release.sh
StevenTobin: I've consolidated the code so each file is manipulated at once to make it clearer

scripts/prepare-release.sh
StevenTobin: I've removed this for now, if @philbrookes says we should be setting the version as we were I can put it back

scripts/prepare-release.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
+set -e
+set -o pipefail
+
+PREVIOUS_VERSION="$(cat deploy/olm-catalog/integreatly-operator/integreatly-operator.package.yaml | grep integreatly-operator | awk -F v '{print $2}')"
+
+create_new_csv() {
+  operator-sdk generate csv --csv-version "$VERSION" --default-channel --csv-channel=rhmi --update-crds --from-version "$PREVIOUS_VERSION"
```
wei-lee, line 8: need to explicitly specify the operator name here using the `--operator-name` flag. otherwise the name of the directory will be used as the operator name.

Makefile
```diff
@@ -5,7 +5,6 @@ NAMESPACE=redhat-rhmi-operator
 PROJECT=integreatly-operator
 REG=quay.io
 SHELL=/bin/bash
-PREVIOUS_TAG ?=2.0.0
```
mikenairn, line 8: You've removed this but not updated the gen/csv target that uses it https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L279-L280

Can you have that make target call the new script? or remove it.

scripts/prepare-release.sh
mikenairn: @StevenTobin If we are going this route I'd update this to point to master https://github.com/integr8ly/integreatly-operator/blob/master/deploy/operator.yaml#L19

@philbrookes @maleck13 Any reason not to do this? Keep master pointing to the master quay.io image?

Makefile
mikenairn: Also need to make sure this is updated https://github.com/integr8ly/integreatly-operator/blob/master/README.md#release

scripts/prepare-release.sh
```diff
@@ -0,0 +1,74 @@
...
+  create_new_csv
+  set_csv
+  set_makefile
+  set_version_go
+fi
+
+# New tag and no new version
+if [[ -n "${TAG}" && -z "${VERSION}" ]]; then
+  set_csv
+fi
```
mikenairn, line 74: Would it not be easier to have "set_tag" and "set_version" functions that just set what they need to set and based on the inputs set VERSION and TAG to appropriate values?

```
set_version() {
     sed -i "s/$PREVIOUS_VERSION/$VERSION/g" version/version.go
     sed -i "s/$PREVIOUS_VERSION/$VERSION/g" Makefile
}

set_tag() {
    sed -i "s/containerImage:.*/containerImage: quay\.io\/integreatly\/integreatly-operator:vTAG/g" deploy/olm-catalog/integreatly-operator/${VERSION}/integreatly-operator.v${VERSION}.clusterserviceversion.yaml
    sed -i "s/image:.*/image: quay\.io\/integreatly\/integreatly-operator:v$TAG/g" deploy/olm-catalog/integreatly-operator/${VERSION}/integreatly-operator.v${VERSION}.clusterserviceversion.yaml
}

if [[ -z "${TAG}" && -z "${VERSION}" ]]; then
  print_usage
fi

if [[ -z "${TAG}" ]]; then
  //Set TAG HERE
fi

if [[ -z "${VERSION}" ]]; then
  //Set VERSION HERE
fi

create_new_csv
update_tag
update_version
```

scripts/prepare-release.sh
```diff
@@ -0,0 +1,74 @@
...
+
+create_new_csv() {
+  operator-sdk generate csv --csv-version "$VERSION" --default-channel --operator-name integreatly-operator --csv-channel=rhmi --update-crds --from-version "$PREVIOUS_VERSION"
+}
+
+set_csv() {
+  if [[ -n "$VERSION" && -n "$TAG" ]]; then
+    sed -i "s/containerImage:.*/containerImage: quay\.io\/integreatly\/integreatly-operator:v$VERSION/g" deploy/olm-catalog/integreatly-operator/${VERSION}/integreatly-operator.v${VERSION}.clusterserviceversion.yaml
+    sed -i "s/image:.*/image: quay\.io\/integreatly\/integreatly-operator:v$VERSION-$TAG/g" deploy/olm-catalog/integreatly-operator/${VERSION}/integreatly-operator.v${VERSION}.clusterserviceversion.yaml
+    sed -i "s/version: $VERSION.*/version: $VERSION-$TAG/g" deploy/olm-catalog/integreatly-operator/${VERSION}/integreatly-operator.v${VERSION}.clusterserviceversion.yaml
```
mikenairn, line 15: What's this for? The version in the CSV should be set by create_new_csv no?

StevenTobin: It does, but I'm treating tag and version seperatly, otherwise the csv would be generated as `integreatly-operator.v9.9.9-<tag>.clusterserviceversion.yaml` which is incorrect

mikenairn: Hmm, i see this has been set to include er1 alright, but i'm not sure why https://github.com/integr8ly/integreatly-operator/blob/master/deploy/olm-catalog/integreatly-operator/integreatly-operator-2.1.0/integreatly-operator.v2.1.0.clusterserviceversion.yaml#L404

@philbrookes @maleck13 Any reason we are setting this version to include a suffix for the build?

philbrookes: I didn't do that for 2.0.0, I think that happened in 2.1.0 as a result of @wei-lee demo. I don't feel that this needs to be here.

mikenairn: @StevenTobin We can drop this part of the update as it's not required, which will let us simplify this script as described here https://github.com/integr8ly/integreatly-operator/pull/629/files#r413598609

scripts/prepare-release.sh
mikenairn: From chat, we might also want to consider making it even more opinionated and only take in 1 input "SEMVER" (1.1.1 or 1.1.1-er1, 1.1.1-cr1 etc..) and infer everything from that:

```
VERSION = ${SEMVER} - any suffix
TAG = v${SEMVER}
```

StevenTobin: I've simplified the script as you suggested, and changed `make gen/csv` to use the new script.

Makefile
```diff
@@ -5,8 +5,8 @@ NAMESPACE=redhat-rhmi-operator
...
 TAG ?= 2.1.0
+SEMVER ?= 2.1.0-rc1
```
mikenairn, line 9: Could TAG just be replaced entirely by SEMVER? There is no difference in the usage of it in this file, and TAG does not actually appear to represent a tag anyway since we apply the v prefix where it is actually used as a tag.

scripts/prepare-release.sh
```diff
@@ -0,0 +1,47 @@
...
+  exit 1
+}
+
+if [[ -z "$SEMVER" ]]; then
+ echo "ERROR: no SEMVER value set"
+ exit 1
+fi
+
+VERSION=$(echo $SEMVER | awk -F - '{print $1}')
+TAG=$(echo $SEMVER | awk -F - '{print $2}')
```
mikenairn, line 32: would TAG not be "v${SEMVER}" ?

Makefile
StevenTobin: Yeah it's used in the INTEGREATLY_OPERATOR_IMAGE var in the Makefile. That doesn't include the -rc* part of the semver, I don't know if it should though

scripts/prepare-release.sh
```diff
@@ -0,0 +1,47 @@
...
+
+if [[ -z "${TAG}" && -z "${VERSION}" ]]; then
```
mikenairn, line 34: Should this be or?

scripts/prepare-release.sh
```diff
@@ -0,0 +1,47 @@
...
+}
+
+set_version() {
+  sed -i "s/$PREVIOUS_VERSION/$VERSION/g" Makefile
+  sed -i "s/$PREVIOUS_VERSION/$VERSION/g" version/version.go
+}
+
+set_images() {
+  sed -i "s/image:.*/image: quay\.io\/integreatly\/integreatly-operator:v$SEMVER/g" deploy/olm-catalog/integreatly-operator/${VERSION}/integreatly-operator.v${VERSION}.clusterserviceversion.yaml
+  sed -i "s/containerImage:.*/containerImage: quay\.io\/integreatly\/integreatly-operator:v$SEMVER/g" deploy/olm-catalog/integreatly-operator/${VERSION}/integreatly-operator.v${VERSION}.clusterserviceversion.yaml
```
mikenairn, line 18: I think we shoudl use TAG here and make sure TAG is set to exactly what we expect a tag to be v1.1.1-er1, v1.1.1 etc..

scripts/prepare-release.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+
+# We have a new version so generate the csv
+if [[ "$VERSION" != "$PREVIOUS_VERSION" ]]; then
+  create_new_csv
+  set_version
+  set_images
+fi
+
+if [[ -n "${TAG}" ]]; then
```
mikenairn, line 45: Should this not always be set to something?

Makefile
mikenairn: The problem here is that this variable was getting used for 2 different things, a VERSION during gen/scv and a TAG for some old commands.that built and pushed an image to quay.

Since nobody should be using those commands to push images to quay.io, i'd be in favour of removing them entirely, but that is probably out of scope of this work and i don't want to delay it any longer.

@philbrookes @maleck13 You any opinion on this? Does anyone actually use these commands https://github.com/integr8ly/integreatly-operator/blob/v2.1.0-er1/Makefile#L117-L130. They shouldn't be for anything going back to our quay registry anyway.

mikenairn:
Tested it locally and worked as expected:

```
$ make gen/csv
ERROR: no SEMVER value set
make: *** [Makefile:280: gen/csv] Error 1
$ make gen/csv SEMVER=wqwqwq
Error: Invalid version string: wqwqwq
make: *** [Makefile:280: gen/csv] Error 1
$ make gen/csv SEMVER=2.1.0-cr1
Valid version string: 2.1.0-cr1
mnairn@deacon integreatly-operator (pr/629) $ git diff deploy/olm-catalog/integreatly-operator/ | grep 2.1.0
diff --git a/deploy/olm-catalog/integreatly-operator/2.1.0/integreatly-operator.v2.1.0.clusterserviceversion.yaml b/deploy/olm-catalog/integreatly-operator/2.1.0/integreatly-operator.v2.1.0.clusterserviceversion.yaml
index da5c0c54..d5814f32 100644
--- a/deploy/olm-catalog/integreatly-operator/2.1.0/integreatly-operator.v2.1.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/integreatly-operator/2.1.0/integreatly-operator.v2.1.0.clusterserviceversion.yaml
-    containerImage: quay.io/integreatly/integreatly-operator:v2.1.0-rc1
+    containerImage: quay.io/integreatly/integreatly-operator:v2.1.0-cr1
-                image: quay.io/integreatly/integreatly-operator:v2.1.0-rc1
+                image: quay.io/integreatly/integreatly-operator:v2.1.0-cr1
$ make gen/csv SEMVER=2.1.0-cr2
Valid version string: 2.1.0-cr2
$ git diff deploy/olm-catalog/integreatly-operator/ | grep 2.1.0
diff --git a/deploy/olm-catalog/integreatly-operator/2.1.0/integreatly-operator.v2.1.0.clusterserviceversion.yaml b/deploy/olm-catalog/integreatly-operator/2.1.0/integreatly-operator.v2.1.0.clusterserviceversion.yaml
--- a/deploy/olm-catalog/integreatly-operator/2.1.0/integreatly-operator.v2.1.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/integreatly-operator/2.1.0/integreatly-operator.v2.1.0.clusterserviceversion.yaml
-    containerImage: quay.io/integreatly/integreatly-operator:v2.1.0-rc1
+    containerImage: quay.io/integreatly/integreatly-operator:v2.1.0-cr2
-                image: quay.io/integreatly/integreatly-operator:v2.1.0-rc1
+                image: quay.io/integreatly/integreatly-operator:v2.1.0-cr2
$ make gen/csv SEMVER=2.1.1-er1
Valid version string: 2.1.1-er1
INFO[0000] Generating CSV manifest version 2.1.1        
INFO[0014] Created deploy/olm-catalog/integreatly-operator/2.1.1/integreatly-operator.v2.1.1.clusterserviceversion.yaml 
$ git diff | grep 2.1.1
+TAG ?= 2.1.1
+- currentCSV: integreatly-operator.v2.1.1
+       Version            = "2.1.1"
+       IntegreatlyVersion = "2.1.1"
```

/lgtm

scripts/prepare-release.sh
philbrookes: I think this is probably fine. Though I have a suspicion this file is used in some e2e tests @psturc @pawelpaszki any thoughts on this?

mikenairn: I'd imagine in that scenario the e2e tests should be run with --image

mikenairn: Actually, the default e2e target always uses image https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L143

philbrookes: ok :+1:

psturc: We're using a pipeline library functions (in RHMI master installation & RHMI upgrade pipelines) for modifying `deploy/operator.yaml` and generating a new CSV, so it shouldn't be affected by this change.
This looks great, we could actually make use of this new script in pipelines. Thanks for letting me know about this.

Makefile
```diff
@@ -278,16 +277,7 @@ endif
 
 .PHONY: gen/csv
 gen/csv:
```
wei-lee, line 279: does it still make sense to call it `gen/csv`, as it looks like the script is doing more than just that?

mikenairn: `gen/csv` was always doing more than just generating a csv, but i agree it is named badly. Maybe `release/prepare` to match the script, `release` or `release/create` might make more sense?

@StevenTobin wdyt?

Makefile
philbrookes: I don't think they are used much anyway, I have no problem with these commands being modified.

Makefile
StevenTobin: Yeah `release/prepare` makes sense, I think `release` or `release/create` implies it does more than just create and edit files locally.

I'll make that change

mikenairn: @StevenTobin Yeah true, `release/prepare` seems fine to me. @wei-lee That work for you?

When you're updating can rebase master into your branch as well. Lot of flake e2e tests at the moment, so if by some miracle they pass i'd like it to merge asap.

PR #633, opened 2020-04-21
INTLY-6568 - namespace recreation e2e test

KevFan:
# Description

Adds product namespace deletion and check recreation as part on an e2e test. As this tests takes a while and should not be run as part of each PR, a `DESTRUCTIVE` flag was added to include or exclude these tests from certain pipelines

Associated PR to update test cases:
* https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases/merge_requests/164

Jira:
* https://issues.redhat.com/browse/INTLY-6568

**NOTE:** there is a manual step necessary for recreation of the 3scale namespace when deleted so this was excluded for now

## Verification
Running e2e test:
* `ORG=<your_quay_registry> DESTRUCTIVE=true make image/build/push make test/e2e`
  * Check that the namespaces gets deleted and recreated when test is executed

Running against existing cluster:
* `DESTRUCTIVE=true  make test/functional`
  * **NOTE:** test would fail at the aws check for resources test if ran against a non-byoc cluster
## Type of change

- [x] Test Automation
## Checklist


test/common/tests.go
```diff
@@ -33,4 +33,9 @@ var (
 		{"F06 - Verify Replicas Scale correctly in Apicurito", TestReplicasInApicurito},
 		{"F08 - Verify Replicas Scale correctly in RHSSO and user SSO", TestReplicasInRHSSOAndUserSSO},
 	}
+
+	NIGHTLY_TESTS = []TestCase{
```
b1zzu, line 37: I like the concept of spit these tests form the rest, but I wouldn't call them nightly, as nightly indicate when the test will be executed and not what kind of test is it. I propose to call them `SYSTEM_TESTS` or even `DESTRUCTIVE_TESTS`. To go even further it could make sense to rename the `AFTER_INSTALL_TESTS` to `HAPPY_PATH_TESTS` or something else

test/functional/integreatly_test.go
```diff
@@ -41,4 +42,23 @@ func TestIntegreatly(t *testing.T) {
 			})
 		}
 	})
+
+	// Do not execute these tests unless NIGHTLY is set to true
+	if os.Getenv("NIGHTLY") != "true" {
+
+		t.Logf("Skipping Nightly tests as NIGHTLY env var is not set to true")
```
b1zzu, line 49: Use `t.Skip` to report that that the suite has been skipped

test/e2e/integreatly_test.go
```diff
@@ -109,6 +109,23 @@ func TestIntegreatly(t *testing.T) {
...
+						testingContext, err = common.NewTestingContext(f.KubeConfig)
+						if err != nil {
+							t.Fatal("failed to create testing context", err)
+						}
+						test.Test(t, testingContext)
+					})
+				}
+			})
+		} else {
+			t.Logf("Skipping Nightly tests as NIGHTLY env var is not set to true")
```
b1zzu, line 127: Same here we should use `t.Skip`

test/common/tests.go
matskiv: I think main differentiator of this test is how long it will take to execute it, and that's why we didn't want it to run on each PR. 
But you have a good point about it being a destructive test... 🤔

test/functional/integreatly_test.go
matskiv: +1, I wanted to suggest the same, docs ;) - https://golang.org/pkg/testing/#T.Skip

test/common/namespace_restoration.go
```diff
@@ -0,0 +1,282 @@
...
+
+			if err != nil {
+				t.Fatalf("Error deleting the namespace - %s: %s", nameSpace, err)
+			}
+
+			t.Logf("Deleted %s namesapce", nameSpace)
+		}
+
+		// Sleep for 10 seconds before deleting finalizers as they can be re-added if namespace is not in a fully terminating state
+		time.Sleep(10 * time.Second)
```
b1zzu, line 117: I can already imagine this may become a problem in the future, is there an easier way to pull the namespace status instead of relying on a fix time sleep?

https://github.com/integr8ly/integreatly-operator/blob/master/test/BEST_PRACTICES.md#dont-sleep

test/common/namespace_restoration.go
```diff
@@ -0,0 +1,282 @@
...
+				},
+			}
+
+			err := ctx.Client.Delete(goctx.TODO(), nameSpaceForDeletion)
+
+			if err != nil {
+				t.Fatalf("Error deleting the namespace - %s: %s", nameSpace, err)
+			}
+
+			t.Logf("Deleted %s namesapce", nameSpace)
```
matskiv, line 113: typo :)

test/common/namespace_restoration.go
KevFan: Yeah, good point, I'll try some things either trying to get the namespace status or do a poll for deleting the finalizers :thinking:

test/common/tests.go
KevFan: Yeah, I'm good with changing the name of the flag to either or, but yeah the idea was to allow longer / certain tests on specifc pipelines (nightly in this case). Do you guys any preference of  `SYSTEM_TESTS` or `DESTRUCTIVE_TESTS` ?

matskiv: IMHO `DESTRUCTIVE_TESTS` is better than `SYSTEM_TESTS`.

b1zzu: Yes `DESTRUCTIVE_TESTS` sounds better, also `LONGRUNNING_TESTS` or `MARATHON_TESTS` if we want to highlight the fact that they take a long time :D

b1zzu:
@KevFan @CathalOConnorRH @matskiv  do you guys know how does the test behaves with a BYOC cluster?

Because the byoc cluster use AWS resources when the cloud resource namespace is deleted it doesn't remove the AWS database/redis/... (as far as I know) , so the question here is what happens when the cloud resource operator is recreated? Would it reattach to the existing resources?

matskiv:
@b1zzu For this test I don't expect any difference between BYOC and non-BYOC.

b1zzu:
@matskiv sound good.. but just out of curiosity, the AWS resources will be removed and recreated or just reconciled?

matskiv:
@b1zzu AWS should not be effected by this. AWS CR are in redhat-rhmi-operator ns, and that one is not removed in this test. Operator will just copy secrets with RDS and elastic credentials into re-created namespaces.

test/common/tests.go
KevFan: Cool, i'll go with `DESTRUCTIVE_TESTS` and having a `DESTRUCTIVE` flag so :+1:

KevFan:
All comments should be addressed now if you guys want to have another look :smiley:

matskiv:
LGTM, I testing the skipping:
`   integreatly_test.go:48: Skipping Destructive tests as DESTRUCTIVE env var is not set to true`
👍

PR #637, opened 2020-04-22
Intly-4776 - Test network policy

jjaferson:
https://issues.redhat.com/browse/INTLY-4776

Test network policy allows cross namespace access to services, so basically a project is created with an endpoint by a user in the dedicated-admin group and a request is made from the 3scale apicast pod to the created endpoint.


davidffrench:
/approve

Haven't verified personally, approving based on verification from other members.

PR #638, opened 2020-04-22
[W.I.P]INTLY-6737-e2e-verify-user-permissions-test-flake

HubertStefanski:
# Description
Fix  *B03 - Verify RHMI Developer User Permissions are Correct* test flake
- Increase timeout to 5 minutes to prevent timeout test flake
- Reverse logic and change subdomain on approval page check function to prevent `Form not found matching expr "[action=approve]" ` test flake
- Add some logging to help troubleshoot if further flakes exist
- Add error handling for io.EOF error returned when checking for openshiftIDP, this error handling is needed to prevent the test flake, the check function will finish and not run again if an error is returned at any stage, this means that in this case the error is returned before the IDP becomes fully available (or such is the theory)
## Type of change
- [:heavy_check_mark: ] Bug fix (non-breaking change which fixes an Issue

## Verification
- Run `make test/e2e/prow` on OSD cluster
- Ensure *B03 - Verify RHMI Developer User Permissions are Correct* passes
- Test flakes may still exists with other tests, this PR only aims to fix B03
- If tests are run more than once in a row on the same cluster with `make test/e2e/prow` be advised of https://issues.redhat.com/browse/INTLY-7026 and ensure that all users are deleted between test runs (Get users by `oc get users` and delete them with `oc delete user xxxxx`)  and  `make cluster/cleanup` has sucessfully cleaned up all resources

### DIFF
--- a/test/resources/openshift_auth.go
+++ b/test/resources/openshift_auth.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"testing"
 	"time"
 
 	"github.com/PuerkitoBio/goquery"
@@ -61,9 +62,10 @@ func DoAuthOpenshiftUser(authPageURL string, username string, password string, h
 	return nil
 }
 
-func OpenshiftIDPCheck(url string, client *http.Client, idp string) (bool, error) {
+func OpenshiftIDPCheck(t *testing.T, url string, client *http.Client, idp string) (bool, error) {
 	browser := surf.NewBrowser()
 	browser.SetTransport(client.Transport)
+	t.Logf("Attempting to open browser with url: %s", url)
 	if err := browser.Open(url); err != nil {
 		return false, fmt.Errorf("failed to open browser url: %w", err)
 	}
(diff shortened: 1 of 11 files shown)


david-martin:
/approve
Thanks @HubertStefanski 
Good to sort out these flakes before the get out of control

austinmartinh:
Verified

test/resources/openshift_auth.go
```diff
@@ -151,7 +153,8 @@ func openshiftClientSetup(url, username, password string, client *http.Client, i
 		return fmt.Errorf("failed to submit login form on oauth proxy screen: %w", err)
 	}
 	//sometimes we'll reach an accept permissions page for the user if they haven't accepted these scope requests before.
-	if strings.Contains(browser.Url().Host, openshiftOauthSubdomain) {
+	//refactored, this approach assumes that if the redirected page is not the console then it looks for an approve action, previous approach would cause e2e test flakes
+	if !strings.Contains(browser.Url().Host, openshiftConsoleSubdomain) {
```
aidenkeating, line 157: @HubertStefanski @david-martin This will break any test that is not performing an auth request that redirects to the OpenShift console or that requires a permissions check. This won't always be the case. If two tests login to the same service that performs a permissions check, the second test will fail as the redirect page:

- Won't be the OpenShift console
- Won't be the permissions request page

It'll be the product homepage, as the permissions check was performed in the first test and isn't repeated. I seem to be able to reproduce this with Grafana pretty consistently. Tests only pass when the first Grafana test (currently a flake) "fails" so the user never performs the permissions flow.

TLDR: For Grafana, I think this will only pass when the auth flow in the "skipped" Grafana test fails.

PR #649, opened 2020-04-28
Add new CRD for customer configuration

briangallagher:
# Description
Add new CRD for customer configuration. This will allow the customer to mange the timing of their upgrades of RHMI, amongst other things

## Type of change
- New feature

## How to Verify
Ensure the RHMI operator is running

Wait a short while for the RHMIConfig reconciler to run. During the bootstrap phase a default customer config CR should be created.

Retrieve the RHMIConfig resource: 
`oc get RHMIConfig rhmi-configuration -o yaml`

Ensure the CR exists and has set the necessary default values:
```
spec.upgrade.alwaysImmediately = false
spec.upgrade.duringNextMaintenance = false
```


test/common/rhmi_config.go
```diff
@@ -0,0 +1,55 @@
...
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"testing"
+
+	goctx "context"
+)
+
+const (
+	RHMIConfigCRName  = "rhmi-config"
+	OperatorNamespace = "redhat-rhmi-operator"
```
KevFan, line 14: There's a `RHMIOperatorNamespace` if you want to use that instead
https://github.com/integr8ly/integreatly-operator/blob/df1dcb34ab3656f994993c4808ba3cfa7f431e11/test/common/types.go#L18

test/common/rhmi_config.go
```diff
@@ -0,0 +1,55 @@
...
+
+// TestIntegreatlyRoutesExist tests that the routes for all the products are created
+func TestRHMIConfigCRs(t *testing.T, ctx *TestingContext) {
+	t.Log("Test rhmi config cr creation")
+
+	rhmiConfig := &v1alpha1.RHMIConfig{
+		TypeMeta: v1.TypeMeta{},
+		ObjectMeta: v1.ObjectMeta{
+			Name:      RHMIConfigCRName,
+			Namespace: OperatorNamespace,
```
KevFan, line 25: There's a `RHMIOperatorNamespace` if you want to use that instead to prevent duplication
https://github.com/integr8ly/integreatly-operator/blob/df1dcb34ab3656f994993c4808ba3cfa7f431e11/test/common/types.go#L18

test/common/rhmi_config.go
```diff
@@ -0,0 +1,55 @@
...
+		},
+	}
+
+	if err := ctx.Client.Create(goctx.TODO(), rhmiConfig); err != nil {
+		if err != nil {
+			t.Fatalf("Failed to create RHMI Config resource %v", err)
+			return
+		}
+	}
```
KevFan, line 34:
```suggestion
	if err := ctx.Client.Create(goctx.TODO(), rhmiConfig); err != nil {
		t.Fatalf("Failed to create RHMI Config resource %v", err)
	}
```
I'm not sure is the inner `if` and `return` neccessary here? :thinking:

test/common/rhmi_config.go
```diff
@@ -0,0 +1,55 @@
...
+	verifyCr(t, ctx)
+}
+
+func verifyCr(t *testing.T, ctx *TestingContext) {
+	t.Log("Verify rhmi config cr creation")
+
+	rhmiConfig := &v1alpha1.RHMIConfig{}
+
+	err := ctx.Client.Get(goctx.TODO(), types.NamespacedName{Name: RHMIConfigCRName, Namespace: OperatorNamespace}, rhmiConfig)
```
KevFan, line 43: Same, just could use `RHMIOperatorNamespace` instead here

pkg/controller/rhmiconfig/rhmiconfig_controller.go
```diff
@@ -0,0 +1,110 @@
+/*
+Copyright YEAR Red Hat, Inc.
```
damienomurchu, line 2:
```suggestion
 Copyright 2020 Red Hat, Inc. 
```

test/common/rhmi_config.go
KevFan: Also what do you think on deleting the resource after the test so the tests can be re-run? Would fail on an `already exists` error if re-run again :thinking:

briangallagher: Yep, will do that. thanks

briangallagher:
@KevFan I've updated the code based on your comments. Also, I've added code to the RHMI bootstrap reconciler to create a default customer config cr. I've also update the verifications steps above. Can you take another look, thanks

KevFan:
@briangallagher Looks good and verified :+1: . Just need an approver to approve the PR
/lgtm

pkg/controller/installation/bootstrapReconciler.go
```diff
@@ -110,6 +120,24 @@ func (r *Reconciler) checkCloudResourcesConfig(ctx context.Context, serverClient
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
 
+func (r *Reconciler) reconcilerRHMIConfigCR(ctx context.Context, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
+	rhmiConfig := &integreatlyv1alpha1.RHMIConfig{
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rhmi-config",
+			Namespace: RHMICustomerConfigNamespace,
```
KevFan, line 128: Just a question, is this always going to be in the RHMI operator namespace? If it is, would probably be better to use `r.ConfigManager.GetOperatorNamespace()` instead of defining the namespace, otherwise do ignore :thinking:

briangallagher: Yes, I think it will remain in the RHMI operator namespace. I'll change that so. @KevFan

philbrookes:
looks like a new flake in that e2e test, see instructions on procedure here:  
https://issues.redhat.com/browse/INTLY-7197

philbrookes:
/override ci/prow/e2e
verified locally by @KevFan

philbrookes:
/override ci/prow/e2e
verified locally by @KevFan

PR #652, opened 2020-04-28
INTLY-6728 Add support for a webhook server running with the operator.

sergioifg94:
# Description
In order to add the ability to implement custom validation for the new customer config CR, set up a Webhook server running with the operator.

Kubernetes allows to create validation webhooks pointing to a service in the cluster. The `controller-runtime` library contains the logic to register webhooks in the server provided by the `Manager` interface, but in order to set up the webhook a `ValidatingWebhookConfiguration` CR must be created pointing to the webhook service. 

These changes create a single source of truth for both the Webhooks endpoint and the CRs that are reconciled, therefore there's no need for manual synchronization between the webhook server and the `ValidatingWebhookConfiguration` CRs.

The webhook server requires TLS certificates to run. The following changes were made to allow the operator to use the certificates following [1]:

* Create a Service pointing to the operator, with an annotation `service.beta.openshift.io/serving-cert-secret-name` that triggers the creation of a secret with a key-pair that can be injected as a volume to the Pod.
* Update the operator Deployment to listen to the `8090` port and mount the certificates.
* Reconcile a ConfigMap where the CA certificate is injected, and obtain it from the operator in order to validate the certificates via the `CABundle` field of `ValidatingWebhookConfiguration`

[1] https://docs.openshift.com/container-platform/4.1/authentication/certificates/service-serving-certificate.html

## Adding webhooks

A singleton instance of the webhook configuration is defined and contains a list of the webhooks to be registered and managed. Adding elements to the list before the startup of the server will guarantee that the webhook will be created and the CR will be reconciled. In order to add a webhook, an implementation of the `WebhookRegister` interface is required. This manages the registration of the webhook against the server. This PR includes two implementations:

1. `ObjectWebhookRegister`, encapsules a webhook that implements the `Validator` or `Defaulter` interface.
2. `AdmissionWebhookRegister` to register webhooks directly to the webhook server with a custom function and path (as opposed to the auto-generated path and validation logic for `ObjectWebhookRegister`)

The following is an example of adding a webhook for the `RHMI` CR (provided that it implements the `Validator` interface):

```go
webhooks.Config.AddWebhook(webhooks.IntegreatlyWebhook{
	Name:      "rhmi",
	Register: webhooks.ObjectWebhookRegister{
		Object: &integreatlyv1alpha1.RHMI{},
	},
	Rule: webhooks.NewRule().
		OneResource("integreatly.org", "v1alpha1", "rhmis").
		ForCreate().
		ForUpdate().
		NamespacedScope(),
})
```

## Verification

In order to verify this, you'll need to run the operator as a Pod. Build your own image and edit the `Deployment` to point to the image.

Create the webhooks service that points to the operator pod:

```sh
make cluster/prepare/service
```

Run the operator

```sh
oc create -f deploy/operator.yaml
```

The operator pod will be created. Verify that:

1. The operator starts successfully and triggers the installation.
2. A `ValidatingWebhookConfiguration` CR is reconciled referencing the `RHMIConfig` CR and the webhooks service that points to the operator pod.
3. Modify the RHMIConfig CR and verify that no error occurs and no error is displayed in the operator log.

## Type of change

## Checklist

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -13,10 +13,16 @@ spec:
         name: rhmi-operator
     spec:
       serviceAccountName: rhmi-operator
+      volumes:
+      - name: webhook-certs
+        secret:
+          secretName: rhmi-webhook-cert
       containers:
         - name: rhmi-operator
           # Replace this with the built image name
           image: quay.io/integreatly/integreatly-operator:master
+          ports:
+            - containerPort: 8090
           command:
           - rhmi-operator
           imagePullPolicy: Always
@@ -27,6 +33,10 @@ spec:
             requests:
               cpu: 40m
               memory: 64Mi
+          volumeMounts:
+          - name: webhook-certs
+            mountPath: "/etc/ssl/certs/webhook"
+            readOnly: true
           env:
             - name: WATCH_NAMESPACE
               valueFrom:
--- a/deploy/webhook-service.yaml
+++ b/deploy/webhook-service.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: rhmi-webhooks
+  namespace: redhat-rhmi-operator
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: rhmi-webhook-cert
+spec:
+  selector:
+    name: rhmi-operator
+  ports:
+    - protocol: TCP
+      port: 443
+      targetPort: 8090
--- a/Makefile
+++ b/Makefile
@@ -171,7 +171,7 @@ cluster/deploy/integreatly-rhmi-cr.yml: deploy/integreatly-rhmi-cr.yml
 	$(call wait_command, oc get RHMI $(INSTALLATION_NAME) -n $(NAMESPACE) --output=json -o jsonpath='{.status.stages.solution-explorer.phase}' | grep -q completed, solution-explorer phase, 10m, 30)
 
 .PHONY: cluster/prepare
-cluster/prepare: cluster/prepare/project cluster/prepare/osrc cluster/prepare/configmaps cluster/prepare/smtp cluster/prepare/dms cluster/prepare/pagerduty cluster/prepare/delorean
+cluster/prepare: cluster/prepare/project cluster/prepare/osrc cluster/prepare/configmaps cluster/prepare/service cluster/prepare/smtp cluster/prepare/dms cluster/prepare/pagerduty cluster/prepare/delorean
 
 .PHONY: cluster/prepare/project
 cluster/prepare/project:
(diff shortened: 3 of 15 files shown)


deploy/webhook-service.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
```
davidffrench, line 1: @sergioifg94 Will this automatically be included in a new CSV generation since it is under the deploy folder?

deploy/operator.yaml
```diff
@@ -13,13 +13,23 @@ spec:
         name: rhmi-operator
     spec:
       serviceAccountName: rhmi-operator
+      volumes:
+      - name: webhook-certs
```
davidffrench, line 17: I am not as familiar with the cluster certs and the usage within pods. Is there any side affects of adding this in?

sergioifg94: @davidffrench 
1. Yes, I'm planning to define those methods so that the new CR implements the `Validator` interface once the PR for the CRD is merged.
2. That's a good question. I'm not too familiar with OLM yet so I'm not sure about that, but the service must be created before the operator pod, as the pod mounts the certificates from the secret created after the service is annotated.

sergioifg94:
@davidffrench 
1. Yes, I'm planning to define those methods so that the new CR implements the `Validator` interface once the PR for the CRD is merged.
2. That's a good question. I'm not too familiar with OLM yet so I'm not sure about that, but the service must be created before the operator pod, as the pod mounts the certificates from the secret created after the service is annotated.

deploy/operator.yaml
sergioifg94: It seems to be supported by Openshift 4. I took this from [the documentation](https://docs.openshift.com/container-platform/4.1/authentication/certificates/service-serving-certificate.html) and it's working very well on the clusters I've tested it with. The use case matches 100% what I needed, as the webhook server from controller-runtime is very restrictive, all you can configure is the directory where the certificates are, so mounting it as a volume does the job well

davidffrench: Perfect, thanks Sergio

davidffrench:
> @davidffrench
> 
> 1. Yes, I'm planning to define those methods so that the new CR implements the `Validator` interface once the PR for the CRD is merged.
Perfect, looking forward to seeing it
> 2. That's a good question. I'm not too familiar with OLM yet so I'm not sure about that, but the service must be created before the operator pod, as the pod mounts the certificates from the secret created after the service is annotated.
This is worth some investigation tomorrow. Two areas to investigate:
1. How to get the new service automatically included when generating the csv
2. How to ensure the service is created before the deployment

CathalOConnorRH:
Install works and no errors 
ValidatingWebhookConfiguration is reconciled and there are no errors when modifying the RHMIConfig. So far looks good to me

PR #654, opened 2020-04-29
add release doc

wei-lee:
# Description

https://issues.redhat.com/browse/DEL-235

## Type of change

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ OPERATOR_SDK_VERSION=0.15.1
 AUTH_TOKEN=$(shell curl -sH "Content-Type: application/json" -XPOST https://quay.io/cnr/api/v1/users/login -d '{"user": {"username": "$(QUAY_USERNAME)", "password": "${QUAY_PASSWORD}"}}' | jq -r '.token')
 TEMPLATE_PATH="$(shell pwd)/templates/monitoring"
 INTEGREATLY_OPERATOR_IMAGE ?= $(REG)/$(ORG)/$(PROJECT):v$(TAG)
+CONTAINER_ENGINE ?= docker
 
 # If openapi-gen is available on the path, use that; otherwise use it through
 # "go run" (slower)
@@ -282,6 +283,11 @@ else
 	rm deploy/integreatly-rhmi-cr.yml.bak
 endif
 	@-oc create -f deploy/integreatly-rhmi-cr.yml
+
+.PHONY: prepare-patch-release
+prepare-patch-release:
+	$(CONTAINER_ENGINE) pull quay.io/integreatly/delorean-cli:master
+	$(CONTAINER_ENGINE) run --rm -e KUBECONFIG=/kube.config -v "${HOME}/.kube/config":/kube.config:z -v "${HOME}/.delorean.yaml:/.delorean.yaml" quay.io/integreatly/delorean-cli:master delorean release openshift-ci-release --config /.delorean.yaml --version $(TAG)
 
 .PHONY: release/prepare
 release/prepare:
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -18,10 +18,8 @@ NOTE: Red Hat VPN is required to access.
 In order to create patch releases, we need to make sure a release branch is created first and it is configured properly. Run the following command:
    
 ```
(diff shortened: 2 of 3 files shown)


RELEASE.md
```diff
@@ -0,0 +1,57 @@
...
+
+NOTE: Red Hat VPN is required to access.
+
+# Prepare for Patch Releases
+
+In order to create patch releases, we need to make sure a release branch is created first and it is configured properly. Run the following command:
+   
+```
+delorean release openshift-ci-release --version 2.1.1-er1
+```
```
wei-lee, line 22: @mikenairn do you think if we can move this to a jenkins pipeline too?

mikenairn: As discussed previously i don't think we should add it to the pipeline, yet anyway. We can look at doing that later i guess, but it might require changes to the way we are triggering certain things that use "docker run".

For now I'd leave it as a step here, i would say though we should probably add it as a make target that calls the containerised delorean cli instead.

```
CONTAINER_ENGINE ?= docker

prepare-patch-release:
	$(CONTAINER_ENGINE) pull quay.io/integreatly/delorean-cli:latest:latest
	$(CONTAINER_ENGINE) run --rm quay.io/integreatly/delorean-cli:latest:latest delorean release openshift-ci-release --version 2.1.1-er1
```

wdyt?

RELEASE.md
```diff
@@ -0,0 +1,47 @@
...
+2. For the job parameters:
+   1. Fill in the `version` value. It should follow the semver spec, something like `2.1.0`, `2.1.0-er1`, `2.1.1-rc1` etc.
+   2. Change the value of `integreatlyOperatorBaseBranch` to the release branch if it's a patch release.
+   3. Check the `updateManagedTenantsOnly` option if you only need to update the `managed-tenants` repo to push a release to edge or stable channels 
+   4. Leave the rest as they are
+3. Click on the *Build* button to start the release. Switch to the blue ocean view to see the progress of the build.
+4. At some point, you will be prompted to review the PR. You should then open the PR link, review it and merge it (make any additional changes if required).
+5. Once done, go back to the Jenkins blue ocean view to confirm the PR is merged to continue.
+6. Wait for the build to finish, and at the end you should see a merge request link to the managed-tenant repo.
+7. Review the MR (and make any additional changes if required). Ping reviewers on the PR once it's ready.
```
mikenairn, line 45: The release pipeline will not currently update the "latest-staging" tag which should always point to the latest release that we create, or whatever the managed tenants repo is updated to anyway. This should obviously be done as part of the release pipeline to do this after the MR is merged (@mikenairn reminder to create a jira), but maybe we should add this in here as a step until that gets automated.

@psturc Do we need to tag both the operator and test harness image here?

psturc: Thanks for mentioning this @mikenairn, the step was actually already in the readme [here](https://github.com/integr8ly/integreatly-operator/pull/654/files/571ff0ab08f4024d5a7e517118b600b1194dd65b#diff-04c6e90faac2675aa89e2176d2eec7d8L246)
We need to tag **only the test harness image** with `latest-staging` tag, because the addon-flow installation (which we're using in nightly pipeline) doesn't need any tag to be specified (it always installs the latest version available in managed-tenants repo).

RELEASE.md
wei-lee: yeah, good idea

wei-lee: I have added the make command now, mind taking another look?

Makefile
```diff
@@ -283,6 +284,11 @@ else
 endif
 	@-oc create -f deploy/integreatly-rhmi-cr.yml
 
+.PHONY: prepare-patch-release
+prepare-patch-release:
+	$(CONTAINER_ENGINE) pull quay.io/integreatly/delorean-cli:master
+	$(CONTAINER_ENGINE) run --rm -e KUBECONFIG=/kube.config -v "${HOME}/.kube/config":/kube.config:z -v "${HOME}/.delorean.yaml:/.delorean.yaml" quay.io/integreatly/delorean-cli:master delorean release openshift-ci-release --config /.delorean.yaml --version $(TAG)
```
mikenairn, line 290: Note this command won;t work until https://github.com/integr8ly/delorean/pull/33 is merged.

PR #660, opened 2020-05-01
INTLY-7158 Semantic validation for RHMIConfig

sergioifg94:
# Description

> Link to JIRA: https://issues.redhat.com/browse/INTLY-7158

Following the implementation of the webhooks server, implement validation for the RHMIConfig.Spec.Upgrade section, and add logic to e2e tests to verify that the validation
is being performed successfully via the `ValidatingWebhookConfiguration` reconciled by
the operator

## Verification steps

1. Follow the verification steps of #652 
2. On the console, find the RHMIConfig CR and attempt to perform the following changes:

| Change | Expected result after save |
| ----------- | ----------------------------------- |
| Set the `applyOn` value to an invalid date format | Error |
| Set the `applyOn` value to a past date (example `2 Jan 2006 15:04`) | Error |
| Set the `applyOn` value to a future date and leave the other values as `false` | Success |
| Set the `applyOn` value to a future date and leave any (or both) values for `applyImmediately` and `duringNextMaintenance` as `true` | Error |

## Type of change

## Checklist

### DIFF
--- a/deploy/crds/integreatly.org_rhmiconfigs_crd.yaml
+++ b/deploy/crds/integreatly.org_rhmiconfigs_crd.yaml
@@ -34,11 +34,15 @@ spec:
             backup:
               properties:
                 applyOn:
+                  description: 'apply-on: string, day time. Format: "DDD hh:mm" >
+                    "wed 20:00". UTC time'
                   type: string
               type: object
             maintenance:
               properties:
                 applyFrom:
+                  description: 'apply-from: string, day time. Currently this is a
+                    6 hour window. Format: "DDD hh:mm" > "sun 23:00". UTC time'
                   type: string
               type: object
             upgrade:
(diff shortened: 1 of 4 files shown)


CathalOConnorRH:
Validated the install and tests.
Two notes. One is that the time appears to be based on the pods cluster time and not the time shown in the dashboard so in the case of my test. I was able to set a time in the past and it was valid due to timezone differences. Not sure what the best way to handle this is. 
Second is, is it possible to match the error timestamp format to match the required format?

`admission webhook "rhmiconfig-validating-config.integreatly.org" denied the request: Invalid value for spec.Upgrade.ApplyOn: 2020-04-30 15:10:00 +0000 UTC. It must be a future date`

CathalOConnorRH:
Thanks for the fix on the time format. 
All tested and working
/lgtm

matskiv:
I only had time to skim through this very quickly, so apologies for asking about something that can be deducted from code...

@sergioifg94 are we expecting UTC time to be inputted into RHMIConfig fields? 
If yes, can you pleease update comments in `pkg/apis/integreatly/v1alpha1/rhmiconfig_types.go` to say so (now it says `Timezone local to cluster`).

sergioifg94:
@matskiv Yes, the times are meant to be UTC. Thanks for the suggestion, I just pushed the update to the comments

matskiv:
@sergioifg94 I noticed you are setting `failurePolicy  = v1beta1.Fail`
So if we run operator locally, any edit of RHMIConfig will fail because it can't find the pod referenced by the webhooks Service?

sergioifg94:
@matskiv I added some logic to disable the webhooks when the operator is running locally. You can see it in [this commit](https://github.com/integr8ly/integreatly-operator/pull/660/commits/5ff9297f2a37414ccf090929256426ac1f67059d)

So when the operator is run locally the ValidatingWebhookConfiguration won't be created and every modification is allowed

matskiv:
@sergioifg94 oh, I didn't notice that, sorry. Awesome! Thank you.

sergioifg94:
Seems like there's a new flaky test. I'll create a JIRA
/test e2e

philbrookes:
/override ci/prow/e2e
e2e is hitting throttling flake in prow
Verified locally by  @CathalOConnorRH

CathalOConnorRH:
@philbrookes I've tested locally and validated it

PR #661, opened 2020-05-01
New product in the monitoring stage

Rajagopalan-Ranganathan:
Jira link:  https://issues.redhat.com/browse/INTLY-7034
Clones service monitors from other namespaces that have "middleware-monitoring=true" in the new
redhat-rhmi-monitoring namespace. Creates rolebindings in the other namespaces (where servicemonitors were cloned)
for prometheus to scrape metrics from the corresponding services.

# Description

## Type of change

## Checklist

### DIFF
--- a/pkg/apis/integreatly/v1alpha1/rhmi_types.go
+++ b/pkg/apis/integreatly/v1alpha1/rhmi_types.go
@@ -49,6 +49,7 @@ var (
 	ProductMonitoring          ProductName = "middleware-monitoring"
 	ProductCloudResources      ProductName = "cloud-resources"
 	ProductDataSync            ProductName = "datasync"
+	ProductMonitoringSpec      ProductName = "monitoring-spec"
 
 	// Could not find a way to determine these versions dynamically, so they are hard-coded
 	// It is preferable to determine the version of a product dynamically (from a CR, or configmap, etc)
@@ -66,6 +67,7 @@ var (
 	VersionDataSync            ProductVersion = "0.9.4"
 	VersionRHSSO               ProductVersion = "7.3"
 	VersionRHSSOUser           ProductVersion = "7.3"
+	VersionMonitoringSpec      ProductVersion = "1.0"
 
 	// Versioning for Fuse on OpenShift does not follow a similar pattern to other products.
 	// It is currently implicitly tied to version 7.6 of Fuse, hence the 7.6 value for VersionFuseOnOpenshift above
@@ -91,6 +93,7 @@ var (
 	OperatorVersionCloudResources      OperatorVersion = "0.15.1"
 	OperatorVersionUPS                 OperatorVersion = "0.5.0"
 	OperatorVersionApicurito           OperatorVersion = "1.6.0"
+	OperatorVersionMonitoringSpec      OperatorVersion = "1.0"
 
 	// Event reasons to be used when emitting events
 	EventProcessingError       string = "ProcessingError"
--- a/pkg/config/monitoringspec.go
+++ b/pkg/config/monitoringspec.go
@@ -3,7 +3,9 @@ package config
 import (
 	"errors"
 
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
@@ -16,7 +18,15 @@ func NewMonitoringSpec(config ProductConfig) *MonitoringSpec {
 }
 
 func (m *MonitoringSpec) GetWatchableCRDs() []runtime.Object {
-	return []runtime.Object{}
+
+	return []runtime.Object{
+		&monitoringv1.ServiceMonitor{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: monitoringv1.SchemeGroupVersion.String(),
+				Kind:       monitoringv1.ServiceMonitorsKind,
+			},
+		},
+	}
 }
 
 func (m *MonitoringSpec) GetNamespace() string {
--- a/pkg/config/monitoringspec.go
+++ b/pkg/config/monitoringspec.go
@@ -1,0 +1,80 @@
+package config
+
+import (
+	"errors"
+
+	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type MonitoringSpec struct {
+	Config ProductConfig
+}
+
+func NewMonitoringSpec(config ProductConfig) *MonitoringSpec {
+	return &MonitoringSpec{Config: config}
+}
+
+func (m *MonitoringSpec) GetWatchableCRDs() []runtime.Object {
+	return []runtime.Object{}
+}
+
+func (m *MonitoringSpec) GetNamespace() string {
+	return m.Config["NAMESPACE"]
+}
+
+func (m *MonitoringSpec) SetNamespace(newNamespace string) {
+	m.Config["NAMESPACE"] = newNamespace
+}
+
+func (m *MonitoringSpec) GetNamespacePrefix() string {
+	return m.Config["NAMESPACE_PREFIX"]
+}
+
+func (m *MonitoringSpec) SetNamespacePrefix(newNamespacePrefix string) {
+	m.Config["NAMESPACE_PREFIX"] = newNamespacePrefix
+}
+
+func (m *MonitoringSpec) GetHost() string {
+	return m.Config["HOST"]
+}
+
+func (m *MonitoringSpec) SetHost(newHost string) {
+	m.Config["HOST"] = newHost
+}
+
+func (m *MonitoringSpec) Read() ProductConfig {
+	return m.Config
+}
+
+func (m *MonitoringSpec) GetProductName() integreatlyv1alpha1.ProductName {
+	return integreatlyv1alpha1.ProductMonitoringSpec
+}
+
+func (m *MonitoringSpec) GetProductVersion() integreatlyv1alpha1.ProductVersion {
+	return integreatlyv1alpha1.VersionMonitoringSpec
+}
+
+func (m *MonitoringSpec) GetOperatorVersion() integreatlyv1alpha1.OperatorVersion {
+	return integreatlyv1alpha1.OperatorVersionMonitoringSpec
+}
+
+func (m *MonitoringSpec) SetProductVersion(version string) {
+	m.Config["VERSION"] = version
+}
+
+func (m *MonitoringSpec) Validate() error {
+	if m.GetProductName() == "" {
+		return errors.New("config product name is not defined")
+	}
+
+	if m.GetNamespace() == "" {
+		return errors.New("config namespace is not defined")
+	}
+
+	if m.GetProductVersion() == "" {
+		return errors.New("config product version is not defined")
+	}
+
+	return nil
+}
--- a/pkg/products/monitoringspec/reconciler.go
+++ b/pkg/products/monitoringspec/reconciler.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"fmt"
 
-	v1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/integr8ly/integreatly-operator/pkg/resources/events"
 	"github.com/operator-framework/operator-registry/pkg/lib/bundle"
 
@@ -154,7 +154,7 @@ func (r *Reconciler) createNamespace(ctx context.Context, serverClient k8sclient
 		return integreatlyv1alpha1.PhaseCompleted, nil
 	}
 
-	resources.PrepareObject(namespace, installation, false, true)
+	resources.PrepareObject(namespace, installation, false, false)
 	err = serverClient.Update(ctx, namespace)
 	if err != nil {
 		return integreatlyv1alpha1.PhaseFailed, err
@@ -230,14 +230,14 @@ func (r *Reconciler) reconcileMonitoring(ctx context.Context, serverClient k8scl
 }
 
 func (r *Reconciler) reconcileServiceMonitor(ctx context.Context,
-	serverClient k8sclient.Client, serviceMonitor *v1.ServiceMonitor) (err error) {
+	serverClient k8sclient.Client, serviceMonitor *monitoringv1.ServiceMonitor) (err error) {
 
 	if serviceMonitor.Spec.NamespaceSelector.Any {
 		logrus.Warnf("servicemonitor : %s cannot be copied to %s namespace. Namespace selector has been set to any",
 			serviceMonitor.Name, r.Config.GetNamespace())
 		return nil
 	}
-	sm := &v1.ServiceMonitor{
+	sm := &monitoringv1.ServiceMonitor{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      serviceMonitor.Namespace + `-` + serviceMonitor.Name,
 			Namespace: r.Config.GetNamespace(),
@@ -270,7 +270,7 @@ func (r *Reconciler) reconcileServiceMonitor(ctx context.Context,
 func (r *Reconciler) reconcileRoleBindingsForServiceMonitor(ctx context.Context,
 	serverClient k8sclient.Client, serviceMonitorName string) (err error) {
 	//Get the service monitor - that was created/updated
-	sermon := &v1.ServiceMonitor{}
+	sermon := &monitoringv1.ServiceMonitor{}
 	err = serverClient.Get(ctx, k8sclient.ObjectKey{Name: serviceMonitorName, Namespace: r.Config.GetNamespace()}, sermon)
 	if err != nil {
 		return err
@@ -317,7 +317,7 @@ func (r *Reconciler) reconcileRoleBinding(ctx context.Context,
 
 func (r *Reconciler) removeServiceMonitor(ctx context.Context,
 	serverClient k8sclient.Client, namespace, name string) (err error) {
-	sm := &v1.ServiceMonitor{
+	sm := &monitoringv1.ServiceMonitor{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
 			Namespace: namespace,
@@ -368,17 +368,17 @@ func (r *Reconciler) removeRoleBinding(ctx context.Context,
 
 func (r *Reconciler) getServiceMonitors(ctx context.Context,
 	serverClient k8sclient.Client,
-	listOpts []k8sclient.ListOption) (serviceMonitorsMap map[string]*v1.ServiceMonitor, err error) {
+	listOpts []k8sclient.ListOption) (serviceMonitorsMap map[string]*monitoringv1.ServiceMonitor, err error) {
 
 	if len(listOpts) == 0 {
 		return serviceMonitorsMap, fmt.Errorf("List options is empty")
 	}
-	serviceMonitors := &v1.ServiceMonitorList{}
+	serviceMonitors := &monitoringv1.ServiceMonitorList{}
 	err = serverClient.List(ctx, serviceMonitors, listOpts...)
 	if err != nil {
 		return serviceMonitorsMap, err
 	}
-	serviceMonitorsMap = make(map[string]*v1.ServiceMonitor)
+	serviceMonitorsMap = make(map[string]*monitoringv1.ServiceMonitor)
 	for _, sm := range serviceMonitors.Items {
 		serviceMonitorsMap[sm.Name] = sm
 	}
--- a/pkg/products/monitoringspec/reconciler.go
+++ b/pkg/products/monitoringspec/reconciler.go
@@ -145,16 +145,15 @@ func (r *Reconciler) createNamespace(ctx context.Context, serverClient k8sclient
 		if !k8serr.IsNotFound(err) {
 			return integreatlyv1alpha1.PhaseFailed, err
 		}
-		//TODO - addClusterMonitoringLabel flipped back to true
 		_, err := resources.CreateNSWithProjectRequest(ctx, r.Config.GetNamespace(),
-			serverClient, installation, false, false)
+			serverClient, installation, false, true)
 		if err != nil {
 			return integreatlyv1alpha1.PhaseFailed, err
 		}
 		return integreatlyv1alpha1.PhaseCompleted, nil
 	}
 
-	resources.PrepareObject(namespace, installation, false, false)
+	resources.PrepareObject(namespace, installation, false, true)
 	err = serverClient.Update(ctx, namespace)
 	if err != nil {
 		return integreatlyv1alpha1.PhaseFailed, err
--- a/pkg/products/monitoringspec/reconciler.go
+++ b/pkg/products/monitoringspec/reconciler.go
@@ -1,0 +1,390 @@
+package monitoringspec
+
+import (
+	"context"
+	"fmt"
+
+	v1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/events"
+	"github.com/operator-framework/operator-registry/pkg/lib/bundle"
+
+	"github.com/sirupsen/logrus"
+	rbac "k8s.io/api/rbac/v1"
+
+	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+	"github.com/integr8ly/integreatly-operator/pkg/config"
+	"github.com/integr8ly/integreatly-operator/pkg/resources"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/marketplace"
+
+	corev1 "k8s.io/api/core/v1"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	defaultInstallationNamespace              = "monitoring"
+	packageName                               = "monitoringspec"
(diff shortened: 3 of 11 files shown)


pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+		_, err := resources.CreateNSWithProjectRequest(ctx, r.Config.GetNamespace(),
+			serverClient, installation, false, true)
+		if err != nil {
+			return integreatlyv1alpha1.PhaseFailed, err
+		}
+		return integreatlyv1alpha1.PhaseCompleted, nil
+	}
+
+	//TODO - Cluster monitoring flipped back to true
+	resources.PrepareObject(namespace, installation, false, false)
```
david-martin, line 156: @alanmoran We can set the last arg to `true` here once we're happy to add the `openshift.io/cluster-monitoring=true` label & start scraping product metrics into the cluster prometheus

alanmoran: Perfect!

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+	opRes, err := controllerutil.CreateOrUpdate(ctx, serverClient, sm, func() error {
+		// Check if the servicemonitor has no  namespace selectors defined,
+		// if not add the namespace
+		sm.Spec = serviceMonitor.Spec
+		if len(sm.Spec.NamespaceSelector.MatchNames) == 0 {
+			sm.Spec.NamespaceSelector.MatchNames = []string{serviceMonitor.Namespace}
+		}
+		//Add all the original labels and append cloned servicemonitor label
+		sm.Labels = serviceMonitor.Labels
+		if len(sm.Labels) == 0 {
```
grdryn, line 253: Should this be a `nil` check rather than a length check?

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+			sm.Labels = make(map[string]string)
+		}
+		sm.Labels[clonedServiceMonitorLabelKey] = clonedServiceMonitorLabelValue
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	r.Logger.Infof("operation result of creating servicemonitor %v was %v", sm.Name, opRes)
```
grdryn, line 262:
```suggestion
	if opRes != controllerutil.OperationResultNone {
		r.Logger.Infof("ServiceMonitor %v CreateOrUpdate OperationResult: %v", sm.Name, opRes)
	}
```

^For the vast majority of cases it should be `controllerutil.OperationResultNone`, and logging in that case is not really useful.

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+
+	//Get the service monitor - that was created/updated
+	sermon := &v1.ServiceMonitor{}
+	err = serverClient.Get(ctx, k8sclient.ObjectKey{Name: sm.Name, Namespace: r.Config.GetNamespace()}, sermon)
+	if err != nil {
+		return err
+	}
+	//Create role binding for each of the namespace label selectors
+	for _, namespace := range sermon.Spec.NamespaceSelector.MatchNames {
```
grdryn, line 271:
```suggestion
	//Create role binding for each of the namespace label selectors
	for _, namespace := range sm.Spec.NamespaceSelector.MatchNames {
```

I don't think it's needed to do a `serverClient.Get` again here -- everything you need should be already in the `sm` variable, shouldn't it?

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+			},
+		}
+		roleBinding.RoleRef = rbac.RoleRef{
+			APIGroup: roleRefAPIGroup,
+			Kind:     bundle.ClusterRoleKind,
+			Name:     roleRefName,
+		}
+		return nil
+	})
+	r.Logger.Infof("operation result of creating rolebinding: %v was %v", roleBindingName, opRes)
```
grdryn, line 304: Same as above, only log when it's not `controllerutil.OperationResultNone`

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+		return nil
+	}
+	return err
+}
+
+func (r *Reconciler) removeRoleBinding(ctx context.Context,
+	serverClient k8sclient.Client, namespace, name string) (err error) {
+
+	// Check if the namespace has service monitors
+	// if so donot delete the rolebinding
```
grdryn, line 332:
```suggestion
	// if so do not delete the rolebinding
```

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+	err = serverClient.Get(ctx, k8sclient.ObjectKey{Name: name, Namespace: namespace}, sm)
+	if err != nil && k8serr.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	//Delete the servicemonitor
+	err = serverClient.Delete(ctx, sm)
```
grdryn, line 321:
```suggestion
	sm := &v1.ServiceMonitor{ObjectMeta: {Name: name, Namespace: namespace}}
	err = serverClient.Delete(ctx, sm)
```

I don't think it's necessary to do a Get on the ServiceMonitor just so that it can be deleted. You're checking the error for an `IsNotFound` case on the Delete call anyway.

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+
+	//Delete the rolebinding
+	err = serverClient.Delete(ctx, rb)
+	if err != nil && k8serr.IsNotFound(err) {
+		return nil
+	}
+	return err
+}
+
+func (r *Reconciler) getServiceMonitors(ctx context.Context,
```
grdryn, line 363: There are a couple of places in this change where listing ServiceMonitors is done without using this function. I think either this function should be split or those other places should learn to use it. What do you think?

test/common/monitoringspec.go
```diff
@@ -0,0 +1,145 @@
+package common
+
+import (
+	goctx "context"
+	"testing"
+
+	v1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: Agreed. Fixed.

test/common/monitoringspec.go
```diff
@@ -0,0 +1,145 @@
...
+	"redhat-rhmi-middleware-monitoring-operator-application-monitoring-operator-metrics",
+	"redhat-rhmi-middleware-monitoring-operator-grafana-servicemonitor",
+	"redhat-rhmi-middleware-monitoring-operator-prometheus-servicemonitor",
+	"redhat-rhmi-rhsso-keycloak-service-monitor",
+	"redhat-rhmi-ups-operator-unifiedpush-operator-metrics",
+	"redhat-rhmi-ups-unifiedpush",
+	"redhat-rhmi-user-sso-keycloak-service-monitor",
+}
+
+// TestServiceMonitorsCloneAndRolebindingsExist monitirng spec testcase
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: You missed an "i" still in your suggestion. Agreed and typo is corrected.

test/common/monitoringspec.go
```diff
@@ -0,0 +1,145 @@
...
+// Verifies the list of servicemonitors that are cloned in monitoring namespace
+// Verifies the rolebindings exist
+// Verifies if there are any stale service monitors in the monitoring namespace
+func TestServiceMonitorsCloneAndRolebindingsExist(t *testing.T, ctx *TestingContext) {
+	//Get list of service monitors in the monitoring namespace
+	monSermonMap, err := getServiceMonitors(ctx, MonitoringSpecNamespace)
+	if err != nil {
+		t.Fatal("failed to list servicemonitors in monitoring namespace:", err, "length is empty")
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: No. Previously i had both the checkof length and error in the same line. forgot to update the log. It is now corrected.

test/common/monitoringspec.go
```diff
@@ -0,0 +1,145 @@
...
+	for _, ns := range namespaces.Items {
+		//Get list of service monitors in each name space
+		listOpts := []k8sclient.ListOption{
+			k8sclient.InNamespace(ns.Name),
+		}
+		serviceMonitors := &v1.ServiceMonitorList{}
+		err := ctx.Client.List(goctx.TODO(), serviceMonitors, listOpts...)
+		if err != nil {
+			t.Fatal("failed to list servicemonitors", err)
+		}
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: The getServicemonitors() , filters out also based on label. And creates a map. Even though the code looks similar, i prefer not to re-use the original one. I might need to add check in the method for empty label being passed, and creating a map for 1 servicemonitor (which is mostly) the case might not be ideal. I would prefer to leave this piece of code as it is.

test/common/monitoringspec.go
```diff
@@ -0,0 +1,145 @@
...
+		clonedServiceMonitorLabelKey: clonedServiceMonitorLabelValue,
+	}
+}
+
+func checkRoleBindingExists(ctx *TestingContext, name, namespace string) (err error) {
+	rb := &rbac.RoleBinding{}
+	err = ctx.Client.Get(goctx.TODO(), k8sclient.ObjectKey{Name: name, Namespace: namespace}, rb)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: Agreed. Fixed.

pkg/products/monitoringspec/reconciler_test.go
```diff
@@ -0,0 +1,516 @@
+package monitoringspec
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	v1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: Kept the one in Line 13 and removed this.

pkg/products/monitoringspec/reconciler_test.go
```diff
@@ -0,0 +1,516 @@
...
+
+	"github.com/integr8ly/integreatly-operator/pkg/config"
+	"github.com/integr8ly/integreatly-operator/pkg/resources"
+
+	prometheusmonitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: Removed the one in Line 8. Fixed.

pkg/products/monitoringspec/reconciler_test.go
```diff
@@ -0,0 +1,516 @@
...
+
+	monitoringv1 "github.com/integr8ly/application-monitoring-operator/pkg/apis/applicationmonitoring/v1alpha1"
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: The import is used to add the scheme and during install plan. Fixed the name.

pkg/products/monitoringspec/reconciler_test.go
```diff
@@ -0,0 +1,516 @@
...
+	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/marketplace"
+
+	moqclient "github.com/integr8ly/integreatly-operator/pkg/client"
+	projectv1 "github.com/openshift/api/project/v1"
+
+	coreosv1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1"
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: Agreed. Fixed.

pkg/products/monitoringspec/reconciler_test.go
```diff
@@ -0,0 +1,516 @@
...
+		},
+		Spec: integreatlyv1alpha1.RHMISpec{
+			SMTPSecret:           mockSMTPSecretName,
+			PagerDutySecret:      mockPagerdutySecretName,
+			DeadMansSnitchSecret: mockDMSSecretName,
+		},
+	}
+}
+
+func creatServicemonitor(name, namepsace string) *v1.ServiceMonitor {
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: Agreed. Fixed.

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
+package monitoringspec
+
+import (
+	"context"
+	"fmt"
+
+	v1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: This has been removed. Fixed.

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+				//namespace is gone, return complete
+				return integreatlyv1alpha1.PhaseCompleted, nil
+			}
+			phase, err := resources.RemoveNamespace(ctx, installation, serverClient, r.Config.GetNamespace())
+			if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+				logrus.Infof("Spec phase removal failure: %v", err)
+				return phase, err
+			}
+			return integreatlyv1alpha1.PhaseInProgress, nil
+		})
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: Agreed. Fixed.

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+			serverClient, installation, false, true)
+		if err != nil {
+			return integreatlyv1alpha1.PhaseFailed, err
+		}
+		return integreatlyv1alpha1.PhaseCompleted, nil
+	}
+
+	//TODO - Cluster monitoring flipped back to true
+	resources.PrepareObject(namespace, installation, false, false)
+	err = serverClient.Update(ctx, namespace)
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: This seems to be common pattern, when creating namespace with project. I am not sure about using ```controllerutil.CreateOrUpdate``` here in this case.

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+// make the federation namespace discoverable by cluster monitoring
+func (r *Reconciler) createNamespace(ctx context.Context, serverClient k8sclient.Client,
+	installation *integreatlyv1alpha1.RHMI) (integreatlyv1alpha1.StatusPhase, error) {
+	namespace, err := resources.GetNS(ctx, r.Config.GetNamespace(), serverClient)
+	if err != nil {
+		if !k8serr.IsNotFound(err) {
+			return integreatlyv1alpha1.PhaseFailed, err
+		}
+		_, err := resources.CreateNSWithProjectRequest(ctx, r.Config.GetNamespace(),
+			serverClient, installation, false, true)
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: Agreed. Fixed.

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+	opts := &k8sclient.ListOptions{
+		LabelSelector: ls,
+	}
+
+	//Get the list of namespaces with the given label selector
+	namespaces := &corev1.NamespaceList{}
+	err = serverClient.List(ctx, namespaces, opts)
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, err
+	}
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: Agreed. Moved it to separate method.

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+	opts := &k8sclient.ListOptions{
+		LabelSelector: ls,
+	}
+
+	//Get the list of namespaces with the given label selector
+	namespaces := &corev1.NamespaceList{}
+	err = serverClient.List(ctx, namespaces, opts)
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: The signature of the method is: ```func (k8sclient.Reader).List(ctx context.Context, list runtime.Object, opts ...k8sclient.ListOption) error``` . When the last argument is an array, we need to expand it using ```...```, and in the case where it is not an array (already expanded) we can use the argument as it is.

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+			Kind:     bundle.ClusterRoleKind,
+			Name:     roleRefName,
+		}
+		return nil
+	})
+	r.Logger.Infof("operation result of creating rolebinding: %v was %v", roleBindingName, opRes)
+	return err
+}
+
+func (r *Reconciler) removeServicemonitor(ctx context.Context,
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: Agreed. Fixed.

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+		err := serverClient.List(ctx, serviceMonitors, listOpts...)
+		if err != nil {
+			return integreatlyv1alpha1.PhaseFailed, err
+		}
+		for _, sm := range serviceMonitors.Items {
+			//Create a copy of service monitors in the monitoring namespace
+			//Create the corresponding rolebindings at each of the service namespace
+			key := sm.Namespace + `-` + sm.Name
+			delete(monSermonMap, key) // Servicemonitor exists, remove it from the local map
+			err := r.reconcileServiceMonitor(ctx, serverClient, sm)
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: Agreed. Split the method to reconcile rolebindings separately. Fixed.

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+				Kind:      rbac.ServiceAccountKind,
+				Name:      clusterMonitoringPrometheusServiceAccount,
+				Namespace: clusterMonitoringNamespace,
+			},
+		}
+		roleBinding.RoleRef = rbac.RoleRef{
+			APIGroup: roleRefAPIGroup,
+			Kind:     bundle.ClusterRoleKind,
+			Name:     roleRefName,
+		}
```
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: The role should be as minimal as possible , that just serves the essential purpose and nothing more in my opinion. Will wait for further comments.

pkg/products/monitoringspec/reconciler.go
Rajagopalan-Ranganathan: Empty map, length can be checked in golang.  ```package main

import "fmt"

func main() {
	var m map[string]string
	fmt.Println(len(m))
	if m == nil {
		fmt.Println("map is nil")
	}
}```  output: ```

1

package main

2

​

3

import "fmt"

4

​

5

func main() {

6

    var m map[string]string

7

    fmt.Println(len(m))

8

    if m == nil {

9

        fmt.Println("map is nil")

10

    }

11

}

12

​

0
map is nil```

pkg/products/monitoringspec/reconciler.go
Rajagopalan-Ranganathan: Agreed. Fixed.

pkg/products/monitoringspec/reconciler.go
Rajagopalan-Ranganathan: The mutate method which updates the servicemonitor object (the changes that is makes) is not available to the parent method that calls , createorupdae(). Never the less, the logic to create role bindings are split now, so it has get the servicemonitor and iterate over namespace selectors and reconcile the rolebindings for them.

pkg/products/monitoringspec/reconciler.go
Rajagopalan-Ranganathan: agreed. Fixed.

pkg/products/monitoringspec/reconciler.go
Rajagopalan-Ranganathan: Agreed. Fixed.

pkg/products/monitoringspec/reconciler.go
Rajagopalan-Ranganathan: Agreed and fixed.

pkg/products/monitoringspec/reconciler.go
Rajagopalan-Ranganathan: Agreed. Re factored the getServicemonitors to take a list options as argument. Fixed.

pkg/products/monitoringspec/reconciler.go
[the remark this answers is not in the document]
grdryn: Looks to be still there to me!

pkg/products/monitoringspec/reconciler.go
```diff
@@ -0,0 +1,407 @@
...
+		//TODO - addClusterMonitoringLabel flipped back to true
+		_, err := resources.CreateNSWithProjectRequest(ctx, r.Config.GetNamespace(),
+			serverClient, installation, false, false)
+		if err != nil {
+			return integreatlyv1alpha1.PhaseFailed, err
+		}
+		return integreatlyv1alpha1.PhaseCompleted, nil
+	}
+
+	resources.PrepareObject(namespace, installation, false, true)
```
grdryn, line 157:
```suggestion
	resources.PrepareObject(namespace, installation, false, false)
```

This got changed in the most recent commit somehow

pkg/products/monitoringspec/reconciler_test.go
```diff
@@ -0,0 +1,514 @@
...
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/integr8ly/integreatly-operator/pkg/config"
+	"github.com/integr8ly/integreatly-operator/pkg/resources"
+
+	prometheusmonitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+
+	applicationmonitoringv1alpha1 "github.com/integr8ly/application-monitoring-operator/pkg/apis/applicationmonitoring/v1alpha1"
```
grdryn, line 13: I think what I was saying about this not being needed might still be valid: yes it's referenced here in a couple of places, but I think it can be removed from those places:

* I _think_ it shouldn't be needed in the scheme, since we're not going to be making any requests related to types from that apigroup in this test
* Where it's referenced in the `TypeMeta` of `InstallPlanList` instances, I think that's actually incorrect. I'd suggest just completely removing the `TypeMeta` field from those instances, as it should get set correctly implicitly.

pkg/products/monitoringspec/reconciler.go
Rajagopalan-Ranganathan: I had flipped to true during testing. flipped back.

Rajagopalan-Ranganathan: VS Code needs to be smart enough , it liked having "v1" instead of "monitoringv1". fixed it now.

pkg/products/monitoringspec/reconciler_test.go
Rajagopalan-Ranganathan: The basic things were first copied from monitoring-middleware unit tests. I removed the import and corrected the scheme.
```TypeMeta``` in the install plan list has been removed as suggested.

Comment (Rajagopalan-Ranganathan) [reply pkg/products/monitoringspec/reconciler_test.go:13]: Also removed, the unnecessary scheme additions.

File: pkg/products/monitoringspec/reconciler.go (hunk 28 above)
Comment (grdryn) [reply pkg/products/monitoringspec/reconciler.go]: > VS Code needs to be smart enough , it liked having "v1" instead of "monitoringv1". fixed it now.

What's worse is that when there are two or more imports with the package name "v1", then the second one will get the automatic alias "v12", the third "v13" and so on. I'm not sure if it's VS Code, or Goland or gopls or something else that does that.

File: pkg/apis/integreatly/v1alpha1/rhmi_types.go [hunk 30]
```diff
@@ -91,6 +93,7 @@ var (
 	OperatorVersionCloudResources      OperatorVersion = "0.14.0"
 	OperatorVersionUPS                 OperatorVersion = "0.5.0"
 	OperatorVersionApicurito           OperatorVersion = "1.6.0"
+	OperatorVersionMonitoringSpec      OperatorVersion = "1.0"
```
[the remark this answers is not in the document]
david-martin: @philbrookes Would appreciate your input here.

@Rajagopalan-Ranganathan or @grdryn what happens if we don't specify an operator version?

pkg/config/monitoringspec.go
```diff
@@ -0,0 +1,80 @@
...
+
+type MonitoringSpec struct {
+	Config ProductConfig
+}
+
+func NewMonitoringSpec(config ProductConfig) *MonitoringSpec {
+	return &MonitoringSpec{Config: config}
+}
+
+func (m *MonitoringSpec) GetWatchableCRDs() []runtime.Object {
```
david-martin, line 18: @Rajagopalan-Ranganathan @grdryn did we decide on watching ServiceMonitors (across all rhmi namespaces) for changes & requeue?
If yes, should that go here, is is there a better place to put it.

pkg/apis/integreatly/v1alpha1/rhmi_types.go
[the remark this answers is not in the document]
Rajagopalan-Ranganathan: As far i can remember, this is boiler plate. We do not specify it will error out. During the installation, getOperatorVersion() is called/expected.

pkg/products/monitoringspec/reconciler.go
david-martin: @Rajagopalan-Ranganathan Can you set the value to true.
We are good to go on scraping these metrics.

pkg/config/monitoringspec.go
Rajagopalan-Ranganathan: @grdryn @david-martin As per the design, we first get a list of namespaces that have ```"monitoring-key=middleware"``` . Then get the corresponding service monitors in each of those name spaces. So IMO we are watching them for changes.  This looks like more a static approach rather then the dynamic one we have now. Correct me if i am wrong.

pkg/products/monitoringspec/reconciler.go
Rajagopalan-Ranganathan: Set the cluster monitoring label to true.

pkg/config/monitoringspec.go
Rajagopalan-Ranganathan: Discussed and agreed that, we will watch for servicemonitors.

PR #673, opened 2020-05-05
INTLY-6123 Feat(e2e): Checking recreating CR's in AMQ Online namespace

Boomatang:
# Description

e2e tests add to check the recreation of CR's in the AMQ Online namespace. Related to closed PR #567 

**Jira** https://issues.redhat.com/browse/INTLY-6123

## Actions Preformed
- Check if an edited CR field is reconciled by RHMI if the field is in the create-or-update function 
- Check if a delete CR field is reconciled by RHMI if the field is in the create-or-update function 
- Check if a added CR field is **not** reconciled by RHMI if the field is **not** in the create-or-update function 

# Type of change

[ ] Bug fix (non-breaking change which fixes an issue)
[ ] New feature (non-breaking change which adds functionality)
[ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist

[ ] This change requires a documentation update
[ ] I have added tests that prove my fix is effective or that my feature works
[ ] Verified independently on a cluster by reviewer

### DIFF
--- a/test/common/crs_modifided.go
+++ b/test/common/crs_modifided.go
@@ -1,0 +1,1973 @@
+package common
+
+import (
+	goctx "context"
+	"fmt"
+	enmasseadminv1beta1 "github.com/integr8ly/integreatly-operator/pkg/apis-products/enmasse/admin/v1beta1"
+	enmassev1beta1 "github.com/integr8ly/integreatly-operator/pkg/apis-products/enmasse/v1beta1"
+	enmasse "github.com/integr8ly/integreatly-operator/pkg/apis-products/enmasse/v1beta2"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sync"
+	"testing"
+	"time"
+)
+
+const (
+	crRetryInterval      = time.Second * 5
+	crTimeout            = time.Second * 60 * 5
+	integreatlyName      = "integreatly-name"
+	integreatlyNamespace = "integreatly-namespace"
+	amqOnline            = "redhat-rhmi-amq-online"
+	crFieldEdit          = true
+	crFieldDelete        = true
+	crFieldAdd           = true
+)
+
+type compareResult struct {
+	Type  string
+	Name  string
+	Key   string
+	Error string
+}
+
+func TestResetCRs(t *testing.T, ctx *TestingContext) {
+	var wg sync.WaitGroup
+	testAMQOnline(t, ctx, &wg)
+
+	wg.Wait()
+}
+
+//========================================================================================================
+// enmasse
+//========================================================================================================
+
+func testAMQOnline(t *testing.T, ctx *TestingContext, wg *sync.WaitGroup) {
+	testAddressSpacePlan(wg, t, ctx)
+	testAddressPlan(wg, t, ctx)
+	testAuthenticationServiceCr(wg, t, ctx)
+	testBrokeredInfraConfigCr(wg, t, ctx)
+	testStandardInfraConfigCr(wg, t, ctx)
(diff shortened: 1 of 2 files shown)


Boomatang:
@briangallagher Good Point. When I first started on this ticket, about two and a half sprints ago, the ticket was more focused on the CR's that are been created by amq and not a specific test case.

One other ask at the time was to see how easy it would be to added the testing in for different product CR's. 

 Does this answer your concerns?

briangallagher:
@Boomatang Still not really clear if Role, RoleRef and ConsoleService should be present or not? Either way, it's probably best that the automated test matches the test case i.e. the same types should be present in both.

briangallagher:
@Boomatang crs_modifided.go is quite long and there seems to be a lot of repetition in there for example, modifyExistingValues(), deleteExistingValues() etc 
We could possibly overcome this with Interfaces and type assertions. I've put together [a gist to demostrate](https://gist.github.com/briangallagher/a53220e75c0247fa4903f4fcbc651014): 
There may be a more elegant way to do this but in general the file is quite long and a bit repetitive so consolidating in some form would be good.

test/common/crs_modifided.go
```diff
@@ -0,0 +1,1973 @@
...
+	resourceVersion := cr.ResourceVersion
+	err = wait.Poll(crRetryInterval, crTimeout, func() (done bool, err error) {
+		err = ctx.Client.Get(goctx.TODO(), k8sclient.ObjectKey{Name: cr.Name, Namespace: cr.Namespace}, &cr)
+		if err != nil {
+			return false, err
+		}
+
+		if resourceVersion != cr.ResourceVersion {
+			return true, nil
+		} else {
```
briangallagher, line 275: else not required, just:
`return false, nil`

test/common/crs_modifided.go
```diff
@@ -0,0 +1,1973 @@
...
+			return false, nil
+		}
+	})
+	if err != nil {
+		return false, err
+	} else {
```
briangallagher, line 281: else not required, just:
`return true, nil`

test/common/crs_modifided.go
```diff
@@ -0,0 +1,1973 @@
...
+				Name:  cr.Name,
+				Key:   "subjects.[].Kind",
+				Error: err.Error(),
+			})
+		}
+	}
+
+	if len(values) > 0 {
+		return &values
+	} else {
```
briangallagher, line 331: else not required, just:
`return nil`

Boomatang:
> @Boomatang crs_modifided.go is quite long and there seems to be a lot of repetition in there for example, modifyExistingValues(), deleteExistingValues() etc
> We could possibly overcome this with Interfaces and type assertions. I've put together [a gist to demostrate](https://gist.github.com/briangallagher/a53220e75c0247fa4903f4fcbc651014):
> There may be a more elegant way to do this but in general the file is quite long and a bit repetitive so consolidating in some form would be good.

This example is great and what I was missing. I didn't know about the Container object but this should really tidy up what I have done. 

I will track down clearer information on what should be covered by this pr today.

test/common/crs_modifided.go
Boomatang: What would be standard practice in this case. Personal I like having the else, its make the next step much clearer to me.

briangallagher: It's not a big deal either way IMO so I'll leave it up to you. AFAIK, our code linting standards are basically the default from golint, so if its not highlighted by that, then assume it's fine. However, it is slightly superfluous code.

philbrookes:
@Boomatang  @briangallagher The best way to be sure this automated test is testing the minimum requirements, is if it has covered everything we currently do manually for that test case. If we do more than that - this is fine, but we must at least do the same.

I hope that has helped!

Boomatang:
Closing as there is a refactor branch been worked on and scope of the jira needs to be refined

PR #675, opened 2020-05-05
add solution explorer 0.0.52 manifest

aidenkeating:
- Include Solution Explorer manifest for 0.0.52
- Remove manifest for 0.0.47
- Ensure 0.0.52 manifest "replaces" 0.0.48

verification:
- ensure installation result sin 0.0.52 of the solution explorer
- ensure upgrade from rhmi 2.1.0 works as expected


aidenkeating:
@philbrookes @JameelB I'm not overly familiar with the upgrade scenario. Is the `replaces` flag required here? Does this structure look ok to you?

JameelB:
@aidenkeating yeah this looks right to me. `replaces` is required here if we're keeping `0.0.48`

PR #678, opened 2020-05-06
INTLY-7427 Add test for PVC check

CathalOConnorRH:
# Description
Add test to check for PVC being available and bound. This test is to replace manual testing. 

## Type of change

## Checklist

### DIFF
--- a/test/common/pvc_validate.go
+++ b/test/common/pvc_validate.go
@@ -1,0 +1,84 @@
+package common
+
+import (
+	goctx "context"
+	"fmt"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//Define list of namespaces and their claims
+var (
+	pvcNamespaces = []PersistentVolumeClaim{
+		{
+
+			Namespace: "redhat-rhmi-fuse",
+			PersistentVolumeClaimNames: []string{
+				"syndesis-meta",
+				"syndesis-prometheus",
+			},
+		},
+
+		{
+
+			Namespace: "redhat-rhmi-middleware-monitoring-operator",
+			PersistentVolumeClaimNames: []string{
+				"prometheus-application-monitoring-db-prometheus-application-monitoring-0",
+			},
+		},
+
+		{
+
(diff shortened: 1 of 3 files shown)


test/common/pvc_validate.go
```diff
@@ -0,0 +1,83 @@
...
+
+func TestPVClaims(t *testing.T, ctx *TestingContext) {
+
+	//get full list of volume claim items
+	pvcs := &corev1.PersistentVolumeClaimList{}
+
+	for _, pvcNamespace := range pvcNamespaces {
+		err := ctx.Client.List(goctx.TODO(), pvcs, &k8sclient.ListOptions{Namespace: pvcNamespace.Namespace})
+		if err != nil {
+			t.Errorf("Error getting PVCs for namespace: %v %w", pvcNamespace.Namespace, err)
```
sergioifg94, line 57:
```suggestion
			t.Errorf("Error getting PVCs for namespace: %v %w", pvcNamespace.Namespace, err)
			continue
```

`t.Errorf` won't stop the test execution, so if there's an error here, the value of `pvcs` won't be set, which is later used in `checkForClaim`. I'd add a `continue` to safely skip this namespace after the error.

test/common/pvc_validate.go
```diff
@@ -0,0 +1,83 @@
...
+			_, err := checkForClaim(claim, pvcs)
+			if err != nil {
+				t.Errorf("Persistant Volume Claim: %v %v", claim, err)
+			}
+
+		}
+	}
+}
+
+func checkForClaim(claim string, pvcs *corev1.PersistentVolumeClaimList) (bool, error) {
```
sergioifg94, line 70:
```suggestion
func checkForClaim(claim string, pvcs *corev1.PersistentVolumeClaimList) error {
```

As the `bool` result is not used when the function is called, I'd just remove it

test/common/pvc_validate.go
CathalOConnorRH: updated,  Thanks

test/common/pvc_validate.go
CathalOConnorRH: good call. Updated.

PR #680, opened 2020-05-06
add instructions for how to request OCP4 clusters on PSI

wei-lee:
# Description
https://issues.redhat.com/browse/DEL-270

## Type of change

## Checklist

### DIFF
--- a/guides/USE_PSI_CLUSTERS.md
+++ b/guides/USE_PSI_CLUSTERS.md
@@ -19,17 +19,10 @@ In order to trigger the job, you need to request build permission to the instanc
 
 ## Request build permission
 
-You need to (The instruction is also available on the Jenkins job page):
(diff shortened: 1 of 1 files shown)


mikenairn:
/lgtm
/approve

Approving since it's just docs updates. I requested the required LDAP access, but wasn't able to actually run the pipeline yet.

I think there might be some benefit to scripting/adding to the cli the ability to trigger the Jenkins jobs to create/destroy the clusters, but having the doc initially seems like a good first step.

matskiv:
I verified the process, all worked.
Just a thought, would it be better to add this into help repo instead?

jasonmadigan:
@wei-lee not sure docs like this should have a home in a public repo. A link is fine (maybe to integreatly-help?)

wei-lee:
@jasonmadigan ok, will do.

PR #691, opened 2020-05-08
update 3scale provider route selector to be more specific

aidenkeating:
when multiple tenants are in the 3scale namespace, the current
approach to selecting routes will return the first route found.
this may not be the managed tenant.

verification:
- run an installation
- create an additional 3scale tenant 'evals'
- ensure the 3scale hostname provided to the solution explorer is
  the correct hostname
- ensure blackbox targets for 3scale are pointed at the main
  developer route

### DIFF
--- a/pkg/products/threescale/objects_test.go
+++ b/pkg/products/threescale/objects_test.go
@@ -251,6 +251,7 @@ var threescaleRoute1 = &v1.Route{
 	},
 }
 
+// Have two system-developer routes, the reconcile should pick up on 3scale.
 var threescaleRoute2 = &v1.Route{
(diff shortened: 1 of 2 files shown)


aidenkeating:
Verified against PDS (end of cluster URLs blocked out)

Multiple tenants:
Screenshot 2020-05-08 at 15 02 40

Webapp CR:
Screenshot 2020-05-08 at 15 05 01

Blackbox targets:
Screenshot 2020-05-08 at 15 12 00
Screenshot 2020-05-08 at 15 11 02
Screenshot 2020-05-08 at 15 10 21

davidffrench:
/approve

code looks good to me

PR #702, opened 2020-05-11
update edge subscription to manual

philbrookes:
# Description
Also update edge subscription to manual

## Type of change


matskiv:
@philbrookes Is this something we still want to do?

PR #703, opened 2020-05-11
Add test to check all stages completed

CathalOConnorRH:
add test to verify that all stages in the integreatly-operator CR report completed

# Description

Added a test to check that all RHMI stages are in a completed status. Product status is tested in another test. 

## Type of change

- [ X] New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/test/common/integreatly_stages.go
+++ b/test/common/integreatly_stages.go
@@ -1,0 +1,34 @@
+package common
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+)
+
+func TestIntegreatlyStagesStatus(t *testing.T, ctx *TestingContext) {
+
+	//get RHMI
+	rhmi, err := getRHMI(ctx.Client)
+	if err != nil {
+		t.Fatalf("error getting RHMI CR: %v", err)
+	}
+
+	//iterate stages and check their status
+	// This does not check the pro
+	for _, stage := range rhmi.Status.Stages {
+		err := checkStatus(stage)
+		if err != nil {
+			t.Errorf("Error: Stage %v not completed. It's current status is %v", stage.Name, err)
+		}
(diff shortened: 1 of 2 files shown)


test/common/integreatly_stages.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+)
+
+var (
+	products = []v1alpha1.ProductName{
+
```
JameelB, line 12: @CathalOConnorRH Do we also need to check for bootstrap here?

JameelB:
@CathalOConnorRH Can you also create a PR to update the test case itself with further details on what actually is being tested?

test/common/integreatly_stages.go
CathalOConnorRH: Good spot. That actually isn't needed anymore. That was left over from a previous attempt at writing this. This test only tests the stages and not the product status. That's covered in another test.

test/common/integreatly_stages.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		t.Fatalf("error getting RHMI CR: %v", err)
+	}
+
+	//iterate stages and check their status
+	// This does not check the pro
+	for _, stage := range rhmi.Status.Stages {
+		err := checkStatus(stage)
+		if err != nil {
+			t.Logf("Error: Product %v not completed. It's current status is %v", stage.Name, err)
```
JameelB, line 23: Should this be `t.Fatalf` instead of a log?

CathalOConnorRH: yeah. it should. Will push an update

CathalOConnorRH:
Closing this as it's covered in more detail in https://issues.redhat.com/browse/INTLY-7424

PR #707, opened 2020-05-12
INTLY-7424 Verify RHMI products status

sergioifg94:
# Description

> Link to JIRA: https://issues.redhat.com/browse/INTLY-7424

Extend tests for RHMI stages completion (#703) with verification of the products status within each stage.

## Type of change

## Checklist

### DIFF
--- a/test/common/integreatly_stages.go
+++ b/test/common/integreatly_stages.go
@@ -1,9 +1,12 @@
 package common
 
 import (
+	"fmt"
 	"testing"
+	"time"
 
 	"github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


CathalOConnorRH:
Tested and verified. 
/lgtm

test/common/integreatly_stages.go
```diff
@@ -0,0 +1,124 @@
...
+		rhmi, err := getRHMI(ctx.Client)
+		if err != nil {
+			return false, fmt.Errorf("error getting RHMI CR: %v", err)
+		}
+
+		//iterate stages and check their status
+		for stageName, productNames := range expectedStageProducts {
+			stage, ok := rhmi.Status.Stages[v1alpha1.StageName(stageName)]
+			if !ok {
+				t.Errorf("Error checking stage %s. Not found", stageName)
```
JameelB, line 60: Should it not exit out of the poll as soon as it gets an error? You never actually return an error here for this line to be called: https://github.com/integr8ly/integreatly-operator/pull/707/files#diff-0db09d831d48ec632033bf219619b029R99

What's the reason for continuing the test once it fails here?

sergioifg94: I don't exit the poll immediately in order to continue testing for other possible failed products or stages. However I set the `done` variable to `true`, so the poll will finish after testing all the stages

sergioifg94:
@JameelB I think this all comes from the fact that there's two separate jiras for them, so @CathalOConnorRH  and myself were working on them separately until he noticed that they're related, then we decided to incorporate my work in his test.

CathalOConnorRH:
@JameelB I'm happy to close my PR and update the jira to link to PR @sergioifg94 is doing.

JameelB:
@CathalOConnorRH I think that makes more sense, no point in getting the other one in if it's going to be changed in this PR anyway. @sergioifg94 this can be the PR linked for both INTLY-7423 AND intly-7424

thanks @CathalOConnorRH!

test/common/integreatly_stages.go
JameelB: Explained - Will continue the tests here to report all errors for all stages and products.

PR #708, opened 2020-05-12
INTLY-7425 - move e2e namespace creation test to functional test

KevFan:
# Description
Move e2e test for checking namespace are successully created as part of functional tests

Assoicated gitlab test case PR:
* https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases/merge_requests/217

Jira:
* https://issues.redhat.com/browse/INTLY-7425

## Type of change

- [x] Test Automation
## Checklist

### DIFF
--- a/test/common/namespace_created.go
+++ b/test/common/namespace_created.go
@@ -1,0 +1,45 @@
+package common
+
+import (
+	goctx "context"
+	corev1 "k8s.io/api/core/v1"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
+)
+
+var (
+	namespacesCreated = []string{
+		MonitoringOperatorNamespace,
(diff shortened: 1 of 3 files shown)


matskiv:
Hmm, that label doesn't seem to be read by anything in the integreatly-operator. So we can probably remove it from the test, just as you did.
/lgtm

KevFan:
> Hmm, that label doesn't seem to be read by anything in the integreatly-operator. So we can probably remove it from the test, just as you did.
> /lgtm

Hmm, yeah I don't think so either, but the label may have originally been used for monitoring, not too sure is this still the case :man_shrugging: I can make a follow up PR to add the check for the label if needed

matskiv:
@KevFan no, we most likely don't need that label, so don't waste your time on the check...

PR #710, opened 2020-05-13
Fix for nil pointer exception in amqonline reconciler.

Rajagopalan-Ranganathan:
The code ordering was fixed to check for the error first and then access the postgres object returned.

# Description
Error check was done after accessing the variable returned. Ordering has been changed to check for errors first.

## Type of change

## Checklist

### DIFF
--- a/pkg/products/amqonline/reconciler.go
+++ b/pkg/products/amqonline/reconciler.go
@@ -312,13 +312,12 @@ func (r *Reconciler) reconcileStandardAuthenticationService(ctx context.Context,
 			return nil
 		},
 	)
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("could not create postgresql for standard auth service: %w", err)
+	}
 
 	if postgres.Status.Phase != cro1types.PhaseComplete {
 		return integreatlyv1alpha1.PhaseAwaitingComponents, nil
-	}
-
-	if err != nil {
-		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("could not create postgresql for standard auth service: %w", err)
 	}
 
 	// Read the CRO secret, to get values to copy to enmasse namespace.


Rajagopalan-Ranganathan:
cc @david-martin  @pb82

pkg/products/amqonline/reconciler.go
```diff
@@ -312,15 +312,14 @@ func (r *Reconciler) reconcileStandardAuthenticationService(ctx context.Context,
 			return nil
 		},
 	)
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("could not create postgresql for standard auth service: %w", err)
+	}
 
 	if postgres.Status.Phase != cro1types.PhaseComplete {
```
david-martin, line 319: @austincunningham Would appreciate your review here as I think this check was added as part of the amq online backup work.
It looks like this check on the status was added above the err nil check (nil check existed first).
Seems like a simple fix to avoid nil pointer, unless we're missing something

pkg/products/amqonline/reconciler.go
```diff
@@ -312,15 +312,14 @@ func (r *Reconciler) reconcileStandardAuthenticationService(ctx context.Context,
 			return nil
 		},
 	)
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("could not create postgresql for standard auth service: %w", err)
```
grdryn, line 316: @Rajagopalan-Ranganathan with this change, I'm interested to know what the error returned from the `croUtil.ReconcilePostgres` function was. Have you managed to hit that since you made the change?

Rajagopalan-Ranganathan: @grdryn - I have not reproduced the issue since the first time. I will keep working on it and let you know if i manage to reproduce it.

pkg/products/amqonline/reconciler.go
austincunningham: @david-martin, err check before checking status I agree with your take on it. Should catch nil pointers .

Rajagopalan-Ranganathan:
> /lgtm
> 
> Happy to verify this too, do you have a list of verification steps I can follow to confirm we're not hitting this bug anymore?

There are no, specific procedures to reproduce it. It occurred randomly when running the operator locally. And the code was misplaced, which was corrected.

PR #716, opened 2020-05-14
few tweaks to upgrade logic

philbrookes:
# Description
Few issues fixed that I noticed setting up the demo

## Type of change

## Checklist

### DIFF
--- a/pkg/controller/subscription/rhmiConfigs/main.go
+++ b/pkg/controller/subscription/rhmiConfigs/main.go
@@ -12,6 +12,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
 
@@ -73,13 +74,14 @@ func UpdateStatus(ctx context.Context, client k8sclient.Client, config *integrea
 		config.Status.Maintenance.Duration = strconv.Itoa(WINDOW) + "hrs"
 	}
 
-	// Calculate the upgrade window
-	if installplan.Spec.Approved {
-		config.Status.Upgrade.Window = ""
-	} else {
-		upStart := installplan.ObjectMeta.CreationTimestamp.Time.Format("2 Jan 2006")
-		upEnd := installplan.ObjectMeta.CreationTimestamp.Time.Add(daysDuration(14)).Format("2 Jan 2006")
-		config.Status.Upgrade.Window = upStart + " - " + upEnd
+	if installplan != nil {
+		if installplan.Spec.Approved {
+			config.Status.Upgrade.Window = ""
+		} else {
+			upStart := installplan.ObjectMeta.CreationTimestamp.Time.Format("2 Jan 2006")
+			upEnd := installplan.ObjectMeta.CreationTimestamp.Time.Add(daysDuration(14)).Format("2 Jan 2006")
+			config.Status.Upgrade.Window = upStart + " - " + upEnd
+		}
 	}
 
 	// Calculate the upgrade schedule based on the spec:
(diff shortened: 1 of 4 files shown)


pkg/controller/subscription/rhmiConfigs/main.go
```diff
@@ -164,18 +170,24 @@ func IsUpgradeServiceAffecting(csv *olmv1alpha1.ClusterServiceVersion) bool {
...
 	if installPlan.Status.Phase == olmv1alpha1.InstallPlanPhaseInstalling {
 		return nil
 	}
 
 	eventRecorder.Eventf(installPlan, "Normal", integreatlyv1alpha1.EventUpgradeApproved,
 		"Approving %s install plan: %s", installPlan.Name, installPlan.Spec.ClusterServiceVersionNames[0])
 
-	installPlan.Spec.Approved = true
-	err := client.Update(ctx, installPlan)
+	_, err := controllerutil.CreateOrUpdate(ctx, client, installPlan, func() error {
```
philbrookes, line 183: A lot of stuff modifies this CR at the same time, and it basically guaranteed an "object has been modified" error, doing the update this way stops that happening.

pkg/controller/subscription/subscription_controller.go
```diff
@@ -101,13 +101,16 @@ func (r *ReconcileSubscription) HandleUpgrades(ctx context.Context, rhmiSubscrip
...
 	}
 
-	logrus.Infof("RHMI upgrade available")
-
 	latestRHMIInstallPlan, err := rhmiConfigs.GetLatestInstallPlan(ctx, rhmiSubscription, r.client)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 
+	if latestRHMIInstallPlan.Spec.Approved {
```
philbrookes, line 109: The upgrade is already approved, move on.

philbrookes:
@davidffrench no hurry, take a look after your PTO.

pkg/controller/subscription/rhmiConfigs/main.go
```diff
@@ -164,18 +170,24 @@ func IsUpgradeServiceAffecting(csv *olmv1alpha1.ClusterServiceVersion) bool {
 }
 
 func ApproveUpgrade(ctx context.Context, client k8sclient.Client, installPlan *olmv1alpha1.InstallPlan, eventRecorder record.EventRecorder) error {
+	if installPlan.Spec.Approved == true {
+		return nil
+	}
```
davidffrench, line 175: This should not be needed. The check below is also checking if the installplan install is in progress.

```	
if installPlan.Status.Phase == olmv1alpha1.InstallPlanPhaseInstalling {
	return nil			
}
```

pkg/controller/subscription/rhmiConfigs/main.go
davidffrench: Agreed

pkg/controller/subscription/subscription_controller.go
davidffrench: The above function `rhmiConfigs.IsUpgradeAvailable()` should be checking if there is an upgrade available. Is there a timing issue here where it will report an upgrade and the installplan is approved?

pkg/controller/subscription/rhmiConfigs/main.go
philbrookes: I found if I removed this check that I was getting quite a lot of noise in the logs in the form of: "the object has been updated", adding this check kept everything functional but stopped that noise.

pkg/controller/subscription/subscription_controller.go
philbrookes: The problem seems to be that the installplan can be approved = true, but not yet processing, as all we are interested in is approving the installplan, checking on that field seemed more sensible.

Or in other words, once the installplan is approved, we have nothing left to do, so let OLM get on with it's thing now.

pkg/controller/subscription/rhmiConfigs/main.go
davidffrench: That makes sense, thanks @philbrookes

matskiv:
@davidffrench are you happy with the code here, or do some comments still need to be addressed?

davidffrench:
> @davidffrench are you happy with the code here, or do some comments still need to be addressed?

Happy with this @matskiv

philbrookes:
flakey prow issue logged: https://issues.redhat.com/browse/INTLY-7197
/retest

PR #722, opened 2020-05-14
always regenerate the csv files

wei-lee:
# Description

At the moment, when the pre-release script is executed, it will only regenerate the csv files when the minor versions are changed. This means if there are changes made to the `operator.yaml` file between RC releases, these changes are not reflected in the csv files.

The changes here will make sure the CSV file will always be regenerated.
## Type of change

## Checklist

### DIFF
--- a/scripts/prepare-release.sh
+++ b/scripts/prepare-release.sh
@@ -8,6 +8,10 @@ create_new_csv() {
   operator-sdk generate csv --csv-version "$VERSION" --default-channel --operator-name integreatly-operator --csv-channel=rhmi --update-crds --from-version "$PREVIOUS_VERSION"
 }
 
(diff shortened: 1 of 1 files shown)


wei-lee:
ping @philbrookes this is why we are not seeing the changes

wei-lee:
@philbrookes @StevenTobin actually, just did a bit of testing here. If we remove the "from" part, the `replaces` field won't be updated correctly. So it's probably not a good idea to remove it.

wei-lee:
@philbrookes @StevenTobin actually, the solution is simpler than I thought. We just need to update the csv file if the current version and previous version is the same. This will ensure all changes made to `operator.yml` files between RC releases are included.

PR #728, opened 2020-05-14
Adds alerting email address to RHMI installation CRD

jjaferson:
# Description

Adds `AlertingEmailAddress` field to RHMI installation CRD and uses the value present in `ALERTING_EMAIL_ADDRESS` env variable as default when the operator creates its own CRD.

https://issues.redhat.com/browse/INTLY-7406

## Verification steps
1) Run the operator against a cluster using `make cluster/prepare/local && make code/run` 
2) Verify if the RHMI CRD has the `AlertingEmailAddress` field 
3) Check if the email address was added in the `alertmanager-application-monitoring` secret in the `redhat-rhmi-middleware-monitoring-operator` namespace

```
receivers:
  - name: default
    email_configs:
      - send_resolved: true
        to: noreply-test@rhmi.com.br
```

## Type of change

- [ x] New feature (non-breaking change which adds functionality)

## Checklist


KevFan:
Verified
/lgtm

pkg/controller/installation/installation_controller_test.go
```diff
@@ -99,6 +99,64 @@ func testCreateInstallationCR_useClusterStorage(t *testing.T, envValue string, a
...
+	defer os.Unsetenv("WATCH_NAMESPACE")
+
+	// Function to test
+	err := createInstallationCR(ctx, mockClient)
+
+	if err != nil {
+		t.Fatalf("Error creating installation CR: %v", err)
+	}
+
+	installation, err := getInstallationCR(ctx, mockClient, t)
```
dimitraz, line 122: Should `err` be handled? The test will panic if a `nil` cr is returned for whatever reason from this function

pkg/controller/installation/installation_controller_test.go
```diff
@@ -99,6 +99,64 @@ func testCreateInstallationCR_useClusterStorage(t *testing.T, envValue string, a
...
```
dimitraz, line 150: Same as above

pkg/controller/installation/installation_controller_test.go
jjaferson: Good catch, thanks

jjaferson:
/test e2e

timed out during installation waiting for completion of cloud-resources

PR #733, opened 2020-05-15
INTLY-7778-OOMKill-Alerts

HubertStefanski:
# Description

This PR adds new warning and critical alerts for OOMKilled pods
Do not merge before SOP is created and merged

## SOP
https://github.com/RHCloudServices/integreatly-help/pull/514

## JIRA 
 https://issues.redhat.com/browse/INTLY-7778

## Type of change

## Checklist

## Verification Steps 
TODO

### DIFF
--- a/pkg/products/monitoring/prometheusRules.go
+++ b/pkg/products/monitoring/prometheusRules.go
@@ -157,7 +157,7 @@ func (r *Reconciler) reconcileKubeStateMetricsAlerts(ctx context.Context, server
 				"sop_url": "TBA",
 				"message": "Pod {{ $labels.namespace }} / {{  $labels.pod  }} has been killed due to being Out Of Memory.",
 			},
-			Expr:   intstr.FromString("sum by(pod, namespace) (kube_pod_container_status_last_terminated_reason {reason='OOMKilled'} * on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'}) = 1"),
+			Expr:   intstr.FromString("count(kube_pod_container_status_last_terminated_reason {reason='OOMKilled'} * on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'}) > 1"),
 			For:    "5m", //TBD
 			Labels: map[string]string{"severity": "warning"},
 		},
@@ -165,10 +165,10 @@ func (r *Reconciler) reconcileKubeStateMetricsAlerts(ctx context.Context, server
 			Alert: "KubePodOutOfMemoryKilledCritical",
 			Annotations: map[string]string{
 				"sop_url": "TBA",
-				"message": "Multiple Pods in the {{ $labels.namespace }} are being killed of do to Out Of Memory ",
+				"message": "Multiple Pods in the {{ $labels.namespace }} are being killed off due to Out Of Memory ",
 			},
-			Expr:   intstr.FromString("sum by(pod, namespace) (kube_pod_container_status_last_terminated_reason {reason='OOMKilled'} * on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'}) > 1"), //TODO discuss the threshold of the no. of pods that are considered critical with SRE
-			For:    "5m",                                                                                                                                                                                                                                          //TBD
+			Expr:   intstr.FromString("count(rate(kube_pod_container_status_last_terminated_reason {reason='OOMKilled'}[10m]) * on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'}) by (reason) > 20"), //TODO discuss the threshold of the no. of pods that are considered critical with SRE
+			For:    "5m",                                                                                                                                                                                                                                                //TBD
 			Labels: map[string]string{"severity": "critical"},
 		},
 		{
--- a/pkg/products/monitoring/prometheusRules.go
+++ b/pkg/products/monitoring/prometheusRules.go
@@ -104,9 +104,9 @@ func (r *Reconciler) reconcileKubeStateMetricsAlerts(ctx context.Context, server
 			Alert: "KubePodCrashLooping",
 			Annotations: map[string]string{
 				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
-				"message": "Pod {{  $labels.namespace  }} / {{  $labels.pod  }} ({{  $labels.container  }}) is restarting {{  $value  }} times every 5 minutes; for the last 15 minutes",
+				"message": "Pod {{  $labels.namespace  }} / {{  $labels.pod  }} ({{  $labels.container  }}) is restarting {{  $value  }} times every 10 minutes; for the last 15 minutes",
 			},
-			Expr:   intstr.FromString("rate(kube_pod_container_status_restarts_total{job='kube-state-metrics'}[10m]) * on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'} * 60 * 5 > 0"),
+			Expr:   intstr.FromString("rate(kube_pod_container_status_restarts_total{job='kube-state-metrics'}[10m]) * on (namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'} * 60 * 5 > 0"),
 			For:    "15m",
 			Labels: map[string]string{"severity": "critical"},
 		},
--- a/pkg/products/monitoring/prometheusRules.go
+++ b/pkg/products/monitoring/prometheusRules.go
@@ -107,7 +107,7 @@ func (r *Reconciler) reconcileKubeStateMetricsAlerts(ctx context.Context, server
 				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
 				"message": "Pod {{  $labels.namespace  }} / {{  $labels.pod  }} ({{  $labels.container  }}) is restarting {{  $value  }} times every 5 minutes; for the last 15 minutes",
 			},
-			Expr:   intstr.FromString("rate(kube_pod_container_status_restarts_total{job='kube-state-metrics'}[15m]) * on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'} * 60 * 5 > 0"),
+			Expr:   intstr.FromString("rate(kube_pod_container_status_restarts_total{job='kube-state-metrics'}[10m]) * on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'} * 60 * 5 > 0"),
 			For:    "15m",
 			Labels: map[string]string{"severity": "critical"},
 		},
@@ -149,26 +149,6 @@ func (r *Reconciler) reconcileKubeStateMetricsAlerts(ctx context.Context, server
 			},
 			Expr:   intstr.FromString("(kube_pod_container_status_waiting_reason{reason='ContainerCreating'} * on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'}) > 0"),
 			For:    "15m",
-			Labels: map[string]string{"severity": "critical"},
-		},
-		{
-			Alert: "KubePodOutOfMemoryKilledWarning",
-			Annotations: map[string]string{
-				"sop_url": "TBA",
-				"message": "Pod {{ $labels.namespace }} / {{  $labels.pod  }} has been killed due to being Out Of Memory.",
-			},
-			Expr:   intstr.FromString("(((sum_over_time(kube_pod_container_status_last_terminated_reason {reason='OOMKilled'}[5m]) / count_over_time(kube_pod_container_status_last_terminated_reason {reason='OOMKilled'}[5m]))* on (namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'}) * 100) > 90"),
-			For:    "5m",
-			Labels: map[string]string{"severity": "warning"},
-		},
-		{
-			Alert: "KubePodOutOfMemoryKilledCritical",
-			Annotations: map[string]string{
-				"sop_url": "TBA",
-				"message": "Multiple Pods in the {{ $labels.namespace }} are being killed off due to Out Of Memory ",
-			},
-			Expr:   intstr.FromString("(((sum_over_time(kube_pod_container_status_last_terminated_reason {reason='OOMKilled'}[5m]) / count_over_time(kube_pod_container_status_last_terminated_reason {reason='OOMKilled'}[5m]))* on (namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'}) * 100) > 90"),
-			For:    "10m",
 			Labels: map[string]string{"severity": "critical"},
 		},
 		{
--- a/pkg/products/monitoring/prometheusRules.go
+++ b/pkg/products/monitoring/prometheusRules.go
@@ -152,6 +152,26 @@ func (r *Reconciler) reconcileKubeStateMetricsAlerts(ctx context.Context, server
 			Labels: map[string]string{"severity": "critical"},
 		},
 		{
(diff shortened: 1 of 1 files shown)


pkg/products/monitoring/prometheusRules.go
```diff
@@ -151,6 +151,26 @@ func (r *Reconciler) reconcileKubeStateMetricsAlerts(ctx context.Context, server
 			For:    "15m",
 			Labels: map[string]string{"severity": "critical"},
 		},
+		{
+			Alert: "KubePodOutOfMemoryKilledWarning",
+			Annotations: map[string]string{
+				"sop_url": "TBA",
+				"message": "Pod {{ $labels.namespace }} / {{  $labels.pod  }} has been killed due to being Out Of Memory.",
+			},
+			Expr:   intstr.FromString("count(kube_pod_container_status_last_terminated_reason {reason='OOMKilled'} * on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'}) > 1"),
```
HubertStefanski, line 160: @alanmoran  @brizrobbo @david-martin, I've implemented some really crude queries to check for OOMKilled pods over the last however many minutes, there's a few approaches I see we could implement this depending on what is really required, we could go by `rate over N minutes` or we could do `sum of OOMKilled pods > X over N minutes` or get an average over N minutes and check if that exceeds a threshold, there's obviously also other ways of querying for this, Looking for some feedback on what you guys would preffer, I would also discourage using the queries I've pushed  in production as I don't actually think they're that informative, looking forward on your guys thoughts, also look at line 170 for the critical alert

alanmoran: @HubertStefanski I don't think this check is working as you expect. It seems to be just counting all the results including zero's. For example, if I run this in my prod cluster I get 99 results but the true no. of OOM pods is 1. 

For warning I would recommend: (it will return any OOM issues in a 5 minute period
```
rate(kube_pod_container_status_last_terminated_reason {reason='OOMKilled'}[5m])* on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'} > 0
```

For critical I think it should be the same check but the duration (For) should be 10minutes. That should mean we have an OOM that is not going away.

@brizrobbo and @mciccone this would need CS SRE input.

HubertStefanski: Thanks @alanmoran , I half expected this query to be broken, wasn't quite sure of what was exactly required, yours is wayy more indivative :+1:, I'll push the changes to that query as soon as I gather feedback from the rest.

HubertStefanski: Using the `rate` query I haven't had much success in getting the alert to trigger, although the query without the `rate()` function works fine, I've instead gotten some results using `sum_over_time(kube_pod_container_status_last_terminated_reason {reason='OOMKilled'}[5m])* on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'} > 0`

brizrobbo:
@pepedocs Can you also please review during APAC time and offer viewpoint?

pkg/products/monitoring/prometheusRules.go
pepedocs: @HubertStefanski I think that the `kube_pod_container_status_last_terminated_reason` returns a gauge value which returns (looks like it) 1 or 0 for a given `reason`. Getting the rate for values such as [1,0,1,1,1,0] for example wouldn't work as expected. Using `sum_over_time` will work but it can be very strict/sensitive of an alert, that alert you've specified will basically assert `If there is at least 1 pod that got OOM killed in the last 5 minutes then fire`. 

One thing we may also consider is to base the alert on the logic `If N% of the time a pod got OOM killed in the last T minutes then fire` something like the below query:
> Note that I haven't really quite tested this query and the hypothesis I mentioned above regarding using the `rate` function, but Prom says that `rate`'s use is intended for `Counters`. Please perform tests if you consider the query below.

N=90% T=5m
```
(((sum_over_time(kube_pod_container_status_last_terminated_reason {reason='OOMKilled'}[5m]) 
/ 
count_over_time(kube_pod_container_status_last_terminated_reason {reason='OOMKilled'}[5m]))
* 
on (namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'}) * 100) > 90
```
The left clause sums all the 1's and the right clause counts all the samples taken per T.

HubertStefanski: I've tested your query and it seems to be working correctly @pepedocs , I'm happy to have this in place for the alert, unless there's any objections from anyone else?

pepedocs:
> @pepedocs Can you also please review during APAC time and offer viewpoint?

Hi @brizrobbo, @HubertStefanski proposed the changes above setting the alert to:

```
 If N% of the time a pod got OOM killed in the last T minutes then fire 
```

The given the other inputs above, we set the following:
**Warning**: N=90%, T=5 minutes
**Critical**: N=90%, T=10 minutes

WDYT?

brizrobbo:
@pepedocs Those thresholds LGTM.

HubertStefanski:
@pepedocs  Sure, I'm happy with the way they work, will push the changes asap

pkg/products/monitoring/prometheusRules.go
```diff
@@ -151,6 +151,26 @@ func (r *Reconciler) reconcileKubeStateMetricsAlerts(ctx context.Context, server
 			For:    "15m",
 			Labels: map[string]string{"severity": "critical"},
 		},
+		{
+			Alert: "KubePodOutOfMemoryKilledWarning",
+			Annotations: map[string]string{
+				"sop_url": "TBA",
+				"message": "Pod {{ $labels.namespace }} / {{  $labels.pod  }} has been killed due to being Out Of Memory.",
```
david-martin, line 158: Should the message be the same on both these alerts?

HubertStefanski: Yes, that's a remainder from the previous approach that the first query had, something closer to `Pods in {{ $labels.namespace }} are being killed off due to OOM` would probably fit better, will change in the next commit :+1:

pkg/products/monitoring/prometheusRules.go
david-martin: @pepedocs @HubertStefanski  After trying out this query in a cluster & getting a better understanding of it, it doesn't work as expected.
In fact, it will continuously evaluate true as long as the last pod terminated reason is `OOMKilled` (which will be forever unless the pod restarts for a different reason). There is nothing SRE could do to 'fix' the alert once it starts triggering in a cluster.

I believe there's an alert in place already that, with a small tweak, can pick up on what we want here, but also covers any other terminated state.
See https://github.com/integr8ly/integreatly-operator/blob/d1e5b688bd4b34c97f4e13864be8d57713ec3009/pkg/products/monitoring/prometheusRules.go#L110

I think the query in that alert is not quite right, and If we change that to the below, it will give us what we need.
```
rate(kube_pod_container_status_restarts_total{job='kube-state-metrics'}[5m]) * on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'} * 60 * 5 > 0
```
i.e. A pod in an rhmi namespace is restarting at least once over a 5 minute range, and continues to do so for a total of 15 minutes ( `for: 15m` ).

Separately, I think we need something to debug once off restarts. An alert *doesn't* make sense for that as it's a once off. I'll follow up separately on how best to report & action those scenarios.

pepedocs: @david-martin Thanks for testing the query I've mentioned, I didn't have the exact environment to test it. I think the reaosn why it didn't work is because the pods were not matched and summed together to form a single timeline of termination states. I did some experiments with my local prom setup by simulating OOMKills/restarts, to observe the behavior of   `kube_pod_container_status_last_terminated_reason` and found out that it was returning pods that had termination states, regardless if those pods are already replaced by a newer pod (depends on filters of course and interval). Therefore for those pods that got terminated with OOMKill but got replaced by newer pods, they will likely to return a near 100% value because their timelines end there. For example, (0 is not OOMKilled 1 otherwise)
```
1. Pod A's OOMKill state for the last 5 mins was [0,0,1,1,1,1,1,1,1,1]                
2. OOMKill continues for pod A after another 5 mins, state was [1,1,1,1,1,1,1,1,1,1]
3. SRE fixed something for pod A
4. New pod called pod B was created to replace pod A, pod B's state was after 5 mins [0,0,0,0,0,0,0,0,0,0]
```
Using the query that you have tested, at #1 prom will not fire (80% OOMKill), at #2 prom will fire (100% OOMKill), at #4 prom will still fire because the aforementioned query still returns pod A's OOMKill status as a separate metric. Therefore these pods should be matched and summed up by container names so that the resulting query will return one timeline of pod termination statuses such as [0,0,1,1,1,1,1,1,1,1]  + [1,1,1,1,1,1,1,1,1,1] + [0,0,0,0,0,0,0,0,0,0]. This will then return a 0% OOMKilsl given 5 mins interval, 50% OOMKills for 10 mins interval and 60% OOMKills infor 30 mins interval.

```
(sum by(container) (sum_over_time(kube_pod_container_status_last_terminated_reason{reason='OOMKilled',namespace='default'}[5m])) /
sum by(container) (count_over_time(kube_pod_container_status_last_terminated_reason{reason='OOMKilled',namespace='default'}[5m])) *
on (namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'}) * 100 > 90
```
@HubertStefanski Can you possibly test this query with the summation? I wasn't able to test the specific part of the query - `on clause`. My setup for testing this is very generic.

@david-martin Thanks for suggesting another option. I tried this in my setup and it seems to be difficult to determin which pods were terminated by OOMKill as it only returns the number of restarts. The restarts could be due to other non-OOMKill states.

david-martin: > New pod called pod B was created to replace pod A

@pepedocs This won't always be true. Pod A may be fine after a restart, but the `last_terminated_reason` will still be set. i.e  Pod B may never be created or have to be created.
I'm not happy to have the alert you are suggesting added as it could lead to indefinite alerts firing.
However, if you still want to still go ahead, I firmly believe you're adding toil and forcing a new pod to be created (by deleting the old one) just to get the alert to no longer evaluate true.
Hopefully that makes sense.






> The restarts could be due to other non-OOMKill states

Sure, and they may be valid reasons.
I don't see why OOMKill needs to be called out in it's own alert.
Surely any reason for a pod restarting often is a valid reason to alert, and then look at the last terminated state to see why it terminated.

pepedocs: > > New pod called pod B was created to replace pod A
> 
> @pepedocs This won't always be true. Pod A may be fine after a restart, but the `last_terminated_reason` will still be set. i.e  Pod B may never be created or have to be created.
> I'm not happy to have the alert you are suggesting added as it could lead to indefinite alerts firing.
> However, if you still want to still go ahead, I firmly believe you're adding toil and forcing a new pod to be created (by deleting the old one) just to get the alert to no longer evaluate true.
> Hopefully that makes sense.
> 
> 
> 
> 
> 
> 
> > The restarts could be due to other non-OOMKill states
> 
> Sure, and they may be valid reasons.
> I don't see why OOMKill needs to be called out in it's own alert.
> Surely any reason for a pod restarting often is a valid reason to alert, and then look at the last terminated state to see why it terminated.

Hi @david-martin That was just one example, and the query isnt fixed to that. If pod A gets restarted and not replaced by pod B, then still it will recover from the alert because the number of OOMKill 0's will increase. It does not require the pod at all to be deleted so it does not as well require toil. Unless the pod infintely gets OOMKilled, I don't understand why it will cause indefinite firing as any type of pod recovery (manual or self heal) will allow the alert to recover, number of 0"s will increase, hence sum_over_time will return a smaller value as the time increases.

But if you think it will cause issues please explain the logic when you get a chance I would like to understand as well. I'm really just following the intial requirement for this PR which is to identify OOMKills, if this requirement can be tweaked then we should update it. And probably use the query you've mentioned. By the way in that query what counts as a pod restart? If we use this, we still have to match it to termination status so we can get the reason right?

@HubertStefanski can you please test the queries?

HubertStefanski: @pepedocs Sure, I'll spin up a cluster and try it out :+1:

david-martin: @pepedocs I hope I'm just missing something obvious here.
Here's some graphs where I'm trying to break up the query to understand it.
These graphs are from a test cluster where a single pod (system-sidekiq-2-qnqxs) was OOMKilled once on the 9th May.
image
The pod got killed, and restarted once. Since then, it's been running fine.
However, there's still a record of this termination, which continuously gets reported in the kube_pod_container_status_last_terminated_reason metric. (It's still the last reason it terminated, even though it's running fine)

Using the suggest expression, the value hits 100% and stays at 100%.
Am I missing something here?

HubertStefanski: I tested the Query, It does work, (with the removal of the `namespace:"default"` condition) However I think we probably could do without it, If a pod is getting OOMkilled that often, then I think we're better off just using `kubePodCrashLooping` instead, it watches all crashes in all pods(And over a greater interval), and it will return a reason and the container in which it's happening so we're probably better off having a look at redefining the requirement for this JIRA, to maybe just involve a SOP that deals with OOMKilled `KubePodCrashLooping` alerts

pepedocs: @david-martin Thanks for breakdown the queries and that particular pod example. I see now what you mean and I think you are right. I wasn't able to test this scenario, where a pod recovers and stays in there the last termination state stayed at OOMKilled, therefore the `kube_pod_container_status_last_terminated_reason` -based query will never work.

@HubertStefanski Thanks for testing it, the query seems to work only for pods that get replaced with newer pods or if it gets restarted manually (requires toil).

pkg/products/monitoring/prometheusRules.go
```diff
@@ -99,15 +99,14 @@ func (r *Reconciler) reconcileKubeStateMetricsAlerts(ctx context.Context, server
...
-
 	rules := []monitoringv1.Rule{
 		{
 			Alert: "KubePodCrashLooping",
 			Annotations: map[string]string{
 				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
 				"message": "Pod {{  $labels.namespace  }} / {{  $labels.pod  }} ({{  $labels.container  }}) is restarting {{  $value  }} times every 5 minutes; for the last 15 minutes",
 			},
-			Expr:   intstr.FromString("rate(kube_pod_container_status_restarts_total{job='kube-state-metrics'}[15m]) * on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'} * 60 * 5 > 0"),
+			Expr:   intstr.FromString("rate(kube_pod_container_status_restarts_total{job='kube-state-metrics'}[10m]) * on (namespace, namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'} * 60 * 5 > 0"),
```
david-martin, line 109: Would it be better to lower the `for` threshold to 10m, and set this range to `[5m]` to match the multiplier at the end?

HubertStefanski: I'll test that out :+1:

HubertStefanski: Seems like it's also never reaching the firing state with `[5m]` and `for:10`, keeps alternating between green and pending :thinking:  so it's the same problem that was occuring with `[5m]` and `for:15`, going to test it once more with the current change

pepedocs: @david-martin Out of curiosity, what does the right side exactly mean `* 60 * 5 ` ?

david-martin: @pepedocs `kube_pod_container_status_restarts_total{}[10m]` gives the rate of restarts per second over a 10m range. The `* 60 * 5` is to scale that figure up to get a more user friendly number for the alert message i.e. `restarting {{  $value  }} times every 5 minutes `

@HubertStefanski Can you update the message text as it has `15 minutes` hardcoded

david-martin:
/approve

@alanmoran or @pepedocs would appreciate an lgtm on this tweak to existing alert

pepedocs:
> /approve
> 
> @alanmoran or @pepedocs would appreciate an lgtm on this tweak to existing alert

/lgtm

PR #734, opened 2020-05-15
DEL-243 Add a new make target to integreatly-operator to run the test…

StanleyKaleta:
To verify:

- Make sure a RHMI cluster is available and running, and you have access to the kubeconfig file and it is saved locally somewhere.

- Check out the branch and build it. Run the following command:
`make test/products`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ AUTH_TOKEN=$(shell curl -sH "Content-Type: application/json" -XPOST https://quay
 TEMPLATE_PATH="$(shell pwd)/templates/monitoring"
 INTEGREATLY_OPERATOR_IMAGE ?= $(REG)/$(ORG)/$(PROJECT):v$(TAG)
 CONTAINER_ENGINE ?= docker
+TEST_RESULTS_DIR ?= "test-results"
 
 # If openapi-gen is available on the path, use that; otherwise use it through
 # "go run" (slower)
@@ -159,6 +160,21 @@ test/functional:  export SURF_DEBUG_HEADERS=1
 test/functional:
 	# Run the functional tests against an existing cluster. Make sure you have logged in to the cluster.
 	go clean -testcache && go test -v ./test/functional -timeout=80m
+
+.PHONY: test/products/local
+test/products/local:
+	# Running the products tests against an existing cluster inside a container. Make sure you have logged in to the cluster.
+	# Using 'test-containers.yaml' as config and 'test-results' as output dir
+	mkdir -p "test-results"
+	$(CONTAINER_ENGINE) pull quay.io/integreatly/delorean-cli:master
+	$(CONTAINER_ENGINE) run --rm -e KUBECONFIG=/kube.config -v "${HOME}/.kube/config":/kube.config:z -v $(shell pwd)/test-containers.yaml:/test-containers.yaml -v $(shell pwd)/test-results:/test-results quay.io/integreatly/delorean-cli:master delorean tests run --test-config ./test-containers.yaml --output /test-results --namespace test-products
+
+.PHONY: test/products
+test/products:
+	# Running the products tests against an existing cluster. Make sure you have logged in to the cluster.
+	# Using "test-containers.yaml" as config and $(TEST_RESULTS_DIR) as output dir
+	mkdir -p $(TEST_RESULTS_DIR)
+	delorean tests run --test-config ./test-containers.yaml --output $(TEST_RESULTS_DIR) --namespace test-products
 
 .PHONY: install/olm
 install/olm: cluster/cleanup/olm cluster/cleanup/crds cluster/prepare cluster/prepare/olm/subscription deploy/integreatly-rhmi-cr.yml cluster/check/operator/deployment cluster/prepare/dms cluster/prepare/pagerduty
--- a/test-containers.yaml
+++ b/test-containers.yaml
@@ -1,0 +1,14 @@
+---
+tests:
+  - name: codeready-test
+    image: quay.io/crw/osd-e2e
+    timeout: 3600
+    envVars:
+      - name: CODEREADY_NAMESPACE
+        value: redhat-rhmi-codeready-workspaces
+  - name: 3scale-test
+    image: quay.io/integreatly/3scale-py-testsuite:0.1.0
+    timeout: 3600
+    envVars:
+      - name: NAMESPACE
+        value: redhat-rhmi-3scale
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ tmp/_output/bin/integreatly-operator
 cmd/manager/debug
 build/_output
 build/_test
+test-results
 # Created by https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
 ### Emacs ###
 # -*- mode: gitignore; -*-
--- a/README.md
+++ b/README.md
@@ -178,6 +178,13 @@ To run E2E tests against an existing RHMI cluster:
 make test/functional
 ```
 
+### Products tests
+
+To run products tests against an existing RHMI cluster
+```
+make test/products/local
+```
+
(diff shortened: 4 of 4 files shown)


test-containers.yaml
```diff
@@ -0,0 +1,14 @@
+---
```
merlante, line 1: @wei-lee @mikenairn Will it always make sense to add product tests with a PR to the integreatly-operator? It might make more sense to have a separate repo for the product teams to add a PR against to decouple the integreatly operator and its core tests from the product tests.

Makefile
```diff
@@ -160,6 +160,12 @@ test/functional:
 	# Run the functional tests against an existing cluster. Make sure you have logged in to the cluster.
 	go clean -testcache && go test -v ./test/functional -timeout=80m
 
+.PHONY: test/products
+test/products:
+	# Running the products tests against an existing cluster. Make sure you have logged in to the cluster.
+	$(CONTAINER_ENGINE) pull quay.io/integreatly/delorean-cli:master
+	$(CONTAINER_ENGINE) run -e KUBECONFIG=/kube.config -v "${HOME}/.kube/config":/kube.config:z -v $(shell pwd)/test-containers.yaml:/test-containers.yaml quay.io/integreatly/delorean-cli:master delorean tests run --test-config ./test-containers.yaml --output ./test-results/ --namespace products-tests
+
```
wei-lee, line 168: we should create a tmp dir and attached it to the container so that we can view the results after the command is finished executing.

test-containers.yaml
wei-lee: I can't say always, but at least for now I don't think there's benefit to store this in a separate repo. It will make it more difficult to run both locally or in the pipeline, without many obvious upside.  I don't see a problem with this coupled to the integreatly-operator. It should be, as different versions of integreatly operator may require different versions of test containers. Also if we keep it in the same repo, in the future, when this file is updated, we can run e2e tests against the new test images without any additional work.

Makefile
mikenairn: One concern i have about doing this is how it will get executed on Jenkins/Prow? This will only work locally where someone isn't already running in k8s.

mikenairn: If we can make this command optionally use the containerised delorean cli or just run the command locally it would probably be a better approach. This would allow it to be used in both CI where delorean cli is made available by some other means and local development where we distribute the tooling in the container.

StanleyKaleta: @mikenairn what do you mean by "containerised delorean cli"?

mikenairn: @StanleyKaleta This is running the command in this container quay.io/integreatly/delorean-cli:master which is useful to allow developers to run it without requiring installing dependencies other than docker.

This won't work in CI though, so you need to be able to just run the command as if the delorean cli is installed locally for those envs:

```
delorean tests run --test-config ./test-containers.yaml --output ./test-results/ --namespace products-tests
```

StanleyKaleta: ok, so for this I'll need to know where delorean cli is installed and then just run delorean command from that dir, right?

mikenairn: You can probably assume it's on the PATH, this will be the case for CI. Should be using the container anywhere else.

Makefile
```diff
@@ -160,6 +160,12 @@ test/functional:
 	# Run the functional tests against an existing cluster. Make sure you have logged in to the cluster.
 	go clean -testcache && go test -v ./test/functional -timeout=80m
 
+.PHONY: test/products
+test/products:
+	# Running the products tests against an existing cluster. Make sure you have logged in to the cluster.
+	$(CONTAINER_ENGINE) pull quay.io/integreatly/delorean-cli:master
+	$(CONTAINER_ENGINE) run -e KUBECONFIG=/kube.config -v "${HOME}/.kube/config":/kube.config:z -v $(shell pwd)/test-containers.yaml:/test-containers.yaml quay.io/integreatly/delorean-cli:master delorean tests run --test-config ./test-containers.yaml --output ./test-results/ --namespace products-tests
```
mikenairn, line 167: Add --rm to any "run" commands or you end up with lots of containers left around https://docs.docker.com/engine/reference/run/#clean-up---rm

Makefile
StanleyKaleta: @mikenairn @wei-lee updated first command to include mounted dir and added second command for CI

Makefile
```diff
@@ -160,6 +160,19 @@ test/functional:
 	# Run the functional tests against an existing cluster. Make sure you have logged in to the cluster.
 	go clean -testcache && go test -v ./test/functional -timeout=80m
 
+.PHONY: test/products/local
+test/products/local:
+	# Running the products tests against an existing cluster inside a container. Make sure you have logged in to the cluster.
+	# Using 'test-containers.yaml' as config and 'tmp' as output dir
+	$(CONTAINER_ENGINE) pull quay.io/integreatly/delorean-cli:master
+	$(CONTAINER_ENGINE) run --rm -e KUBECONFIG=/kube.config -v "${HOME}/.kube/config":/kube.config:z -v $(shell pwd)/test-containers.yaml:/test-containers.yaml -v $(shell pwd)/tmp:/test-results quay.io/integreatly/delorean-cli:master delorean tests run --test-config ./test-containers.yaml --output /test-results --namespace test-products
```
wei-lee, line 168: is the `$(shell pwd)/tmp` exist locally? Do we need to create it first?

StanleyKaleta: @wei-lee updated

Makefile
```diff
@@ -160,6 +160,21 @@ test/functional:
...
+	mkdir "test-results"
+	$(CONTAINER_ENGINE) pull quay.io/integreatly/delorean-cli:master
+	$(CONTAINER_ENGINE) run --rm -e KUBECONFIG=/kube.config -v "${HOME}/.kube/config":/kube.config:z -v $(shell pwd)/test-containers.yaml:/test-containers.yaml -v $(shell pwd)/test-results:/test-results quay.io/integreatly/delorean-cli:master delorean tests run --test-config ./test-containers.yaml --output /test-results --namespace test-products
+
+.PHONY: test/products
+test/products:
+	# Running the products tests against an existing cluster. Make sure you have logged in to the cluster.
+	# Using 'test-containers.yaml' as config and 'test-results' as output dir
+	mkdir "test-results"
+	delorean tests run --test-config ./test-containers.yaml --output ./test-results --namespace test-products
```
wei-lee, line 176: when we run this inside CI, I don't think we should create the directory here. Let's make the output dir configurable, so that we can control the output directory in the pipeline.

Makefile
```diff
@@ -160,6 +160,21 @@ test/functional:
 	# Run the functional tests against an existing cluster. Make sure you have logged in to the cluster.
 	go clean -testcache && go test -v ./test/functional -timeout=80m
 
+.PHONY: test/products/local
+test/products/local:
+	# Running the products tests against an existing cluster inside a container. Make sure you have logged in to the cluster.
+	# Using 'test-containers.yaml' as config and 'test-results' as output dir
+	mkdir "test-results"
```
wei-lee, line 167: make sure this is added to the `.gitignore` file

Makefile
StanleyKaleta: @wei-lee updated

Makefile
StanleyKaleta: added

Makefile
mikenairn: I'm not sure we have to deal with this right now, but we should endeavour to remove the need for 2 separate make targets. It should be possible to have the command we want to run, run inside the container if we aren't already running in a container, and using the binary directly otherwise.

mikenairn:
I'd much rather see the e2e tests pass when making changes to the Makefile since changes here could quite easily cause errors. However, the e2e test failed on our own test suite, so any changes here have effectively been tested. Also, it's likely going to have to pass again for tide anyway before it will actually merge.

/override ci/prow/e2e

PR #745, opened 2020-05-19
Adding in additional permissions to resolve metrics error in the logs

laurafitzgerald:
# Description
https://issues.redhat.com/browse/DEL-261

Error faced
```
failed to initialize service object for metrics: replicasets.apps \"rhmi-operator-5b7847c84\" is forbidden: User \"system:serviceaccount:redhat-rhmi-operator:rhmi-operator\" cannot get resource \"replicasets\" in API group \"apps\" in the namespace \"redhat-rhmi-operator\""}
```

Note: I didn't deploy this using olm.
I used `make cluster/prepare` `make cluster/prepare/crd` and oc deployed the required resources in the rhmi namespace

## Type of change

## Checklist

Please contact me for the cluster where I have this deployed.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -101,6 +101,19 @@ rules:
       - list
       - watch
 
+  # Permissions needed to be able to grant to prometheus-k8s SA from openshift-monitoring NS
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+      - services
+      - endpoints
+    verbs:
+      - get
+      - list
+      - watch
+  # END Permissions needed to be able to grant to prometheus-k8s SA from openshift-monitoring NS
+
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -48,8 +48,10 @@ rules:
       - apps
     resources:
       - deployments/finalizers
+      - replicasets
     verbs:
       - update
+      - get
   # END For monitoring functions by operator-sdk
 
   # Our own Subscription needs to be switched into Manual mode after initial installation


laurafitzgerald:
btw. I'm not sure if this is the correct place to put this permission. There seems to be a more defined structure to this role file than previous operators I've worked on. @matskiv I see you've contributed to this file a lot. Could you take a look and I can move it somewhere else if required.

matskiv:
@laurafitzgerald The failure happens in `redhat-rhmi-operator` namespace. Adding replicaset permission to operators Role, instead of ClusterRole is prefferable. The Role is in the same file, just above the ClusterRole.

laurafitzgerald:
@matskiv made the change. You can see the metric exposed by integreatly-operator is present here:
Screenshot 2020-05-21 at 12 24 42

https://prometheus-route-redhat-rhmi-middleware-monitoring-operator.apps.lfitzger.l1o6.s1.devshift.org/graph?g0.range_input=1h&g0.expr=integreatly_version_info&g0.tab=1



Steps I took to reverify - `oc apply -f deploy/role.yaml` - delete operator-pod, logs confirm that metrics service is running and the ServiceMonitor is present. 
Screenshot 2020-05-21 at 12 24 51

https://console-openshift-console.apps.lfitzger.l1o6.s1.devshift.org/k8s/ns/redhat-rhmi-operator/monitoring.coreos.com~v1~ServiceMonitor/rhmi-operator-metrics

cluster will be deleted this evening. contact me for credentials if required...

matskiv:
@laurafitzgerald Thanks for the verification steps in your comment 👍

grdryn:
@matskiv you spelled /approve as /approvee, so the label didn't get added as expected.

I tried to see if it would update if I modified your commit (I'm an org admin but not an approver here). It would have been a useful hack, but thankfully it didn't work! :smile: 

It seems you'll need to /approve in a new comment yourself, if that was your original intention.

matskiv:
Ooops.. That was not intentional :)

/approvе

matskiv:
Ok, in that last comment I was messing with you 😅 

/approve

grdryn:
@matskiv what's the special character?! :see_no_evil: 

```python
[gryan@localhost integreatly-operator]$ python
Python 3.8.2 (default, Feb 28 2020, 00:00:00) 
[GCC 10.0.1 20200216 (Red Hat 10.0.1-0.8)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> "/approvе" == "/approve"
False
>>> "/approve" == "/approve"
True
>>> 
```

laurafitzgerald:
is the failure here still related to the previous failures or has that been resolved?

matskiv:
@grdryn Last character is Cyrillic `е`

@laurafitzgerald Looks like install failed terribly. There will be more info on "Artifacts" page of the build, in `artifacts/e2e/` folder. Ping me if you need further help investigating this.

grdryn:
@laurafitzgerald it's still the same issue as before. Raj's PR to fix that got merged _after_ this latest run started, so it should have more success on a new attempt.

/retest

matskiv:
Hmm, if it's not in retesting stage in Tide, won't it use just Laura's branch as is?
Don't we need to do a rebase to pick up that fix from Raj?

grdryn:
@matskiv I might be wrong as there's a lot I don't know about prow/tide, but I think it can be lazy about retesting.

My understanding is that it's never just testing on e.g. Laura's branch as-is, but rather on the state of merging Laura's branch with master.

If that had _passed_ but Raj's changes had been merged before it could merge this (and all other merge conditions were satisfied) then I think it would kick off a retest (it's trying to get to a merge, but it needs to ensure that the state at merge is a tested one). If there were other conditions that weren't yet satisfied (missing lgtm for example) then I think it would lazily wait until that was there before trying to run again.

As this _failed_, I don't think it eagerly would re-run just because master has changed -- it's up to a human to decide if a new run with the new master would succeed.

@mikenairn does that sound like how it works? A lot if it is just guesses based on experience.

matskiv:
@grdryn I am not 100% sure about this myself.
I thought that on first test it tested the branch as is.

But you are right about "lazy re-tests".

grdryn:
@matskiv If for the first run the branch is already up-to-date with master, then it might be incidentally as-is (merging master with something that already contains master shouldn't result in a merge commit, so it would be the branch as-is)?

grdryn:
So the error that this is failing with, I don't understand how only this PR is failing with it. How did it not get caught in #755 for example (which was supposed to fix the issue, but didn't completely), or on master since then?

@matskiv @philbrookes @mikenairn any ideas?

grdryn:
@david-martin @matskiv this is now passing. Since I added changes, could one of you lgtm?

matskiv:
@grdryn In which namespace was the latest permission problem?

david-martin:
/lgtm
Missing replicaset perm makes sense, given the error.
And missing pods, services, endpoints perms makes sense so operator can delegate them

grdryn:
@matskiv I'm not actually sure what namespace(s) it's failing on, but here's the log message:

```
time="2020-05-21T14:48:40Z" level=error msg="failed to reconcile: roles.rbac.authorization.k8s.io \"rhmi-prometheus-k8s\" is forbidden: user \"system:serviceaccount:redhat-rhmi-operator:rhmi-operator\" (groups=[\"system:serviceaccounts\" \"system:serviceaccounts:redhat-rhmi-operator\" \"system:authenticated\"]) is attempting to grant RBAC permissions not currently held:\n{APIGroups:[\"\"], Resources:[\"endpoints\"], Verbs:[\"get\" \"list\" \"watch\"]}"
```

grdryn:
Here's the "event" that's emitted when the operator hits this error:

```json
{
            "apiVersion": "v1",
            "count": 119,
            "eventTime": null,
            "firstTimestamp": "2020-05-21T14:15:33Z",
            "involvedObject": {
                "apiVersion": "integreatly.org/v1alpha1",
                "kind": "RHMI",
                "name": "rhmi",
                "namespace": "redhat-rhmi-operator",
                "resourceVersion": "18457",
                "uid": "cb425674-8188-4b38-8c45-e9f3b440de5f"
            },
            "kind": "Event",
            "lastTimestamp": "2020-05-21T14:45:36Z",
            "message": "Failed to reconcile::\nroles.rbac.authorization.k8s.io \"rhmi-prometheus-k8s\" is forbidden: user \"system:serviceaccount:redhat-rhmi-operator:rhmi-operator\" (groups=[\"system:serviceaccounts\" \"system:serviceaccounts:redhat-rhmi-operator\" \"system:authenticated\"]) is attempting to grant RBAC permissions not currently held:\n{APIGroups:[\"\"], Resources:[\"endpoints\"], Verbs:[\"get\" \"list\" \"watch\"]}",
            "metadata": {
                "creationTimestamp": "2020-05-21T14:15:33Z",
                "name": "rhmi.161110b9c833838f",
                "namespace": "redhat-rhmi-operator",
                "resourceVersion": "27488",
                "selfLink": "/api/v1/namespaces/redhat-rhmi-operator/events/rhmi.161110b9c833838f",
                "uid": "25dc0d5e-badd-4e26-8627-4051e33e7747"
            },
            "reason": "ProcessingError",
            "reportingComponent": "",
            "reportingInstance": "",
            "source": {
                "component": "monitoring-spec"
            },
            "type": "Warning"
        }
```

It has the `redhat-rhmi-operator` listed as the "namespace" in the metadata there. I'm not sure if that's where it's failing to apply the Role/RoleBinding though. I'll try to dig around on an existing cluster, and maybe try with moving the permissions to the Role here rather than the ClusterRole.

matskiv:
@grdryn This looks like it's happening in redhat-rhmi-operator namespace.
Not because it's in the event, but because it's failing because of just on permission(endpoints).
Operator should have that permission in product namespaces by being admin of those.
Adding it to the Role should fix this. :)

grdryn:
I've moved the permissions up to the Role rather than the RoleBinding, and did a quick sanity check to make sure that the rhmi-operator has all of those permissions on the other RHMI namespaces. Here's what it's like without this change, and we can see that the permissions only need to be added to the redhat-rhmi-operator namespace, so using the Role rather than RoleBinding is more appropriate:

```
[gryan@localhost pagerduty-operator]$ oc access-matrix -n redhat-rhmi-operator | grep -E '^NAME |^services |^endpoints |^pods '
NAME                                                          LIST  CREATE  UPDATE  DELETE
endpoints                                                     ✖     ✖       ✖       ✖
pods                                                          ✔     ✔       ✔       ✖
services                                                      ✔     ✔       ✔       ✖
[gryan@localhost pagerduty-operator]$ oc access-matrix -n redhat-rhmi-3scale | grep -E '^NAME |^services |^endpoints |^pods '
NAME                                                          LIST  CREATE  UPDATE  DELETE
endpoints                                                     ✔     ✔       ✔       ✔
pods                                                          ✔     ✔       ✔       ✔
services                                                      ✔     ✔       ✔       ✔
[gryan@localhost pagerduty-operator]$ oc access-matrix -n redhat-rhmi-3scale-operator | grep -E '^NAME |^services |^endpoints |^pods '
NAME                                                          LIST  CREATE  UPDATE  DELETE
endpoints                                                     ✔     ✔       ✔       ✔
pods                                                          ✔     ✔       ✔       ✔
services                                                      ✔     ✔       ✔       ✔
```

grdryn:
/unhold

@matskiv any idea why this wouldn't have been hit in other PRs or the master e2e job? It should have been hit in #755 at the very least.

matskiv:
@grdryn I am not sure. I don't see RHMI CR or rhmi-operator pod logs in artifacts of successful e2e build on master or on 755 PR. :/

grdryn:
/retest

^test flake, being addressed in https://github.com/integr8ly/integreatly-operator/pull/780.

grdryn:
@matskiv Mike points out that those only get copied over on failure: https://github.com/integr8ly/integreatly-operator/blob/master/test/e2e/integreatly_test.go#L130 Maybe we should always archive them?

matskiv:
@grdryn yes, that could be useful in cases like this :)

deploy/role.yaml
```diff
@@ -99,6 +101,19 @@ rules:
       - list
       - watch
 
+  # Permissions needed to be able to grant to prometheus-k8s SA from openshift-monitoring NS
+  - apiGroups:
+      - ""
```
laurafitzgerald, line 106:
```suggestion
      - rhmi-prometheus-k8s
```
I was seeing the same error when trying to upgrade form one version to another. It got stuck reconciling on monitoring. Applied this change but had to make the above suggestion to make it work.

matskiv: This doesn't make much sense to me.
There is no APIGroup "rhmi-prometheus-k8s".

You can read more about these groups here: https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups

grdryn: As discussed in chat, this would be incorrect as the apiGroup, but the act of modifying the role on the cluster may have caused something else to reconcile and fix your issue.

PR #754, opened 2020-05-20
INTLY-7238 Switched from update to patch

CathalOConnorRH:
Switched from update to patch and check for readyreplicas before passing test.

# Description
Switched from update to patch and check for readyreplicas before passing test.
Tests were reported as failing in these two tickets. 
- https://issues.redhat.com/browse/INTLY-7238 
- https://issues.redhat.com/browse/INTLY-7556

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/test/common/scale_up_replicas_apicurito.go
+++ b/test/common/scale_up_replicas_apicurito.go
@@ -7,8 +7,8 @@ import (
 	"time"
 
 	apicuritov1alpha1 "github.com/apicurio/apicurio-operators/apicurito/pkg/apis/apicur/v1alpha1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -22,6 +22,8 @@ const (
 	apicuritoNamespace         = "redhat-rhmi-apicurito"
 	retryIntervalApicurito     = time.Second * 20
 	timeoutApicurito           = time.Minute * 7
+	requestURLApicturito       = "/apis/apicur.io/v1alpha1"
+	kindApicurito              = "apicuritos"
 )
 
 func TestReplicasInApicurito(t *testing.T, ctx *TestingContext) {
@@ -36,7 +38,7 @@ func TestReplicasInApicurito(t *testing.T, ctx *TestingContext) {
 		t.Fatalf("Incorrect number of replicas to start : %v", err)
 	}
 
-	apicuritoCR, err = updateApicurito(ctx.Client, scaleUpApicuritoReplicas, t)
+	apicuritoCR, err = updateApicurito(ctx, scaleUpApicuritoReplicas, t)
 	if err != nil {
 		t.Fatalf("Unable to update : %v", err)
 	}
@@ -46,7 +48,7 @@ func TestReplicasInApicurito(t *testing.T, ctx *TestingContext) {
 		t.Fatalf("Incorrect number of replicas : %v", err)
 	}
 
-	apicuritoCR, err = updateApicurito(ctx.Client, scaleDownApicuritoReplicas, t)
+	apicuritoCR, err = updateApicurito(ctx, scaleDownApicuritoReplicas, t)
 	if err != nil {
 		t.Fatalf("Unable to update : %v", err)
 	}
@@ -56,6 +58,31 @@ func TestReplicasInApicurito(t *testing.T, ctx *TestingContext) {
 		t.Fatalf("Incorrect number of replicas : %v", err)
 	}
 
+	t.Log("Checking replicas are ready")
+	if err := checkReplicasAreReady(ctx, t, numberOfApicuritoReplicas); err != nil {
+		t.Fatalf("Replicas weren't ready within timeout")
+	}
+
+}
+
+func checkReplicasAreReady(dynClient *TestingContext, t *testing.T, replicas int32) error {
+
+	err := wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+
+		deployment, err := dynClient.KubeClient.AppsV1().Deployments(apicuritoNamespace).Get(apicuritoName, metav1.GetOptions{})
+		if err != nil {
+			t.Errorf("Failed to get Deployment %s in namespace %s with error: %s", apicuritoName, apicuritoNamespace, err)
+		}
+		if deployment.Status.ReadyReplicas == replicas {
+			t.Logf("Replicas Ready %v", deployment.Status.ReadyReplicas)
+			return true, nil
+		}
+		return false, fmt.Errorf("%v", deployment.Status.ReadyReplicas)
+	})
+	if err != nil {
+		return fmt.Errorf("Number of replicas for apicurito.Spec.Size is not correct : Replicas - %v, Expected - %v", err, replicas)
+	}
+	return nil
 }
 
 func getApicurito(dynClient k8sclient.Client) (apicuritov1alpha1.Apicurito, error) {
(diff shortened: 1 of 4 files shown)


test/common/scale_up_replicas_apicurito.go
```diff
@@ -46,16 +48,52 @@ func TestReplicasInApicurito(t *testing.T, ctx *TestingContext) {
 		t.Fatalf("Incorrect number of replicas : %v", err)
 	}
 
-	apicuritoCR, err = updateApicurito(ctx.Client, scaleDownApicuritoReplicas, t)
+	t.Log("Chekcing replicas are ready")
```
pawelpaszki, line 51:
```suggestion
	t.Log("Checking replicas are ready")
```

test/common/scale_up_replicas_apicurito.go
```diff
@@ -46,16 +48,52 @@ func TestReplicasInApicurito(t *testing.T, ctx *TestingContext) {
...
+	if err := checkReplicasAreReady(ctx, t, scaleUpApicuritoReplicas); err != nil {
+		t.Fatalf("Replicas weren't ready within timeout")
+	}
+
+	apicuritoCR, err = updateApicurito(ctx, scaleDownApicuritoReplicas, t)
 	if err != nil {
 		t.Fatalf("Unable to update : %v", err)
 	}
 
+	t.Log("Chekcing replicas are ready")
```
pawelpaszki, line 61:
```suggestion
	t.Log("Checking replicas are ready")
```

test/common/scale_up_replicas_apicurito.go
```diff
@@ -46,16 +48,52 @@ func TestReplicasInApicurito(t *testing.T, ctx *TestingContext) {
...
+	if err := checkReplicasAreReady(ctx, t, scaleDownApicuritoReplicas); err != nil {
+		t.Fatalf("Replicas weren't ready within timeout")
+	}
+
 	t.Log("Checking correct number of updated replicas are set")
 	if err := checkNumberOfReplicasAgainstValueApicurito(apicuritoCR, ctx, numberOfApicuritoReplicas, retryIntervalApicurito, timeoutApicurito, t); err != nil {
 		t.Fatalf("Incorrect number of replicas : %v", err)
 	}
 
+	t.Log("Chekcing replicas are ready")
```
pawelpaszki, line 71:
```suggestion
	t.Log("Checking replicas are ready")
```

test/common/scale_up_replicas_rhsso.go
```diff
@@ -38,7 +41,12 @@ func checkScalingOfKeycloakReplicas(t *testing.T, ctx *TestingContext, keycloakC
 		t.Fatalf("Incorrect number of replicas to start : %v", err)
 	}
 
-	keycloakCR, err = updateKeycloakCR(ctx.Client, scaleUpRhssoReplicas, keycloakCRName, keycloakCRNamespace)
+	t.Log("Chekcing replicas are ready")
```
pawelpaszki, line 44:
```suggestion
	t.Log("Checking replicas are ready")
```

test/common/scale_up_replicas_rhsso.go
```diff
@@ -48,17 +56,27 @@ func checkScalingOfKeycloakReplicas(t *testing.T, ctx *TestingContext, keycloakC
 		t.Fatalf("Incorrect number of replicas : %v", err)
 	}
 
-	keycloakCR, err = updateKeycloakCR(ctx.Client, scaleDownRhssoReplicas, keycloakCRName, keycloakCRNamespace)
+	t.Log("Chekcing replicas are ready")
```
pawelpaszki, line 59:
```suggestion
	t.Log("Checking replicas are ready")
```

test/common/scale_up_replicas_rhsso.go
```diff
@@ -67,6 +85,11 @@ func checkScalingOfKeycloakReplicas(t *testing.T, ctx *TestingContext, keycloakC
 	if err := checkNumberOfReplicasAgainstValueRhsso(keycloakCR, ctx, numberOfRhssoReplicas, retryInterval, timeout, t); err != nil {
 		t.Fatalf("Incorrect number of replicas : %v", err)
 	}
+
+	t.Log("Chekcing replicas are ready")
```
pawelpaszki, line 89:
```suggestion
	t.Log("Checking replicas are ready")
```

CathalOConnorRH: Thanks for those. will push a new version now.

pawelpaszki:
ran the tests locally, but would be interested to see what @davidffrench has to say.

test/common/scale_up_replicas_apicurito.go
```diff
@@ -46,16 +48,52 @@ func TestReplicasInApicurito(t *testing.T, ctx *TestingContext) {
...
 
+	t.Log("Checking replicas are ready")
+	if err := checkReplicasAreReady(ctx, t, numberOfApicuritoReplicas); err != nil {
+		t.Fatalf("Replicas weren't ready within timeout")
+	}
+
+}
+
+func checkReplicasAreReady(dynClient *TestingContext, t *testing.T, replicas int32) error {
+	deployment, err := dynClient.KubeClient.AppsV1().Deployments(apicuritoNamespace).Get(apicuritoName, metav1.GetOptions{})
```
davidffrench, line 79: This should not be required with the same get inside the poll on line 83 https://github.com/integr8ly/integreatly-operator/pull/754/files#diff-c390ece00887543e97f361f7e04d648aR83

test/common/scale_up_replicas_apicurito.go
```diff
@@ -68,26 +106,27 @@ func getApicurito(dynClient k8sclient.Client) (apicuritov1alpha1.Apicurito, erro
 	return *apicuritoCR, nil
 }
 
-func updateApicurito(dynClient k8sclient.Client, replicas int32, t *testing.T) (apicuritov1alpha1.Apicurito, error) {
+func updateApicurito(dynClient *TestingContext, replicas int32, t *testing.T) (apicuritov1alpha1.Apicurito, error) {
```
davidffrench, line 109:
```suggestion
func updateApicurito(ctx *TestingContext, replicas int32, t *testing.T) (apicuritov1alpha1.Apicurito, error) {
```

Small name change but keep it consistent with other functions.

CathalOConnorRH:
> Good work @CathalOConnorRH .
> 
> Could you tell me again what was happening when the deployment was not in a ready state?

When this test completes a subsequent test fails due to the pods not being ready when the next tests tries to access them. By adding in the check here we try to ensure it's safe to proceed when the pods are ready.

davidffrench:
> > Good work @CathalOConnorRH .
> > Could you tell me again what was happening when the deployment was not in a ready state?
> 
> When this test completes a subsequent test fails due to the pods not being ready when the next tests tries to access them. By adding in the check here we try to ensure it's safe to proceed when the pods are ready.

Perfect, thank you. Do you think it is worth just adding the ready check at the end of the tests?

CathalOConnorRH:
> > > Good work @CathalOConnorRH .
> > > Could you tell me again what was happening when the deployment was not in a ready state?
> > 
> > 
> > When this test completes a subsequent test fails due to the pods not being ready when the next tests tries to access them. By adding in the check here we try to ensure it's safe to proceed when the pods are ready.
> 
> Perfect, thank you. Do you think it is worth just adding the ready check at the end of the tests? This might reduce the time it is taking to run

That depends. are we happy to assume that the scaling up is ok regardless of the pod status once the CR is updated ? If we are then just checking the pods are ready on the last reset test should be ok.

davidffrench:
> > > > Good work @CathalOConnorRH .
> > > > Could you tell me again what was happening when the deployment was not in a ready state?
> > > 
> > > 
> > > When this test completes a subsequent test fails due to the pods not being ready when the next tests tries to access them. By adding in the check here we try to ensure it's safe to proceed when the pods are ready.
> > 
> > 
> > Perfect, thank you. Do you think it is worth just adding the ready check at the end of the tests? This might reduce the time it is taking to run
> 
> That depends. are we happy to assume that the scaling up is ok regardless of the pod status once the CR is updated ? If we are then just checking the pods are ready on the last reset test should be ok.

I would say yes, this would then be testing that 3sacle itself for example is correctly updating the deployment and testing the deployment is ready but this test should just be that we can modify the replicas in the CR

CathalOConnorRH:
> > > > > Good work @CathalOConnorRH .
> > > > > Could you tell me again what was happening when the deployment was not in a ready state?
> > > > 
> > > > 
> > > > When this test completes a subsequent test fails due to the pods not being ready when the next tests tries to access them. By adding in the check here we try to ensure it's safe to proceed when the pods are ready.
> > > 
> > > 
> > > Perfect, thank you. Do you think it is worth just adding the ready check at the end of the tests? This might reduce the time it is taking to run
> > 
> > 
> > That depends. are we happy to assume that the scaling up is ok regardless of the pod status once the CR is updated ? If we are then just checking the pods are ready on the last reset test should be ok.
> 
> I would say yes, this would then be testing that 3sacle itself for example is correctly updating the deployment and testing the deployment is ready but this test should just be that we can modify the replicas in the CR

ok perfect, I'll update the tests to only check the pods at the last step.

matskiv:
@davidffrench @pawelpaszki Hey guys, you reviewed this already, can you give final look and approve it all looks good? Thank you.

PR #755, opened 2020-05-20
Fix the rolebinding permission issue

Rajagopalan-Ranganathan:
# Description
JIRA Link: https://issues.redhat.com/browse/INTLY-7854

Operator did not have permissions to create cluster role bindings, since
it runs with service account privileges. Code has been modified to create
a new role with the minimal required permissions and create a corresponding rolebinding
for them, so that prometheus can scrape metrics from the services.

## Type of change

## Checklist

### DIFF
--- a/pkg/products/monitoringspec/reconciler.go
+++ b/pkg/products/monitoringspec/reconciler.go
@@ -321,7 +321,7 @@ func (r *Reconciler) reconcileRole(ctx context.Context,
 		return nil
 	})
 	if opRes != controllerutil.OperationResultNone {
-		r.Logger.Infof("operation result of creating role: %v was %v", roleBindingName, opRes)
+		r.Logger.Infof("operation result of creating role: %v was %v", roleRefName, opRes)
 	}
 	return err
 }
--- a/pkg/products/monitoringspec/reconciler.go
+++ b/pkg/products/monitoringspec/reconciler.go
@@ -293,7 +293,7 @@ func (r *Reconciler) reconcileRole(ctx context.Context,
 
 	role := &rbac.Role{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      roleBindingName,
+			Name:      roleRefName,
 			Namespace: namespace,
 		},
 	}
--- a/pkg/products/monitoringspec/reconciler.go
+++ b/pkg/products/monitoringspec/reconciler.go
@@ -29,11 +29,11 @@ import (
 const (
 	defaultInstallationNamespace              = "monitoring"
 	packageName                               = "monitoringspec"
-	roleBindingName                           = "prometheus-k8s"
+	roleBindingName                           = "rhmi-prometheus-k8s"
 	clusterMonitoringPrometheusServiceAccount = "prometheus-k8s"
 	clusterMonitoringNamespace                = "openshift-monitoring"
 	roleRefAPIGroup                           = "rbac.authorization.k8s.io"
-	roleRefName                               = "view"
+	roleRefName                               = "rhmi-prometheus-k8s"
 	labelSelector                             = "monitoring-key=middleware"
 	clonedServiceMonitorLabelKey              = "integreatly.org/cloned-servicemonitor"
 	clonedServiceMonitorLabelValue            = "true"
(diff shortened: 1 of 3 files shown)


pkg/products/monitoringspec/reconciler.go
```diff
@@ -276,14 +276,56 @@ func (r *Reconciler) reconcileRoleBindingsForServiceMonitor(ctx context.Context,
...
 	}
 	return err
 }
 
+func (r *Reconciler) reconcileRole(ctx context.Context,
+	serverClient k8sclient.Client, namespace string) (err error) {
+
+	role := &rbac.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      roleBindingName,
```
grdryn, line 296: Given that the variable `roleBindingName` is now used also for a "Role" here (so not just a RoleBinding), I think maybe the variable should be renamed to something like `prometheusK8s` or something.

test/common/monitoringspec.go
```diff
@@ -61,7 +62,11 @@ func TestServiceMonitorsCloneAndRolebindingsExist(t *testing.T, ctx *TestingCont
 	//Check if rolebindings exists
 	for _, sm := range monSermonMap {
 		for _, namespace := range sm.Spec.NamespaceSelector.MatchNames {
-			err := checkRoleBindingExists(ctx, roleBindingName, namespace)
+			err := checkRoleExists(ctx, roleRefName, namespace)
+			if err != nil {
+				t.Fatal("Error retrieving rolebinding: ", err, "in namespace:", namespace)
```
grdryn, line 67:
```suggestion
				t.Fatal("Error retrieving role: ", err, "in namespace:", namespace)
```

pkg/products/monitoringspec/reconciler.go
```diff
@@ -276,14 +276,56 @@ func (r *Reconciler) reconcileRoleBindingsForServiceMonitor(ctx context.Context,
...
+			"services",
+			"endpoints",
+			"pods",
+		}
+
+		verbs := []string{
+			"get",
+			"list",
+			"watch",
+		}
```
grdryn, line 311: Is the rhmi-operator SA able to grant these permissions in all namespaces? Even if it happens to be able to do it by default, I think maybe we should add those permissions to the ClusterRole specified in deploy/role.yaml. WDYT?

pkg/products/monitoringspec/reconciler.go
Rajagopalan-Ranganathan: Sorry, I had another constant for rolename, updated it with the correct variable. Fixed.

test/common/monitoringspec.go
Rajagopalan-Ranganathan: Fixed.

pkg/products/monitoringspec/reconciler.go
Rajagopalan-Ranganathan: Yes, it can grant permissions in all namespaces. Discussed today morning with @matskiv and verified that role.yaml doesnt needed to be updated at this time.

matskiv: not in "all namespaces" actually,
only in the namespaces created by rhmi-operator SA.

It is fairly safe set of permissions IMHO.
But we were trying to keep cluster wide permissions minimal, so I would not add additional cluster permissions if they are not required at the moment.

pkg/products/monitoringspec/reconciler.go
```diff
@@ -276,10 +276,52 @@ func (r *Reconciler) reconcileRoleBindingsForServiceMonitor(ctx context.Context,
...
+			{
+				APIGroups: apiGroups,
+				Resources: resources,
+				Verbs:     verbs,
+			},
+		}
+		return nil
+	})
+	if opRes != controllerutil.OperationResultNone {
+		r.Logger.Infof("operation result of creating role: %v was %v", roleBindingName, opRes)
```
grdryn, line 324:
```suggestion
		r.Logger.Infof("operation result of creating role: %v was %v", roleRefName, opRes)
```

Rajagopalan-Ranganathan: fixed.

PR #756, opened 2020-05-20
INTLY-7829 Remove dependency on Service

sergioifg94:
# Description

Due to the errors when upgrading the operator through the OLM, remove the creation of the
webhooks Service prior to the operator initialization and add logic in the operator to create the
service before the webhooks service is initialized

# Verification

## Verify that the validating webhook still works

Same verification steps as #660, excluding the `make cluster/prepare/service` step

## Verify that the upgrade succeeds

1. Checkout to the `v2.2.0` tag and install RHMI through OLM
2. Back to this PR, create a `2.2.1` release from it:
    1. Run `SEMVER=2.2.1 ./scripts/prepare-release.sh`
    2. Edit the ORG in Makefile to your quay.io username
    3. Check that the 2.2.1 CSV points to your image or modify it if it does not
    4. Push the CSV to quay: `QUAY_USERNAME=<your username> QUAY_PASSWORD=<your password> REPO=<your username> make push/csv`
3. Delete the rhmi pod in `openshift-marketplace`
4. A new InstallPlan for 2.2.1 will be created, update the `.scec.upgrade.alwaysImmediately` in the RHMIConfig instance to `true`. The operator will automatically approve the upgrade
5. Verify that the pod with the latest image is created successfully and no errors in the InstallPlan
6. Check that the validating webhooks are working correctly.
## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


matskiv:
@sergioifg94 this probably needs to be rebased to pick up https://github.com/integr8ly/integreatly-operator/pull/768
otherwise e2e will probably fail

sergioifg94:
@matskiv thanks for the tip, will do so

pkg/webhooks/operator_webhooks.go
```diff
@@ -147,6 +173,65 @@ func (webhookConfig *IntegreatlyWebhookConfig) Reconcile(ctx context.Context, cl
...
+// ReconcileService creates or updates the service that points to the Pod
+func (webhookConfig *IntegreatlyWebhookConfig) ReconcileService(ctx context.Context, client k8sclient.Client) error {
+	service := &corev1.Service{
+		ObjectMeta: v1.ObjectMeta{
+			Name:      operatorPodServiceName,
+			Namespace: "redhat-rhmi-operator",
+		},
+	}
+
+	_, err := controllerutil.CreateOrUpdate(ctx, client, service, func() error {
```
pb82, line 185: Won't this fail when the service already exists because we're not setting the clusterIP? Or is the old value of the clusterIP implicitely copied here? What about always setting it to `None`?

matskiv: It probably won't fail createOrUpdate applies changes to the object it first `Get`s.
So the clusterIP field in the `Update` call will be set to previous value.

But I raised a related concern in [a comment in JIRA](https://issues.redhat.com/browse/INTLY-7829?focusedCommentId=14108855&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14108855), so we might need to set it to `None` for that purpose.

sergioifg94: No, it will keep whatever value it had before, or leave it empty it it's being created. The problem of setting it to `None` is that if the existing service has a different value it will fail.
This way the upgrade from 2.2.0 won't fail as the existing service has an automatically assigned clusterIP

matskiv: We can do Get before the CreateOrUpdate, and Delete Service if clusterIP is not empty.
This would get all clusters into a consistent state.
@sergioifg94 @pb82  Wdyt?

sergioifg94: That sounds good to me. Deleting it would give us more control over the clusterIP

pkg/webhooks/operator_webhooks.go
```diff
@@ -147,6 +173,65 @@ func (webhookConfig *IntegreatlyWebhookConfig) Reconcile(ctx context.Context, cl
 	return nil
 }
 
+// ReconcileService creates or updates the service that points to the Pod
+func (webhookConfig *IntegreatlyWebhookConfig) ReconcileService(ctx context.Context, client k8sclient.Client) error {
+	service := &corev1.Service{
```
JameelB, line 178: @sergioifg94 should this have an owner ref to rhmi cr? I notice on previous installations that this doesn't get cleaned up when rhmi is uninstalled.

matskiv: +1, nice catch 👍

sergioifg94: I didn't know about this owner ref feature, I'll give it a look. Thanks!

sergioifg94:
Just updated the PR with:
* [The deletion of the service if the clusterIP is not "None"](https://github.com/integr8ly/integreatly-operator/pull/756/files#diff-c69a764ac2f1bd658edfeb6149aa8825R193-R199)
* [Setting the ownerReference to the RHMI instance](https://github.com/integr8ly/integreatly-operator/pull/756/files#diff-c69a764ac2f1bd658edfeb6149aa8825R211-R219)

matskiv:
@sergioifg94 I am curious, why didn't you keep the implementation with `controllerutil.CreateOrUpdate` ?

btw: there is handy util for owner: 
https://github.com/integr8ly/integreatly-operator/blob/2d867e5e23a10ae6c6b87ddf5f38c67c52d1e907/pkg/controller/installation/bootstrapReconciler.go#L242

sergioifg94:
@matskiv In order to introduce the possibility of deletion if the clusterIP isn't "None". I don't think that would've worked with `CreateOrUpdate`

Thanks for that util. I did a quick search in the project for OwnerReference but didn't find anything. I'm updating the PR to use this and also to add the ownerRef to the ValidatingWebhookConfiguration resources so they're deleted too when the RHMI is deleted

pkg/webhooks/reconciller.go
```diff
@@ -71,6 +72,7 @@ func (reconciler *MutatingWebhookReconciler) Reconcile(ctx context.Context, clie
 	}
 
 	_, err := controllerutil.CreateOrUpdate(ctx, client, cr, func() error {
+		ownerutil.EnsureOwner(cr, owner)
```
matskiv, line 75: I suppose RHMI CR would act as owner here.
We should not set that because `MutatingWebhookConfiguration` [is not namespaced](https://github.com/kubernetes/api/blob/dccc90724807c47108487bab2c317ece185246f5/admissionregistration/v1beta1/types.go#L153-L162),
and non-namespaced(cluster scoped) object can not be owned by namespaced objects(e.g. RHMI), as per docs:
```
Note: Cross-namespace owner references are disallowed by design. This means: 
1) Namespace-scoped dependents can only specify owners in the same namespace, and owners that are cluster-scoped. 
2) Cluster-scoped dependents can only specify cluster-scoped owners, but not namespace-scoped owners.
```
From: https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

pkg/webhooks/reconciller.go
```diff
@@ -122,6 +124,7 @@ func (reconciler *ValidatingWebhookReconciler) Reconcile(ctx context.Context, cl
...
```
matskiv, line 127: Same as for MutatingWebhookConfiguration above, don't set owner reference.
Removal of cluster scoped objects should be handled via finalizers, but we can solve this in a follow up task.
Ping me in GChat if you need more details. :)

matskiv:
Thanks for the changes.

/lgtm

trdoyle81:
/lgtm
Verified on OSD 

1. Upgrade was installed successfully from 2.2.0 -> 2.2.1
2. Service for RHMI webhooks was updated with the correct owner reference.
3. Validating webhooks are still working correctly.
Screenshot 2020-05-22 at 12 36 16
Screenshot 2020-05-22 at 12 37 12
Screenshot 2020-05-22 at 12 38 18
Screenshot 2020-05-22 at 12 38 55
Screenshot 2020-05-22 at 12 39 45

PR #759, opened 2020-05-20
add Skip to EO1, verify grafana route

HubertStefanski:
Adding skip to verify grafana route test, to avoid test flakes


matskiv:
@HubertStefanski have you checked why e2e failed?

btw: you can type `/test e2e` to retest only that particular test, instead of all of them

matskiv:
It's a new, legitimate, issue - https://issues.redhat.com/browse/INTLY-7878
Please do not retest without investigating the cause of the failure. :)

matskiv:
I checked the test results, there is just one failed test, which is unrelated to this change.
I think it's worth getting this fix in to prevent potentially slowing down merges of other PR's, I'll override e2e.

/override ci/prow/e2e

PR #760, opened 2020-05-20
Updating to RHSSO 7.4 and keycloak operator to 10.0.0

davidffrench:
# Description
https://issues.redhat.com/browse/INTLY-7020

Update Keycloak operator to 10.0.0 and RHSSO to 7.4. 

**Verification:**
Login to products as test-user and customer-admin to products.
- **Install Cluster:** https://console-openshift-console.apps.dffrench-instal.n7j5.s1.devshift.org/
- **Upgrade Cluster:** https://console-openshift-console.apps.dffrench-upgrad.o0u6.s1.devshift.org/

Please send me a message and I will provide password details.

## Type of change

## Checklist


davidffrench:
@matskiv @pb82 You are the RHSSO experts. Would you mind taking a look at verifying on cluster when you have time.

version/version.go
```diff
@@ -1,6 +1,6 @@
 package version
 
 var (
-	Version            = "2.2.0"
-	IntegreatlyVersion = "2.2.0"
+	Version            = "2.2.1"
+	IntegreatlyVersion = "2.2.1"
```
matskiv, line 5: This is usually done by Delorean.
I am guessing that you didn't mean to commit this change.

matskiv:
Looks okay. Just one comment there^.
I should have time to verify this tomorrow.

version/version.go
davidffrench: I did not :-)

matskiv:
I checked Installplans, CRDs, SSO image in SS, login into 3Scale via new plugin - all good.
I noticed few issues in these places:
1) all Keycloak CR, on both clusters have:
```
status:
  phase: initialising
  ready: false
```
No errors about this in operator logs.

2) On install cluster, in RHMI CR:
```
status:
  lastError: >-
    product installation errors : failed installation of rhssouser: failed to
    reconcile dedicated-admins members: not found
```
Haven't looked into this too much.
RHMI operator logs: https://gist.github.com/matskiv/b871eef6cb1d3960aafb7e29fdedf2a3

3) On install cluster, "generated-customer-admin01..." KeycloakUser CR in redhat-rhmi-user-sso ns:
```
status:
  message: 'failed to create user: (500) 500 Internal Server Error'
  phase: failing
```
This seems to be related to the group we create here: https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/rhssouser/reconciler.go#L1200
I logged in as admin and check groups - all created, same way as on the upgrade cluster(see screenshot below). But there was no user "customer-admin01".
Possibly same issue as below?


4) On upgrade cluster, "generated-customer-admin01..." KeycloakUser CR in redhat-rhmi-user-sso ns:
```
status:
  message: 'failed to create user: (409) 409 Conflict'
  phase: failing
```
Nothing much in operator logs: `FAILED create user customer-admin01`
I checked the console, user is actually created, with correct federated identity.
There is no `spec.user.id` in KeycloakUser CR, so I thought this might be related to this change: https://github.com/keycloak/keycloak-operator/pull/150 , but likely this isn't the main issue here, but can be causing problems as well. 


but seeing this in SSO pod logs:
```[0m[31m14:49:22,310 ERROR [org.keycloak.services.error.KeycloakErrorHandler] (default task-3) Uncaught server error: java.lang.RuntimeException: Unable to find group specified by path: realm-managers```
A "realm-managers" group actually exists, but it's nested under "dedicated-admin":
Screenshot 2020-05-21 at 12 51 04
I think we have been adding users to this group by API calls(implemented by Sergio btw) because .spec.user.groups field was ignored. This was fixed in the new SSO release, see https://issues.redhat.com/browse/KEYCLOAK-13099
See this also: https://github.com/integr8ly/integreatly-operator/blob/34b666c44b73209d20527462f582c7aa56ee5f3f/pkg/products/rhssouser/reconciler.go#L1281-L1288

matskiv:
Verified on the provided clusters.
/lgtm

davidffrench:
@matskiv I have amended the previous commit with the official Keycloak CSV from https://github.com/keycloak/keycloak-operator/pull/201.

After my diff, the CRD names are different but the content is the same. The only real changes to the CSV are a removal of a role permission duplication for poddisruptionbudgets and changing support for allNamespaces to false

PR #761, opened 2020-05-20
[WIP] INTLY-6738 Grafana route 403 response testing PR

HubertStefanski:
# Description
Testing new `wait.pollimmediate()` logic to try and circumvent 403 response code, the theory is that the route returns 403 even after the initial expected 403 response, new test approach is to wait 5 minutes and test every 25 seconds to see if the response has changed and returns status 200

## Type of change

### DIFF
--- a/test/common/grafana_routes.go
+++ b/test/common/grafana_routes.go
@@ -4,9 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"net/http"
-	"net/http/httputil"
 	"testing"
+	"time"
 
 	"github.com/integr8ly/integreatly-operator/test/resources"
 	v1 "github.com/openshift/api/route/v1"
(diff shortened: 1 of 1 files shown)


HubertStefanski:
Unrelated failure to what is being tested, counting this as a successful run for e2e E01-verify-grafana-route-is-accessible 

/retest e2e

HubertStefanski:
Test failed on unrelated flake, retesting, counting this as a successful pass

/test e2e

HubertStefanski:
@mikenairn @jhellar @matskiv, E01 hasn't failed in any of the runs on this PR, I think it might've pushed down the flake down to a reasonable level. Any thoughts on these changes? are you guys happy to merge?

matskiv:
@HubertStefanski If the 403 problem happened on your PR, I would expect to see this in the logs:
`Expecting statusCode 200, got: 403, retrying`

I didn't see that^ on in any `artifacts/e2e/container-logs/test.log` file from the jobs run on the latest "74f9d76" commit - https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/pr-history/?org=integr8ly&repo=integreatly-operator&pr=761

I might be reading this wrong of course. But to me it seems like the problem wasn't reproduced, and the retry you are introducing didn't get executed. Wdyt?

matskiv:
As for the fix approach:
I don't think this would effective because we are retrying with the same http client. I suspect that the problem isn't with grafana, but with the authentication cookies set on this client.
IMHO only retry of the whole authentication process could be effective here.

And another point I'd like to raise - tests already take long time, and in case of a problem this potentially adds another 5 minutes. We should be careful about increasing test execution time because it really adds up, and we can have problems if there are many PRs that need to be merged(like before code freeze).

HubertStefanski:
@matskiv 
Yup I've noticed that, At this point I'm starting to think whether this flake is enough of an issue to warrant any more work, I think we're reaching a point where we're not encountering it all that much, and most of it seems to be periodical anyways but Ideally I'd like to have this solved as well.

As for retrying the authentication process, Could we maybe refactor the logic to run the test N amount of times, if the test fails, retry the authentication process, and after N times if we're still failing then fail the test?, Would this be a better approach?

matskiv:
@HubertStefanski over the last 14days this issue happened 21/531 runs ([search query with original error and with skip message](https://search.apps.build01.ci.devcluster.openshift.com/?search=%28skipping+due+to+known+flaky+behaviour+https%3A%2F%2Fissues.redhat.com%2Fbrowse%2FINTLY-6738%7Cgrafana_routes.go%3A60%3A+unexpected+status+code+on+success+request%2C+got%3D%26%5C%7BStatus%3A403+Forbidden+StatusCode%3A403%29&maxAge=336h&context=5&type=build-log&name=pull-ci-integr8ly-integreatly-operator-master-e2e&maxMatches=5&maxBytes=20971520&groupBy=job)). Doesn't sound like that much, but we also had a number of issues that caused failures that prevented installation, so the number of all runs is inflated. So I am not very keen on just ignoring this.
If we don't know what's causing this, we should look at data we have or gather more. E.g. we need to look at pods logs of grafana, oauth proxy, openshift oauth server, etc. If you haven't done this already, you can open one job from the search query I linked above, and look at the pod logs in `artifacts/e2e/pods/`.

As for retrying, do you mean for tests, for some of them or just for this one?
I would be against doing this for all tests because this could significantly increase the amount of time tests take to execute(if there is an issue).

matskiv:
@HubertStefanski I noticed when this issue happens, there is always OAuth proxy page in that 403 response, with this button in body: 
`<button ...>Log in with OpenShift</button>`
Hope this helps get down to the bottom of it.

HubertStefanski:
Closing off this PR, taking a different approach as discussed with @matskiv

PR #677, opened 2020-05-06
INTLY-7392 RHMIConfig mutating webhook

sergioifg94:
# Description

> Link to JIRA: https://issues.redhat.com/browse/INTLY-7392

Register a mutating webhook for the RHMIConfig CRD that adds to the CR the following annotations

| Annotation | Value |
| ------------- | --------- |
| `lastEditUsername` | Name of the user that performed the update to the CR (ignoring the RHMI operator as it would override the value at reconciliation) |
| `lastEditTimestamp` | Time (in UTC) when the update is performed |

Add test to log in as a customer admin and verify that their name is stored when performing a change to the CR, as well as the timestamp.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist
- [ ] This change requires a documentation update 
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] Verified independently on a cluster by reviewer

PR #776, opened 2020-05-21
expose upgrade metrics

briangallagher:
# Description
Update RHMI CR to reflect current version and upgrading version, meaning, the version RHMI is attempting to upgrade to. Add prometheus metrics around this data.  Explained in detail in JIRA
[INTLY-7394](https://issues.redhat.com/browse/INTLY-7394)

## Why
**Track the following**
- If an upgrade is happening
- how long an upgrade took
- what version we are upgrading to

## Type of Change
- New feature (non-breaking change which adds functionality)

## Verification

**NOTE:** Verifying this JIRA is dependant on the below JIRAs. 
These PRs do not block this PR as this change is independant.
I've added work arounds below also. https://github.com/integr8ly/integreatly-operator/pull/756 https://github.com/integr8ly/integreatly-operator/pull/776 https://github.com/integr8ly/integreatly-operator/pull/677

### How to Verify (short)
- Deploy 2.2.0 of RHMI via OLM 
- Update RHMIConfig to immediately install upgrades
- Upgrade RHMI to version 2.2.1 via OLM 
- Verify CR updates and Metrics Created
- Upgrade RHMI to version 2.2.2 via OLM 
- Verify CR updates and Metrics Created

**NOTE:** The reason for upgrading twice is:
The first upgrade is from a CR without the new fields to one with the new fields (toVersion, Version). The second upgrade is from a CR with the new fields.
Different parts of the code base are used in these 2 scenarios, hence the need to test both.

### Install 2.2.0 via OLM (use master branch)
* Change makefile `ORG` to your own quay.io registry
  * https://github.com/integr8ly/integreatly-operator/blob/8eee7ace6c571569d62eddca93cc8d18b2daced6/Makefile#L3
* Build and push RHMI image to your own registry
  * `make image/build/push`
* Replace image in csv with built image
  * https://github.com/integr8ly/integreatly-operator/blob/8eee7ace6c571569d62eddca93cc8d18b2daced6/deploy/olm-catalog/integreatly-operator/integreatly-operator-2.0.0/integreatly-operator.v2.0.0.clusterserviceversion.yaml#L44
  * https://github.com/integr8ly/integreatly-operator/blob/8eee7ace6c571569d62eddca93cc8d18b2daced6/deploy/olm-catalog/integreatly-operator/integreatly-operator-2.0.0/integreatly-operator.v2.0.0.clusterserviceversion.yaml#L273
* Push csv to your registry
  * `make push/csv  REPO=<your_registry> QUAY_USERNAME=<your_registry> QUAY_PASSWORD=<password>`
* After pushing CSV make sure to make the Application Repo in Quay public
* Install operator source with your registry
```
kind: OperatorSource
metadata:
  name: RHMI
spec:
  authorizationToken: {}
  displayName: RHMI Operators
  registryNamespace: ${REPO}
  type: appregistry
```
* Create operator namespace and setup secrets
  * `make cluster/prepare/project && make cluster/prepare/smtp && make cluster/prepare/dms && make cluster/prepare/pagerduty`
* Install RHMI from operator hub to the `redhat-rhmi-operator` namespace

### Upgrade to 2.2.1 via OLM (use INTLY-7394-expose-upgrade-metrics branch)

**NOTE:** Before starting check if workarounds below required.

Switch branch to INTLY-7394-expose-upgrade-metrics branch
Change the 
* Update versions in version.go to 2.2.1
* Build and push RHMI image to your own registry
  * `make image/build/push TAG=2.2.1 ORG=<yourQuayOrg>`
* Copy and Paste ./deploy/olm-catalog/integreatly-operator/2.2.0 and rename to 2.2.1
  * Rename csv file to 2.2.1
  * Change the following fields in the CSV file:
    * verison to 2.2.1
    * replaces: integreatly-operator.v2.2.0
    * Change 2 quay image references same as above
    * name: integreatly-operator.v2.2.1
* Update integreatly-operator.package.yaml to reference 2.2.1
* Push csv to your registry
  * make push/csv TAG=2.2.1 REPO=<REPO> QUAY_USERNAME=<user> QUAY_PASSWORD=<password>
* After pushing CSV make sure to make the Application Repo in Quay public
* Update the RHMIConfig CR spec.upgrade.alwaysImmediately = true
* Delete the Status Block from the OperatorSource created above to force the operator to check Quay for updates.
* Start watching the RHMI CR at this point
* Check the RHMI operator, after a period of time the new version should be available and start automatically upgrading due to the RHMIConfig CR setting (alwaysImmediately)
* When the upgrade starts the CR should add a new field Status.toVersion: 2.2.1
* After a successful reconcile the CR should have Status.Version: 2.2.1. toVersion will be removed. 
* Verify Prometheus using the API and a command similar to below. Tweak start and end query parameters as appropriate. 
```/api/v1/query_range?query=rhmi_version&start=2020-05-20T05:10:30.781Z&end=2020-05-21T20:11:00.781Z&step=15s```
There should be at least 2 entries in prometheus, one with toVersion: 2.2.1 and the other with Version: 2.2.1 

### Upgrade to 2.2.2 via OLM (use INTLY-7394-expose-upgrade-metrics branch)
* Repeat the previous steps changing values from 2.2.1 to 2.2.2 where appropriate.
* Verify the CR. In this case you should initially see:
  * Status.version: 2.2.1 and Status.toVersion: 2.2.2, then:
  * Status.version: 2.2.0 
* There should be another 2 entries in Prometheus reflecting the CR values.

### PR Workarounds
For 756, **after every install of RHMI delete** the **_rhmi-webhooks_** service in the redhat-rhmi-operator ns. Also, delete **_ValidatingWebhookConfiguration_** and **_MutatingWebhookConfiguration_**
For 776, run: 
`oc policy add-role-to-user admin system:serviceaccount:redhat-rhmi-operator:rhmi-operator
`

### DIFF
--- a/pkg/controller/installation/installation_controller.go
+++ b/pkg/controller/installation/installation_controller.go
@@ -3,11 +3,11 @@ package installation
 import (
 	"context"
 	"fmt"
+	"github.com/integr8ly/integreatly-operator/pkg/webhooks"
+	"github.com/integr8ly/integreatly-operator/version"
 	"os"
 	"strings"
 	"time"
-
-	"github.com/integr8ly/integreatly-operator/pkg/webhooks"
 
 	"k8s.io/apimachinery/pkg/types"
 
@@ -305,6 +305,20 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		return retryRequeue, nil
 	}
 
+	// If no current or target version is set this is the first installation of rhmi.
+	if installation.Status.Version == "" && installation.Status.ToVersion == "" {
+		installation.Status.ToVersion = version.IntegreatlyVersion
+		logrus.Infof("Setting installation.Status.ToVersion on initial install %s", version.IntegreatlyVersion)
+		if err := r.client.Status().Update(r.context, installation); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	// It's important to set the metric values at this point to account for the upgrade scenario. The ToVersion
+	// is set on the CR when the install plan is approved, however, the operator pod is terminated shortly
+	// after this point which may not be enough time for prometheus to scrape the metric.
+	metrics.SetRhmiVersions(string(installation.Status.Stage), installation.Status.Version, installation.Status.ToVersion)
+
 	// Reconcile the webhooks
 	if err := webhooks.Config.Reconcile(r.context, r.client); err != nil {
 		return reconcile.Result{}, err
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by briangallagher** [comment]
e2e test failed for the following reason:
https://issues.redhat.com/browse/INTLY-7932

**Comment by Boomatang** [comment]
I have not reviewed the source yet. But I have seen the `version` and `toVersion` change in the CR as expected. What I have not seen is the metric been in Prometheus. I was trying the find this metric in the rhmi Prometheus instances, was this where I should have seen it or is it in the cluster Prometheus instance I should have looked?

File: pkg/controller/subscription/subscription_controller.go [hunk 1]
```diff
@@ -136,7 +136,14 @@ func (r *ReconcileSubscription) HandleUpgrades(ctx context.Context, rhmiSubscrip
 
 	if !rhmiConfigs.IsUpgradeServiceAffecting(latestRHMICSV) || canUpgradeNow {
 		eventRecorder := r.mgr.GetEventRecorderFor("RHMI Upgrade")
-		err = rhmiConfigs.ApproveUpgrade(ctx, r.client, latestRHMIInstallPlan, eventRecorder)
+
+		// TODO: investigate a better approach to getting RHMI rather than hardcoding values
+		installation := &integreatlyv1alpha1.RHMI{}
+		err = r.client.Get(ctx, k8sclient.ObjectKey{Name: "rhmi", Namespace: "redhat-rhmi-operator"}, installation)
```
Boomatang, line 142: Is it possible for a user to install the operator in a different namespace? If it is possible then we would never get to do an upgrade. 

On a side note if it is possible for the user to install the operator in a different namespace, what namespaces would the products get installed in? I am guessing that would stay the same.

pkg/controller/installation/installation_controller.go
```diff
@@ -377,12 +391,265 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+		verifyCodeReadyWorkspacesVersions(stages) &&
+		verifyFuseOnOpenshiftVersions(stages) &&
+		verifyMonitoringVersions(stages) &&
+		verify3ScaleVersions(stages) &&
+		verifyFuseOnlineVersions(stages) &&
+		verifyDataSyncVersions(stages) &&
+		verifyRHSSOUserVersions(stages) &&
+		verifyUPSVersions(stages) &&
+		verifySolutionExplorerVersions(stages))
+}
```
Boomatang, line 435: Just a thought, would we be better getting out of this check as soon as possible. If any product verify function returns a  false  the whole function output will be a false. 
```
if !verifySolutionExplorerVersions(stages){
    return false
}
if !verifyUPSVersions(stages) {
    return false
}
return true
```

pkg/controller/installation/installation_controller.go
```diff
@@ -377,12 +391,265 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+
+func verifySolutionExplorerVersions(stages map[integreatlyv1alpha1.StageName]integreatlyv1alpha1.RHMIStageStatus) bool {
```
Boomatang, line 437: This comment applies to all the product verify functions. If the product uses floating tags and there is a patch release, does this function still work as expected? I feel this will still compare the old versions and not see the newer version.

damienomurchu:
@briangallagher can you add a testcase under https://github.com/integr8ly/integreatly-operator/tree/master/test-cases/tests to reflect the automated testing of these changes and any manual testing required if the automated testing doesn't cover everything.

matskiv:
This is handled in #798

pkg/controller/subscription/subscription_controller.go
jjaferson: Good question, I don't think the operator can be installed in another namespace, but would be good to find a way of not having to it hardcoded

https://github.com/integr8ly/integreatly-operator/blob/master/pkg/controller/subscription/subscription_controller.go#L38

pkg/controller/installation/installation_controller.go
jjaferson: this is called Short circuit evaluation, it will get out as sooner one of the function return false

 https://kuree.gitbooks.io/the-go-programming-language-report/content/22/text.html

Boomatang: That's good to hear, and thanks for the link too. It looks like there maybe a few other answers in there to questions I have.

PR #784, opened 2020-05-22
Expose metrics for the rhmi installation CR

jjaferson:
# Description
Exposes two metrics for the RHMI installation CR

rhmi_status and rhmi_info 

https://issues.redhat.com/browse/INTLY-6667

## Verification steps

1) Install operator via OLM
2) Go to the rhmi-operator pod
3) In the terminal tab curl the metric endpoint to verify if metrics are present 
- `curl 0.0.0.0:8383/metrics | grep rhmi_status{` 
-  `curl 0.0.0.0:8383/metrics | grep rhmi_info{`

4) Check if metrics have all the labels created and if their values match with installation CR
```
rhmi_info{installation_type="", master_url="", namespace="", operator_name="", use_cluster_storage=""}
rhmi_status{namespace="", last_error="", operator_name="", preflight_message="", preflight_status="", stage="", <product_name>-status="" ...}
```
5) Go to redhat-rhmi-middleware-monitoring-operator namespace 
6) Access Prometheus and verify if rhmi_status and rhmi_info are available  

## Type of change

## Checklist

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -1,9 +1,14 @@
 package metrics
 
 import (
+	"fmt"
+	"strings"
+
+	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
 	"github.com/integr8ly/integreatly-operator/version"
 
 	"github.com/prometheus/client_golang/prometheus"
+	customMetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
 // Custom metrics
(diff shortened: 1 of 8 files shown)


deploy/olm-catalog/integreatly-operator/2.2.0/integreatly-operator.v2.2.0.clusterserviceversion.yaml
```diff
@@ -365,8 +365,10 @@ spec:
           - apps
           resources:
           - deployments/finalizers
+          - replicasets
```
davidffrench, line 368: This and the verb addition below will need to be removed, the existing CSVs shouldn't be modified.

pkg/metrics/metrics.go
```diff
@@ -25,4 +30,99 @@ var (
...
+		"last_error",
+		"preflight_message",
+		"preflight_status",
+		"stage",
+	}
 )
+
+// SetRHMIInfo exposes rhmi info metrics with labels from the installation CR
+func SetRHMIInfo(installation *integreatlyv1alpha1.RHMI) {
+	RHMIInfo.WithLabelValues(installation.Spec.UseClusterStorage,
```
davidffrench, line 62: There are additional fields included in the CR that might be useful here as well (namespacePrefix, operatorsInProductNamespace, routingSubdomain, selfSignedCerts). Unless the decision was made to only include these 5.

jjaferson:
> Looks fantastic @jjaferson . Reviewed code and verified on cluster. One requested change on the addition to the CSV. this should be changed in `deploy/role.yaml`.
> 
> ## rhmi_status - Looks Good
> image
> 
> ## rhmi_status_available - Looks Good
> image
> 
> ## rhmi_info - Question
> image
> 
> One question on the rhmi_info metric. You can see from the screenshot there are two metrics being exposed for this. Do you know why this is?

I noticed that but I think it's something that happens by default as I saw the same metric for other CRDs like `enmasse_addressplan_info` and `blackboxtarget_info` with the same labels 

*enmasse_addressplan_info*
Screenshot from 2020-05-25 14-59-16

*blackboxtarget_info*
Screenshot from 2020-05-25 15-00-16

deploy/olm-catalog/integreatly-operator/2.2.0/integreatly-operator.v2.2.0.clusterserviceversion.yaml
jjaferson: Sure, they already added in role.yaml and I removed them from the csv.

pkg/metrics/metrics.go
jjaferson: David Martin suggested these fields but I don't see any problem in adding more, what do you think @david-martin ?

david-martin:
> I noticed that but I think it's something that happens by default as I saw the same metric for other CRDs like `enmasse_addressplan_info` and `blackboxtarget_info` with the same labels


It looks like the metric is being scraped from 2 different places/endpoints. `cr-metrics` and `http-metrics`.
Is there a duplicate metric being exposed?

pkg/metrics/metrics.go
david-martin: The storage overhead tends to be with the values for each scrape time (lots of scrapes) rather than the labels, 
unless the labels change very frequently.
It should be safe, from a storage point of view, to include more labels as long as they don't change very often.

davidffrench: Perfect, thanks @david-martin

jjaferson:
> > I noticed that but I think it's something that happens by default as I saw the same metric for other CRDs like `enmasse_addressplan_info` and `blackboxtarget_info` with the same labels
> 
> It looks like the metric is being scraped from 2 different places/endpoints. `cr-metrics` and `http-metrics`.
> Is there a duplicate metric being exposed?

I was doing some investigation and I found out that there is a gauge metric being exposed on port 8686 of the operator

```
curl 0.0.0.0:8686/metrics
# HELP rhmi_info Information about the RHMI custom resource.
# TYPE rhmi_info gauge
rhmi_info{namespace="redhat-rhmi-operator",rhmi="rhmi"} 1
```

https://github.com/integr8ly/integreatly-operator/blob/master/cmd/manager/main.go#L148

Not sure where the metric is created

jjaferson:
> > > I noticed that but I think it's something that happens by default as I saw the same metric for other CRDs like `enmasse_addressplan_info` and `blackboxtarget_info` with the same labels
> > 
> > 
> > It looks like the metric is being scraped from 2 different places/endpoints. `cr-metrics` and `http-metrics`.
> > Is there a duplicate metric being exposed?
> 
> I was doing some investigation and I found out that there is a gauge metric being exposed on port 8686 of the operator
> 
> ```
> curl 0.0.0.0:8686/metrics
> # HELP rhmi_info Information about the RHMI custom resource.
> # TYPE rhmi_info gauge
> rhmi_info{namespace="redhat-rhmi-operator",rhmi="rhmi"} 1
> ```
> 
> https://github.com/integr8ly/integreatly-operator/blob/master/cmd/manager/main.go#L148
> 
> Not sure where the metric is created

Metric is created by default by the operator-sdk 

Custom resource specific metrics

By default operator will expose info metrics based on the number of the current instances of an operator’s custom resources in the cluster. It leverages kube-state-metrics as a library to generate those metrics. Metrics initialization lives in the cmd/manager/main.go file of the operator in the serveCRMetrics function. Its arguments are a custom resource’s group, version, and kind to generate the metrics. The metrics are served on 0.0.0.0:8686/metrics by default.

https://sdk.operatorframework.io/docs/golang/monitoring/prometheus/

david-martin:
> Not sure where the metric is created

Looks like a CR metric exposed by the operator sdk for any/all CRD instances that exist.

I think we need to change the name of the metric that we're in control of (not the sdk generated one)

jjaferson:
> > Not sure where the metric is created
> 
> Looks like a CR metric exposed by the operator sdk for any/all CRD instances that exist.
> 
> I think we need to change the name of the metric that we're in control of (not the sdk generated one)

Yup, what about `rhmi_installation_info` ?

pkg/metrics/metrics.go
jjaferson: Cool, I don't think these values should ever change once the installation is completed, I'll add them to the metric.

jjaferson:
/test e2e

failed to create e2e pod

> "failed to fetch Cluster: failed to fetch dependency of \"Cluster\": failed to generate asset \"Platform Permissions Check\": validate AWS credentials: mint credentials check: error simulating policy: Throttling: Rate exceeded\n\tstatus code: 400, request id: 557a3694-d525-4c45-a76b-e95ce8ed683e"

davidffrench:
/lgtm
/approve

@jjaferson Great work. Please make sure to update the JIRA with the name change from rhmi_info to rhmi_spec

david-martin:
Unable to tag @openshift/openshift-team-monitoring,
so leaving a comment that these are the code changes that expose `rhmi_status` metric. 
See #795 for cluster-monitoring-operator PR to whitelist this metric.

PR #804, opened 2020-05-27
Test case for wt3 solution-pattern updates

tonyxrmdavidson:
# Description
Test-case for wt3 solution-pattern update

### DIFF
--- a/test-cases/tests/walkthroughs/g13-verify-wt3-solution-pattern-updates.md
+++ b/test-cases/tests/walkthroughs/g13-verify-wt3-solution-pattern-updates.md
@@ -1,0 +1,20 @@
+---
+estimate: 15m
+tags:
+---
+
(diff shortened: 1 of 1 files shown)


pawelpaszki:
It looks good, but `npm run lint` fails. Can you please fix it?

test-cases/package.json
```diff
@@ -17,7 +17,7 @@
     "form-data": "3.0.0",
     "googleapis": "47.0.0",
     "gray-matter": "4.0.2",
-    "handlebars": "4.1.2",
+    "handlebars": "^4.7.6",
```
damienomurchu, line 20: @tonyxrmdavidson is this required?

pawelpaszki:
we need to merge one more change before this can be merged. I will approve it after that
/hold

PR #819, opened 2020-05-28
Add Delorean CLI and skopeo to test base image

mikenairn:
# Description
Add Delorean CLI and skopeo to base test image
Add test-containers checks to test the format of the test-containers.yaml file (test-containers/check/format) and that all images exist (test-containers/check/images).
All test-containers checks can be executed with test-containers/check

```
make test-containers/check
```

Jiras:

* https://issues.redhat.com/browse/DEL-289
* https://issues.redhat.com/browse/DEL-287

## Type of change

## Checklist

### DIFF
--- a/make/test-containers.mk
+++ b/make/test-containers.mk
@@ -1,0 +1,11 @@
+
+.PHONY: test-containers/check/format
+test-containers/check/format:
+	yq v test-containers.yaml
+
+.PHONY: test-containers/check/images
+test-containers/check/images:
+	yq r test-containers.yaml 'tests[*].image' | xargs -i skopeo inspect --config docker://'{}'
+
+.PHONY: test-containers/check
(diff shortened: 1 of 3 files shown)


make/test-containers.mk
```diff
@@ -0,0 +1,11 @@
+
+.PHONY: test-containers/check/format
+test-containers/check/format:
+	yq v test-containers.yaml
```
b1zzu, line 4: Should we document somewhere how to fix the format of the test-containers.yaml if this job fails?

mikenairn: :man_shrugging: It's just checking it's valid yaml tbh, if it fails make it valid yaml

b1zzu: I thought it was a linter

PR #825, opened 2020-05-29
removed duplicated test case - cluster prometheus

jhellar:
# Description

I believe that this test case is duplicated with [A20](https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/installation/a20-verify-cluster-prometheus-has-rhmi-targets.md)

## Checklist


jhellar:
@matskiv Hi, can you please lgtm again, needed to run prettier for the test to pass.

PR #829, opened 2020-05-29
fix backup maintenance window overlap check

aidenkeating:
currently the backup and maintenance windows are allowed to end
and start on the same minute. this is not permitted by the cloud
resource operator and validation on the rhmi operator side should
fail.

verification:
- run the operator on-cluster to ensure webhooks work as expected
- update rhmiconfig '.spec.backup.applyOn' to '01:00' and
  '.spec.maintenance.applyFrom' to '02:00' and ensure an error is
  returned by openshift through the validation webhook.
- update rhmiconfig '.spec.backup.applyOn' to '01:00' and
  '.spec.maintenance.applyFrom' to '02:01' and ensure no error is
  returned by openshift through the validation webhook
- update rhmiconfig '.spec.backup.applyOn' to '03:00' and
  '.spec.maintenance.applyFrom' to '02:00' and ensure an error is
  returned by openshift through the validation webhook.
- update rhmiconfig '.spec.backup.applyOn' to '03:01' and
  '.spec.maintenance.applyFrom' to '02:00' and ensure no error is
  returned by openshift through the validation webhook

### DIFF
--- a/pkg/apis/integreatly/v1alpha1/rhmiconfig_types_test.go
+++ b/pkg/apis/integreatly/v1alpha1/rhmiconfig_types_test.go
@@ -1,0 +1,67 @@
+package v1alpha1
+
+import "testing"
+
+func TestValidateBackupAndMaintenance(t *testing.T) {
+	type args struct {
+		backupApplyOn        string
+		maintenanceApplyFrom string
+	}
+	tests := []struct {
+		name            string
+		args            args
+		wantBackup      string
+		wantMaintenance string
+		wantErr         bool
+	}{
+		{
+			name: "test exact overlapping window fails",
+			args: args{
+				backupApplyOn:        "01:00",
+				maintenanceApplyFrom: "mon 02:00",
+			},
+			wantErr: true,
+		},
+		{
+			name: "test overlapping window fails",
+			args: args{
+				backupApplyOn:        "01:30",
+				maintenanceApplyFrom: "mon 01:00",
+			},
+			wantErr: true,
+		},
+		{
+			name: "test non-overlapping window succeeds with backup before maintenance",
+			args: args{
+				backupApplyOn:        "01:00",
+				maintenanceApplyFrom: "mon 02:01",
+			},
+			wantBackup:      "01:00",
+			wantMaintenance: "mon 02:01",
+		},
+		{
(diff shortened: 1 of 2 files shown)


darahayes:
I'm still super new to the codebase and I haven't run the unit tests locally but I think it should be pretty simple and worthwhile to add a unit test case for the values provided in the verification steps here: https://github.com/integr8ly/integreatly-operator/blob/master/test/common/rhmi_config.go#L35-L110

darahayes:
Verified that the changes overlap checks work 
Screenshot 2020-05-29 at 16 57 32

pkg/apis/integreatly/v1alpha1/rhmiconfig_types_test.go
```diff
@@ -0,0 +1,67 @@
+package v1alpha1
+
+import "testing"
+
+func TestValidateBackupAndMaintenance(t *testing.T) {
```
darahayes, line 5: Just a heads up: there is already an `rhmiconfig_types` test file under `test/common` here: https://github.com/integr8ly/integreatly-operator/blob/master/test/common/rhmi_config.go#L35-L110

It might not be necessary to add a new one here.

darahayes: In fact there are already some cases for overlap checks so you might be able to just add a few extra there. 

https://github.com/integr8ly/integreatly-operator/blob/master/test/common/rhmi_config.go#L54-L72

aidenkeating: @darahayes Those tests, any any in `test/common`, are run as part of the automated e2e test suite and are designed to be run against a real OpenShift cluster.

I think having additional unit tests that can be run locally without an OpenShift cluster can be useful for development

aidenkeating:
@matskiv FYI

resolves https://issues.redhat.com/browse/INTLY-8141

pkg/apis/integreatly/v1alpha1/rhmiconfig_types_test.go
darahayes: Good to know!

PR #831, opened 2020-05-29
removed a19 test case

pawelpaszki:
# Description
removed a19 test case


pawelpaszki:
@aidenkeating - am I right in saying that this test case is obsolete?

aidenkeating:
@pawelpaszki Is the reason this is obsolete because we're now using `/23` by default in the test cluster pipeline? If so, and we're testing against /23 clusters, then yes this is obsolete.

pawelpaszki:
@aidenkeating - that's correct

tonyxrmdavidson:
lgtm

PR #838, opened 2020-06-02
Update webapp 2.25.0

pb82:
update solution explorer operator to 0.0.56 to include https://github.com/integr8ly/tutorial-web-app-operator/pull/98

update webapp to 2.25.0 to fix some issues with the navigation

### DIFF
--- a/manifests/integreatly-solution-explorer/0.0.56/solution-explorer-operator.v0.0.56.clusterserviceversion.yaml
+++ b/manifests/integreatly-solution-explorer/0.0.56/solution-explorer-operator.v0.0.56.clusterserviceversion.yaml
@@ -1,0 +1,145 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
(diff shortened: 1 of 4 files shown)


matskiv:
@pb82 would it make sense to skip this version, as we still need to get this change in:
https://github.com/integr8ly/tutorial-web-app/pull/589 ?

pb82:
@matskiv Yes, just wanted to say. I'll get Kevin's changes verified and then update this PR.

matskiv:
@pb82 Okay
Also, 2.2.1 is using 0.0.53, so we could replace 0.0.55 with the new version (0.0.57) to avoid double upgrade. 
I don't know how we done this in the past releases though.

pb82:
@matskiv I'd skip 0.0.56 and upgrade from 0.0.53 to 0.0.57.

matskiv:
@pb82 Just to clarify, I was suggesting that we can skip 0.0.55 also, because it wasn't part of GA release.
But I am okay with leaving it in, as long as we know that integreatly-operator will handle two upgrades in a row.

PR #840, opened 2020-06-02
INTLY-7831 Fix attempt to index first element of possibly empty list

sergioifg94:
# Description

The `get3ScaleAPIcasePod` function in the `TestNetworkPolicyAccessNSToSVC` test retrieves
a list of the APICast pods in the 3scale namespace, and returns the first element on the
list. This has caused some tests to fail, possibly because the test is executed before
any pod has been created.

Modify the `get3ScaleAPIcasePod` function to wait for the pod list to have at least one
element, or fail with timeout.

## Type of change

## Checklist


sergioifg94:
@matskiv Just pushed those changes. Removed the polling so the test will directly fail if the pod list is empty

matskiv:
Thank you!
/lgtm
/approve

PR #844, opened 2020-06-03
do not modify unrelated api manager properties for test

pb82:
Only modify properties of the APIManager CR that are relevant for the test. Specifically, do not modify the value of the high availability flag because that would cause extra deployments in the 3Scale namespace.

### DIFF
--- a/test/common/scale_up_replicas_threescale.go
+++ b/test/common/scale_up_replicas_threescale.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	threescalev1 "github.com/3scale/3scale-operator/pkg/apis/apps/v1alpha1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


test/common/scale_up_replicas_threescale.go
```diff
@@ -76,48 +75,24 @@ func getAPIManager(dynClient k8sclient.Client) (threescalev1.APIManager, error)
...
-			Zync: &threescalev1.ZyncSpec{
-				AppSpec: &threescalev1.ZyncAppSpec{Replicas: &[]int64{replicas}[0]},
-				QueSpec: &threescalev1.ZyncQueSpec{Replicas: &[]int64{replicas}[0]},
-			},
-		},
-	}
+
+	apim.Name = name
+	apim.Namespace = namespace
+	apim.ResourceVersion = apim.GetResourceVersion()
```
matskiv, line 85: Is the name, namespace and resourceversion added intentionally here?
I would expect those field to already be set in the CR we Get. 🤔

matskiv: This is minor, and can be left as is.
I am approving, but it will merge after the release.

pb82: I would also expect them to be present already. I can try to remove and rerun the tests.

matskiv:
Cathal implemented patch in #754 
So seems like we can close this one.
Is that okay with you @pb82

pb82:
@matskiv Yes, that's ok. Closing this one.

PR #849, opened 2020-06-03
Make maintenance window for RHMIConfig case insensitive

sergioifg94:
# Description

The days difference calculated from a date to the next maintenance window uses a
`map[string]int` with the different week days in lowercase. If the value
from the maintenance window is not in lowercase it will fail silently to
match a value from that `map`, so it'll default to 0 as the zero value
of the map, which is Sunday, resulting in the day difference until the
next Sunday.

Convert the week day to lower case before indexing the map.

## Type of change

## Checklist


maleck13:
is there a  unit test we can update or add so we don't regress in the future?

sergioifg94:
@maleck13 I just added some unit tests now

PR #853, opened 2020-06-04
INTLY-7964  e2e test flake: TestIntegreatly/integreatly/Cluster

Boomatang:
# Description

e2e test updated to quit if any part of the Cluster test fails. This causes the test run to exit early and not cause any failures later test due to the cluster install not been fully complete.

**JIRA** https://issues.redhat.com/browse/INTLY-7964

## Type of change

- [x] e2e test flake fix
 
## Checklist

### DIFF
--- a/test/e2e/integreatly_test.go
+++ b/test/e2e/integreatly_test.go
@@ -93,9 +93,19 @@ func TestIntegreatly(t *testing.T) {
 			})
 		}
 
+		var err error
 		t.Run("Cluster", func(t *testing.T) {
-			IntegreatlyCluster(t, f, ctx)
+			err = IntegreatlyCluster(t, f, ctx)
+			if err != nil {
+				t.Log(err)
+				t.Fail()
+			}
 		})
+
+		if err != nil {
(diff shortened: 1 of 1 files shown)


Boomatang:
In the Jira on of the quick fix is to increase the time out delay. This Pr does not do this for the following reason. The test that is affected in this PR uses multiply time outs at different stages. With not having a clear picture of what time outs need to be adjusted I feel its better we track the failures again to find the correct time out.

Time outs used:
- timeout                                    = time.Second * 75
- deploymentTimeout                 = time.Minute * 25
- bootStrapStageTimeout           = time.Minute * 5
- cloudResourcesStageTimeout = time.Minute * 10
- monitoringStageTimeout          = time.Minute * 10
- authenticationStageTimeout     = time.Minute * 30
- productsStageTimout                = time.Minute * 30
- solutionExplorerStageTimeout  = time.Minute * 10

CathalOConnorRH:
Looks good, I agree, once this fix is in we should monitor for timeouts and adjust 
/lgtm

PR #869, opened 2020-06-10
remove system-provider from memory check as it has a memory leak that…

maleck13:
… is not being addressed

# Description
Removes system-provider from alert check for 3scale
## Type of change

## Checklist


david-martin:
/approve
Approving to get this merged back, with upgrade fix landing in https://issues.redhat.com/browse/INTLY-6599 that will move location of the alert rules

PR #872, opened 2020-06-10
added automation-jira tag

jhellar:
# Description

Added `automation-jira-created` tag, so that it is easier to filter manual test cases for which automation jira does not exist yet. This could be improved by updating tooling to allow creating jiras for test cases that are not covered yet by jira ticket.

### DIFF
--- a/test-cases/tests/alerts/c05-verify-that-no-alerts-were-firing-during-and-after-installat.md
+++ b/test-cases/tests/alerts/c05-verify-that-no-alerts-were-firing-during-and-after-installat.md
@@ -1,5 +1,7 @@
 ---
 estimate: 15m
+automation_jiras:
+  - INTLY-7413
 ---
 
 # C05 - Verify that no alerts were firing during and after installation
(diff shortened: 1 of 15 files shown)


test-cases/tests/alerts/c03-verify-that-alerting-mechanism-works.md
```diff
@@ -2,6 +2,9 @@
 estimate: 30m
 tags:
   - happy-path
+  - automation-jira-created
```
b1zzu, line 5: @jhellar I think that if we have the `automation_jiras` we don't need also a tag, we can modify the tool to automatically add the tag if it finds an `automation_jiras`

jhellar: ok, should be fixed now

b1zzu:
@jhellar do we have some follow up tasks to create a new cmd in the tools to automatically create the tasks for automating the test cases?

b1zzu:
@jhellar I forgot about this, but we should at least but a small note in the README.md

jhellar:
@b1zzu I've created the task in https://issues.redhat.com/browse/INTLY-8430. Will send another PR for the readme.

b1zzu:
Thanks @jhellar  :)

PR #874, opened 2020-06-10
Grafana dashboards are reconciled with create/update method through code

Rajagopalan-Ranganathan:
# Description
JIRA Link: https://issues.redhat.com/browse/INTLY-6831

It fixes the following issues as well:
https://issues.redhat.com/browse/INTLY-5962
https://issues.redhat.com/browse/INTLY-7869

Grafana dashboards were defined statically in the template yaml files. Code changes to
make them reconcile using code has been made. Upgrade/update should work now
without having to remove the CR. Template files are removed.

## Type of change

## Checklist


pkg/products/monitoring/dashboardsHelper.go
```diff
@@ -0,0 +1,38 @@
+package monitoring
+
+import (
+	"fmt"
+
+	monitoring "github.com/integr8ly/integreatly-operator/pkg/products/monitoring/dashboards"
+)
+
+func getSpecDetailsForDashBoard(dbName string) (string, string, error) {
```
laurafitzgerald, line 9: minor naming thing here: I read db initially as database.

pkg/products/monitoring/reconciler.go
```diff
@@ -457,19 +459,51 @@ func (r *Reconciler) reconcileScrapeConfigs(ctx context.Context, serverClient k8
...
+
+	for _, dashBoard := range r.Config.GetDashBoards() {
+		err := r.reconcileGrafanaDashboards(ctx, serverClient, dashBoard)
 		if err != nil {
-			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to create/update monitoring template %s: %w", template, err)
+			logrus.Errorf("Error reconciling dashboards: %v", err)
+			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to create/update grafana dashboard %s: %w", dashBoard, err)
 		}
-		r.Logger.Infof("Reconciling the monitoring template %s was successful", template)
+		r.Logger.Infof("Reconciling the grafana dashboard  %s was successful", dashBoard)
```
laurafitzgerald, line 470: again another minor thing here. In some cases dashboard is `dashBoard, dashboard or db` would be good to settle on one I think.

pkg/config/monitoring.go
```diff
@@ -123,15 +123,15 @@ func (m *Monitoring) GetPrometheusStorageRequest() string {
...
-		"resources-by-pod.yaml",
-		"cluster-resources.yaml",
-		"critical_slo_alerts.yaml",
+		"endpointsdetailed",
+		"endpointsreport",
+		"endpointssummary",
+		"resources-by-namespace",
+		"resources-by-pod",
+		"cluster-resources",
+		"critical-slo-alerts",
```
laurafitzgerald, line 134: I notice you refactored the `critical-slo-alerts` syntax here. would it be worth making them all follow the same syntax? e.g. the three endpoints ones? Not sure if that would trigger some upgrade issues, e.g. having to delete ones with the older names?

pkg/products/monitoring/dashboardsHelper.go
Rajagopalan-Ranganathan: Fixed.

pkg/config/monitoring.go
Rajagopalan-Ranganathan: In the previous code, just the file name had underscores. The name inside the yaml definition was "critical-slo-alerts". So i renamed the method to return the dashboard names. Yes, once the name is changed the upgrade doesnt work or in the sense, there might be a stale old dashboard there. Also, the name displayed in the grafana dashboard is not the same as the one we define here. They have pretty names, defined in the "Title" section inside the spec. I would not change the CR names for now.

pkg/products/monitoring/reconciler.go
Rajagopalan-Ranganathan: Agreed. Will settle for "dashboard". fixed.

pkg/config/monitoring.go
laurafitzgerald: @Rajagopalan-Ranganathan yes I was thinking that could be the case. It's a minor consistency thing so not necessary to fix.

pkg/products/monitoring/reconciler.go
```diff
@@ -457,19 +459,51 @@ func (r *Reconciler) reconcileScrapeConfigs(ctx context.Context, serverClient k8
...
 }
 
+func (r *Reconciler) reconcileGrafanaDashboards(ctx context.Context, serverClient k8sclient.Client, dashBoard string) (err error) {
+
+	grafanaDB := &grafanav1alpha1.GrafanaDashboard{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      dashBoard,
+			Namespace: r.Config.GetOperatorNamespace(),
+			Labels: map[string]string{
+				"monitoring-key": r.Config.GetLabelSelector(),
```
grdryn, line 482: I think the setting of this label should be moved to the mutate function that's passed to the `controllerutil.CreateOrUpdate` (the final func parameter there), since otherwise there's a chance that it won't be re-added if it's removed.

pkg/products/monitoring/reconciler.go
```diff
@@ -457,19 +459,51 @@ func (r *Reconciler) reconcileScrapeConfigs(ctx context.Context, serverClient k8
...
+			logrus.Errorf("Error reconciling dashboards: %v", err)
+			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to create/update grafana dashboard %s: %w", dashboard, err)
 		}
-		r.Logger.Infof("Reconciling the monitoring template %s was successful", template)
+		r.Logger.Infof("Reconciling the grafana dashboard  %s was successful", dashboard)
 	}
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
 
+func (r *Reconciler) reconcileGrafanaDashboards(ctx context.Context, serverClient k8sclient.Client, dashBoard string) (err error) {
```
grdryn, line 475:
```suggestion
func (r *Reconciler) reconcileGrafanaDashboards(ctx context.Context, serverClient k8sclient.Client, dashboard string) (err error) {
```

pkg/products/monitoring/reconciler_test.go
```diff
@@ -441,6 +446,15 @@ func TestReconciler_fullReconcile(t *testing.T) {
 			if status != tc.ExpectedStatus {
 				t.Fatalf("Expected status: '%v', got: '%v'", tc.ExpectedStatus, status)
 			}
+
+			//Verify that grafana dashboards are created
+			for _, dashBoard := range reconciler.Config.GetDashboards() {
```
grdryn, line 451:
```suggestion
			for _, dashboard := range reconciler.Config.GetDashboards() {
```

pkg/products/monitoring/reconciler.go
```diff
@@ -457,19 +459,51 @@ func (r *Reconciler) reconcileScrapeConfigs(ctx context.Context, serverClient k8
...
-	for _, template := range r.Config.GetTemplateList() {
-		// create it
-		_, err := r.createResource(ctx, template, serverClient)
+func (r *Reconciler) reconcileDashboards(ctx context.Context, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
+
+	for _, dashboard := range r.Config.GetDashboards() {
+		err := r.reconcileGrafanaDashboards(ctx, serverClient, dashboard)
 		if err != nil {
-			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to create/update monitoring template %s: %w", template, err)
+			logrus.Errorf("Error reconciling dashboards: %v", err)
```
grdryn, line 467: This looks to be a new error log. It looks like there's more info in the error passed to the return call on the next line, so I'm wondering if this log call would be better placed above line 242, WDYT?

pkg/products/monitoring/reconciler.go
Rajagopalan-Ranganathan: Agreed. Fixed.

pkg/products/monitoring/reconciler.go
Rajagopalan-Ranganathan: agreed. fixed.

pkg/products/monitoring/reconciler_test.go
Rajagopalan-Ranganathan: Agreed. Fixed

pkg/products/monitoring/reconciler.go
Rajagopalan-Ranganathan: Agreed.Fixed.

pkg/products/monitoring/reconciler.go
```diff
@@ -457,19 +460,51 @@ func (r *Reconciler) reconcileScrapeConfigs(ctx context.Context, serverClient k8
...
+			Name:      dashboard,
+			Namespace: r.Config.GetOperatorNamespace(),
+		},
+	}
+
+	opRes, err := controllerutil.CreateOrUpdate(ctx, serverClient, grafanaDB, func() error {
+		specJSON, name, err := getSpecDetailsForDashboard(dashboard)
+		if err != nil {
+			return err
+		}
```
grdryn, line 488: Minor/optional thing: this could maybe be moved to the top of the `reconcileGrafanaDashboards` function rather than here in the mutate function that's passed to `controllerutil.CreateOrUpdate`. Doing so would avoid the extra k8s client GET that `controllerutil.CreateOrUpdate` will do before running this mutate function (we know at a much earlier stage that an error may occur here, and we don't get any extra info by waiting until `controllerutil.CreateOrUpdate` calls the mutate function).

Rajagopalan-Ranganathan: This is the main portion of JSON definition that needs to be updated if there are any changes. From our code base, anything that isnt objectMeta, and that needs to be updates are usually placed inside the createorupdate(). This method "getSpecDetailsForDashboard" is not a client operation , but rather a static helper method that returns the JSON spec and "name" as strings.

grdryn: I think it could still be pulled out of the mutate function here, into the `reconcileGrafanaDashboards` function though: `specJSON` and `name` should still be in scope in here in the mutate function for when they're used in L494/L495.

Like I said though, it's no big deal, it's an optional change :slightly_smiling_face:

Rajagopalan-Ranganathan: I understood now your point. Agreed. And Fixed!

PR #875, opened 2020-06-10
Creation of alerts for all RHMI service endpoints

obrienrobert:
## Description
https://issues.redhat.com/browse/INTLY-7403

The naming conventions for each alert along with the requirements for the alerts expressions can be found in [INTLY-7402](https://issues.redhat.com/browse/INTLY-7402).

## Verification Steps
- Install `RHMI` using this branch
- Ensure that none of the alerts are firing during install
- Ensure all of the required alerts are created. [This](https://docs.google.com/spreadsheets/d/1Wt6qrRskOQLd5DndMXvMvocsfsxcIOk7uk8TWsX4MBo/edit#gid=0) Google Sheet contains all of the alerts that were created (please note that there are 3 tabs)
- To test an individual alert, go to the product/operator namespace in which the alerts resides, and search for the `Endpoints` type resource (`home` -> `search` -> `Endpoints`). Within the YAML definitition of the resource, delete everything from the word `subsets` down
- After a couple of minutes, the corresponding alert should fire. Please note that the operator will eventually reconcile the `Endpoints` resource, so the alert will return to normal.
- Ensure that a SOP URL is also available for all of the newly created alerts. The SOP is still a WiP so the link will be dead until the SOP is completed and merged.

## Testing
The `alerts_exist.go` test file has been updated to include the alerts created in this PR.
- Install `RHMI` using this branch.
- Run the tests against your cluster using  `make test/functional`

## Type of change

## Checklist

### DIFF
--- a/pkg/products/amqonline/prometheusRules.go
+++ b/pkg/products/amqonline/prometheusRules.go
@@ -6,6 +6,7 @@ import (
 	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
 	"github.com/integr8ly/integreatly-operator/pkg/config"
+	"github.com/integr8ly/integreatly-operator/pkg/resources"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -25,9 +26,9 @@ func (r *Reconciler) reconcileSloAlerts(ctx context.Context, serverClient k8scli
 
 	rules := []monitoringv1.Rule{
 		{
-			Alert: fmt.Sprintf("AMQOnlineConsoleAvailable"),
+			Alert: "AMQOnlineConsoleAvailable",
 			Annotations: map[string]string{
-				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
+				"sop_url": resources.SopUrlAlertsAndTroubleshooting,
 				"message": fmt.Sprintf("AMQ-SLO-1.1: AMQ Online console is not available in namespace '%s'", r.Config.GetNamespace()),
 			},
 			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint='console',namespace='%s'}==2)", r.Config.GetNamespace())),
@@ -35,9 +36,9 @@ func (r *Reconciler) reconcileSloAlerts(ctx context.Context, serverClient k8scli
 			Labels: map[string]string{"severity": "critical"},
 		},
 		{
-			Alert: fmt.Sprintf("AMQOnlineKeycloakAvailable"),
+			Alert: "AMQOnlineKeycloakAvailable",
 			Annotations: map[string]string{
-				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
+				"sop_url": resources.SopUrlAlertsAndTroubleshooting,
 				"message": fmt.Sprintf("AMQ-SLO-1.4: Keycloak is not available in namespace %s", r.Config.GetNamespace()),
 			},
 			Expr:   intstr.FromString(fmt.Sprintf("absent(kube_endpoint_address_available{endpoint='standard-authservice',namespace='%s'}==%v)", r.Config.GetNamespace(), keycloakServicePortCount)),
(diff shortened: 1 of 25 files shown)


R-Lawton:
:eyes:

pkg/products/amqonline/prometheusRules.go
```diff
@@ -0,0 +1,166 @@
...
+	rule := &monitoringv1.PrometheusRule{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rhmi-amq-online-slo",
+			Namespace: r.Config.GetNamespace(),
+		},
+	}
+
+	rules := []monitoringv1.Rule{
+		{
+			Alert: fmt.Sprintf("AMQOnlineConsoleAvailable"),
```
Rajagopalan-Ranganathan, line 28: Is there a reason to use sprintf? when we are not building any string? Applicable in other places as well.

pkg/products/amqonline/prometheusRules.go
```diff
@@ -0,0 +1,166 @@
...
+			Annotations: map[string]string{
+				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
```
Rajagopalan-Ranganathan, line 30: Could this be made as a global variable (const) and referred? It is the same string used in places for now, wouldbe easier to make changes in future.

pkg/products/amqonline/prometheusRules.go
```diff
@@ -0,0 +1,166 @@
...
+			Annotations: map[string]string{
+				"sop_url": "https://github.com/RHCloudServices/integreatly-help/tree/master/sops/2.x/alerts/service_endpoint_down.asciidoc",
+				"message": fmt.Sprintf("No {{  $labels.endpoint  }} endpoints in namespace %s. Expected at least 1.", r.Config.GetNamespace()),
+			},
+			Expr:   intstr.FromString("kube_endpoint_address_available{endpoint='enmasse-operator-metrics'} * on (namespace) group_left kube_namespace_labels{label_monitoring_key='middleware'} < 1"),
+			For:    "5m",
+			Labels: map[string]string{"severity": "critical"},
+		}}
+
+	_, err := controllerutil.CreateOrUpdate(ctx, serverClient, rule, func() error {
```
Rajagopalan-Ranganathan, line 149: It is good to get the operation result and log it for anything other than "controllerutil.OperationResultNone".

pkg/products/amqonline/prometheusRules.go
obrienrobert: @Rajagopalan-Ranganathan I've removed the use of `sprintf`. I'm not sure why is was there, but this is an existing alert that I simply moved into the newly created `prometheusRule.go` file, which is a file I've created for each product and contains all of the products related Prometheus rules.

The `reconcileKubeStateMetricsEndpointAvailableAlerts` and `reconcileKubeStateMetricsOperatorEndpointAvailableAlerts` functions within each products `prometheusRule.go` file are the new alerts that I have added.

pkg/products/amqonline/prometheusRules.go
obrienrobert: Good catch. Again, this is an existing alert that I simply moved into the `prometheusRule.go` file, but I've updated it to use a variable rather than be hard-coded.

test/common/alerts_exist.go
```diff
@@ -239,6 +239,157 @@ var expectedRules = []alertsTestRule{
...
+			"RHMIFuseOnlineSyndesisPrometheusServiceEndpointDown",
+			"RHMIFuseOnlineSyndesisServerServiceEndpointDown",
+			"RHMIFuseOnlineSyndesisUiServiceEndpointDown",
+		},
+	},
+	{
+		File: NamespacePrefix + "ups-ksm-endpoint-alerts.yaml",
+		Rules: []string{
+			"RHMIUpsUnifiedPushServiceEndpointDown",
+			"RHMIUpsUnifiedpushProxyServiceEndpointDown",
```
david-martin, line 390: @obrienrobert the naming of these looks a little odd, and I can see why it might from the service names of `ups-unifiedpush` & `ups-unifiedpush-proxy`.
Perhaps `RHMIUPSUnifiedPushServiceEndpointDown` & `RHMIUPSUnifiedPushProxyServiceEndpointDown`

test/common/alerts_exist.go
```diff
@@ -239,6 +239,157 @@ var expectedRules = []alertsTestRule{
 		File:  "redhat-rhmi-amq-online-enmasse-console-rules.yaml",
 		Rules: []string{},
 	},
+	{
+		File: NamespacePrefix + "apicurito-ksm-endpoint-alerts.yaml",
+		Rules: []string{
+			"RHMIApicuritoServiceEndpointDown",
```
david-martin, line 245: @Nanyte25 @alanmoran fyi, ~50 new alerts to cover all Services having at least 1 Endpoint Address available.
They all use a similar query, and will all have the same SOP.
The reason for having individual alerts rather than 1 generic one was to allow for a couple of edge cases where a generic alert would be tricky to allow for (& maintain), and to allow tweaking individually per product & per service in future.

test/common/alerts_exist.go
obrienrobert: @david-martin agreed. I've updated the names of the `UPS` alerts.

PR #879, opened 2020-06-11
adds logic to delete installplan when CSVs are published

jjaferson:
# Description
We need to be able to add patch releases to RHMI operator even when there is an upgrade scheduled so this PR deletes an installplan with service affecting upgrade, in case it's scheduled, and recreates the upgrade once the patch release is installed

https://issues.redhat.com/browse/INTLY-8185

## Verication Steps

1) Create three CSVs:
 1.1 CSV 2.3.0 removing the replaces field.
 1.2 CSV 2.4.0 with the replaces field pointing to CSV 2.3.0
 1.3 CSV 2.3.1 with the replaces field pointing to CSV 2.3.0 and add the service affecting annotation to the csv metadata,  like here: https://github.com/integr8ly/integreatly-operator/blob/master/deploy/olm-catalog/integreatly-operator/2.2.1/integreatly-operator.v2.2.1.clusterserviceversion.yaml#L51
 1.4 Change CSV 2.4.0 to point the replaces field to CSV 2.3.1 and tag the CSV with another name

2) Install & Upgrade
   2.1 Install the operator from CSV 2.3.0
   2.2 Change catalog source to point to CSV 2.4.0
   * An installplan will be created and pending approval
   * Verify if the RHMIConfig CR has got a TargetVersion pointing to CSV 2.4.0

   2.3 Change catalog source to point to CSV 2.4.0 that have the replaces field pointing to CSV 2.3.1
   * Verify if the installplan was deleted and the subscription state changed to UpgradeAvailable 

   2.4 Change catalog source to point to CSV 2.3.1  
    * Verify if the upgrade was performed 

   2.5 Change the catalog source to point to CSV 2.4.0 with the replaces field pointing to CSV 2.3.1
    * Verify if a new installplan is created and it's waiting for approval
    * Check if the RHMIConfig CR status block has not changed and have the TargetVersion set

 

## Type of change

## Checklist


pkg/controller/subscription/rhmiConfigs/main.go
```diff
@@ -73,6 +85,8 @@ func UpdateStatus(ctx context.Context, client k8sclient.Client, config *integrea
 		config.Status.Maintenance.Duration = strconv.Itoa(WINDOW) + "hrs"
 	}
 
+	config.Status.TargetVersion = targetVersion
+
 	// Calculate the upgrade window
 	if installplan.Spec.Approved {
 		config.Status.Upgrade.Window = ""
```
sergioifg94, line 92:
```suggestion
		config.Status.Upgrade.Window = ""
		config.Status.TargetVersion = ""
```
I'd clear the targetVersion field once the upgrade is approved

pkg/controller/subscription/subscription_controller.go
```diff
@@ -29,29 +32,49 @@ import (
 const (
 	// IntegreatlyPackage - package name is used for Subsription name
 	IntegreatlyPackage = "integreatly"
+	CSVNamePrefix      = "integreatly-operator"
```
philbrookes, line 35: Do we know this is the name used in managed-tenants too?

philbrookes: I just checked, it is.

pkg/controller/subscription/subscription_controller.go
```diff
@@ -71,13 +94,14 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 type ReconcileSubscription struct {
-	client            k8sclient.Client
-	restConfig        *rest.Config
-	context           context.Context
-	cancel            context.CancelFunc
-	scheme            *runtime.Scheme
-	operatorNamespace string
-	mgr               manager.Manager
+	client              k8sclient.Client
+	context             context.Context
```
philbrookes, line 98: We shouldn't store handles to contexts in structs, see the last paragraph in this overview: https://golang.org/pkg/context/

pkg/controller/subscription/webapp/main.go
```diff
@@ -13,12 +13,17 @@ import (
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-type UpgradeNotifier struct {
+type UpgradeNotifier interface {
```
philbrookes, line 16: Why is this stuff in here?

sergioifg94: I can answer this one. As @jjaferson added unit tests that would hit the `NotifyUpgrade` call, this would fail the unit tests as the subscription controller was initially instantiating the `UpgradeNotifier` struct with a `rest.Config` object, which can't be mocked in unit tests. So we had to refactor it here to make `UpgradeNotifier` an interface that can be mocked in the unit tests

pkg/controller/subscription/subscription_controller.go
jjaferson: I see, it got added in after rebasing off master. I'll see what are the impacts of removing the context from the structs and make the changes. 

thanks

pkg/controller/subscription/subscription_controller.go
```diff
@@ -146,29 +163,68 @@ func (r *ReconcileSubscription) HandleUpgrades(rhmiSubscription *operatorsv1alph
...
+		return reconcile.Result{}, fmt.Errorf("error getting the csv from catalogsource %w", err)
 	}
 
-	canUpgradeNow, err := rhmiConfigs.CanUpgradeNow(config, installation)
-	if err != nil {
-		return reconcile.Result{}, err
+	currentOperatorVersionName := fmt.Sprintf("%s.v%s", CSVNamePrefix, version.Version)
+	if csvFromCatalogSource.Spec.Replaces != currentOperatorVersionName {
+
+		if csvFromCatalogSource.Spec.Replaces != latestRHMICSV.Spec.Replaces {
```
matskiv, line 183: What if an instance is at 2.3.0, then we release 2.4.0 (replaces 2.3.0) but before the upgrade happens we release 2.3.1. (also replaces 2.3.0) => this condition would resolve to false and IP wouldn't be deleted?

philbrookes: in this situation I'd expect the 2.4.0 to change replaces field to 2.3.1?

sergioifg94: In that case even if we release 2.3.1, the `csvFromCatalogSource` is still 2.4.0, only that now the `replaces` has been updated to 2.3.1, so in that case both conditions would be true, as the current available CSV (latestRHMICSV) is also 2.4.0 but still replaces 2.3.0

philbrookes: but if latestRHMICSV replaces != csvFromCatalogSource replaces, then delete the install plan, right?

jjaferson: @philbrookes yes, we would need to push another csv for 2.4.0 with the replaces field pointing to 2.3.1

matskiv: Ok, thanks for explanation guys.
I didn't know we would change 2.4.0 replaces field.

PR #882, opened 2020-06-15
fix test flake - grafana route

jhellar:
# Description

https://issues.redhat.com/browse/INTLY-6738

This change uses new http client to authenticate and send request to grafana to avoid possible issue with cookies remaining from last session.

As suggested by @HubertStefanski I've left the test as skipped if failed as it is hard to verify that this change really fixed the issue. I will observe prow tests and if the test flake won't show up in ~1 week, I will change it so that it won't be skipped in case of error.


matskiv:
Looks good 👍 
/approve

PR #889, opened 2020-06-16
Intly 6599 convert remaing alert templates to code

R-Lawton:
# Description
The following PR is for converting Prometheus rules (alerts) that were once implemented by templates are now implemented by code

# Verifcation
1. install rhmi from this branch
2. run the following scripts to make alerts fire
```
NS_PREFIX="redhat-rhmi"

namespaces=("3scale" "amq-online" "codeready-workspaces" "fuse" "solution-explorer")         

while true; do
  echo "scaling down"

  for proj in ${namespaces[@]}; do

    project="$NS_PREFIX-$proj"

    for deployment in `oc get deployment -n $project -o json |  jq -r '.items[].metadata.name' | grep -v operator`; do
      if oc get deployment $deployment -n $project -o json | jq '.spec.replicas' | grep -v 0; then
        echo "scaling down $deployment"
        oc scale deployment $deployment --replicas=0 -n $project
      fi
    done

    for dc in `oc get dc -n $project -o json |  jq -r '.items[].metadata.name' | grep -v operator`; do
      if oc get dc $dc -n $project -o json | jq '.spec.replicas' | grep -v 0; then
        echo "scaling down $dc"
        oc scale dc $dc --replicas=0 -n $project
      fi
    done

  done
done
```

```
NS_PREFIX="redhat-rhmi"

namespaces=("3scale-operator" "amq-online" "codeready-workspaces-operator" "fuse-operator"  "solution-explorer-operator") 

while true; do
  echo "scaling down"

  for proj in ${namespaces[@]}; do

    project="$NS_PREFIX-$proj"

    for deployment in `oc get deployment -n $project -o json |  jq -r '.items[].metadata.name' | grep operator`; do
      if oc get deployment $deployment -n $project -o json | jq '.spec.replicas' | grep -v 0; then
        echo "scaling down $deployment"
        oc scale deployment $deployment --replicas=0 -n $project
      fi
    done

    for dc in `oc get dc -n $project -o json |  jq -r '.items[].metadata.name' | grep operator`; do
      if oc get dc $dc -n $project -o json | jq '.spec.replicas' | grep -v 0; then
        echo "scaling down $dc"
        oc scale dc $dc --replicas=0 -n $project
      fi
    done

  done
done
```
3. The following tests will not fire as they are not affected by the scripts above:

The following can be checked by removing the absent in the expression in Prometheus this should give you back a value for each expression
- ThreeScaleDeveloperUIBBT 
- ThreeScaleSystemAdminUIBBT  
- ThreeScaleAdminUIBBT

The following have to be checked by comparing the expression from the scripts against the expression in the code to make sure they are the same
- ThreeScaleContainerHighCPU 
- ThreeScaleContainerHighMemory
- AMQOnlineContainerHighMemory

4. Ensure the converted alerts are them same as the templates


pkg/products/amqonline/prometheusRules.go
```diff
@@ -0,0 +1,70 @@
...
+		},
+}
+
+	var namespace = r.Config.GetNamespace()
+
+	rules := []monitoringv1.Rule{
+		{
+			Alert: "AMQOnlinePodCount",
+			Annotations: map[string]string{
+				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
```
Rajagopalan-Ranganathan, line 32: The SOP url, can be moved to global const, making changes and maintenance would be easy. WDYT?

pkg/products/amqonline/prometheusRules.go
```diff
@@ -0,0 +1,70 @@
...
+				"message": "The {{  $labels.container  }} Container in the {{ $labels.pod  }} Pod has been using {{  $value }}% of available memory for longer than 15 minutes.",
+			},
+			Expr:   intstr.FromString(fmt.Sprintf("sum by(container, pod) (container_memory_usage_bytes{container!='',namespace='%s'}) / sum by(container, pod) (kube_pod_container_resource_limits_memory_bytes{namespace='%s'}) * 100 > 90", namespace, namespace)),
+			For:    "15m",
+			Labels: map[string]string{"severity": "warning"},
+		},
+	}
+	_, err := controllerutil.CreateOrUpdate(ctx, serverClient, rule, func() error {
```
Rajagopalan-Ranganathan, line 50: Would recommend getting the operation result and logging it in the events of ```controllerutil.OperationResultNone```. Would be easier to debug if it had created or updated, during a reconcile.

pkg/products/amqonline/prometheusRules.go
```diff
@@ -0,0 +1,70 @@
...
+					Name:  "general.rules",
+					Rules: rules,
+				},
+			},
+		}
+		return nil
+	})
+
+	if err != nil {
+		logrus.Infof("Phase: %s reconcilePrometheusAlerts", err)
```
Rajagopalan-Ranganathan, line 64: I think errors, should be at error level. ```logrus.Errorf```. WDYT?

pkg/products/codeready/prometheusRules.go
```diff
@@ -0,0 +1,58 @@
...
+		},
+	}
+
+	var namespace = r.Config.GetNamespace()
+
+	rules := []monitoringv1.Rule{
+		{
+			Alert: "CodeReadyPodCount",
+			Annotations: map[string]string{
+				"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md",
```
Rajagopalan-Ranganathan, line 31: Similar comment, can SOP URL and other repeated constants be declared as global constants, so that making changes in the future might be easier?

pkg/products/codeready/prometheusRules.go
```diff
@@ -0,0 +1,58 @@
...
+				"message": "Pod count for namespace {{  $labels.namespace  }} is {{  $value  }}. Expected at least 2 pods.",
+			},
+			Expr:   intstr.FromString(fmt.Sprintf("(1-absent(kube_pod_status_ready{condition='true', namespace='%s'})) or sum(kube_pod_status_ready{condition='true', namespace='%s'}) < 2", namespace, namespace)),
+			For:    "5m",
+			Labels: map[string]string{"severity": "critical"},
+		},
+}
+_, err := controllerutil.CreateOrUpdate(ctx, serverClient, rule, func() error {
```
Rajagopalan-Ranganathan, line 39: Same as previous comments, if possible please get operaiton result and log it.

pkg/products/codeready/prometheusRules.go
```diff
@@ -0,0 +1,58 @@
...
+					Name:  "general.rules",
+					Rules: rules,
+				},
+			},
+		}
+		return nil
+	})
+
+	if err != nil {
+		logrus.Infof("Phase: %s reconcilePrometheusAlerts", err)
```
Rajagopalan-Ranganathan, line 53: should be logrus.Errorf, since it is an error?

Rajagopalan-Ranganathan:
@R-Lawton : General comments/observations: SOP url, can be made a global constant and referred. easier to maintain and make changes. For createorupdate methods, would be better to get the operation result and log it when ever there is a change (Create or an update). Error logs should be at error level (logrus.Errorf), rather than info level.

pkg/products/fuse/reconciler.go
```diff
@@ -177,13 +177,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
-	phase, err = r.reconcileTemplates(ctx, serverClient)
-	logrus.Infof("Phase: %s reconcileTemplates", phase)
+	// phase, err = r.reconcileTemplates(ctx, serverClient)
```
Rajagopalan-Ranganathan, line 180: Can we remove these commented lines? or are they kept intentionally?

pkg/products/fuse/reconciler.go
```diff
@@ -224,18 +232,18 @@ func (r *Reconciler) createResource(ctx context.Context, resourceName string, se
...
-		// create it
-		_, err := r.createResource(ctx, template, serverClient)
-		if err != nil {
-			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to create/update monitoring template %s: %w", template, err)
-		}
-		logrus.Infof("Reconciling the monitoring template %s was successful", template)
-	}
-	return integreatlyv1alpha1.PhaseCompleted, nil
-}
+// func (r *Reconciler) reconcileTemplates(ctx context.Context, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
```
Rajagopalan-Ranganathan, line 235: Can we remove these commented out lines? if they are left intentioanlly, would it be better have a comment why they are commented? WDYT?

pkg/products/fuse/reconciler.go
R-Lawton: oops that should have been removed my mistake

pkg/products/amqonline/prometheusRules.go
R-Lawton: Okey dokes ill change them

pkg/products/amqonline/prometheusRules.go
R-Lawton: i was holding off to rebase from Robs Pr and convert them but i can do it now

pkg/products/amqonline/prometheusRules.go
R-Lawton: I think the error message that is returned should be enough to figure out why an error is occurring I dont really see the need to know if it happened during and update or create

Rajagopalan-Ranganathan: The information of either a create or update, being logged is crucial because: when we upgrade an existing cluster and expect a dashboard to be updated (rather than created ) during a reconcile, it would be nice to know that it was updated. In a similar way for creation. It is not about the error , rather than a log for us to track if it was created or updated as expected. And code changes like this would be ideal, so that it doesnt flood the messages during each reconcile: ```	if opRes != controllerutil.OperationResultNone {
		r.Logger.Infof("operation result of XXXXXX  %v was %v",, opRes)
	}```. It would log only created or updated cases, which ideally should occur only once.

R-Lawton: Scratch the above comment looking through how other create or updates functions are done in the code this morning implement the log.  I can no see why its needed so im happy to add it in

PR #896, opened 2020-06-18
Update n02-measure-downtime-during-rhmi-upgrade.md

wei-lee:
# Description

## Type of change

## Checklist

### DIFF
--- a/test-cases/tests/upgrade/n02-measure-downtime-during-rhmi-upgrade.md
+++ b/test-cases/tests/upgrade/n02-measure-downtime-during-rhmi-upgrade.md
@@ -23,6 +23,7 @@ Mesure the downtime of the RHMI components during the RHMI upgrade (not to be co
    ```
    oc login --token=<TOKEN> --server=https://api.<CLUSTER_NAME>.s1.devshift.org:6443
    ```
+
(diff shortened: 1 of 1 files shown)


b1zzu:
/lgtm 

But we have to update this document:
https://github.com/RHCloudServices/integreatly-help/blob/master/guides/manual-release-tests-selection.md#pre-upgrade-work
and remove the H21 from the Pre-upgrade Work section

we should also update the N04 and remove the H21 requirements, and update the H21 and remove the note about upgrades.. or more generic what should we do with the H21?
N04:https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/upgrade/n04-verify-customer-config-reconcile-upgrade-logic-and-trigger-t.md
H21: https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/products/h21-verify-all-products-using-the-workloadwebapp.md

wei-lee:
@b1zzu I think we should remove H21. Looks like N02 is covering the upgrade already.

b1zzu:
@wei-lee the idea of H21 is that it could be used also for a rhmi installation to easily test 3scale, sso, and amq, so I think is good to keep it, but we should clean it from all upgrade info

PR #899, opened 2020-06-18
testcase: update testcases

damienomurchu:
# Description

## Type of change

## Checklist


test-cases/tests/walkthroughs/g06-verify-user-access-to-settings-menu.md
```diff
@@ -3,7 +3,7 @@ estimate: 15m
 tags:
 ---
 
-# G06 - Verify user role can be determined in solution-explorer
+# G06 - Verify user access to settings menu
```
trdoyle81, line 6: This will fail on checks because the title has to be the same as the file name.

PR #915, opened 2020-06-23
INLTY-7774 Code refactor and added http timeouts to try fix memory leak

CathalOConnorRH:
# Description
Refactor of code and added http timeouts to fix a memory leak caused by connections being left open and idle

## Type of change

## Checklist


pb82:
:eyes:

pkg/controller/installation/installation_controller.go
```diff
@@ -241,14 +228,13 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 	}
 	metrics.SetRHMIStatus(installation)
 
-	configManager, err := config.NewManager(r.context, r.client, request.NamespacedName.Namespace, installationCfgMap, installation)
+	configManager, err := config.NewManager(context.TODO(), r.client, request.NamespacedName.Namespace, installationCfgMap, installation)
```
pb82, line 231: Any particular reason for no longer using a long lived context?

pkg/controller/installation/installation_controller.go
```diff
@@ -386,51 +315,157 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
 	// Check if the version needs to be updated
-	if (isFirstInstallReconcile(installation) || isUpgradeReconcile(installation)) && allProductsReconciled {
+	if (isFirstInstallReconcile(installation) || isUpgradeReconcile(installation)) && allProductsReconciled(installation) {
+		installation.Status.Version = installation.Status.ToVersion
+		installation.Status.ToVersion = ""
+		metrics.SetRhmiVersions(string(installation.Status.Stage), installation.Status.Version, installation.Status.ToVersion, installation.CreationTimestamp.Unix())
+	}
+	metrics.SetRHMIStatus(installation)
+
+	if (isFirstInstallReconcile(installation) || isUpgradeReconcile(installation)) && allProductsReconciled(installation) {
```
pb82, line 325: redundant? This seems to be the same condition as above with the same actions when true.

pkg/controller/installation/installation_controller.go
```diff
@@ -386,51 +315,157 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
+		phase, err := reconciler.Reconcile(context.TODO(), installation, product, serverClient)
+		if err != nil {
+			merr.Add(fmt.Errorf("Failed to reconcile product %s: %w", product.Name, err))
+		}
+		logrus.Infof("current phase for %s is: %s", product.Name, phase)
 	}
-	if installInProgress {
+
+	//any products left to uninstall, or any errors during product uninstall, add to CR and return
+	if len(installation.Finalizers) > 1 || len(merr.errors) > 0 {
```
pb82, line 417: why `> 1` here? Is 1 finalizer reserved for the intly operator itself or should it be > 0?

pkg/controller/installation/installation_controller.go
philbrookes: it's stated explicitly that storing a context can cause a memory leak.

pkg/controller/installation/installation_controller.go
philbrookes: yes, there is one reserved finalizer for the operator itself.

pkg/controller/installation/installation_controller.go
philbrookes: this looks like a rebasing error. @CathalOConnorRH can we get the duplicate removed here?

pkg/controller/installation/installation_controller.go
philbrookes: last paragraph in overview here, for more info: https://golang.org/pkg/context/

pkg/controller/installation/installation_controller.go
```diff
@@ -445,6 +480,240 @@ func isUpgradeReconcile(installation *integreatlyv1alpha1.RHMI) bool {
 	return false
 }
 
+func allProductsReconciled(installation *integreatlyv1alpha1.RHMI) bool {
+	stages := installation.Status.Stages
+
+	return (verifyClusterRHSSOVersions(stages) &&
```
pb82, line 486: I wonder if these functions could be generic, e.g. `verifyProductVersion(stages, stage, product, operatorVersion, productVersion)`

philbrookes: this is done in another PR. This could should no longer exist. I wonder if this is another rebasing error @CathalOConnorRH ?

pkg/controller/installation/installation_controller.go
```diff
@@ -445,6 +480,240 @@ func isUpgradeReconcile(installation *integreatlyv1alpha1.RHMI) bool {
...
+		verifyMonitoringVersions(stages) &&
+		verify3ScaleVersions(stages) &&
+		verifyFuseOnlineVersions(stages) &&
+		verifyDataSyncVersions(stages) &&
+		verifyRHSSOUserVersions(stages) &&
+		verifyUPSVersions(stages) &&
+		verifySolutionExplorerVersions(stages))
+}
+
+func verifySolutionExplorerVersions(stages map[integreatlyv1alpha1.StageName]integreatlyv1alpha1.RHMIStageStatus) bool {
```
philbrookes, line 501: All of these functions should be gone too

pkg/controller/installation/installation_controller.go
```diff
@@ -586,19 +855,16 @@ func (r *ReconcileInstallation) processStage(installation *integreatlyv1alpha1.R
 	installation.Status.Stage = stage.Name
 
 	for _, product := range stage.Products {
-
 		reconciler, err := products.NewReconciler(product.Name, r.restConfig, configManager, installation, r.mgr)
 		if err != nil {
 			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to build a reconciler for %s: %w", product.Name, err)
 		}
-
-		allProductsReconciled = reconciler.VerifyVersion(installation)
```
philbrookes, line 595: This should be back in

pkg/products/datasync/reconciler.go
```diff
@@ -78,6 +79,8 @@ func NewReconciler(configManager config.ConfigReadWriter, installation *integrea
 
 	logger := logrus.NewEntry(logrus.StandardLogger())
 	var httpClient http.Client
+	httpClient.Timeout = time.Second * 10
+	httpClient.Transport = &http.Transport{DisableKeepAlives: true, IdleConnTimeout: time.Second * 10}
```
pb82, line 83: Do we need `IdleConnTimeout` if keep alive connections are disabled? If I understand correctly, that timeout is only used to close idle connections, according to the doc:

```
    // IdleConnTimeout is the maximum amount of time an idle
    // (keep-alive) connection will remain idle before closing
    // itself.
    // Zero means no limit.
```

philbrookes:
@CathalOConnorRH is on PTO, I will re-open this PR from my fork

pkg/products/datasync/reconciler.go
```diff
@@ -111,6 +114,7 @@ func (r *Reconciler) reconcileTemplates(ctx context.Context, serverClient k8scli
 		if err != nil {
 			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to get file contents of %s: %w", templateFn, err)
 		}
+		defer fileData.Close()
```
pb82, line 117: We could make sure that `getFileContentFromURL` closes it's request body and instead of a `Reader` just returns a string. Otherwise we always have to be careful to close the reader when using this function.

pkg/products/fuseonopenshift/reconciler.go
```diff
@@ -126,6 +127,8 @@ func NewReconciler(configManager config.ConfigReadWriter, installation *integrea
 
 	logger := logrus.NewEntry(logrus.StandardLogger())
 	var httpClient http.Client
+	httpClient.Timeout = time.Second * 10
+	httpClient.Transport = &http.Transport{DisableKeepAlives: true, IdleConnTimeout: time.Second * 10}
```
pb82, line 131: Same as above, `IdleConnTimeout` redundant but probably not harmful.

pkg/controller/installation/installation_controller.go
pb82: Oh good point. I think we do this in other places too.

philbrookes:
moved: https://github.com/integr8ly/integreatly-operator/pull/951

PR #917, opened 2020-06-23
changing from container_memory_usage_bytes to container_memory_workin…

R-Lawton:
# Description
[Jira](https://issues.redhat.com/browse/INTLY-8591) Change container_memory_usage_bytes to container_memory_working_set_bytes

## Type of change

## Verification
1.Install RHMI from master and compare and take note of values gotten from the memory usage graph in both resource usage by pod and resource usage by namespace against the equivalent values in the openshift console metrics e.g keycloak-0 in rhsso in a dashboard and keycloak-0 in the same namespace in the openshift-console metrics. These values should be different. 
2. When gathering values in resource usage by pod  and namespace take note the values the pod requests% and limits% values. There is no opensift metric equivalent to this but they will be compared against the new values later
3. Install Rhmi from this branch
4. Compare these new values again, the values from the dashboard should now be the same as the openshift console values now and shouldnt have changed much from the original openshift metrics values .
5. Do the same for the request% and limit % but compare them against the older version they should be slightly different either positively or negatively different
## Checklist

### DIFF
--- a/pkg/products/monitoring/dashboards/resourceByNamespace.go
+++ b/pkg/products/monitoring/dashboards/resourceByNamespace.go
@@ -423,7 +423,7 @@ const MonitoringGrafanaDBResourceByNSJSON = `{
 			"stack": false,
 			"steppedLine": false,
 			"targets": [{
-				"expr": "sum(container_memory_usage_bytes{namespace=~'$namespace', container!=''}) by (pod)",
+				"expr": "sum(container_memory_working_set_bytes{namespace=~'$namespace', container=''}) by (pod)",
 				"format": "time_series",
 				"intervalFactor": 2,
 				"legendFormat": "{{pod}}",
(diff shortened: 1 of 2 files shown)


pkg/products/monitoring/dashboards/resourceByNamespace.go
```diff
@@ -423,7 +423,7 @@ const MonitoringGrafanaDBResourceByNSJSON = `{
 			"stack": false,
 			"steppedLine": false,
 			"targets": [{
-				"expr": "sum(container_memory_usage_bytes{namespace=~'$namespace', container!=''}) by (pod)",
+				"expr": "sum(container_memory_working_set_bytes{namespace=~'$namespace', container=''}) by (pod)",
```
HubertStefanski, line 426: Changes look good to me, verification worked with the steps above.

The only question I have (Just to get a better understanding) is why do we remove the `!=` operator? my understanding would be that it doesn't actually make a difference in this query, since the previous way just searched for all containers that don't have an empty name  i.e. `!=''` , while with your changes we search for all containers ( `''` being a syntactical equivalent of a non name specific search?) essentially looking for all containers no matter what their label is within this namespace? correct me if I'm wrong, thanks!

/lgtm 
/Approve

R-Lawton: Good question, So both the expressions with != or just = will get you similar results as they are essentially doing the same thing. The difference between them is with the != your manually adding up all the memory resources needed individually to get the value. Whats removed with the !=  is a result that has all these resources added up already but is also bringing in certain memory resources being used that we are missing when summing the values manually. So using the = rather then != is the value coming back is more accurate as its including certain resources we are not getting.

HubertStefanski:
Re-verified :+1: 

/lgtm

PR #920, opened 2020-06-24
Docs Template for Test Cases

joaedwar:
# Description

## Type of change

## Checklist

### DIFF
--- a/test-cases/tests/documentation/template.md
+++ b/test-cases/tests/documentation/template.md
@@ -6,6 +6,7 @@
 
 <Summary>
 
+* Staging doc link:
 * Jira ref link:
 * Peer contact: 
 
--- a/test-cases/tests/documentation/template.md
+++ b/test-cases/tests/documentation/template.md
@@ -1,0 +1,20 @@
+<Test Case Number> - <Jira Title>
+
+* Sprint # 
+
+## Description
+
+<Summary>
+
+* Jira ref link:
+* Peer contact: 
+
+
+## Prerequisites
+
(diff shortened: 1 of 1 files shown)


test-cases/tests/documentation/template.md
```diff
@@ -0,0 +1,20 @@
+<Test Case Number> - <Jira Title>
```
jciroli-rh, line 1: I think we need to use the i<number> format (e.g., i07: Verify Network Policy Settings in Admin Guide).

test-cases/tests/documentation/template.md
```diff
@@ -0,0 +1,20 @@
...
+
```
jciroli-rh, line 2: Not sure if we need to include the time estimate like the others have?

test-cases/tests/documentation/template.md
```diff
@@ -0,0 +1,20 @@
...
+* Sprint # 
+
+## Description
+
+<Summary>
+
+* Jira ref link:
```
jciroli-rh, line 9: This should be a link to the module or modules on our staging server. We can also add the JIRA link if you think it will be helpful to have that relationship preserved.

test-cases/tests/documentation/template.md
```diff
@@ -0,0 +1,20 @@
...
+* Peer contact: 
+
+
+## Prerequisites
+
+* <What is needed to complete this procedure>
+
```
jciroli-rh, line 16: What do you envision here? Contacting the writer for a pair testing session? Something else?

PR #928, opened 2020-06-25
Create user-permissions.md

maleck13:
# Description

## Type of change

## Checklist

### DIFF
--- a/docs/user-permissions.md
+++ b/docs/user-permissions.md
@@ -1,0 +1,55 @@
+# Integreatly Users and Permissions on OSD
+
+Within an RHMI cluster there are two main groups of users that the operator interacts with these are: 
+
+- Dedicated Admins Group
+- RHMI Developers Group
+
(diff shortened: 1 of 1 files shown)


maleck13:
@matskiv ok to review

matskiv:
@maleck13 Are you aware of this spreadsheet, that describes the permissions - https://docs.google.com/spreadsheets/d/1SplYIdK7Z4SBdUzceDyM24XL2pkeD7EDTbL4-Et7nZQ ?
Depending on the target audience, we might want to stick to one place for this information. Wdyt?

maleck13:
@matskiv I was not aware of that. hmm I am torn as I feel there is value in it being in the upstream also

matskiv:
Maybe we keep general information in the upstream, but don't go into details on per product level?
Linking to a spreadsheet would be nice on one hand, but it's not public, so not sure if we should add it 🤔

PR #933, opened 2020-06-26
INTLY-6750-Enable ops addon in Fuse Online and remove alerts/dashboards/servicemonitors created by RHMI for Fuse Online

HubertStefanski:
# Description
Due to a bug in Fuse Online 7.6 (https://github.com/syndesisio/syndesis/issues/8289) we had to create a subset of the Fuse Online alerts and dashboards in the RHMI Operator, instead of just enabling the Fuse Online ops addon.

- Ensure the ops addon for the latest release of Fuse Online does not create alerts for a database when external databases are enabled
- Remove the Fuse Online alerts/dashboards created from the RHMI operator (https://github.com/integr8ly/integreatly-operator/tree/19ffc3907adc138a0ef76b687bb2e88a4a821933/templates/monitoring/fuseonline) & update the reconcile logic to remove the template creation
- Enable the Fuse Online ops addon (here https://github.com/integr8ly/integreatly-operator/blob/19ffc3907adc138a0ef76b687bb2e88a4a821933/pkg/products/fuse/reconciler.go#L342-L352)
- As part of upgrade, add removal logic to delete the old resources from the cluster
- Create a follow up issue to remove the removal logic

## JIRA
https://issues.redhat.com/browse/INTLY-6750

## Type of change

## Checklist

## Verification
- Checkout master and install integreatly on the cluster
- Wait for the integreatly install to finish
- Check that there aren't any owner references in the yaml of the resources created from the [linked templates](https://github.com/integr8ly/integreatly-operator/tree/19ffc3907adc138a0ef76b687bb2e88a4a821933/templates/monitoring/fuseonline)
- Stop the operator
- Checkout this branch and start the operator
- After a few minutes return to the resources and ensure that they now have owner references and don't have any integreatly labels ( apart from the owner reference, which references a `syndesis` named `integreatly`


pkg/products/monitoringspec/reconciler.go
```diff
@@ -225,6 +299,18 @@ func (r *Reconciler) reconcileMonitoring(ctx context.Context, serverClient k8scl
 			}
 		}
 	}
+	//TODO remove from here vvvvvvvvvvvvvvvvvvvvv
+	for _, object := range obsoleteObjects {
+		err := r.removeOldResources(ctx, serverClient, object.objectType, object.name, object.namespace)
+		if k8serr.IsNotFound(err) {
+			//ignore not found, We expect to hit this on every reconcile apart from the first after upgrade
```
HubertStefanski, line 306: ***Note to self:*** come up with some better way of confirming that the resources have been deleted

pkg/products/monitoringspec/reconciler.go
```diff
@@ -39,6 +40,79 @@ const (
...
+type obsoleteObjectStruct struct {
+	name       string
+	namespace  string
+	objectType string
+}
+
+var obsoleteObjects = []obsoleteObjectStruct{
+	{
+		name:       "syndesis-infra-api-dashboard",
+		namespace:  "redhat-rhmi-middleware-monitoring",
```
HubertStefanski, line 54: Not entirely sure if this is the correct namespace/ will be able to test this once we can spin up clusters again.

grdryn:
/cc

:eyes:

HubertStefanski:
FuseOnlinePostgresExporterDown alert firing, holding until resolved

/hold

pkg/products/fuse/reconciler.go
```diff
@@ -510,3 +586,29 @@ func (r *Reconciler) reconcileSubscription(ctx context.Context, serverClient k8s
 		catalogSourceReconciler,
 	)
 }
+
+//this is only a temporary function included to remove obsolete resources left from a workaround for a fuse syndesis bug, all these objects should now be generated through the ops addon
+//TODO Remove from here vvvvvvvvvvvvvvvvvv
+func (r *Reconciler) removeOldResources(ctx context.Context, serverClient k8sclient.Client, obsoleteObjects []runtime.Object) (integreatlyv1alpha1.StatusPhase, error) {
```
grdryn, line 592: `obsoleteObjects` is currently a global variable that's captured in the `Reconcile` function to pass to this function, even though this function would already have access to it. That's fine and won't cause any issues, there's just a level of indirection that might not be necessary and in other cases could cause confusion, especially when the name is re-used: `obsoleteObjects` inside this function is whatever is passed into this function, and that may shadow the global variable if it's different.

Given that `obsoleteObjects` is actually only needed inside this function, my suggestion would be to declare it _inside_ this function, rather than having it be global or passed in. What do you think?

pkg/products/fuse/reconciler.go
```diff
@@ -49,8 +53,90 @@ const (
 	manifestPackage              = "integreatly-fuse-online"
 	syndesisPrometheusPVC        = "10Gi"
 	syndesisPrometheus           = "syndesis-prometheus"
+	//TODO remove this var vvvv
+	obsoleteResourceNS = "redhat-rhmi-fuse"
 )
 
+//Remove this section, it's used for a temporary function to clean up obsolete resources
+//TODO remove from here ---------------------------------------------------
```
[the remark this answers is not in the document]
HubertStefanski: I was waiting till this gets merged so that I can reference these specific TODOs line by line, but I like your suggestion better :+1:

HubertStefanski:
holding long term : https://chat.google.com/room/AAAA9w3UtHY/bhG3f6lltRA , current version of fuse still doesn't have the fix we were hoping for, this causes the alert to be created as described above

PR #937, opened 2020-06-29
[INTLY-8110] Standalone VPC functional test

dimitraz:
# Description

## Type of change

## Checklist

### DIFF
--- a/test/functional/aws_standalone_vpc.go
+++ b/test/functional/aws_standalone_vpc.go
@@ -1,0 +1,534 @@
+package functional
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net"
+	"strings"
+	"testing"
+
+	"github.com/integr8ly/cloud-resource-operator/pkg/resources"
+
+	"github.com/aws/aws-sdk-go/service/elasticache"
+
+	v1 "k8s.io/api/core/v1"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/aws/aws-sdk-go/service/rds"
+
+	"github.com/aws/aws-sdk-go/aws"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/integr8ly/integreatly-operator/test/common"
+)
+
+var (
+	resourceType    = "_network"
+	tier            = "development"
+	strategyMapName = "cloud-resources-aws-strategies"
+
+	vpcTagName           = "RHMI Cloud Resource VPC"
+	subnetTagName        = "RHMI Cloud Resource Subnet"
+	securityGroupTagName = "RHMI Cloud Resource Security Group"
+	peeringTagName       = "RHMI Cloud Resource Peering Connection"
+	routeTableTagName    = "RHMI Cloud Resource Route Table"
+)
+
+type strategyMap struct {
+	CreateStrategy json.RawMessage `json:"createStrategy"`
+}
+
+// a custom error for reporting errors for each
+// network component
+type networkConfigTestError struct {
+	vpcError                  []string
+	subnetsError              []string
+	securityGroupError        []string
+	peeringConnError          []string
+	standaloneRouteTableError []string
+	clusterRouteTablesError   []string
+	rdsSubnetGroupsError      []string
+	cacheSubnetGroupsError    []string
+}
+
+// pretty print the error message if there are any errors
+// this looks ugly but the output looks good!
+func (e *networkConfigTestError) Error() string {
+	var str strings.Builder
+	if len(e.vpcError) != 0 {
+		str.WriteString("\nVPC errors:")
+		for _, item := range e.vpcError {
+			str.WriteString(fmt.Sprintf("\n\t%s", item))
+		}
+	}
+	if len(e.subnetsError) != 0 {
+		str.WriteString("\nSubnet errors:")
+		for _, item := range e.subnetsError {
+			str.WriteString(fmt.Sprintf("\n\t%s", item))
+		}
+	}
+	if len(e.securityGroupError) != 0 {
+		str.WriteString("\nSecurity Group errors:")
+		for _, item := range e.securityGroupError {
+			str.WriteString(fmt.Sprintf("\n\t%s", item))
+		}
+	}
+	if len(e.peeringConnError) != 0 {
+		str.WriteString("\nPeering Connection errors:")
+		for _, item := range e.peeringConnError {
+			str.WriteString(fmt.Sprintf("\n\t%s", item))
+		}
+	}
+	if len(e.standaloneRouteTableError) != 0 {
+		str.WriteString("\nStandalone Route Table errors:")
+		for _, item := range e.standaloneRouteTableError {
+			str.WriteString(fmt.Sprintf("\n\t%s", item))
+		}
+	}
+	if len(e.clusterRouteTablesError) != 0 {
+		str.WriteString("\nCluster Route Table errors:")
+		for _, item := range e.clusterRouteTablesError {
+			str.WriteString(fmt.Sprintf("\n\t%s", item))
+		}
+	}
+	if len(e.rdsSubnetGroupsError) != 0 {
+		str.WriteString("\nRDS Subnet Groups errors:")
+		for _, item := range e.rdsSubnetGroupsError {
+			str.WriteString(fmt.Sprintf("\n\t%s", item))
+		}
+	}
+	if len(e.cacheSubnetGroupsError) != 0 {
+		str.WriteString("\nElasticache Subnet Groups errors:")
+		for _, item := range e.cacheSubnetGroupsError {
+			str.WriteString(fmt.Sprintf("\n\t%s", item))
+		}
+	}
+	return str.String()
+}
+
+// the error is valid if any of the error slices are not empty
+func (e *networkConfigTestError) isValid() bool {
+	return len(e.vpcError) != 0 ||
+		len(e.subnetsError) != 0 ||
+		len(e.securityGroupError) != 0 ||
+		len(e.peeringConnError) != 0 ||
+		len(e.standaloneRouteTableError) != 0 ||
+		len(e.clusterRouteTablesError) != 0 ||
+		len(e.rdsSubnetGroupsError) != 0 ||
+		len(e.cacheSubnetGroupsError) != 0
+}
+
+// TestStandaloneVPCExists tests that the cloud resource operator network components
+// have been correctly set up and configured
+func TestStandaloneVPCExists(t *testing.T, testingCtx *common.TestingContext) {
+	ctx := context.TODO()
+	testErrors := &networkConfigTestError{}
+
+	// create a new session
+	session, err := CreateAWSSession(ctx, testingCtx.Client)
+	if err != nil {
+		t.Fatal("could not create aws session", err)
+	}
+	ec2Sess := ec2.New(session)
+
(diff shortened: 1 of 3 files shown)


test/functional/tests.go
```diff
@@ -7,6 +7,7 @@ var (
 		{Description: "F01 - Verify AWS rds resources exist and are in expected state", Test: AWSRDSResourcesExistTest},
 		{Description: "F03 - Verify AWS elasticache resources exist and are in expected state", Test: AWSElasticacheResourcesExistTest},
 		{Description: "A21 - Verify AWS maintenance and backup windows", Test: CROStrategyOverrideAWSResourceTest},
+		{Description: "Verify AWS Standalone VPC exists", Test: TestStandaloneVPCExists},
```
aidenkeating, line 10: We should probably make sure this has an ID/code associated with it in the test cases

aidenkeating:
@psturc FYI it'd be great to have your input on this

test/functional/tests.go
dimitraz: Sure, @psturc is there a test case already associated to this?

test/functional/aws_standalone_vpc.go
```diff
@@ -0,0 +1,573 @@
...
+	if len(vpcs) != 1 {
+		newErr.vpcError = append(newErr.vpcError, fmt.Sprintf("expected 1 vpc but found %d", len(vpcs)))
+		return nil, newErr
+	}
+
+	// cidr blocks should match
+	vpc := vpcs[0]
+	foundCidr := aws.StringValue(vpc.CidrBlock)
+	if foundCidr != expectedCidr {
+		newErr.vpcError = append(newErr.vpcError, fmt.Sprintf("cidr blocks not equal, expected %s but got %s", expectedCidr, foundCidr))
```
ciaranRoche, line 249: Should we expand that error message to explain what the cidr block should equal too. Or maybe some extra developer docs to call out we expect the cidr block of the vpc to be equal to the cidr block from the strategy map.

test/functional/aws_standalone_vpc.go
```diff
@@ -0,0 +1,573 @@
...
+	// than the vpc subnet mask
+	for _, subnet := range subnets {
+		_, subnetCidr, err := net.ParseCIDR(aws.StringValue(subnet.CidrBlock))
+		if err != nil {
+			newErr.subnetsError = append(newErr.subnetsError, fmt.Sprintf("could not parse subnet mask for vpc subnets: %v", err))
+			return subnets, newErr
+		}
+		subnetCidrMask, _ := subnetCidr.Mask.Size()
+		if subnetCidrMask != cidrMask+1 {
+			newErr.subnetsError = append(newErr.subnetsError, fmt.Sprintf("unexpected subnet mask size for vpc subnets: %v", err))
```
ciaranRoche, line 298: I think we should call out what we expected to find here, something like
```suggestion
			newErr.subnetsError = append(newErr.subnetsError, fmt.Sprintf("subnet mask expect to be 1 bit greater than vpc subnet mask, found : %d, expected %d, subnetCidrMask, cidrMask + 1))
```

test/functional/aws_standalone_vpc.go
```diff
@@ -0,0 +1,573 @@
...
+		newErr.securityGroupError = append(newErr.securityGroupError, fmt.Sprintf("could not find vpc security group: %v", err))
+	}
+
+	// expect 1 security group
+	secGroups := describeGroups.SecurityGroups
+	if len(secGroups) != 1 {
+		newErr.securityGroupError = append(newErr.securityGroupError, fmt.Sprintf("unexpected number of security groups: %d", len(secGroups)))
+		return newErr
+	}
+
```
ciaranRoche, line 334: Should we be validating the ingress rules here, or is it okay to valid the security group exists? 
Goes to what we talked about earlier, if the connectivity metrics fail, it would point to this.

test/functional/aws_standalone_vpc.go
```diff
@@ -0,0 +1,573 @@
...
+		newErr.peeringConnError = append(newErr.peeringConnError, fmt.Sprintf("unexpected number of vpc peering connections: %d", len(conns)))
+		return nil, newErr
+	}
+
+	// verify that the requester info is correct
+	conn := conns[0]
+	if aws.StringValue(conn.RequesterVpcInfo.CidrBlock) != expectedCidr && aws.StringValue(conn.RequesterVpcInfo.VpcId) != vpcID {
+		newErr.peeringConnError = append(newErr.peeringConnError, fmt.Sprintf("unexpected accepter vpc cidr block: %d", len(conns)))
+	}
+
```
ciaranRoche, line 447: Should we add a check to verify the state of the peering connection. We expect `VpcPeeringConnectionStateReasonCodeActive` if it is not active I think we should fail and report back the state of the peering connection. 
Again like above the connection metric should be failing on this, but having the test error would help in debugging the issue IMO

test/functional/aws_standalone_vpc.go
```diff
@@ -0,0 +1,573 @@
+package functional
```
ciaranRoche, line 1: Will this link to a test plan? 
If it does we should leave some developer docs here pointing to the test plan, if not we should add developer docs here outlining what resources we expect to find in a standalone vpc. I think it would help with debugging issues here. What do you think?

aidenkeating: This will/should be linked to a test plan ID, see comment above

test/functional/aws_standalone_vpc.go
```diff
@@ -0,0 +1,573 @@
...
+		peeringConnError: []string{},
+	}
+
+	// filter the peering connections by name and integreatly tag
+	peeringConn, err := session.DescribeVpcPeeringConnections(&ec2.DescribeVpcPeeringConnectionsInput{
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("tag:Name"),
+				Values: []*string{aws.String(peeringTagName)},
+			},
```
aidenkeating, line 423: As this tag is basically just a display name, I think we can ignore it in the test. Our provision logic doesn't depend on it, and it could easily change in the future

dimitraz: Ok, does that apply to all of the other resources as well then?

aidenkeating: Yes, I think so. I'd be a bit concerned that our tests become so exact that we can't make metadata updates without having to update an e2e test in RHMI. It's nothing major, but I don't think these are worth testing.

dimitraz: Sure, makes sense

test/functional/aws_standalone_vpc.go
```diff
@@ -0,0 +1,573 @@
...
+	err = verifyRdsSubnetGroups(rdsSvc, name, subnetIDs)
+	testErrors.rdsSubnetGroupsError = err.(*networkConfigTestError).rdsSubnetGroupsError
+
+	// verify elasticache subnet groups
+	cacheSvc := elasticache.New(session)
+	err = verifyCacheSubnetGroups(cacheSvc, name, subnetIDs)
+	testErrors.rdsSubnetGroupsError = err.(*networkConfigTestError).rdsSubnetGroupsError
+
+	// peering connection and route tables
+	if vpc == nil {
```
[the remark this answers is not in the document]
dimitraz: Sure, good point

psturc:
👀

test/functional/tests.go
psturc: @dimitraz I'll sent a PR for the test case and ping you on it

test/functional/aws_standalone_vpc.go
```diff
@@ -0,0 +1,582 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/integr8ly/integreatly-operator/test/common"
+)
+
+var (
+	resourceType    = "_network"
+	tier            = "development"
```
psturc, line 31: Is the `development` tier valid for BYOC clusters as well?

dimitraz: No good catch, this should be changed!

test/functional/aws_standalone_vpc.go
```diff
@@ -0,0 +1,582 @@
...
+	// verify standalone vpc route table
+	err = verifyStandaloneRouteTable(ec2Sess, clusterTag, conn)
+	testErrors.standaloneRouteTableError = err.(*networkConfigTestError).standaloneRouteTableError
+
+	// verify cluster route table
+	err = verifyClusterRouteTables(ec2Sess, clusterTag, expectedCidr, conn)
+	testErrors.clusterRouteTablesError = err.(*networkConfigTestError).clusterRouteTablesError
+
+	// if any error was found, fail the test
+	if testErrors.isValid() {
```
darahayes, line 214: nit: `isValid` is a potentially confusing in the context of running tests where we want to not have errors. Maybe `hasErrors` or something similar would be a bit more obvious?

test/functional/aws_standalone_vpc.go
```diff
@@ -0,0 +1,582 @@
...
+	subnets, err := verifySubnets(ec2Sess, clusterTag, expectedCidr)
+	testErrors.subnetsError = err.(*networkConfigTestError).subnetsError
+
+	// verify security groups
+	err = verifySecurityGroup(ec2Sess, clusterTag)
+	testErrors.securityGroupError = err.(*networkConfigTestError).securityGroupError
+
+	// we have to manually construct the subnet group names for rds and elasticache,
+	// since tag filtering isnt currently available
+	name := resources.ShortenString(fmt.Sprintf("%s-%s", clusterTag, "subnet-group"), 40)
```
darahayes, line 183: nit: `name` is a bit generic

test/functional/aws_standalone_vpc.go
dimitraz: Sure, let me change this :+1:

test/functional/aws_standalone_vpc.go
dimitraz: Good point :+1:

test/functional/aws_standalone_vpc.go
```diff
@@ -0,0 +1,582 @@
...
+		},
+	})
+	if err != nil {
+		newErr.vpcError = append(newErr.vpcError, fmt.Errorf("could not find vpc: %v", err))
+		return nil, newErr
+	}
+
+	// only one vpc is expected
+	vpcs := describeVpcs.Vpcs
+	if len(vpcs) != 1 {
```
psturc, line 241: Will there be always only 1 vpc present, even if we have multiple clusters in the same AWS account?

dimitraz: It should be, since we're filtering by the cluster tag here - https://github.com/integr8ly/integreatly-operator/pull/937/files/850cece6bbca39fe59c8eea1d0be9ec7fdf5d90d#diff-d235060cf6f899b2b7562835f9853f33R230

test/functional/tests.go
psturc, line 10:
```suggestion
		{Description: "A25 - Verify standalone RHMI VPC exists and is configured properly", Test: TestStandaloneVPCExists},
```

dimitraz:
Prow e2e test failure:
```   
could not run steps: step e2e failed: failed to acquire lease: resources not found
time="2020-07-02T16:12:37Z" level=info 
msg="Reporting job state \"failed\" with reason \"executing_graph:step_failed:utilizing_lease:acquiring_lease\""
```
Going to retry 

/retest

dimitraz:
Closing as these changes are contained in https://github.com/integr8ly/integreatly-operator/pull/946

PR #945, opened 2020-07-02
CVE-2019-5736: update runc to v1.0.0-rc91

pb82:
# Description

fixes CVE-2019-5736 by adding a replace rule for affected runc versions

## Type of change


MStokluska:
I've checked it and it looks good to be merged after switching to version rc8.

PR #970, opened 2020-07-08
INTLY-8358-Update RHMI operator to uninstall based on label on namespace

CathalOConnorRH:
# Description

## Type of change

## Checklist

## Validate
1. Install from this branch and wait for installation to be complete. 
2. Set uninstall label by using the command 
`oc label namespace redhat-rhmi-operator api.openshift.com/addon-rhmi-operator-delete=true
`
3. Verify the RHMI CR has a deletion timestamp and that the RHMI operator uninstalls as normal.

### DIFF
--- a/pkg/controller/add_uninstall.go
+++ b/pkg/controller/add_uninstall.go
@@ -1,0 +1,26 @@
+/*
+Copyright YEAR Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


Boomatang:
I have tried this for both local install and catalog source install. After the installation had completed and during the installation process if the label was applied the rhmi cr would have the deletion time stamp added and all resources removed. This looks good to me and the behaviour was as expected. 

Once this is merged, @CathalOConnorRH do you think it would be a good idea to share the "adding of the label to delete" method with the other teams so they are aware it can be done. And in the hope it may highlight any future changes that may happen.

**NOTE** I did not check this one a BYOC

PR #980, opened 2020-07-09
Remove IntegreatlyAnnotations setting products Subscriptions objects

miguelsorianod:
# Description

This PR removes IntegreatlyOwnerAnnotations from products Subscription objects.

Explanation/Motivation:

In marketplace.Manager type the method InstallOperator currently adds the IntegreatlyOwnerAnnotations to
the Subscription that is created for each product. It is created using an Owner provided as a parameter of the
InstallOperator method.

The InstallOperator method is called on each of the products reconciler indirectly via calling the resources.Reconciler ReconcileSubscription method.
The Owner that is sent is the 'Namespace' K8s object belonging to the the product's namespace (which sometimes is the same as the operator's namespace like in the case of
the cloudresources product).

This means that the following annotations are set to the product's Subscription k8s object's object that is created:
"integreatly-namespace": <NamespaceOfTheProductNamespaceK8sObject>
"integreatly-name": <NameOfTheProductNamespaceK8sObject>

Looking at the code for references to this annotations (IntegreatlyOwnerName and IntegreatlyOwnerNamespace constants in Go code), If I have understood correctly, I've seen that there's a custom event handler named EnqueueIntegreatlyOwner in pkg/controller/installation/enqueueIntegreatlyOwner.go.
This EventHandler is used as a Watch for custom Informers that are created in the InstallationController. The custom informers
created are created for CRDs that are specified in each product's Configuration through the GetWatchableCRDs method.
I haven't seen any Subscription object defined in any of the product's Configuration GetWatchableCRDs method.
I have also looked for Watch calls in existing controllers in the operator and the only one I've found that contains a Watch to a Subscription object type is the SubscriptionController
but that controller does not use the EnqueueIntegreatlyOwner custom event handler.
The EnqueueIntegreatlyOwner seems to be used to reenqueue changes on the objects provided by GetWatchableCRDs that contain the IntegreatlyOwnerAnnotations using the name and namespace values of those annotations as a reconciliation Event.

All of this make makes me think this added annotations to the Subscription objects are 'dead code'. I hope I haven't missed any other part of the code.

Because of this I've removed the annotations setting on the Subscriptions.

Also, I've noticed the 'owner' parameter of the InstallOperator method is only used to set this annotations so we could go further
and delete that parameter too. Doing so would require modifying the MarketplaceInterface signature so it would mean changing all calls to that method too for all the products so for the moment I haven't done it until I have confirmation that those annotations in Subscriptions are indeed dead code and there's no point in having them, and in affirmative case whether we want to go further and directly remove the Owner parameter from the InstallOperator signature.

What do you think about all of this? Are my findings/assumptions are correct and the change desirable? If so, do you want me to go further and remove the owner parameter from the InstallOperator method signature updating all the calls to it?

Thank you.

## Type of change

## Checklist

### DIFF
--- a/pkg/products/fuse/reconciler.go
+++ b/pkg/products/fuse/reconciler.go
@@ -483,12 +483,6 @@ func preUpgradeBackupExecutor(rhmi *integreatlyv1alpha1.RHMI) backup.BackupExecu
 }
 
 func (r *Reconciler) reconcileSubscription(ctx context.Context, serverClient k8sclient.Client, inst *integreatlyv1alpha1.RHMI, productNamespace string, operatorNamespace string) (integreatlyv1alpha1.StatusPhase, error) {
-	productNamespaceObj, err := resources.GetNS(ctx, productNamespace, serverClient)
-	if err != nil {
-		events.HandleError(r.recorder, inst, integreatlyv1alpha1.PhaseFailed, fmt.Sprintf("Failed to retrieve %s namespace", productNamespace), err)
-		return integreatlyv1alpha1.PhaseFailed, err
-	}
-
 	target := marketplace.Target{
 		Pkg:       constants.FuseSubscriptionName,
 		Namespace: operatorNamespace,
@@ -502,7 +496,6 @@ func (r *Reconciler) reconcileSubscription(ctx context.Context, serverClient k8s
 	)
 	return r.Reconciler.ReconcileSubscription(
 		ctx,
-		productNamespaceObj,
 		target,
 		[]string{productNamespace},
 		preUpgradeBackupExecutor(inst),
--- a/pkg/products/amqonline/reconciler.go
+++ b/pkg/products/amqonline/reconciler.go
@@ -600,12 +600,6 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation 
 }
 
 func (r *Reconciler) reconcileSubscription(ctx context.Context, serverClient k8sclient.Client, inst *integreatlyv1alpha1.RHMI, productNamespace string, operatorNamespace string) (integreatlyv1alpha1.StatusPhase, error) {
-	productNamespaceObj, err := resources.GetNS(ctx, productNamespace, serverClient)
-	if err != nil {
-		events.HandleError(r.recorder, inst, integreatlyv1alpha1.PhaseFailed, fmt.Sprintf("Failed to retrieve %s namespace", productNamespace), err)
-		return integreatlyv1alpha1.PhaseFailed, err
-	}
-
 	target := marketplace.Target{
 		Pkg:       constants.AMQOnlineSubscriptionName,
 		Namespace: operatorNamespace,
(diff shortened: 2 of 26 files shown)


philbrookes:
@jjaferson can you take a look at this PR, when you get a minute, please?

briangallagher:
Hi @miguelsorianod , yes, your logic and assumptions are correct. 
You can proceed with removing the owner param and I can approve once the code freeze is over.

Thanks.

miguelsorianod:
:+1: I'll remove it from the InstallOperator signature and refactor all calls to it too then.

miguelsorianod:
@briangallagher updated code removing the owner attribute in the InstalllOperator method and all other methods signatures that sent the owner for the only purpose of sending it to the InstallOperator method at the end.

pkg/products/fuse/reconciler.go
```diff
@@ -483,7 +483,7 @@ func preUpgradeBackupExecutor(rhmi *integreatlyv1alpha1.RHMI) backup.BackupExecu
 }
 
 func (r *Reconciler) reconcileSubscription(ctx context.Context, serverClient k8sclient.Client, inst *integreatlyv1alpha1.RHMI, productNamespace string, operatorNamespace string) (integreatlyv1alpha1.StatusPhase, error) {
```
miguelsorianod, line 485: @briangallagher  Notice changes in this method that also apply to other products reconcilers.

Notice how the productNamespaceObj is not needed anymore so we replaced it with a _.

I'm/was hesitant to completely remove the code that performs resources.GetNS because I'm not sure whether this was being done due to owner being sent or because we actually want to check that that namespace exists. Being conservative I haven't removed the GetNS call.

What do you think?

briangallagher: @miguelsorianod I think its fine to remove resources.GetNS. If we need to verify the namespace we can add an explicit function to do so. Everything else looks good

miguelsorianod: Done. I've removed the calls to resources.GetNS in the reconcileSubscription methods of the products reconcilers.
I've not moved the call anywhere else because I've seen in the Reconcile method in the products there are calls to r.ReconcileNamespace so I understand the check of the namespace is implicitly done there.

miguelsorianod:
@philbrookes e2e failed with an error that doesn't seem related to the PR:

```
2020/07/14 14:24:42 Ran for 7m14s
error: some steps failed:
  * could not run steps: step e2e failed: could not wait for template instance to be ready: failed to create objects: rolebindings.authorization.openshift.io "e2e-image-puller" already exists
pods "e2e" already exists
time="2020-07-14T14:24:42Z" level=info msg="Reporting job state \"failed\" with reason \"executing_graph:step_failed:utilizing_lease:executing_test:executing_template\"" 
```

Can you take a look at it?

briangallagher:
@miguelsorianod We have a sporadic issue there which I've created a [JIRA](https://issues.redhat.com/browse/INTLY-9479) for.  It should not block this change.

Incidentally, we have some flakiness with our automated testing which we are trying to address by initially cataloging all failures. More details are in this [JIRA](https://issues.redhat.com/browse/INTLY-7197).

miguelsorianod:
Can this be merged?

briangallagher:
> Can this be merged?

Yes, sorry Miguel

davidffrench:
/approve

Approving based on the review by @briangallagher . Thank you very much for the PR @miguelsorianod

PR #983, opened 2020-07-09
INTLY-9201

david-martin:
Add a limited permission to bypass OLM bug.
The important thing about this change is that the rules for the alertmanager-service-account is different is some way than the rules for the grafana-serviceaccount.
This ensures the generated names of the corresponding clusterRoles & clusterRoleBindings will be different.

Related OLM issue https://github.com/operator-framework/operator-lifecycle-manager/issues/1625

The granted additional permission is to `get` `subjectaccessreviews`,
which will allow the SA to read an individual subjectaccessreviews resource by id.
Although not ideal, this permission is fairly limited.

# Description
https://issues.redhat.com/browse/INTLY-9201

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist
- [ ] This change requires a documentation update 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added a test case that will be used to verify my changes 
- [ ] Verified independently on a cluster by reviewer

PR #987, opened 2020-07-10
workaround olm bug when sas have the exact same rules

pb82:
Similar to https://github.com/integr8ly/integreatly-operator/pull/983 but applies to the AMQ Online manifest. I had to grant `update` permissions because both sas already have get, list and watch.


matskiv:
`Throttling: Rate exceeded` again... :(

matskiv:
AMQ Online installation verified manually
Screenshot 2020-07-10 at 19 41 53

PR #988, opened 2020-07-10
Update time estimate in manual test b03

grdryn:


grdryn:
/cc @jjaferson @wei-lee 

It looks like you've both run this test recently, so maybe you could review the accuracy of this time estimation change?

PR #995, opened 2020-07-14
feat: alert for failed deletion of redis and postgres resources

darahayes:
# Description

Adds new alerts specifically for the failed deletion of Postgres and Redis CRs. This realistically should only happen on teardown of RHMI.

Is a follow on from this PR (still in review) https://github.com/integr8ly/cloud-resource-operator/pull/159

Still todo:

* Create new SOPs for handling failed deletion of Redis and Postgres instances. cc @aidenkeating 

## Verification Steps

Setup:
* Provision the integreatly operator, ensuring that v0.18.0 or the master branch of the cloud-resources-operator is used. This is needed because the cro now exposes some new metrics when AWS resources are being deleted.
* ensure that `useClusterStorage` is `'false'` in the `rhmi` CR.

Verify the alerts were created:

* run the command `oc get prometheusrule -n redhat-rhmi-operator | grep resource-deletion-status-phase-failed-rule`. You should see there are a bunch of prometheus rules for the failed deletion of postgres instances for each product and then the two redis instances that 3scale uses.

Trigger a failed deletion and observe the alerts:

* in the `redhat-rhmi-operator` namespace, find the aws strategies configmap and under the `createStrategy` JSON object for both `'postgres'` and `'redis'` there should be `region` set to `""`. Change the region for both postgres and redis to some dummy value like `"abc123"`. This will cause all AWS operations performed by CRO to fail.
* In the `redhat-rhmi-operator` namespace `oc get postgres` should give you a list of all the various postgres instances being used by the RHMI products. Delete one of those postgres CRs.
* In the alerts tab in the middleware monitoring prometheus you should be able to find the PostgresDeletionFailed alert for the specific product whose postgres instance you tried to delete. It should show up as pending and eventually after about 5 minutes it should start firing. You can also search in the metrics tab for `cro_postgres_deletion_timestamp` and you should get back some metrics.
* repeat the same process for one of the threescale redis instances and observe that there are some `cro_redis_deletion_timestamp` metrics and also observe the alerts moving from a pending state, eventually into a firing state.

## Type of change

- [ x ] New feature (non-breaking change which adds functionality)

## Checklist


aidenkeating:
/lgtm

This was verified in a group. Waiting on SOP update also

pkg/resources/metrics.go
```diff
@@ -190,6 +220,32 @@ func CreateRedisResourceStatusPhaseFailedAlert(ctx context.Context, client k8scl
...
+		return nil, nil
+	}
+	productName := cr.Labels["productName"]
+	redisCRName := strings.Title(strings.Replace(cr.Name, "redis-example-rhmi", "", -1))
+	alertName := redisCRName + "RedisResourceDeletionStatusPhaseFailed"
+	ruleName := fmt.Sprintf("resource-deletion-status-phase-failed-rule-%s", cr.Name)
+	alertExp := intstr.FromString(
+		fmt.Sprintf("%s{exported_namespace='%s',resourceID='%s',productName='%s',statusPhase='failed'}", DefaultRedisDeletionMetricName, cr.Namespace, cr.Name, productName),
+	)
+	alertDescription := fmt.Sprintf("The deletion of the Redus instance has been failing longer than %s. Redis Custom Resource: %s in namespace %s (strategy: %s) for product: %s", alertFor5Mins, cr.Name, cr.Namespace, cr.Status.Strategy, productName)
```
psturc, line 236:
```suggestion
	alertDescription := fmt.Sprintf("The deletion of the Redis instance has been failing longer than %s. Redis Custom Resource: %s in namespace %s (strategy: %s) for product: %s", alertFor5Mins, cr.Name, cr.Namespace, cr.Status.Strategy, productName)
```

psturc:
@darahayes I didn't know how to get CRO v0.18.0 running otherwise than by updating its version in `rhmi_types.go` and adding proper manifests, so I was verifying this PR with [these changes](https://github.com/psturc/integreatly-operator/commit/a5aa3572c6011176ca5e05d9d82a80134ba86638) (should this PR be updated with cro version bump as well?).

Regarding verification: everything worked as expected
Just one small typo, otherwise LGTM

darahayes:
@psturc thanks for the verification. With regards to the manifests, I'm not totally sure what the process is in terms of updating the integreatly-operator to use the latest version of CRO. @aidenkeating should I be including @psturc's changes to the manifests in this PR?

aidenkeating:
@darahayes If you want to include the actual bump in this PR also, the changes from https://github.com/psturc/integreatly-operator/commit/a5aa3572c6011176ca5e05d9d82a80134ba86638 make sense here. The process would be:

- Remove 0.16.1 from https://github.com/integr8ly/integreatly-operator/tree/master/manifests/integreatly-cloud-resources
- Copy in 0.18.0 from CRO and ensure it states that it replaces 0.17.0

We'll have to run a final upgrade verification also

darahayes:
@aidenkeating I've bumped the CRO version as expected. Maybe we can go through a final verification tomorrow!

aidenkeating:
Might need to remove this line from 0.17.0 [1] similar to how we removed the `replaces` line in `0.16.1` when `0.17.0` became the latest version [2].

[1] https://github.com/integr8ly/integreatly-operator/blob/master/manifests/integreatly-cloud-resources/0.17.0/cloud-resources.v0.17.0.clusterserviceversion.yaml#L344
[2] https://github.com/integr8ly/integreatly-operator/commit/c22f0898530ff64e3c650abc3ea8398cd16a7dd5#diff-7870ac29182d861b1f50b0787b3303beL344

PR #999, opened 2020-07-14
Test Case - divides test cases for rhmi alerts in two

jjaferson:
# Description

Divides test case for the rhmi installation alerts in two tests:

1) C11  is a test case to verify `RHMIOperatorInstallDelayed` alert and will be a destructive test

2) C12 is a test case to verify `RHMIInstallationControllerIsNotReconciling` and `RHMIInstallationControllerStoppedReconciling` alerts


test-cases/tests/alerts/c12-verify-rhmi-reconciliation-alerts-are-firing.md
```diff
@@ -0,0 +1,23 @@
...
+
+Verify that RHMI operator alerts are in place and firing
+
+More info: <https://issues.redhat.com/browse/INTLY-7395>
+
+## Steps
+
+1. Verify RHMIInstallationControllerIsNotReconciling and RHMIInstallationControllerStoppedReconciling alerts are present and firing:
+2. Open OpenShift console in your browser and login as admin
+3. Login as admin
```
matskiv, line 18: same as above, duplicate step :)

matskiv:
C12 test case doesn't mention that I also need to scale down to see results for prometheus query.
Is that needed? Because I didn't get any results first time I tried it. 

But reading the graph, it seems like I tried it right after successful reconcile. So waiting a minute would do the trick?

jjaferson:
> C12 test case doesn't mention that I also need to scale down to see results for prometheus query.

You don't need to scale down the rhmi operator pod to see the results, the operator needs to be running for returning the metric `rhmi_status{stage="complete"}` 

I added a line in C12 mentioning that the _query should return, if not wait another minute and try again_, it should cover the case that the reconcile has just run successfully.

matskiv:
@jjaferson cool. Thank you!

/lgtm
/approve

PR #938, opened 2020-06-29
[INTLY 8454] Update alerting mechanism test to accept multiple alerts

obrienrobert:
## Description
https://issues.redhat.com/browse/INTLY-8454

Updated the `TestIntegreatlyAlertsMechanism` test to allow the alerting mechanism for multiple Fuse alerts to be tested, as opposed to a single Fuse alert.

The test checks that the `FuseOnlineSyndesisUIInstanceDown` and `RHMIFuseOnlineSyndesisUiServiceEndpointDown` Fuse alerts go through the following process: `not firing before test execution -> pending -> firing -> not firing`

**This PR should be tested using both a 4.3 and 4.4 OS4 cluster**

## Verification Steps

- Install via this branch
- Run the `TestIntegreatlyAlertsMechanism` locally against your cluster: `go clean -testcache && go test -v ./test/functional -run=//^C03 -timeout=80m`
- Ensure the that test passes and that the logging is clear and concise
- During the execution of the test, when the alerts are in a pending/firing state, open a new tab/terminal and execute the test again. Verify that the test fails due to the alerts already firing, and that this is reflected clearly in the logs.

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist
- [ ] This change requires a documentation update 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added a test case that will be used to verify my changes 
- [x] Verified independently on a cluster by reviewer

PR #1004, opened 2020-07-14
fixed alert-mechanism test

jhellar:
# Description

There was a bug in alerts-mechanism test, that sometimes caused deployments not being scaled back up, so the test failed. The order of deferred calls was not right. Also there could be a situation where calls to scale the deployments up and down were called simultaneously.

The fix makes sure that once it is verified that alert is firing, goroutines that scale deployments down continuously, are canceled, then it waits for the last call to scale down deployment to finish, then scale everything back up.

### DIFF
--- a/test/common/alerts_mechanism.go
+++ b/test/common/alerts_mechanism.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"sync"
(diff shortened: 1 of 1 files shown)


jhellar:
This is already fixed with https://github.com/integr8ly/integreatly-operator/pull/938 -> closing this PR.

PR #1007, opened 2020-07-15
detect profile type for the integration test based on created CR

akoserwal:
# Description

INTLY-8771

Adding a utility method to detect the profile based on the CR instance. It can help in sorting the test cases based on profiles.

## Type of change


matskiv:
/ok-to-test
/lgtm
/approve

Note: we are still in code freeze, so this PR will not be merged until the code freeze ends.

PR #1011, opened 2020-07-15
moving self-managed installationType upstream from RHI

jeremyary:
# Description
Introduction of self-managed installationType with matching (un)install stages

## Type of change


jeremyary:
Hey @philbrookes, here's our base PR for self-managed which will allow us to open new PR's for both Service Registry inclusion and our e2e testing changes so that we can begin working towards INTLY-8771 (test multiple profiles) as well.

philbrookes:
/lgtm

@maleck13 @matskiv @pb82 any thoughts on this?

jeremyary:
updated for format fails

miguelsorianod:
Hi @jeremyary,

I don't see changes related to test profiles classification. Is another PR intended only for that? Keep in mind that service-registry PR probably will have changes related to that.
The changes I'm talking about are:
https://github.com/redhat-integration/rhi-operator/pull/26

And maybe https://github.com/redhat-integration/rhi-operator/pull/34/ but not completely sure about this one as it seems to contain references to apicurio registry.

What do you think?

jeremyary:
> Is another PR intended only for that? 

Exactly that. I put together a local branch with all of it yesterday, but after looking at it for a bit, I got a strong feeling there's going to be some discussion/rework around it to align with RHMI's goals for multiple profiles, so I thought we'd get this through in a shorter term.

matskiv:
LGTM

PR #1014, opened 2020-07-16
Update PV usage alerts to match upstream kubernetes-mixin

grdryn:
# Description

JIRA: https://issues.redhat.com/browse/INTLY-9432

## Type of change

## Checklist


pkg/products/monitoring/prometheusRules.go
```diff
@@ -164,22 +164,22 @@ func (r *Reconciler) newAlertsReconciler() resources.AlertReconciler {
 						Labels: map[string]string{"severity": "warning"},
 					},
 					{
-						Alert: "PVCStorageWillFillIn4Days",
+						Alert: "KubePersistentVolumeFillingUp",
```
HubertStefanski, line 167: Should this and the alert on line `177` be called the same? should we maybe change this one to `KubePersistentVolumeFillingUpPredictiveLongRun` or something along those lines? to have a clear way to distinguish their purpose from their name? I realise that this one is a `warning` level and the other one is `critical` but since they're queries are different then maybe their names should be as well? wdyt @grdryn ?

grdryn: Good question. I think it's fine that they have the same name. The alerts come from upstream [here](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/alerts/storage_alerts.libsonnet). The message annotation will give more clarity than a different name ever could IMHO.

pkg/products/monitoring/prometheusRules.go
```diff
@@ -164,22 +164,22 @@ func (r *Reconciler) newAlertsReconciler() resources.AlertReconciler {
...
-						Expr:   intstr.FromString("(predict_linear(kubelet_volume_stats_available_bytes{job='kubelet'}[6h], 4 * 24 * 3600) <= 0 AND kubelet_volume_stats_available_bytes{job='kubelet'} / kubelet_volume_stats_capacity_bytes{job='kubelet'} < 0.25 )* on(namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'}"),
+						Expr:   intstr.FromString("(predict_linear(kubelet_volume_stats_available_bytes{job='kubelet'}[6h], 4 * 24 * 3600) <= 0 and kubelet_volume_stats_available_bytes{job='kubelet'} / kubelet_volume_stats_capacity_bytes{job='kubelet'} < 0.15) * on(namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key='middleware'}"),
 						For:    "15m",
 						Labels: map[string]string{"severity": "warning"},
 					},
 					{
-						Alert: "PVCStorageWillFillIn4Hours",
+						Alert: "KubePersistentVolumeFillingUp",
 						Annotations: map[string]string{
 							"sop_url": "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/2.x/alerts/pvc_storage.asciidoc#pvcstoragewillfillin4hours",
```
david-martin, line 179: @grdryn Would it make sense to update the sop content as well with the new alert name(s)?
Perhaps have both alert names in the sop to allow for a transitionary period where it could be either

grdryn: I've created https://github.com/RHCloudServices/integreatly-help/pull/624 now with a new section in that SOP file, and updated the link here. Please take a look! (it's mostly a copy of the previous troubleshooting steps

grdryn:
@david-martin this is ready for review again if you're happy with it

PR #1018, opened 2020-07-17
update n07 test case

aidenkeating:
# Description

the current n07 test case is not correct.

update it to ensure that we clarify what we are testing for:

- that cro has lined up maintenance correctly
- forcing the maintenance in aws through a configmap change
- ensuring aws performed the maintenance successfully

## Type of change

## Checklist


aidenkeating:
@psturc @pawelpaszki Mind taking a look? I realise this case is basically being updated towards the end of the only test cycle it's really relevant for. But I still think an update makes sense for context/to avoid confusion

psturc:
Just need to run `npm run prettier` to be able to merge.
Otherwise LGTM

psturc:
Error from the test-cases-lint job:
```
warn: i10-align-docs-with-changes-to-Solution-Explorer-ui-text.md should be renamed to i10-align-docs-with-changes-to-solution-explorer-ui-text.md
error: some files are not named correctly, use --write to rename all of them
```
apparently someone forgot to run `npm run rename` in their PR and merged it anyway... :/

mikenairn:
@aidenkeating @psturc Should this not be in the 2.5 branch? I don't suppose it really matters, but it fixes the test cases lint job so was thinking of cherry picking it back unless there is some reason we wouldn't want to do that?

PR #1022, opened 2020-07-20
Fetch the pager duty secret that the Pager duty operator creates

Rajagopalan-Ranganathan:
# Description
JIRA Link:  https://issues.redhat.com/browse/INTLY-4885

Check first for the secret created by the pagerduty operator, if
exists then use that, else consider the secret created by CSSRE
and use it.

## Type of change

## Checklist

### DIFF
--- a/pkg/products/monitoring/reconciler.go
+++ b/pkg/products/monitoring/reconciler.go
@@ -635,13 +635,10 @@ func (r *Reconciler) reconcileAlertManagerConfigSecret(ctx context.Context, serv
 		logrus.Warnf("could not obtain smtp credentials secret: %v", err)
 	}
 
-	// handle pagerduty credentials
-	pagerdutySecret := &corev1.Secret{}
-	if err := serverClient.Get(ctx, types.NamespacedName{Name: r.installation.Spec.PagerDutySecret, Namespace: rhmiOperatorNs}, pagerdutySecret); err != nil {
-		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("could not obtain pagerduty credentials secret: %w", err)
-	}
-	if len(pagerdutySecret.Data["serviceKey"]) == 0 {
-		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("serviceKey is undefined in pager duty secret")
+	//Get pagerduty credentials
+	pagerDutySecret, err := r.getPagerDutySecret(ctx, serverClient)
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, err
 	}
 
 	// handle dead mans snitch credentials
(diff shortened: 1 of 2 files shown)


pkg/products/monitoring/reconciler_test.go
```diff
@@ -376,7 +376,9 @@ func TestReconciler_fullReconcile(t *testing.T) {
 		{
 			Name:           "test successful reconcile",
 			ExpectedStatus: integreatlyv1alpha1.PhaseCompleted,
-			FakeClient:     moqclient.NewSigsClientMoqWithScheme(scheme, ns, operatorNS, federationNs, grafanadatasourcesecret, installation, smtpSecret, pagerdutySecret, dmsSecret, alertmanagerRoute),
+			FakeClient: moqclient.NewSigsClientMoqWithScheme(scheme, ns, operatorNS, federationNs,
+				grafanadatasourcesecret, installation, smtpSecret, pagerdutySecret,
+				dmsSecret, alertmanagerRoute),
```
grdryn, line 381: Interested in your opinion between that style of wrapping versus the following:

```suggestion
			FakeClient: moqclient.NewSigsClientMoqWithScheme(
				scheme,
				ns,
				operatorNS,
				federationNs,
				grafanadatasourcesecret,
				installation,
				smtpSecret,
				pagerdutySecret,
				dmsSecret,
				alertmanagerRoute,
			),
```

pkg/products/monitoring/reconciler.go
```diff
@@ -776,3 +773,30 @@ func (r *Reconciler) reconcileSubscription(ctx context.Context, serverClient k8s
...
+	err := serverClient.Get(ctx, types.NamespacedName{Name: r.installation.Spec.PagerDutySecret,
+		Namespace: r.installation.Namespace}, pagerdutySecret)
+
+	if err != nil {
+		return "", fmt.Errorf("could not obtain pagerduty credentials secret: %w", err)
+	}
+
+	if len(pagerdutySecret.Data["PAGERDUTY_KEY"]) != 0 {
+		secret = string(pagerdutySecret.Data["PAGERDUTY_KEY"])
+
```
grdryn, line 791: These trailing empty lines here and on L794 look a little odd to me. NBD though

pkg/products/monitoring/reconciler_test.go
Rajagopalan-Ranganathan: This style in my opinion refers to instantiation part (creating an object with initial parameters) rather than a function call for some people (may be most people :)) . What if the method takes another 10 parameters, then it would be scrolling a page vertically to see the parameters (the horizontal problem -- shifted vertically).

pkg/products/monitoring/reconciler.go
Rajagopalan-Ranganathan: Agreed and fixed.I had to google what "NBD" meant :)

pkg/products/monitoring/reconciler_test.go
grdryn: If the method expected another 10 parameters, I'd maybe suggest we have a different problem (it's already borderline absurd) :slightly_smiling_face: 

I'm happy with either option (or allowing the line to run long _where it might make sense_). :+1:

PR #1025, opened 2020-07-21
Refactoring grafana test cases

obrienrobert:
## Description
https://issues.redhat.com/browse/INTLY-6738

Both of the Grafana test cases are flaky, caused by an incorrect cookie is being sent in the GET request intermittently (~2-4% of the time - please see the JIRA for details). As a result, the test cases have been updated as such:

`TestGrafanaExternalRouteAccessible:` this test case has been refactored to check the the `/metrics` Grafana endpoint is accessible. The `/metrics` endpoint is accessable without the need to authenticate due to the following auth-skip in place: `- '-skip-auth-regex=^/metrics'`

`TestGrafanaExternalRouteDashboardExist:` test will now be skipped on failure, and the response from the Grafana GET request printed out to help with debugging. This is a temporary measure and a follow-up issue will address the flakiness of this test.

### DIFF
--- a/test/common/grafana_routes.go
+++ b/test/common/grafana_routes.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/http/httputil"
 	"testing"
 
 	"github.com/integr8ly/integreatly-operator/test/resources"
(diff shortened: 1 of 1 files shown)


obrienrobert:
@david-martin @matskiv @jhellar

david-martin:
/approve
this removes the cause of flakey behaviour for 1 test and marks another as skipped until that can be addressed (in https://issues.redhat.com/browse/INTLY-9480)

PR #1023, opened 2020-07-20
J03 - Add note to comment out unnecessary tests

HubertStefanski:
# Description
Added a note to comment out unnecessary tests, speeding up the time taken to complete this test, as well as a minor note reminding about the need for manual 3scale namespace restoration.

PR #1026, opened 2020-07-21
Added trepel to test-case owners

trepel:
# Description
Added 'trepel' (myself) to the test-case owners.

To verify I think it is enough to check there's no typo there.


trepel:
the test-cases-lint failed due to wrong filename, that is fixed by https://github.com/integr8ly/integreatly-operator/pull/1023 (not merged yet, hence the error)

b1zzu:
we will have to wait for the master to be unfrozen before fixing the test-cases-lint

PR #944, opened 2020-07-01
3scale update to 0.6.0

fh-automatron:
Update 3scale manifest to 0.6.0 

This is an automatically generated branch to track the '0.6.0' version of the '3scale' OLM manifests.
This branch will be automatically updated with any changes related to this version until it finally GAs. 

Changes required to fix installation issues should be applied to this branch following the usual [PR Workflow](https://github.com/integr8ly/delorean/blob/master/docs/integrealy-ci/prow-pr-workflow.md)

More details about the Early Warning System that generated this can be found [here](https://github.com/integr8ly/delorean/tree/master/docs/ews)

## Type of change

- [X] Product Update

## Checklist
- [ ] Older product manifests (N-2) have been removed
- [ ] Operator and Product Versions are updated correctly
- [ ] Go Modules are updated if required

PR #1031, opened 2020-07-22
INTLY-9099 Update configmap CSV to include CRDs from older manifests

CathalOConnorRH:
# Description
change to include older CRDs still being referenced only when there aren't newer versions available. 

This can be verified using the new version of 3scale from https://issues.redhat.com/browse/INTLY-9065 
and checking that the old crds are included with the new ones

## Type of change

## Checklist

### DIFF
--- a/pkg/resources/marketplace/manifestHelper.go
+++ b/pkg/resources/marketplace/manifestHelper.go
@@ -6,11 +6,16 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"reflect"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 
+	"github.com/Masterminds/semver"
 	"github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
+	apiextensionv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 )
 
 const (
@@ -36,7 +41,7 @@ func GenerateRegistryConfigMapFromManifest(manifestPackageName string) (map[stri
 		return nil, err
 	}
 
-	crdStringList, err := GetFilesFromManifestAsStringList(manifestDir, "^*.crd.yaml$", packageStringList)
+	crdStringList, err := GetCRDsFromManifestAsStringList(manifestDir, "^*.crd.yaml$", packageStringList, manifestPackageName)
 	if err != nil {
 		logrus.Fatalf("Error proccessing crds from %s with error: %s", manifestPackageName, err)
 		return nil, err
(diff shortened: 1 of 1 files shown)


mikenairn:
@CathalOConnorRH @philbrookes Let me know when you think this might be nearing ready to merge so we can test it against the 3scale 0.6.0 release branch first https://github.com/integr8ly/integreatly-operator/pull/944

CathalOConnorRH:
Hi @mikenairn I've made a small change and waiting on review. I've tested using 0.6.0, 0.5.4 and 0.5.2 manifests included. I've attached the configmap (renamed .txt to upload) generated here if you want to check over it. there are no errors in the registry pod (logs below) and the threescale scaling functional test works as expected. 

`time="2020-07-22T14:54:14Z" level=info msg="Using in-cluster kube client config"
time="2020-07-22T14:54:14Z" level=info msg="loading CRDs" configmap=registry-cm-redhat-rhmi-3scale-operator ns=redhat-rhmi-3scale-operator
time="2020-07-22T14:54:14Z" level=info msg="loading Bundles" configmap=registry-cm-redhat-rhmi-3scale-operator ns=redhat-rhmi-3scale-operator
time="2020-07-22T14:54:15Z" level=info msg="loading Packages" configmap=registry-cm-redhat-rhmi-3scale-operator ns=redhat-rhmi-3scale-operator
time="2020-07-22T14:54:15Z" level=info msg="serving registry" configMapName=registry-cm-redhat-rhmi-3scale-operator configMapNamespace=redhat-rhmi-3scale-operator port=50051`

[configmap-registry-cm-redhat-rhmi-3scale-operator.txt](https://github.com/integr8ly/integreatly-operator/files/4960724/configmap-registry-cm-redhat-rhmi-3scale-operator.txt)

pkg/resources/marketplace/manifestHelper.go
```diff
@@ -49,6 +74,113 @@ func GenerateRegistryConfigMapFromManifest(manifestPackageName string) (map[stri
...
+		var folders []string
+
+		e = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+			if info.IsDir() && info.Name() != manifestPackageName {
+				folders = append(folders, path)
+			}
+			return nil
+		})
+
+		ReverseSlice(folders)
```
Boomatang, line 94: If I understand what is going on here correctly. The result of the ReverseSlice would give us an order list going from large to small in name ie `[10.0.0, 9.4.1, 9.4.0]`. There maybe an issue with this. The rhsso is moving from the up stream version to down stream version. That would mean the folder name for the latest version is `7.4.1` which is replacing `9.0.3`. In the crds there are different operators referenced and that's how the version numbers can go back awards.

philbrookes: This is a valid concern, but I think it's such an unlikely event, that we would catch in Delorean in any case that I'd prefer to leave this as a known issue, and correct it if it occurs, rather than spending the time addressing it now.

Boomatang: Yap I agree with that. Its much better to have the known issue, than the unknown issue.

CathalOConnorRH: You point brought up a good point on the version numbers so I've updated to pull the version number from the package yaml file rather than the highest number of directory. We'll still have the known issue but we'll be adding the correct CRDs for the current version correctly now.

mikenairn:
The test here passed https://github.com/integr8ly/integreatly-operator/pull/1045, so from the POV of fixing the 3scale issue these changes are required for, it lgtm. Will leave it to @philbrookes and the core team to do the final lgtm and approval.

pkg/resources/marketplace/manifestHelper.go
```diff
@@ -6,18 +6,37 @@ import (
...
 	"github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
 )
 
 const (
 	manifestEnvVarKey  = "MANIFEST_DIR"
 	defaultManifestDir = "manifests"
 )
 
+type CrdConfig struct {
```
matskiv, line 23: Hmm, have you tried importing the CRD type from:
https://github.com/kubernetes/apiextensions-apiserver/blob/4f04d5ce70bb4470e352582ac2d039148b233cd1/pkg/apis/apiextensions/v1/types.go#L359 ?
I am fine with keeping this as is, just curious if there was something that prevented the import (version of the type maybe?).

pkg/resources/marketplace/manifestHelper.go
```diff
@@ -49,6 +68,128 @@ func GenerateRegistryConfigMapFromManifest(manifestPackageName string) (map[stri
...
+	if err != nil {
+		fmt.Printf("Error parsing YAML file: %s\n", err)
+	}
+}
+
+// ReverseSlice reverses the order of the folders list
+func ReverseSlice(data interface{}) {
+	value := reflect.ValueOf(data)
+	if value.Kind() != reflect.Slice {
+		panic(errors.New("data must be a slice type"))
```
matskiv, line 182: Wouldn't panic crash the whole operator?
I think we should return an error rather then panic.

pkg/resources/marketplace/manifestHelper.go
CathalOConnorRH: I had tried to use that but it wasn't working for me. I've just retried and it's working now so I'll switch and push a new update. I was getting an error when debugging which later turned out to be a vscode bug rather than code bug.

pkg/resources/marketplace/manifestHelper.go
CathalOConnorRH: Good point, I've updated that with a new version

matskiv:
Thanks for the changes 👍

PR #1034, opened 2020-07-22
INTLY-9315 - add note to make tester aware of alert firing time range

HubertStefanski:
# Description
Adding a note for future testers to make them aware of the `time range` function in grafana UI, for which if the time range is too high will not reflect the actual and current state of the alerts, possibly showing up false positives for testers.

## Type of change
[:heavy_check_mark: ] Test case


HubertStefanski:
/hold 

Waiting for test-cases-lint to be fixed

david-martin:
@HubertStefanski Should this close, given the changes in #1052 ?

HubertStefanski:
>@HubertStefanski Should this close, given the changes in #1052 ?

Yes, I was just about to do this :+1:

PR #1040, opened 2020-07-23
INTLY-8450 

MStokluska:
# Description
Addressing Jira: https://issues.redhat.com/browse/INTLY-8450 

## What has been done so far
1) New fields in RHMIConfig CRD: 
```
UpgradeAvailable:
availableAt: creationTimestamp
targetVersion: latest avaialable version of the RHMI Operator
```
2) Update status functionality was moved from the subscription controller into RHMIConfig controller
3) Update status function was edited so that it no longer requires installPlan - we used it only to get the creationTime stamp which is now set by the subscription controller and is available in the RHMIConfig.
4) Changed slightly the logic around the `if serviceAffecting` in the subscription controller `handleUpgrades` function
5) Functionality was tested ( although two more steps are left before this PR can be considered as lgtm)
6) RHMIConfig now reflects changes to the user set dates immediately and doesn't need to wait for the subscription reconciler to kick in.

## What needs to be done
1) As the `updateStatus` function has been changed, unit testing needs to be edited as well
2) Seems that I had issues with simulating a patch upgrade, it looks like the current install plan from version N1 to N2 isn't deleted when a catalogSource is pointing at version N1.1 replacing version N1 - There is either an issue with the logic around patching an upgrade ( deleting install plan and recreating it in case of patch upgrade) or I was doing something wrong. 

### Short:
- install RHMI Operator 2.5.0 via catalog source
- simulate an upgrade of the operator to version 2.6.0 
- change `waitForMaintenance= false` and `notBeforeDays=2` in RHMIConfig yaml file
- `upgrade` `scheduled` date should be updated immediately
- change `notBeforeDays=0` 
- upgarde of RHMI Operator should trigger 
- once upgrade is completed, RHMIConfig `upgradeAvailable` and `upgrade` should disappear. 

### Long
1) in integreatly operator repo run:
```
SEMVER=2.5.0 ./scripts/prepare-release.sh
```
2) Make sure to set your own repo name in Makefile https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L3
3) Run `make image/build/push`
4) Navigate to `/deploy/olm-catalog/integreatly-operator/2.5.0`
5) In `integreatlyoperator.clusterserviceversion` delete the `replaces` field and set the https://github.com/integr8ly/integreatly-operator/blob/master/deploy/olm-catalog/integreatly-operator/2.5.0/integreatly-operator.v2.5.0.clusterserviceversion.yaml#L44  and the https://github.com/integr8ly/integreatly-operator/blob/master/deploy/olm-catalog/integreatly-operator/2.5.0/integreatly-operator.v2.5.0.clusterserviceversion.yaml#L314 to your own repo ie: `quay.io/<your repo name>/integreatly-operator:v2.5.0`
6) Run the following commands: 
```
opm alpha bundle generate -d . --channels alpha --package integreatly --output-dir bundle --default alpha
```
```
docker build -f bundle.Dockerfile -t quay.io/<quay username>/integreatly-bundle:2.5.0 . 
```
```
docker push quay.io/<quay username>/integreatly-bundle:2.5.0
```
```
operator-sdk bundle validate quay.io/<quay username>/integreatly-bundle:2.5.0
```
```
opm index add \
    --bundles quay.io/<your org>/integreatly-bundle:2.5.0 \
    --build-tool docker \
    --tag quay.io/<your org>/integreatly-index:2.5.0
```
```
docker push quay.io/<your org>/integreatly-index:2.5.0
```
7) Make sure you are loggind in to correct cluster 
8) Create file `rhmi-operators.yaml` and paste the following contents:
```
kind: CatalogSource
metadata:
  name: rhmi-operators
spec:
  sourceType: grpc
  image: quay.io/<your org>/integreatly-index:2.5.0
```
9) Run `oc apply -f rhmi-operators.yaml`
10) Run 
```
make cluster/prepare/project && \
   make cluster/prepare/smtp && \
   make cluster/prepare/dms && \
   make cluster/prepare/pagerduty
```
11) Navigate to your cluster and in to the Operators Hub, under `openshift-marketplace` you should see RHMI Operator, install it under `rhmi-operator` namespace.
12) Back in your main operator repo run:
```
SEMVER=2.6.0 ./scripts/prepare-release.sh
```
13) Repeat step 2 and 3 but this time navigate to 2.6.0
14) Edit the `replaces` field https://github.com/integr8ly/integreatly-operator/blob/master/deploy/olm-catalog/integreatly-operator/2.5.0/integreatly-operator.v2.5.0.clusterserviceversion.yaml#L459 to `2.5.0` and both images to your own repo as in step 5.
15) Run the following commands:
```
opm alpha bundle generate -d . --channels alpha --package integreatly --output-dir bundle --default alpha
```
```
docker build -f bundle.Dockerfile -t quay.io/<quay username>/integreatly-bundle:2.6.0 . 
```
```
docker push quay.io/<quay username>/integreatly-bundle:2.6.0
```
```
operator-sdk bundle validate quay.io/<quay username>/integreatly-bundle:2.6.0
```
```
opm index add \
    --bundles quay.io/<your org>/integreatly-bundle:2.5.0 \
    --bundles quay.io/<your org>/integreatly-bundle:2.6.0 \
    --build-tool docker \
    --tag quay.io/<your org>/integreatly-index:2.6.0
```
```
docker push quay.io/<your org>/integreatly-index:2.6.0
```
16) Navigate to catalogSource CR under `openshift-marketplace` namespace and change the `  image: quay.io/<your org>/integreatly-index:2.5.0` to `  image: quay.io/<your org>/integreatly-index:2.6.0`
17) Navigate to installed operators and you should see upgrade being triggered soon.
18) Navigate to RHMIConfig CR in RHMI Operator
19) Change the `spec` `notBeforeDays:2` and `waitForMaintenance:false` - changes to RHMIConfig `upgrade scheduled` should happen immediately.
20) Change `notBeforeDays:0` and installation of new version of RHMI Operator should trigger.

## Type of change

## Checklist


pkg/controller/rhmiconfig/helpers/update_status.go
```diff
@@ -0,0 +1,101 @@
...
+	"time"
+
+	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
+
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const WINDOW = 6
+
+func UpdateStatus(ctx context.Context, client k8sclient.Client, config *integreatlyv1alpha1.RHMIConfig, targetVersion string) error {
```
sergioifg94, line 16:
```suggestion
func UpdateStatus(config *integreatlyv1alpha1.RHMIConfig) error {
```

I'd suggest to make this function simply update the status of the RHMIConfig in memory, as `client.Status().Update()` is already being called afterwards, to avoid calling it twice unnecessarily. We can get rid of the context and client parameters then.

You can safely get rid of the `targetVersion` parameter. If you notice you're setting the `UpgradeAvailable.TargetVersion` field with this value, but then passing `UpgradeAvailable.TargetVersion` as the parameter, effectively doing nothing.

pkg/controller/rhmiconfig/helpers/update_status.go
```diff
@@ -0,0 +1,101 @@
...
+			return err
+		}
+
+		config.Status.Maintenance.ApplyFrom = mtStart.Format("2-1-2006 15:04")
+		config.Status.Maintenance.Duration = strconv.Itoa(WINDOW) + "hrs"
+	}
+
+	config.Status.UpgradeAvailable.TargetVersion = targetVersion
+
+	client.Status().Update(ctx, config)
```
sergioifg94, line 30:
```suggestion
```

This can go as it's meant to be set by the subscription controller to trigger a reconciliation of the RHMIConfig controller on change.

pkg/controller/rhmiconfig/rhmiconfig_controller.go
```diff
@@ -113,6 +114,28 @@ func (r *ReconcileRHMIConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+			err = helpers.UpdateStatus(context.TODO(), r.client, rhmiConfig, rhmiConfig.Status.UpgradeAvailable.TargetVersion)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+
+			rhmiConfig.Annotations[integreatlyv1alpha1.RecalculateScheduleAnnotation] = ""
+			err = r.client.Update(context.TODO(), rhmiConfig)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
```
sergioifg94, line 129:
```suggestion
			err = helpers.UpdateStatus(context.TODO(), rhmiConfig)
			if err != nil {
				return reconcile.Result{}, err
			}

			rhmiConfig.Annotations[integreatlyv1alpha1.RecalculateScheduleAnnotation] = ""
			err = r.client.Update(context.TODO(), rhmiConfig)
			if err != nil {
				return reconcile.Result{}, err
			}
```

As a result of the previous suggestion

pkg/controller/rhmiconfig/helpers/update_status.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	// Update the upgrade status
+	config.Status.Upgrade = integreatlyv1alpha1.RHMIConfigStatusUpgrade{
+		Scheduled: &integreatlyv1alpha1.UpgradeSchedule{
+			For: upgradeSchedule.Format(integreatlyv1alpha1.DateFormat),
+		},
+	}
+
+	return client.Status().Update(ctx, config)
```
sergioifg94, line 55:
```suggestion
	return nil
```

As mentioned in the suggestion above, the `client.Status().Update()` call can be performed afterwards, so there's no need to call it in the `UpdateStatus` function

pkg/controller/subscription/subscription_controller.go
```diff
@@ -253,24 +244,27 @@ func (r *ReconcileSubscription) HandleUpgrades(ctx context.Context, rhmiSubscrip
...
-	if reschedule, ok := config.Annotations[v1alpha1.RecalculateScheduleAnnotation]; ok && reschedule == "true" {
-		err = rhmiConfigs.UpdateStatus(ctx, r.client, config, latestRHMIInstallPlan, rhmiSubscription.Status.CurrentCSV)
-		if err != nil {
+	if isServiceAffecting {
+
+		config.Status.UpgradeAvailable = &integreatlyv1alpha1.UpgradeAvailable{
+			TargetVersion: rhmiSubscription.Status.CurrentCSV,
+			AvailableAt:   latestRHMIInstallPlan.CreationTimestamp}
+
+		if err := r.client.Status().Update(context.TODO(), config); err != nil {
```
sergioifg94, line 253:
```suggestion
	if isServiceAffecting {
                newUpgradeAvailable := &integreatlyv1alpha1.UpgradeAvailable{
			TargetVersion: rhmiSubscription.Status.CurrentCSV,
			AvailableAt:   latestRHMIInstallPlan.CreationTimestamp
	        }
	        
	        if !reflect.DeepEquals(newUpgradeAvailable, config.Status.UpgradeAvailable {
	                config.Status.UpgradeAvailable = newUpgradeAvailable
		        if err := r.client.Status().Update(context.TODO(), config); err != nil {
		                return reconcile.Result(), err
		        }
		        return reconcile.Result{RequeueAfter: time.Second}
		}
```

I think there might be a race condition as the RHMIConfig controller (running concurrently) might have not updated the status when the verification for whether the upgrade must be approved or not is performed. I'd like to test this to ensure this, but if there is a race condition this suggestion might solve it. Update the UpgradeAvailable field and if it has changed, requeue the reconciliation after a second, giving the RHMIConfig controller enough time to reconcile the status.

pkg/controller/rhmiconfig/helpers/update_status.go
MStokluska: Thanks, @sergioifg94, your suggested changes around UpdateStatus make a lot of sense!

pkg/controller/rhmiconfig/helpers/update_status.go
MStokluska: Hey Sergio, thank you for very valuable comments. When these changes were applied it seems that the `upgrade` from the RHMIConfig is empty, and in my understanding, because we are not making API call to upgrade the RHMIConfig CR inside the updateStatus function anymore and returning a nil value, we are returning an empty upgrade value. I think the update.Status() should be somewhere within the UpdateStatus or should we return a config with this value updated and then do client.Update in the RHMIConfig reconciler?

pkg/controller/subscription/subscription_controller.go
MStokluska: Thanks Sergio, didn't realize until now that this could be happening! Thanks a lot for clarification!

pkg/controller/rhmiconfig/helpers/update_status.go
sergioifg94: Notice that the return type of the UpdateStatus function is `error`, so by returning `nil` we're confirming that the function was successful. We don't need to return the config because the function mutates the one that's passed as a parameter as a reference

pkg/controller/rhmiconfig/rhmiconfig_controller.go
```diff
@@ -113,6 +114,28 @@ func (r *ReconcileRHMIConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+		}
+	} else {
+		rhmiConfig.Status.Upgrade = integreatlyv1alpha1.RHMIConfigStatusUpgrade{}
+		err = r.client.Update(context.TODO(), rhmiConfig)
```
sergioifg94, line 133:
```suggestion
		rhmiConfig.Status.Upgrade = integreatlyv1alpha1.RHMIConfigStatusUpgrade{
		    Scheduled: nil,
		}
		err = r.client.Status().Update(context.TODO(), rhmiConfig)
```

Mainly need to call `.Status()`, otherwise it won't be updated

pkg/controller/subscription/subscription_controller.go
```diff
@@ -253,24 +245,36 @@ func (r *ReconcileSubscription) HandleUpgrades(ctx context.Context, rhmiSubscrip
...
+
+		newUpgradeAvailable := &integreatlyv1alpha1.UpgradeAvailable{
+			TargetVersion: rhmiSubscription.Status.CurrentCSV,
+			AvailableAt:   latestRHMIInstallPlan.CreationTimestamp,
 		}
 
-		config.Annotations[v1alpha1.RecalculateScheduleAnnotation] = ""
-		err = r.client.Update(ctx, config)
-		if err != nil {
+		if !reflect.DeepEqual(newUpgradeAvailable, config.Status.UpgradeAvailable) {
```
sergioifg94, line 255:
```suggestion
		if !reflect.DeepEqual(newUpgradeAvailable, config.Status.UpgradeAvailable) && !latestRHMIInstallPlan.Spec.Approved {
```

We have to add this condition, otherwise, once the upgrade is approved and the `UpgradeAvailable` field is set to empty, it will be set again in the next reconcilliation

pkg/controller/subscription/subscription_controller.go
```diff
@@ -253,24 +245,36 @@ func (r *ReconcileSubscription) HandleUpgrades(ctx context.Context, rhmiSubscrip
...
+
+			config.Status.UpgradeAvailable = newUpgradeAvailable
+			if err := r.client.Status().Update(context.TODO(), config); err != nil {
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{
+				Requeue:      true,
+				RequeueAfter: 10 * time.Second,
```
sergioifg94, line 263:
```suggestion
				RequeueAfter: 1 * time.Second,
```

1 second should be enough for the RHMIConfig controller to reconcile the status

pkg/controller/subscription/subscription_controller.go
```diff
@@ -253,24 +245,36 @@ func (r *ReconcileSubscription) HandleUpgrades(ctx context.Context, rhmiSubscrip
...
+			}, nil
+		}
+
+		if err := r.client.Status().Update(context.TODO(), config); err != nil {
 			return reconcile.Result{}, err
 		}
 
-	} else if isServiceAffecting && rhmiSubscription.Status.CurrentCSV != config.Status.TargetVersion {
-		err = rhmiConfigs.UpdateStatus(ctx, r.client, config, latestRHMIInstallPlan, rhmiSubscription.Status.CurrentCSV)
+		phase, err := r.webbappNotifier.NotifyUpgrade(config, latestRHMICSV.Spec.Version.String(), isServiceAffecting)
```
sergioifg94, line 271: This method call was copied but not moved so it's being called again outside the `if isServiceAffecting` block

pkg/controller/subscription/subscription_controller.go
```diff
@@ -253,24 +245,36 @@ func (r *ReconcileSubscription) HandleUpgrades(ctx context.Context, rhmiSubscrip
...
+			if err := r.client.Status().Update(context.TODO(), config); err != nil {
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{
+				Requeue:      true,
+				RequeueAfter: 10 * time.Second,
+			}, nil
+		}
+
+		if err := r.client.Status().Update(context.TODO(), config); err != nil {
```
sergioifg94, line 267:
```suggestion
```

The status update is being already called in the `if !reflect.DeepEqual(...)` block. There's no need to call it if the status is not modified, otherwise it'll be unnecessarily called

pkg/controller/rhmiconfig/rhmiconfig_controller.go
pb82: :+1:

pkg/controller/subscription/subscription_controller.go
pb82: :+1:

pkg/controller/rhmiconfig/rhmiconfig_controller.go
```diff
@@ -113,6 +114,28 @@ func (r *ReconcileRHMIConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+
+			rhmiConfig.Annotations[integreatlyv1alpha1.RecalculateScheduleAnnotation] = ""
+			err = r.client.Update(context.TODO(), rhmiConfig)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
```
sergioifg94, line 130:
```suggestion		
			err = helpers.UpdateStatus(context.TODO(), r.client, rhmiConfig)
			if err != nil {
				return reconcile.Result{}, err
			}

			rhmiConfig.Annotations[integreatlyv1alpha1.RecalculateScheduleAnnotation] = ""
			err = r.client.Update(context.TODO(), rhmiConfig)
			if err != nil {
				return reconcile.Result{}, err
			}		
```

I realized that this condition can go. There's no need for that annotation anymore as the schedule will only be recalculated when the UpgradeAvailable field is modified

pkg/controller/subscription/subscription_controller.go
```diff
@@ -253,23 +245,21 @@ func (r *ReconcileSubscription) HandleUpgrades(ctx context.Context, rhmiSubscrip
...
-			return reconcile.Result{}, err
+		if !reflect.DeepEqual(newUpgradeAvailable, config.Status.UpgradeAvailable) && !latestRHMIInstallPlan.Spec.Approved {
+			config.Status.UpgradeAvailable = newUpgradeAvailable
+			if err := r.client.Status().Update(context.TODO(), config); err != nil {
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{
+				Requeue:      true,
+				RequeueAfter: 10 * time.Second,
+			}, nil
```
sergioifg94, line 262: After verifying the patch upgrade recalculation I realized this could be simplified and another condition would be needed in order to avoid recalculation when a patch release is approved:

```go
if isServiceAffecting && !latestRHMIInstallPlan.Spec.Approved && config.Status.UpgradeAvailable == nil {
		config.Status.UpgradeAvailable = &integreatlyv1alpha1.UpgradeAvailable{
			TargetVersion: rhmiSubscription.Status.CurrentCSV,
			AvailableAt:   latestRHMIInstallPlan.CreationTimestamp,
		}
		if err := r.client.Status().Update(context.TODO(), config); err != nil {
		}
		return reconcile.Result{
			Requeue:      true,
			RequeueAfter: 10 * time.Second,
		}, nil
	}
```

Mainly, if the UpgradeAvailable field is already set, there's no need to update it. When the upgrade is approved this field will be set to `nil`, so subsequent upgrades will still be scheduled

pkg/controller/subscription/subscription_controller.go
```diff
@@ -289,6 +279,12 @@ func (r *ReconcileSubscription) HandleUpgrades(ctx context.Context, rhmiSubscrip
 	if !isServiceAffecting || canUpgradeNow {
 		eventRecorder := r.mgr.GetEventRecorderFor("RHMI Upgrade")
 
+		config.Status.UpgradeAvailable = nil
+
+		if err := r.client.Status().Update(context.TODO(), config); err != nil {
+			return reconcile.Result{}, err
+		}
+
```
sergioifg94, line 287:
```suggestion
		if config.Status.UpgradeAvailable != nil && config.Status.UpgradeAvailable.TargetVersion == rhmiSubscription.Status.CurrentCSV {
				config.Status.UpgradeAvailable = nil
				if err := r.client.Status().Update(context.TODO(), config); err != nil {
					return reconcile.Result{}, err
				}
		}
```

We only clear the UpgradeAvailable field when the upgrade that has been approved is the one set in the target version. Otherwise then approving a patch release we will lose the upgrade information from the next release

pkg/controller/subscription/rhmiConfigs/main.go
```diff
@@ -72,53 +72,6 @@ func CreateInstallPlan(ctx context.Context, rhmiSubscription *olmv1alpha1.Subscr
...
-	config.Status.Upgrade = integreatlyv1alpha1.RHMIConfigStatusUpgrade{
-		Scheduled: &integreatlyv1alpha1.UpgradeSchedule{
-			For: upgradeSchedule.Format(integreatlyv1alpha1.DateFormat),
-		},
-	}
-
-	return client.Status().Update(ctx, config)
-}
-
 func CanUpgradeNow(config *integreatlyv1alpha1.RHMIConfig, installation *integreatlyv1alpha1.RHMI) (bool, error) {
```
sergioifg94, line 75: I noticed there's a small possibility of a panic if the UpgradeAvailable field is set and this function is called before the RHMIConfig controller has calculated the schedule. I'd add this condition to return false if this is the case

```go
// If the upgrade schedule hasn't been calculated
if config.Status.Upgrade.Scheduled == nil {
	return false, nil
}
```

pkg/controller/rhmiconfig/rhmiconfig_controller.go
pb82: :+1:

sergioifg94:
Looking good!
/lgtm

PR #1043, opened 2020-07-23
Added test case for Sol Ex Settings banner

tonyxrmdavidson:
# Description
New test-case for sol ex settings

### DIFF
--- a/test-cases/tests/walkthroughs/g16-verify-settings-notifications-pop-up.md
+++ b/test-cases/tests/walkthroughs/g16-verify-settings-notifications-pop-up.md
@@ -16,5 +16,5 @@ Log into solution explorer in a private browser session as a user from the dedic
 4. Refresh the page and confirm the banner does not reappear.
 5. Close the private browser session.
 6. Open another private browser session and log into Solution Explorer as the same user.
-7. Return to the Solution Explorer settings page. 
+7. Return to the Solution Explorer settings page.
 8. Confirm the banner is again visible at the top of the settings page.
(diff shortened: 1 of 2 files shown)


test-cases/tests/walkthroughs/g16-verify-settings-notifications-pop-up.md
```diff
@@ -0,0 +1,19 @@
+---
+estimate: 15m
+tags:
```
trepel, line 3: should be `targets`, not `tags` I think

test-cases/tests/walkthroughs/g16-verify-settings-notifications-pop-up.md
```diff
@@ -0,0 +1,19 @@
...
+  - 2.6.0
+---
+
+# G16 - Verify settings notifications pop up
+
+Acceptance Criteria:
+
+When logged into solution explorer as a customer-admin
```
trepel, line 11: not `customer-admin`, but `dedicated-admin` ? Even more precise would be to say `a user from didicated-admin group` but that's probably an overkill

test-cases/tests/walkthroughs/g16-verify-settings-notifications-pop-up.md
tonyxrmdavidson: Yeah, you're correct.  Fixed it with new commit.

test-cases/tests/walkthroughs/g16-verify-settings-notifications-pop-up.md
tonyxrmdavidson: Agree, updated with commit

PR #1046, opened 2020-07-24
Use jq to extract the auth token from the delorean pull secret

mikenairn:
# Description
Use jq to extract the auth token from the delorean pull secret so it works regardless of the file formatting.

## Type of change

## Checklist

### DIFF
--- a/scripts/setup-delorean-pullsecret.sh
+++ b/scripts/setup-delorean-pullsecret.sh
@@ -21,8 +21,9 @@ if [ ! -f ${DELOREAN_PULLSECRET} ]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


StevenTobin:
Testing locally it looks like there's an extra decode step
```
$ jq -r '.auths|to_entries[0].value.auth' integreatly-delorean-auth.json | base64 --decode | awk -F : '{print $2}' | base64 --decode| awk -F : '{print $2}'

$ jq -r '.auths|to_entries[0].value.auth' integreatly-delorean-auth.json | base64 --decode | awk -F : '{print $2}'
<AN-ACCESS-KEY>
```

```
$jq -r '.auths|to_entries[0].value.auth' integreatly-delorean-auth.json | base64 --decode | awk -F : '{print $1}' | base64 --decode| awk -F : '{print $1}'
base64: invalid input
�{^����\�u�h��
$jq -r '.auths|to_entries[0].value.auth' integreatly-delorean-auth.json | base64 --decode | awk -F : '{print $1}'
<A-USERNAME>
```
```
jq --version
jq-1.6
```

PR #1052, opened 2020-07-24
INTLY-9315 - Refine critical SLO summary dashboard and fix non-instant alert state updates

HubertStefanski:
# Description
- Adding 28 day limit on the queries to adhere to the SLA, especially in %SLO panel, previously the SLO percentage was non-indicative of the actual state, potentially miscalculating the SLA breach.
- Add instant selector to fetch alert states as they change, rather than previous approach of waiting 15 minutes.
- Add default dashboard refresh to 10 seconds, previously no refresh period was set and the dashboard would only update on manual refresh.
- Change default time range to now-5m (previously now-15m) meaning that states only got updated after 15 minutes from the actual change, no longer an issue due to `instant : true`, (now-5m) is still needed to update line graphs, the instant field on that query will break the graph.

## JIRA
https://issues.redhat.com/browse/INTLY-9315

## Type of change

## Checklist

## Verification
- Install this branch on your cluster.
- Run [E04](https://issues.redhat.com/browse/INTLY-9343) and ensure that the dashboard reflects the state of the alerts as seen in Prometheus UI
- Alerts should be updated immediately
- Critical SLO summary graph and % should grab metrics from the last 28 days
- Use the admin secret on the cluster to get an admin view of the graphs


pkg/products/monitoring/dashboards/criticalSLOAlerts.go
```diff
@@ -225,6 +226,7 @@ const MonitoringGrafanaDBCriticalSLOAlertsJSON = `{
 			"targets": [{
 				"expr": "sum(ALERTS{severity='critical', alertstate='firing'}) or vector(0)",
 				"format": "time_series",
+				"instant": false,
```
HubertStefanski, line 229: Added in these instant fields even for places where they're not being used, I think it adds a bit more readability to the already busy Json, but I can take them out if we don't want them here.

HubertStefanski, line 229: Added in these instant fields even for places where they're not being used, I think it adds a bit more readability to the already busy Json, but I can take them out if we don't want them.

pkg/products/monitoring/dashboards/criticalSLOAlerts.go
```diff
@@ -4181,8 +4212,9 @@ const MonitoringGrafanaDBCriticalSLOAlertsJSON = `{
 			}
 		]
 	},
+	"refresh": "10s",
```
HubertStefanski, line 4215: Aforementioned default refresh, previously the dashboard wouldn't update automatically, now it does so every 10 seconds, looking for opinions on this time, should it be lower/higher?

pkg/products/monitoring/dashboards/criticalSLOAlerts.go
```diff
@@ -4181,8 +4212,9 @@ const MonitoringGrafanaDBCriticalSLOAlertsJSON = `{
...
 	"time": {
-		"from": "now-15m",
+		"from": "now-5m",
```
HubertStefanski, line 4217: From what I can tell, this is the default "poll rate" that the graphs will use to update themselves if no `instant` field is set, I might be wrong, just going off my observations.

pkg/products/monitoring/dashboards/criticalSLOAlerts.go
obrienrobert: Is this default value of `instant` false? If so, I'm just thinking if there is a need to add this to each `target` block. I would be thinking along the lines of why are we explicitly adding this variable with it's default value but not other doing the same for other possible values?

obrienrobert: Is the default value of `instant` false? Just thinking if we need to set it is so, as we would be explicitly setting the default value for a single variable, but not for other possible variables.

pkg/products/monitoring/dashboards/criticalSLOAlerts.go
obrienrobert: Was not having a `refresh` value set causing issues/problems? Or is this just more of an enhancement?

pkg/products/monitoring/dashboards/criticalSLOAlerts.go
HubertStefanski: Yes, the default value for instant is false, I get your point, in this case it's probably best to remove them if we're not using those, thanks for the feedback :+1:

pkg/products/monitoring/dashboards/criticalSLOAlerts.go
HubertStefanski: It wasn't causing problems with the metrics themselves right away, it just meant that the dashboard wouldn't be updated unless the refresh button was clicked, so this is a 50/50 bug fix/enhancement, most of the other dashboards have a default automatic refresh, so I think we're probably best off having an automatic refresh for this one as well, seeing as this is a what I would consider "vital" dashboard we should be getting an update quite often. This is also something that people might not be aware of when testing E04 during RC testing, I usually refreshed mine manually and wasn't aware that there was an option to set an automatic refresh until I looked into the JSON.

pkg/products/monitoring/dashboards/criticalSLOAlerts.go
obrienrobert: Looks like this sets the time range value? https://grafana.com/docs/grafana/latest/dashboards/time-range-controls/#common-time-range-controls

from

I'm not sure what is the ideal value is for this, if this is the case.

obrienrobert:
@HubertStefanski I've noticed on some of the panels that the `Last 28 days` text overlaps the panel title. I've attached a screenshot. Not sure if this is new or if there was always text overlapping, but it's just something I observed.

slo_dashboard_alerts

pkg/products/monitoring/dashboards/criticalSLOAlerts.go
obrienrobert: Agree, sounds reasonable. At the minimum it reduces the need to refresh to update the info.

pkg/products/monitoring/dashboards/criticalSLOAlerts.go
HubertStefanski: So I think what this does is that it checks if the alert was firing in any period from now to 5 minutes ago then uses that to determine whether to update dashboards which don't use instant fields, not so much of a problem now though I think with the auto-refresh

HubertStefanski:
> @HubertStefanski I've noticed on some of the panels that the `Last 28 days` text overlaps the panel title. I've attached a screenshot. Not sure if this is new or if there was always text overlapping, but it's just something I observed.

That's a result of setting the `timefrom` field in the dashboard JSON, I'll have a look if I can move it somehow, or at least make the box big enough so that it doesn't obstruct.

HubertStefanski:
image
fixed the issue @obrienrobert pointed out above

PR #1055, opened 2020-07-27
testcase: combine c05 and c12 tc

damienomurchu:
# Description

Merge C05 and C12 alerts testcases

## Type of change

## Checklist


trepel:
@damienomurchu LGTM, just one small comment. I think we can add a note to step 4 that it is only mandatory for RC1, optional otherwise. Few checks failed due to some weird error (to me), you might need to retest.

trepel:
@damienomurchu except for conflict resolving, I think you should add `per-build` tag here (I think this should be done for each RC). Also, adding some `environment` would be good. I can't help thinking this should be done for both fresh install and upgrade, both public and private clusters. Given the capacity I would reduce the matrix somehow though. public and private post upgrade only maybe?

trepel:
@damienomurchu Could you please run prettier to make the test-cases-lint check pass?

PR #1056, opened 2020-07-27
fix: o01 test case

psturc:
# Description
https://issues.redhat.com/browse/INTLY-8272


psturc:
/assign @aidenkeating @darahayes 
guys, just a minor update to the test case as per the discussion we had earlier

PR #1057, opened 2020-07-27
Add the plan cmd to the test-cases tools to bulk update the target version

b1zzu:
# Description

Added two new cmds `./tools.sh plan release ...` and `./tools.sh plan for ...` that would help updating the target version on the test cases

## Type of change

## Checklist

### DIFF
--- a/test-cases/README.md
+++ b/test-cases/README.md
@@ -156,6 +156,24 @@ to pretty print the csv output on the terminal:
 ./tools.sh export csv | column -t -s, | less -S
 ```
 
+## Bulk update the target version on the test cases
+
+> Attention: all changes are applied locally and need to be manually pushed to the integreatly-operator repo
+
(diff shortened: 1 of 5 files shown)


trepel:
It works as expected. I am not very familiar with typescript, but I looks good to me.

test-cases/README.md
```diff
@@ -156,6 +156,24 @@ to pretty print the csv output on the terminal:
 ./tools.sh export csv | column -t -s, | less -S
 ```
 
+## Bulk update the target version on the test cases
+
+> Attention: all changes are applied locally and need to be manually pushed to the integreatly-operator repo
+
+To automatically add the target version to all test cases which target version is older then 3 releases use:
```
trdoyle81, line 163:
```suggestion
To automatically add the target version to all test cases with a target version older than 3 releases, use the following command:
```

PR #1062, opened 2020-07-28
update webapp version to 2.27.0 and SP to 1.0.12

pb82:
# Description

update webapp to 2.27.0 and Solution patterns to v1.0.12


grdryn:
/retest

Not sure if this will work, there seem to be prow issues atm

pb82:
@grdryn Prow seems to be completely down, the link to the build even does not work

grdryn:
@pb82 it's apparently that the routes aren't working after a router restart on that cluster. Which means that it can't receive the webhooks from GH, among other things probably

PR #1070, opened 2020-07-29
Implement CRO PostgreSQL as Apicurio Registry's database

miguelsorianod:
# Description

This PR implements CRO PostgreSQL as Apicurio Registry's database.

Additionally, the other code that assumes AMQStreams was being used
has been moved to its own reconciliation method that as of now is
unused. This means that AMQ Streams for the moment is no longer being used
as Apicurio Registry's data source but the idea is to make the datasource configurable in future
changes.

There are several subjects that I'd like to discuss:

* This PR implements the deployment of Apicurio Registry's own PostgreSQL through CRO. There were talks about whether a shared database should be used between the different products instead. At the moment I've implemented it using its own dedicated database as I've seen the other PostgreSQL database (for rhsso?) is tied to the product's lifecycle so going for the shared database approach would require some additional changes. So, the current implementation has been done taking rhsso and its postgrsql database as the "model". My opinion on this is there's no "ideal" solution that fits for all cases. It would depend on the environment's workload, HA requirements, etc...
* This PR effectively replaces and stops using AMQ Streams usage as apicurio registry's data source. However, Apicurio Registry supports several data sources and we need to support them. As a requirement we need to be able to choose the desired datasource for Apicurio Registry between AMQ Streams, PostgreSQL and others it offers.  We would also be interested, in the case of PostgreSQL, the ability to provide an external endpoint so the database is not managed by the operator at all as a possible case. Is there some product-level configurability mechanism that integreatly operator provides to allow that?We can do this in another PR if it is not trivial / needs additional work/discussion and we want to apply changes incrementally

Let me know about this PR code changes and the subjects I mentioned above.

Thank you.

## Type of change

## Checklist

### DIFF
--- a/pkg/products/apicurioregistry/reconciler.go
+++ b/pkg/products/apicurioregistry/reconciler.go
@@ -3,8 +3,12 @@ package apicurioregistry
 import (
 	"context"
 	"fmt"
+	"net/url"
 
 	apicurioregistry "github.com/Apicurio/apicurio-registry-operator/pkg/apis/apicur/v1alpha1"
+	crov1alpha1 "github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1"
+	"github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1/types"
+	croUtil "github.com/integr8ly/cloud-resource-operator/pkg/client"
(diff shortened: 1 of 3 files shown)


akieling:
The only way I know to configure the datasource for Apicurio Registry is via a new field in the RHMI CR.

PR #1072, opened 2020-07-30
INTLY-9233

MStokluska:
# Description
Refactor of https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/marketplace/configmap_catalog_source_reconciler.go#L60 - addressing ticket https://issues.redhat.com/browse/INTLY-9233.
The function was refactored so that it uses `CreateOrUpdate` function.

# Verification

On a fresh cluster run:
- `make cluster/prepare/local`
- `make code/run`
RHMI Operator should successfully install with all the config maps created. Once installed, change some value in the manifests folder to test update functionality. For example:
- On your cluster open config map YAML for cloud resources 
- On your machine navigate to https://github.com/integr8ly/integreatly-operator/blob/master/manifests/integreatly-cloud-resources/0.18.0/cloud-resources.v0.18.0.clusterserviceversion.yaml#L116 and set the media type to image. 
- Once the cloud resources operator is reconciled you should see the config map update happening. 

## Type of change

## Checklist

### DIFF
--- a/pkg/resources/marketplace/configmap_catalog_source_reconciler.go
+++ b/pkg/resources/marketplace/configmap_catalog_source_reconciler.go
@@ -3,12 +3,10 @@ package marketplace
 import (
 	"context"
 	"fmt"
-	"reflect"
 
 	coreosv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
-	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


pkg/resources/marketplace/configmap_catalog_source_reconciler.go
```diff
@@ -68,26 +66,23 @@ func (r *ConfigMapCatalogSourceReconciler) reconcileRegistryConfigMap(ctx contex
...
-			}
-
-			logrus.Infof("Updated config map %s in namspace %s", configMapName, r.Namespace)
-		}
+	case controllerutil.OperationResultUpdated:
+		logrus.Infof("Updated config map %s in namspace %s", configMapName, r.Namespace)
+	case controllerutil.OperationResultNone:
+		break
+	default:
+		return configMapName, fmt.Errorf("Unknown controllerutil.OperationResult '%v'", or)
```
sergioifg94, line 85: I would absent from returning an error in the `default`. It's very unlikely to reach this statement, but as long as the value of `err` is `nil` we shouldn't fail, maybe just log it. Anyways, it's very **very** unlikely to that this is going to happen so it's not a big deal.

```suggestion
	default:
		logrus.Infof("Unknown controllerutil.OperationResult '%v'", or)
```

MStokluska: Thanks Sergio! That is a very good point! I've applied your suggestion.

MStokluska:
Closed as the Jira got closed with "won't do"

PR #1073, opened 2020-07-30
9500 cr version on upgrade

briangallagher:
# Description
An RHMI CR has a Version field and a ToVersion field. The ToVersion field is set during upgrade. Post upgrade, the version field is set to ToVersion. On automatic upgrades (meaning the InstallPlan is auto approved) the logic to set ToVersion is done during Subscription Reconcile. 
This code change caters for the manual (meaning the InstallPlan is manually approved) scenario. 

## How to verify
Install and then upgrade RHMI manually and automatically. Verify that the version field on the CR is set to the correct version before and after the upgrade.
```
oc -n redhat-rhmi-operator get rhmi rhmi -o yaml | grep ersion
```

### Manually
See [deploying RHMI with bundle format](https://github.com/integr8ly/integreatly-operator/blob/master/README.md#deploying-to-a-cluster-with-olm-and-the-bundle-format) on how to install and then upgrade

### Automatically 
Follow the same procedure as above but do not approve the install plan. Instead, adjust `rhmiconfig` cr to force an automatic install. Set:
```spec.upgrade.notBeforeDays = 1```
```spec.upgrade.waitForMaintenance = false```

Wait for upgrade, see:
```status.upgrade.scheduled.for```

### DIFF
--- a/pkg/controller/installation/installation_controller.go
+++ b/pkg/controller/installation/installation_controller.go
@@ -271,7 +271,7 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 	}
 
 	// If no current or target version is set this is the first installation of rhmi.
-	if installation.Status.Version == "" && installation.Status.ToVersion == "" {
+	if manualUpgradeFirstReconcile(installation) || firstInstallFirstReconcile(installation) {
 		installation.Status.ToVersion = version.IntegreatlyVersion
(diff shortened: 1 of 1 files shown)


MStokluska:
it looks good to merge, tested the functionality on a cluster and it works as expected. Great job!
/lgtm

philbrookes:
/override ci/prow/e2e
Verified locally by @MStokluska

PR #1027, opened 2020-07-21
added test cases for 2.6.0 monitoring

jhellar:
# Description

https://issues.redhat.com/browse/INTLY-9456

PR #1079, opened 2020-07-31
fix: test c03: pagerduty key

psturc:
# Description
https://issues.redhat.com/browse/INTLY-9697
Based on failed test [here](https://master-jenkins-csb-intly.cloud.paas.psi.redhat.com/job/Integreatly/job/rhmi-install-addon-flow/218/console) and [this test case](https://github.com/integr8ly/integreatly-operator/pull/1027/files#diff-231c5ff530657ac84ede9045e779b3a0R33)

### DIFF
--- a/test/common/alerts_mechanism.go
+++ b/test/common/alerts_mechanism.go
@@ -98,6 +98,7 @@ func TestIntegreatlyAlertsMechanism(t *testing.T, ctx *TestingContext) {
 }
 
 func verifySecrets(kubeClient kubernetes.Interface) error {
+	var pagerdutyKey string
 	res, err := kubeClient.CoreV1().Secrets(RHMIOperatorNamespace).Get("redhat-rhmi-deadmanssnitch", metav1.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


jhellar:
@david-martin is this related to the change where the secret is being created by hive? If so, does it mean that the test would fail for installation types other than via addon flow? Should we maybe check for `PAGERDUTY_KEY` and if that is not present then for `serviceKey`?

david-martin:
> Should we maybe check for `PAGERDUTY_KEY` and if that is not present then for `serviceKey`?

@jhellar This sounds reasonable, and is what the rhmi operator code does (backwards compatible)

psturc:
thanks @jhellar @david-martin, I will update the test then

grdryn:
@Rajagopalan-Ranganathan should a similar thing be put in place for the DMS secret, so that we don't hit a similar issue once we move to that?

Rajagopalan-Ranganathan:
The DMS secret logic is also the same, it searches for the secret created by the operator first and then for the one created by CSSRE. If it doesnt exist then we throw an error. Is that what you mean @grdryn

PR #1084, opened 2020-07-31
INTLY-9691 Fix(bug): Add extra uninstall stage to allow CRO be remove before Monitoring but after SSO

Boomatang:
# Description

Issue was caused by pr #1049 which required CRO to be uninstalled after SSO. The fix is to add the CRO to a new uninstall stage that gets removed after the main products but before the monitoring is removed.

**JIRA** https://issues.redhat.com/browse/INTLY-9691

To validated:
The `cloud-resources` namespace should be fully removed before any namespace for `monitoring` is removed during uninstall.

## Type of change

## Checklist


psturc:
/lgtm

verified on osd v4.4.11

PR #1087, opened 2020-08-04
add versions check

StevenTobin:
# Description
Add version check to catch when a CHANGEME string is in the rhmi_types file

## Type of change

## Checklist


mikenairn:
This check will fail if "CHANGME" is found anywhere in this file pkg/apis/integreatly/v1alpha1/rhmi_types.go

We could get more explicit about what shouldn't be set to CHANGME, but probably not worth the effort.

/lgtm
/approve

PR #1088, opened 2020-08-05
Intly 9078

pb82:
Convert nightly results into HTML and include stage results as a table.


pawelpaszki:
@pb82 the script looks good 👍 I forgot to mention that it suppose to be placed here actually: https://github.com/integr8ly/delorean/tree/master/scripts/jenkins-reports (there will be some changes required due to shellcheck syntax warnings). The script from here can be removed

matskiv:
Emails will not be revived because we will use report portal. Closing.

PR #1103, opened 2020-08-12
remove network calls in unit tests

briangallagher:
# Description
remove network calls in unit tests to overcome Travis issue of network timeouts

PR #1110, opened 2020-08-12
bump version on cr after manually approved upgrade

pmccarthy:
# Description
Cherry picking fix from master: https://github.com/integr8ly/integreatly-operator/pull/1073

### DIFF
--- a/deploy/olm-catalog/integreatly-operator/2.6.1/integreatly-operator.v2.6.1.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/integreatly-operator/2.6.1/integreatly-operator.v2.6.1.clusterserviceversion.yaml
@@ -47,6 +47,7 @@ metadata:
     support: RHMI
   name: integreatly-operator.v2.6.1
   namespace: placeholder
+  olm.skipRange: '>=2.4.0 <2.6.1'
 spec:
   apiservicedefinitions: {}
   customresourcedefinitions:
--- a/deploy/olm-catalog/integreatly-operator/2.6.1/integreatly-operator.v2.6.1.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/integreatly-operator/2.6.1/integreatly-operator.v2.6.1.clusterserviceversion.yaml
@@ -457,7 +457,7 @@ spec:
   minKubeVersion: 1.12.0
   provider:
     name: rhmi
-  replaces: integreatly-operator.v2.6.0
+  replaces: integreatly-operator.v2.5.0
   selector:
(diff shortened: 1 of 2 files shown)


deploy/olm-catalog/integreatly-operator/2.6.1/integreatly-operator.v2.6.1.clusterserviceversion.yaml
```diff
@@ -47,6 +47,7 @@ metadata:
     support: RHMI
   name: integreatly-operator.v2.6.1
   namespace: placeholder
+  olm.skipRange: '>=2.4.0 <2.6.1'
```
mikenairn, line 50: Is this going to replace anything greater than or equal to 2.4.0? Any reason we are doing that?

mikenairn: Just noting here that we usually clear up old manifests for products between releases to get round the configmap limitations, only leaving the minimum required to upgrade from the previous version. This assumes that we always apply every release of our operator. Assuming nothing was upgraded between these versions, we might be ok, but if this is something we plan on doing more often, or want to allow for, then we need to maintain the correct OLM graph for every product.

mikenairn:
/override ci/prow/unit

Failing due to a known issue, fix is on the way from https://github.com/integr8ly/integreatly-operator/pull/1103

deploy/olm-catalog/integreatly-operator/2.6.1/integreatly-operator.v2.6.1.clusterserviceversion.yaml
pmccarthy: @mikenairn Yeah this is a unique case. Unfortunately due to issues with 2.5.0 upgrades in production, SRE put a halt to them and so we have clusters on both 2.4.0 and 2.5.0. We want to be able to upgrade both directly to 2.6.1 which is why we need to specify this range

pmccarthy: @mikenairn We'll be looking to remove this following the 2.6.1 upgrade. We just need it here for the reasons I've listed above

pawelpaszki:
@mikenairn can you approve again? Would it be possible to override e2e to speed up the process of merging?

pawelpaszki:
/override ci/prow/unit

Failing due to a known issue, fix is on the way from #1103

PR #1114, opened 2020-08-13
N02 and E04 test case descriptions updated

trepel:
# Description

E04 updates:
Small notes on how to use the scripts and that it might be slightly tricky to bring stuff back up. Based on my experience doing the test case during 2.6.0

N02 updates:
Removed step 8 - CLUSTER_ID is not used in subsequent steps
Added steps 14 and 15 - these describe how to use workload-web-app to see the downtimes. These steps were just copy-pasted from H21.
The rest of the diff is just due to changing the step numbers, no content changes there.

Not sure how to verify this. Probably just read through the changes if that makes sense.

### DIFF
--- a/test-cases/tests/dashboards/e04-verify-slo-dashboard.md
+++ b/test-cases/tests/dashboards/e04-verify-slo-dashboard.md
@@ -31,7 +31,7 @@ Note: this test should only be performed at a time it will not affect other ongo
    > redhat-rhmi-solution-explorer -> Workloads -> Deployment Configs -> Scale to 1
    > redhat-rhmi-ups -> Workloads -> Deployments -> Scale to 1
 
-Execute the scripts below in a separate terminal tab/window and keep them running. It will keep scaling down stuff so that things don't start working automatically.
+Execute the scripts below in a separate terminal tab/window and keep them running. It will keep scaling down resources and prevent from automatic scale up by the operator(s).
 
 Bringing things up might not be so simple. Make sure you have access to the terminal output of the scripts to see how many pods (replicas) were there originally and manually change that back if some alerts don't stop firing.
 
--- a/test-cases/tests/dashboards/e04-verify-slo-dashboard.md
+++ b/test-cases/tests/dashboards/e04-verify-slo-dashboard.md
@@ -18,7 +18,7 @@ Note: this test should only be performed at a time it will not affect other ongo
 
 ## Steps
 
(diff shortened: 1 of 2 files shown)


pawelpaszki:
@trepel actually we are going to stop using `measure-downtime.json` script, so could you remove references to it from N02?

test-cases/tests/dashboards/e04-verify-slo-dashboard.md
```diff
@@ -31,6 +31,10 @@ Note: this test should only be performed at a time it will not affect other ongo
    > redhat-rhmi-solution-explorer -> Workloads -> Deployment Configs -> Scale to 1
    > redhat-rhmi-ups -> Workloads -> Deployments -> Scale to 1
 
+Execute the scripts below in a separate terminal tab/window and keep them running. It will keep scaling down stuff so that things don't start working automatically.
```
pawelpaszki, line 34:
```suggestion
Execute the scripts below in a separate terminal tab/window and keep them running. It will keep scaling down resources and prevent from automatic scale up by the operator(s).
```

test-cases/tests/dashboards/e04-verify-slo-dashboard.md
```diff
@@ -31,6 +31,10 @@ Note: this test should only be performed at a time it will not affect other ongo
...
+
+Bringing things up might not be so simple. Make sure you have access to the terminal output of the scripts to see how many pods (replicas) were there originally and manually change that back if some alerts don't stop firing.
```
pawelpaszki, line 36:
```suggestion
Bringing resources up might not be so simple. Make sure you have access to the terminal output of the scripts to see how many pods (replicas) were there originally and manually change that back if some alerts don't stop firing.
```

test-cases/tests/dashboards/e04-verify-slo-dashboard.md
trepel: hey, resources is much nicer word than stuff :) Was not sure how to call Deployments, DeploymentConfigs, and StatefulSets by general word.

trepel:
> @trepel actually we are going to stop using `measure-downtime.json` script, so could you remove references to it from N02?

Are we? I did not know about that. Ok, I will update the PR accordingly.

Thanks for the review.

trepel:
@pawelpaszki Thanks for the suggestions, both were accepted. As per discussion we will keep this PR small and focused. The measure-downtime.js script removal will be part of separate PR. Would you mind to take another look?

PR #1120, opened 2020-08-17
Added test case to upgrade to simulated version

trepel:
# Description

A new test case based on https://issues.redhat.com/browse/INTLY-9492


trepel:
@pb82 is this important to test this for 2.7.0? If yes, it would be ideal to have this merged before the code freeze?

trepel:
@pb82 due to recent shift from RHMI to RHOAM, do you find this test case still relevant?

PR #1131, opened 2020-08-19
Release v2.6

pmccarthy:
## Description
Pushing v2.6 base branch changes back into master. It has been decided to skip v2.6 and move directly for v2.7 in production

Main change of note is the introduction of the `olm.skipRange` parameter than may be required for future releases, v2.7 in particular as we have some production clusters running v2.4.0 while others are running v2.5.0


deploy/olm-catalog/integreatly-operator/2.6.1/integreatly-operator.v2.6.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,465 @@
...
+            "smtpSecret": "redhat-rhmi-smtp",
+            "type": "managed",
+            "useClusterStorage": "true"
+          }
+        }
+      ]
+    capabilities: Basic Install
+    categories: RHMI integration
+    certified: "false"
+    containerImage: quay.io/integreatly/integreatly-operator:v2.6.1-rc3
```
pmccarthy, line 44: Not sure if it's ok to still reference an RC image here? Testing for v2.6.1 commenced but was then halted when the decision to skip v2.6 was made. As a result we don't have an official `integreatly-operator:v2.6.1` image to reference

wei-lee: I think this is ok. This will only have an impact if you run the operator via OLM, and it doesn't impact production environments at all.

PR #1132, opened 2020-08-19
Delete test case A20 because obsolete

b1zzu:
# Description

The test case A20 is testing a requirement that has been removed

## Type of change

## Checklist


trdoyle81:
My argument for keeping test cases is pertinent to one-off or version specific test cases (eg for bug fixes) and retaining the ability to rerun the test case in the future if needed (if a bug reoccured for example). However, In this case, the feature requirement has removed and therefore the test case should also be deleted (imo).

trdoyle81 (approved):

b1zzu:
Good points @trepel, for the link we can fix it by using the commit sha instant of the branch name when generating the jira tasks

PR #1134, opened 2020-08-19
Docs: How to include manual tests into the next release

b1zzu:
# Description

Document how to include manual test cases into the next release

## Type of change

## Checklist

### DIFF
--- a/test-cases/README.md
+++ b/test-cases/README.md
@@ -58,7 +58,7 @@ Commit everything and open a new PR.
 
 By default we don't execute all manual tests on each release, therefore if we know we have made a change to a component and a specific test should be executed to verify the change or prevent regressions we must manually add the test case to the next release.
 
-To do that the next version needs to added to the `targets`, for example if the last released version was `2.7.0` then the next release version would be `2.8.0`.
+To do that the next version needs to be added to the `targets`, for example if the last released version was `2.7.0` then the next release version would be `2.8.0`.
 
 ```
 ---
--- a/test-cases/README.md
+++ b/test-cases/README.md
@@ -9,6 +9,21 @@
 - Try to write test cases using the standard syntax described in the [test-template.md](./fixtures/test-template.md).
 - General guidelines in `common/general-guidelines.md` are imported on the bottom of every test case.
 
+## Index
+
+- [How to create a test case](#How-to-create-a-test-case)
+- [How to include a manual test case in the next release](#How-to-include-a-manual-test-case-in-the-next-release)
+- [How to bulk update the target version on the test cases](#How-to-bulk-update-the-target-version-on-the-test-cases)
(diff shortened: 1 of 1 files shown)


test-cases/README.md
```diff
@@ -37,6 +52,43 @@ npm run prettier
 
 Commit everything and open a new PR.
 
+## How to include a manual test case in the next release
+
+> Attention: all changes need to be manually pushed to master or to the release branch
+
+By default we don't execute all manual tests on each release, therefore if we know we have made a change to a component and a specific test should be executed verify the change or prevent regressions we must manually add the test case to the next release.
```
trepel, line 59: replace `executed verify` with `executed to verify`

test-cases/README.md
```diff
@@ -37,6 +52,43 @@ npm run prettier
...
+
+To do it we need to add the to the `targets` the next version, for example if the last released version was `2.7.0` then the next release version would be `2.8.0`.
```
trepel, line 61: `the to the` -> I would change the whole sentence:
"To do that the next version needs to added to the `targets`,..."

test-cases/README.md
trepel, line 59:
```suggestion
By default we don't execute all manual tests on each release, therefore if we know we have made a change to a component and a specific test should be executed to verify the change or prevent regressions we must manually add the test case to the next release.
```

test-cases/README.md
trepel, line 61:
```suggestion
To do that the next version needs to be added to the `targets`, for example if the last released version was `2.7.0` then the next release version would be `2.8.0`.
```

test-cases/README.md
```diff
@@ -37,6 +52,43 @@ npm run prettier
 
 Commit everything and open a new PR.
 
+## How to include a manual test case in the next release
+
+> Attention: all changes need to be manually pushed to master or to the release branch
+
+By default we don't execute all manual tests on each release, therefore if we know we have made a change to a component and a specific test should be executed to verify the change or prevent regressions we must manually add the test case to the next release.
+
+To do that the next version needs to added to the `targets`, for example if the last released version was `2.7.0` then the next release version would be `2.8.0`.
```
trepel, line 61:
```suggestion
To do that the next version needs to be added to the `targets`, for example if the last released version was `2.7.0` then the next release version would be `2.8.0`.
```

trepel (changes requested):

test-cases/README.md
trepel: Sorry @b1zzu , in my suggestion there was a mistake :+1: Once this is fixed, I am happy to lgtm and approve this one

b1zzu: np, there was no rush to merge it

b1zzu:
@tremes should be ready now

PR #1143, opened 2020-08-25
INTLY-2486 Add Console Link for API Management.

Boomatang:
The Bounce url is been used to bypass the default login in for 3scale

# Description

**Jira**
https://issues.redhat.com/browse/INTLY-2486

The Bounce URI is been used to bypass the default login in for 3scale when using the link under the console links. Allowing the user to be take straight to the openshift login selection page.

This work does not have any affect on the solution explorer.

When the operator has installed 3scale the new console link can been seen under "API Management". This is link is also removed when the rhmi operator is uninstalled. 
image

## Type of change

## Checklist


pkg/products/threescale/reconciler.go
```diff
@@ -1444,3 +1457,42 @@ func (r *Reconciler) reconcileSubscription(ctx context.Context, serverClient k8s
...
+		cl.Spec.ApplicationMenu.Section = "Red Hat Applications"
```
MStokluska, line 1473: I think 3-Scale would be better under "Red Hat Integration" - I'm not 100% sure but it seems it is a part of our integration products https://www.redhat.com/en/products/integration

pkg/products/threescale/reconciler.go
```diff
@@ -1444,3 +1457,42 @@ func (r *Reconciler) reconcileSubscription(ctx context.Context, serverClient k8s
...
+		cl.Spec.Href = fmt.Sprintf("%v/auth/rhsso/bounce", r.Config.GetHost())
+		cl.Spec.Location = consolev1.ApplicationMenu
+		cl.Spec.Text = "API Management"
```
MStokluska, line 1476: Do you think it would be better to call it "3scale API Management" following https://www.redhat.com/en/products/integration

pkg/products/threescale/reconciler.go
```diff
@@ -1444,3 +1457,42 @@ func (r *Reconciler) reconcileSubscription(ctx context.Context, serverClient k8s
...
...C41MVpNNTAuMSw2Ni43N2EyLjY5LDIuNjksMCwxLDAsMi42OCwyLjY5QTIuNywyLjcsMCwwLDAsNTAuMSw2Ni43N1pNODEuMjUsNDEuMTJhMi43LDIuNywwLDAsMC0yLjY4LDIuNjksMi42NSwyLjY1LDAsMCwwLDIuNjgsMi42OSwyLjY5LDIuNjksMCwwLDAsMC01LjM3Wk00NC40OSw3Ni40N2EzLjczLDMuNzMsMCwwLDAtMy43MywzLjc0LDMuNzcsMy43NywwLDEsMCwzLjczLTMuNzRaTTc5LjA2LDU2LjcyYTQsNCwwLDEsMCw0LDRBNCw0LDAsMCwwLDc5LjA2LDU2LjcyWm0tNiwxMS43OEEzLjA5LDMuMDksMCwwLDAsNzAsNzEuNmEzLDMsMCwwLDAsMy4wOCwzLjA5LDMuMDksMy4wOSwwLDAsMCwwLTYuMTlaTTI4LjMsNjhhNC4xNiw0LjE2LDAsMCwwLTQuMTQsNC4xNUE0LjIxLDQuMjEsMCwwLDAsMjguMyw3Ni4zYTQuMTUsNC4xNSwwLDAsMCwwLTguM1ptLTguMjItOWEzLDMsMCwxLDAsMywzQTMuMDUsMy4wNSwwLDAsMCwyMC4wOCw1OVptMS44NC05Ljc0YTMsMywwLDEsMCwzLDNBMy4wNSwzLjA1LDAsMCwwLDIxLjkxLDQ5LjIyWk0yMi4zNyw0MmEzLjI0LDMuMjQsMCwxLDAtMy4yNCwzLjI2QTMuMjYsMy4yNiwwLDAsMCwyMi4zNyw0MlpNNDMuMTEsNzAuMmEzLjgsMy44LDAsMCwwLTMuODEtMy43NCwzLjczLDMuNzMsMCwwLDAtMy43MywzLjc0QTMuOCwzLjgsMCwwLDAsMzkuMyw3NCwzLjg3LDMuODcsMCwwLDAsNDMuMTEsNzAuMlpNMzcuNTYsNTguNDNhNC42OCw0LjY4LDAsMCwwLTQuNjItNC42NCw0LjYzLDQuNjMsMCwwLDAtNC42Miw0LjY0LDQuNTgsNC41OCwwLDAsMCw0LjYyLDQuNjRBNC42Myw0LjYzLDAsMCwwLDM3LjU2LDU4LjQzWk0yMy4xMSwzMy44MmEyLjUyLDIuNTIsMCwxLDAtMi41MS0yLjUyQTIuNTMsMi41MywwLDAsMCwyMy4xMSwzMy44MloiLz48L2c+PC9zdmc+"
```
MStokluska, line 1472: Do you think it could be better to upload 3Scale image to https://github.com/integr8ly/integreatly-operator/tree/master/assets/icons and use it from there instead, just like it's done with the solution explorer for example - https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/solutionexplorer/reconciler.go#L261. It might be easier to maintain it in case we would need to update icons at some stage?

pkg/products/threescale/reconciler.go
```diff
@@ -1444,3 +1457,42 @@ func (r *Reconciler) reconcileSubscription(ctx context.Context, serverClient k8s
...
+		return nil
+	})
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("error creating or updating 3Scale console link, %s", err)
+	}
+
+	return integreatlyv1alpha1.PhaseCompleted, nil
+}
+
+func (r *Reconciler) deleteConsoleLink(ctx context.Context, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
```
MStokluska, line 1486: I wonder could we possibly remove the console link the same way as it's done with solution explorer? https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/solutionexplorer/reconciler.go#L156 - in my understanding it would save us creating a new function and keep the codebase slightly shorter?

MStokluska:
Great job @Boomatang! I'm still waiting to verify the functionality and left just a couple of comments.

pkg/products/threescale/reconciler.go
Boomatang: From a maintenance point I agree with your comment but I believe the reasons for not doing it in that fashion out weight the maintenance reasons.

- There is a lag on the solution explorer icon when the customer first opens the menu. Which I feel cheapness the experiences. That happens as the cluster needs to call to github to get a copy of the image.
- If github is down the icon can not be retrieved. How often this would happen is questionable but it can happen.
- If the cluster when this is been install is a private cluster by my understanding the cluster may have none or limited access to the internet. Which means the call to github might not be possible.
- The icon is not version-able with the static URL. As example if 2.7 is planned to have a new icon as soon as that change lands in the master branch any new 2.6 clusters would pull that new icon. 
- Most likely the cluster does caching of this icon, so I am not sure if even for maintenances updates would the cluster pull a fresh copy of the image.

With that said I do agree its not the easily find where to change the icon. So I will move it to a const at the top of the file.

pkg/products/threescale/reconciler.go
Boomatang: Personally I don't agree with the style taken in the solution explorer reconciler, I do understand the reasoning behind it. 

With in the 3 scale reconciler the style seem to be have a function/method for a body of work been done (for the most part). To my it would be better to keep the style of code when the changes are been made.

pkg/products/threescale/reconciler.go
Boomatang: Agree, Changes will come shortly

pkg/products/threescale/reconciler.go
Boomatang: This I am very open to change on and as on that product page its under integration, that's the menu I will move it under. Change coming soon.

pkg/products/threescale/reconciler.go
MStokluska: Thanks for explaining it to me @Boomatang, it makes sense!

MStokluska:
Verified on the cluster, and it works, however, I would expect the URL links to be added without the need of refreshing the OSD webpage. This happens for all the links, not only the one addressed in this PR, maybe something worth looking into in the future!
 
Good job!
/lgtm

PR #1146, opened 2020-08-25
fix flake e2e test for 3scale authentication

jjaferson:
# Description

Adds a fix to the e2e test for the 3scale authentication

https://issues.redhat.com/browse/INTLY-8433

## Type of change

### DIFF
--- a/test/common/3scale_crudl.go
+++ b/test/common/3scale_crudl.go
@@ -28,6 +28,9 @@ func Test3ScaleCrudlPermissions(t *testing.T, ctx *TestingContext) {
 
 	// Get the fuse host url from the rhmi status
 	host := rhmi.Status.Stages[v1alpha1.ProductsStage].Products[v1alpha1.Product3Scale].Host
+	if host == "" {
+		host = fmt.Sprintf("https://3scale-admin.%v", rhmi.Spec.RoutingSubdomain)
+	}
 	keycloakHost := rhmi.Status.Stages[v1alpha1.AuthenticationStage].Products[v1alpha1.ProductRHSSO].Host
 	redirectUrl := fmt.Sprintf("%v/p/admin/dashboard", host)
 
@@ -36,8 +39,8 @@ func Test3ScaleCrudlPermissions(t *testing.T, ctx *TestingContext) {
 	// Login to 3Scale
 	err = loginToThreeScale(t, host, threescaleLoginUser, DefaultPassword, "testing-idp", ctx.HttpClient)
 	if err != nil {
-		t.Skip("Skipping due to known flaky behavior, to be fixed ASAP.\nJIRA: https://issues.redhat.com/browse/INTLY-8433")
-		// t.Fatal(err)
+		// t.Skip("Skipping due to known flaky behavior, to be fixed ASAP.\nJIRA: https://issues.redhat.com/browse/INTLY-8433")
+		t.Fatal(err)
 	}
 
 	// Make sure 3Scale is available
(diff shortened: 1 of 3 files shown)


test/common/authdelay_first_broker_login.go
```diff
@@ -177,7 +203,38 @@ func loginToThreeScale(t *testing.T, tsHost, username, password string, idp stri
 			})
 		})
 
-		t.Logf("browser URL %s", browser.Url().RequestURI())
+		// checks if an error happened in the login
+		if browser.StatusCode() == 502 {
```
matskiv, line 207: I am wondering about this part.
a) how did you find out a 502 status code was returned?
b) any chance we can call some function instead of the code below? It seems like something that's probably already implemented elsewhere.

test/common/3scale_crudl.go
```diff
@@ -28,6 +28,9 @@ func Test3ScaleCrudlPermissions(t *testing.T, ctx *TestingContext) {
 
 	// Get the fuse host url from the rhmi status
 	host := rhmi.Status.Stages[v1alpha1.ProductsStage].Products[v1alpha1.Product3Scale].Host
+	if host == "" {
```
matskiv, line 31: hmm, I am wondering how Host could be empty, any ideas why is this happening @jjaferson ?

matskiv:
Looks pretty good overall.
I added two questions.

test/common/3scale_crudl.go
jjaferson: `Test3ScaleSMTPConfig` patches the DC for the `system-app` [1] which modifies the installation status for 3scale to `in progress`, I don't think it would happen in PROW as the `Test3ScaleSMTPConfig` is the last test to run [2] but I spot that issue by running the functional test a few times

[1] https://github.com/integr8ly/integreatly-operator/blob/bbb544ff360c502719983badd181e284b04586bd/test/common/3scale_smtp.go#L294
[2] https://github.com/integr8ly/integreatly-operator/blob/master/test/common/tests.go#L53

test/common/authdelay_first_broker_login.go
jjaferson: Just to give a bit of context on this, when I created the e2e test for the authentication delay plugin I was verifying the authentication of new users, but other tests are using the `loginToThreeScale` function for existing users and as a few tests are using the same user to login to 3scale and this user is already authenticated the test get stuck at an endpoint like the one below 

```/auth/realms/openshift/broker/openshift-v4/endpoint?code=5eoXNSKQ6Fw_uvcYiEsemDU6EOUE90KG0V9Me6PPblQ&state=ZWlDcGsbqD00QL_lVQ1AIKM1MAgn_T06tRNIs5EvMbY.kBoSCwUdIwM.3scale```

a) I logged in to 3scale and got the link with the endpoint that was similar to the one that was getting stuck in the test and it returned 502.

b) I suppose you are right,  some parts of this code is kind repeated and I could actually create some functions, do you reckon it would be used by other tests?

jjaferson:
> Looks pretty good overall.
> I added two questions.

thanks, replied to your questions.

test/common/3scale_crudl.go
matskiv: Oh, I see.
Not an ideal behavior, but it's probably minor enough...

test/common/authdelay_first_broker_login.go
matskiv: Thank you for the explanation.

As for b), probably not. Let's leave it as is.

jjaferson:
failed for the first time in **FAIL: TestIntegreatly/integreatly/C03_-_Verify_that_alerting_mechanism_works**, will retest and in case it happens again report as flaky e2e test

/test e2e

PR #1148, opened 2020-08-27
feat(InstallationController): Adds ability to pass installType as env var

jjaferson:
# Description

Adds ability to pass installType as env var

https://issues.redhat.com/browse/MGDAPI-7

## Verification steps

1) export an env var
`export INSTALLATION_TYPE=managed`
2) install the operator locally
3) verify if the installation type in the RHMI CR is the same as the one in the env var
`oc get rhmi rhmi -o json -n redhat-rhmi-operator | jq .spec.type`

## Type of change

## Checklist


Boomatang:
I was about to have a look over this. There is one question that I would like to ask. If a user states a invalid install type what happens?

PR #1150, opened 2020-08-31
fix grafana routes e2e test - list secrets

jhellar:
# Description

## Type of change

## Checklist

### DIFF
--- a/test/common/grafana_routes.go
+++ b/test/common/grafana_routes.go
@@ -95,37 +95,39 @@ func TestGrafanaExternalRouteDashboardExist(t *testing.T, ctx *TestingContext) {
 		t.Fatal("failed to create clusterRoleBinding", err)
 	}
 	defer ctx.Client.Delete(goctx.TODO(), binding)
-	err = ctx.Client.Get(goctx.TODO(), k8sclient.ObjectKey{Name: serviceAccountName, Namespace: grafanaNamespace}, serviceAccount)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


test/common/grafana_routes.go
```diff
@@ -95,38 +95,40 @@ func TestGrafanaExternalRouteDashboardExist(t *testing.T, ctx *TestingContext) {
...
-	for _, secret := range serviceAccount.Secrets {
-		if strings.Contains(secret.Name, "token") {
-			secretName = secret.Name
+	token := ""
+	secrets, err := ctx.KubeClient.CoreV1().Secrets(grafanaNamespace).List(metav1.ListOptions{})
+	if err != nil {
+		t.Fatal("failed to get secrets", err)
+	}
+	for _, secretsItem := range secrets.Items {
+		if strings.HasPrefix(secretsItem.Name, "test-token-") {
```
david-martin, line 110: @jhellar will there only ever be 1 with this prefix?

jhellar: There are 2 secrets created with this prefix when the service account is created. One of them can be used as the bearer token. By looking at both of them, the only difference I could found was `kubernetes.io/created-by` annotation, that is not present in the secret with the bearer token [1].

[1] https://github.com/integr8ly/integreatly-operator/pull/1150/files#diff-ab3b7c7cf278b4574268aa4ce059a50aR122

PR #1152, opened 2020-09-01
INTLY-9953: Remove permission of dedicated admins over RHMIConfig

sergioifg94:
# Description

> 🔨 Link to JIRA: https://issues.redhat.com/browse/INTLY-9953
> ℹ️ Link to JIRA **that added this permission**: https://issues.redhat.com/browse/INTLY-6744

Remove reconciliation logic of a Role and RoleBinding that granted dedicated admins
permissions to update RHMIConfig CRs, and add logic to ensure that if this
permission exists it is removed.

Update unit tests and e2e tests.

### Pre-requisites

* An image of the operator based on this PR
    ```sh
    git checkout master
    TAG=2.6.1 ORG=<your quay username> make image/build/push
    ```
* An image of the operator based on master
    ```sh
    git checkout INTLY-9953
    TAG=intly9553 ORG=<your quay username> make image/build/push
    ```

### Steps

1. Install RHMI using the master-based operator image
    1. Edit the `deploy/operator.yaml` file to point to `quay.io/<your quay org>/integreatly-operator:v2.6.1`
    2. Run `make cluster/prepare/local`
    3. Deploy the operator
    ```sh
    oc apply -f deploy/operator.yaml
    ```
2. When the installation completes. Create the testing IdP
    ```sh
    ```
3. Log in as a dedicated admin, and check that you can edit the RHMIConfig CR.
4. Modify the operator deployment to point to the image based on this PR
5. Wait for the operator Pod to perform a full reconcile
6. Log in as a dedicated admin, and verify that you can **not** edit the RHMIConfig CR anymore.
   

## Type of change

## Checklist

### DIFF
--- a/pkg/controller/installation/bootstrapReconciler.go
+++ b/pkg/controller/installation/bootstrapReconciler.go
@@ -314,31 +314,6 @@ func (r *Reconciler) reconcilerGithubOauthSecret(ctx context.Context, serverClie
 }
 
 func (r *Reconciler) reconcileRHMIConfigPermissions(ctx context.Context, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
-	configRole := &rbacv1.Role{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "rhmiconfig-dedicated-admins-role",
-			Namespace: r.ConfigManager.GetOperatorNamespace(),
-		},
-	}
-
-	if _, err := controllerutil.CreateOrUpdate(ctx, serverClient, configRole, func() error {
-		configRole.Rules = []rbacv1.PolicyRule{
-			{
-				APIGroups: []string{"integreatly.org"},
-				Resources: []string{"rhmiconfigs"},
-				Verbs:     []string{"update", "get", "list", "watch"},
-			},
-			{
-				APIGroups: []string{"integreatly.org"},
-				Resources: []string{"rhmis"},
-				Verbs:     []string{"watch", "get", "list"},
-			},
-		}
-		return nil
-	}); err != nil {
-		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("error creating RHMI Config dedicated admin role: %w", err)
-	}
-
 	configRoleBinding := &rbacv1.RoleBinding{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "rhmiconfig-dedicated-admins-role-binding",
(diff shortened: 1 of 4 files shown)


pkg/controller/installation/bootstrapReconciler.go
```diff
@@ -310,54 +310,31 @@ func (r *Reconciler) reconcilerGithubOauthSecret(ctx context.Context, serverClie
...
 	}
 
-	if _, err := controllerutil.CreateOrUpdate(ctx, serverClient, configRole, func() error {
-		configRole.Rules = []rbacv1.PolicyRule{
-			{
-				APIGroups: []string{"integreatly.org"},
-				Resources: []string{"rhmiconfigs"},
-				Verbs:     []string{"update", "get", "list", "watch"},
-			},
-			{
```
philbrookes, line 327: does removing this role also remove the dedicated-admins ability to watch, get, lost on the RHMI CR? I'm not sure if that's a problem or not.

philbrookes: @maleck13 do we still need the dedicated-admins to have this permission, or is deleting both acceptable?

sergioifg94: It does, but when installing through OLM it performs all these role aggregations that grant them viewing permissions, so effectively they still can view the RHMIConfig

sergioifg94: Oh sorry I see now you meant RHMI. I'll double check, but I think it's the same, as RHMI is owned by the operator, so OLM will aggregate these roles for the RHMI as well

philbrookes: ah ok, then I think we're fine if that is true.

briangallagher: I ran this locally, as per above steps, and list priv on RHMI was removed from the dedicated-admin. Is this ok? @philbrookes @sergioifg94

sergioifg94: @briangallagher I think the best way to verify the effective permissions is to deploy RHMI through the add-on, then change the CSV to point to an image generated from this PR, and log in as dedicated admin. The reason is that OLM adds some permissions to dedicated admins over the CRs owned by the operators, so it will vary from the permissions you'll see when running locally

briangallagher: @sergioifg94 Thanks, confirming that worked. I'll assume we are ok with the local scenario.

PR #1153, opened 2020-09-01
add 3scale test case

wei-lee:
# Description

## Type of change

## Checklist


b1zzu:
@wei-lee Is still missing many test cases if you run
```
./tools.sh plan for --target 2.7.0 --component product-3scale
```
it will update all of them automatically, and then you just need to commit and push the changes made from the tool

PR #1155, opened 2020-09-01
Mgdapi 31 Updating the prepeare release script 

Boomatang:
# Description

The prepare release script now accepts more arguments with all the user to sepicfiy the `olm-catalog` resource be released. 
Also is the ability to set the quay.io org that will point to the catalogue images. This is useful when doing upgrade testing of bundles. Short readme added with a description of how to use the script and the arguments that can be set, plus the current defaults.

**JIRA** https://issues.redhat.com/browse/MGDAPI-31

**Requires https://github.com/integr8ly/delorean/pull/148 to be merged**

## Type of change

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,8 +19,9 @@ TEST_RESULTS_DIR ?= "test-results"
 TEMP_SERVICEACCOUNT_NAME="rhmi-operator"
 CLUSTER_URL:=$(shell sh -c "oc cluster-info | grep -Eo 'https?://[-a-zA-Z0-9\.:]*'")
 
+# These tags are modified by the prepare-release script.
 RHMI_TAG ?= 2.7.0
-RHOAM_TAG ?= 0.1.0-rc2
+RHOAM_TAG ?= 0.1.0
 
 # If openapi-gen is available on the path, use that; otherwise use it through
 # "go run" (slower)
@@ -46,6 +47,7 @@ else
 endif
 
 export SELF_SIGNED_CERTS   ?= true
+# Setting the INSTALLATION_TYPE to managed-api will configure the values required for RHOAM installs
 export INSTALLATION_TYPE   ?= managed
 export INSTALLATION_NAME   ?= rhmi
 
@@ -70,7 +72,8 @@ ifeq ($(INSTALLATION_TYPE), managed-api)
 	OPERATOR_IMAGE=$(REG)/$(ORG)/$(PROJECT):v$(TAG)
 	NAMESPACE_PREFIX ?= redhat-rhoam-
 	APPLICATION_REPO ?= managed-api-service
-	export INSTALLATION_PREFIX ?= redhat-rhoam
+	# TODO follow on naming of this folder by INSTALLATION_PREFIX and contents of the role_binding.yaml
+	export INSTALLATION_PREFIX ?= redhat-managed-api
 	export OLM_TYPE ?= managed-api-service
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,10 @@
 include ./make/*.mk
 
 ORG ?= integreatly
-NAMESPACE_PREFIX ?= redhat-rhmi-
-NAMESPACE=$(NAMESPACE_PREFIX)operator
-PROJECT=integreatly-operator
+
 REG=quay.io
 SHELL=/bin/bash
-TAG ?= 2.7.0
-MGDAPI_TAG ?= 0.1.0-rc2
+
 PKG=github.com/integr8ly/integreatly-operator
 TEST_DIRS?=$(shell sh -c "find $(TOP_SRC_DIRS) -name \\*_test.go -exec dirname {} \\; | sort | uniq")
 TEST_POD_NAME=integreatly-operator-test
@@ -15,8 +12,10 @@ COMPILE_TARGET=./tmp/_output/bin/$(PROJECT)
 OPERATOR_SDK_VERSION=0.17.1
 AUTH_TOKEN=$(shell curl -sH "Content-Type: application/json" -XPOST https://quay.io/cnr/api/v1/users/login -d '{"user": {"username": "$(QUAY_USERNAME)", "password": "$(QUAY_PASSWORD)"}}' | jq -r '.token')
 TEMPLATE_PATH="$(shell pwd)/templates/monitoring"
+
 INTEGREATLY_OPERATOR_IMAGE ?= $(REG)/$(ORG)/$(PROJECT):v$(TAG)
 MANAGED_API_OPERATOR_IMAGE ?= $(REG)/$(ORG)/managed-api-service:v$(MGDAPI_TAG)
+
 CONTAINER_ENGINE ?= docker
 TEST_RESULTS_DIR ?= "test-results"
 TEMP_SERVICEACCOUNT_NAME="rhmi-operator"
@@ -54,11 +53,35 @@ export OPERATORS_IN_PRODUCT_NAMESPACE ?= false # e2e tests and createInstallatio
 export DELOREAN_PULL_SECRET_NAME ?= integreatly-delorean-pull-secret
 export ALERTING_EMAIL_ADDRESS = noreply-test@rhmi-redhat.com
 
+ifeq ($(INSTALLATION_TYPE), managed)
+	RHMI_TAG ?= 2.7.0
+	PROJECT=integreatly-operator
+	OPERATOR_IMAGE=$(REG)/$(ORG)/$(PROJECT):v$(RHMI_TAG)
+	TAG ?= RHMI_TAG
+	NAMESPACE_PREFIX ?= redhat-rhmi-
+	APPLICATION_REPO ?= integreatly
+	export OLM_TYPE ?= integreatly-operator
+endif
+
+ifeq ($(INSTALLATION_TYPE), managed-api)
+	RHOAM_TAG ?= 0.1.0
+	PROJECT=managed-api-service
+	OPERATOR_IMAGE=$(REG)/$(ORG)/$(PROJECT):v$(RHOAM_TAG)
+	TAG ?= RHOAM_TAG
+	NAMESPACE_PREFIX ?= redhat-rhoam-
+	APPLICATION_REPO ?= managed-api-service
+	export OLM_TYPE ?= managed-api-service
+endif
+
+NAMESPACE=$(NAMESPACE_PREFIX)operator
+
 define wait_command
 	@echo Waiting for $(2) for $(3)...
 	@time timeout --foreground $(3) bash -c "until $(1); do echo $(2) not ready yet, trying again in $(4)s...; sleep $(4); done"
 	@echo $(2) ready!
 endef
+
+
 
 .PHONY: setup/moq
 setup/moq:
@@ -130,18 +153,22 @@ code/fix:
 
 .PHONY: image/build
 image/build: code/compile
-	@$(OPERATOR_SDK) build $(INTEGREATLY_OPERATOR_IMAGE)
+	echo "build image $(OPERATOR_IMAGE)"
+	@$(OPERATOR_SDK) build $(OPERATOR_IMAGE)
 
 .PHONY: image/push
 image/push:
-	docker push $(INTEGREATLY_OPERATOR_IMAGE)
+	echo "push image $(OPERATOR_IMAGE)"
+	docker push $(OPERATOR_IMAGE)
 
 .PHONY: image/build/push
 image/build/push: image/build image/push
 
+#TODO operator-sdk 0.17.1 does not have the `--enable-test` flag
 .PHONY: image/build/test
 image/build/test:
-	$(OPERATOR_SDK) build --enable-tests $(INTEGREATLY_OPERATOR_IMAGE)
+	echo "build test image for $(OPERATOR_IMAGE)"
+	$(OPERATOR_SDK) build --enable-tests $(OPERATOR_IMAGE)
 
 .PHONY: test/unit
 test/unit:
@@ -149,7 +176,7 @@ test/unit:
 
 .PHONY: test/e2e/prow
 test/e2e/prow: export component := integreatly-operator
-test/e2e/prow: export INTEGREATLY_OPERATOR_IMAGE := "${IMAGE_FORMAT}"
+test/e2e/prow: export OPERATOR_IMAGE := "${IMAGE_FORMAT}"
 test/e2e/prow: test/e2e
 
 .PHONY: test/e2e/rhoam/prow
@@ -166,6 +193,8 @@ test/e2e/rhoam:  cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/pr
 test/e2e:  export SURF_DEBUG_HEADERS=1
 test/e2e:  cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/prepare/crd deploy/integreatly-rhmi-cr.yml
 	$(OPERATOR_SDK) --verbose test local ./test/e2e --operator-namespace="$(NAMESPACE)" --go-test-flags "-timeout=120m" --debug --image=$(INTEGREATLY_OPERATOR_IMAGE)
+	 export SURF_DEBUG_HEADERS=1
+	$(OPERATOR_SDK) --verbose test local ./test/e2e --operator-namespace="$(NAMESPACE)" --go-test-flags "-timeout=120m" --debug --image=$(OPERATOR_IMAGE)
 
 .PHONY: test/e2e/local
 test/e2e/local: cluster/cleanup cluster/cleanup/crds cluster/prepare cluster/prepare/crd deploy/integreatly-rhmi-cr.yml
@@ -215,7 +244,9 @@ cluster/deploy/integreatly-rhmi-cr.yml: deploy/integreatly-rhmi-cr.yml
 	$(call wait_command, oc get RHMI $(INSTALLATION_NAME) -n $(NAMESPACE) --output=json -o jsonpath='{.status.stages.monitoring.phase}' | grep -q completed, monitoring phase, 10m, 30)
 	$(call wait_command, oc get RHMI $(INSTALLATION_NAME) -n $(NAMESPACE) --output=json -o jsonpath='{.status.stages.authentication.phase}' | grep -q completed, authentication phase, 10m, 30)
 	$(call wait_command, oc get RHMI $(INSTALLATION_NAME) -n $(NAMESPACE) --output=json -o jsonpath='{.status.stages.products.phase}' | grep -q completed, products phase, 30m, 30)
+ifeq ($(INSTALLATION_TYPE), managed)
 	$(call wait_command, oc get RHMI $(INSTALLATION_NAME) -n $(NAMESPACE) --output=json -o jsonpath='{.status.stages.solution-explorer.phase}' | grep -q completed, solution-explorer phase, 10m, 30)
+endif
 
 .PHONY: cluster/prepare
 cluster/prepare: cluster/prepare/project cluster/prepare/osrc cluster/prepare/configmaps cluster/prepare/smtp cluster/prepare/dms cluster/prepare/pagerduty cluster/prepare/ratelimits cluster/prepare/delorean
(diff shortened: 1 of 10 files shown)


scripts/prepare-release.sh
```diff
@@ -2,30 +2,51 @@
 set -e
 set -o pipefail
 
-PREVIOUS_VERSION=$(grep integreatly-operator deploy/olm-catalog/integreatly-operator/integreatly-operator.package.yaml | awk -F v '{print $2}')
+
+PREVIOUS_VERSION=$(grep $TYPE deploy/olm-catalog/$TYPE/$TYPE.package.yaml | awk -F v '{print $2}') || echo "No previous version"
+
+if [[ -z "$TYPE" ]]; then
```
philbrookes, line 8: Doesn't this need to occur before line 6 which is reading this value?

philbrookes:
e2e rate limit flake
/retest

sergioifg94:
Just a couple of questions about this:

1. I see the version of the CSV is `0.0.0`, shouldn't it be `1.0.0` instead?
2. I see that the CSV name is `managed-3scale`. Are we sure we're going for that name for it, or `managed-api-service`? Maybe @philbrookes can clarify this one

Just confirming to be in sync with the manifests and metadata that I bring into managed tenants

Boomatang:
@sergioifg94 These are good questions that I will try answer.

1. The version `0.0.0` was done as a place holder. I should have been clear on that in my description. This is not a release of the api managed service, that CSV will install the full RHMI. I would be expecting that CSV to be removed when a genuine release is been created.
2. I had spoke to @philbrookes about using the `managed-3scale` and at the time it seemed like the right name. As the initial `0.0.0` CSV is a place holder there will be no issue rerunning the script with the correct values "service name & version" when we are ready to do an official release.

philbrookes:
1. `0.0.0` is fine for now.
2. This should be `managed-api` now, can we rename the directory, if we merge after https://github.com/integr8ly/integreatly-operator/pull/1175 then this PR can use `managed-api` as the install type Param too.

Boomatang:
Thanks, I will keep an eye for #1175 to be merged and then make the changes to suit.

scripts/README.md
```diff
@@ -0,0 +1,17 @@
...
+This script creates a release csv for the operator under the `olm-catalog` using `operator-sdk generate csv`
+
+**Required System Variables**
+
+- SEMVER -> valid x.y.z format. Usage: `SEMVER=2.6.0`
+
+**Optional System Variables**
+
+- TYPE -> Which resource the release csv will be created under. Current Default: `integreatly-operator`.  Usage: `TYPE=managed-api-service`
+- ORG -> The org to where any images maybe stored on quay.io. Setting this will change the affect the location images
```
obrienrobert, line 12: `the affect the`: does this need to be reworded?

Boomatang: Yes, yes it does. Even when I read it not it don't make must sense not.

Boomatang: @obrienrobert I have just push just push an update if you would like to review

scripts/README.md
```diff
@@ -0,0 +1,16 @@
...
+- SEMVER -> valid x.y.z format. Usage: `SEMVER=2.6.0`
+
+**Optional System Variables**
+
+- TYPE -> Which resource the release csv will be created under. Current Default: `integreatly-operator`.  Usage: `TYPE=managed-api-service`
+- ORG -> The org to where any images maybe stored on quay.io. Setting the ORG will change the image locations specified in \<type>.\<version>.clusterserviceversion.yaml under `containerImage` and `image`. Default: `integreatly`. Usage: `ORG=<user-repo>`.
+
+**Usage**
+
+`SEMVER=2.6.0 TPYE=managed-api-service ORG=<user-repo> ./prepare-release.sh`
```
obrienrobert, line 16: Small one: `TPYE` -> `TYPE`.

deploy/olm-catalog/managed-api-service/managed-api.package-service.yaml
```diff
@@ -0,0 +1,5 @@
+channels:
+- currentCSV: managed-api-service.v0.0.0
```
laurafitzgerald, line 2: Interesting that v0.0.0 versioning works. Will we merge it with this?

scripts/prepare-release.sh
```diff
@@ -2,30 +2,50 @@
 set -e
 set -o pipefail
 
-PREVIOUS_VERSION=$(grep integreatly-operator deploy/olm-catalog/integreatly-operator/integreatly-operator.package.yaml | awk -F v '{print $2}')
+if [[ -z "$TYPE" ]]; then
+  TYPE="integreatly-operator"
+fi
+
+PREVIOUS_VERSION=$(grep $TYPE deploy/olm-catalog/$TYPE/$TYPE.package.yaml | awk -F v '{print $2}') || echo "No previous version"
```
laurafitzgerald, line 9: could we use jq here to pick out the currentCSV version? if there were any other instances of the $TYPE entry in the file this would pick up on it as well.

Something like

`yq r <file> channels[0].currentCSV`

I'm not super familiar with the manifests in integreatly-operator but I'm pretty sure we don't ever include historical channels so could fairly safely assume it's the first entry.

scripts/prepare-release.sh
```diff
@@ -2,30 +2,50 @@
...
 
 update_csv() {
-  operator-sdk generate csv --csv-version "$VERSION" --default-channel --operator-name integreatly-operator --csv-channel=rhmi --update-crds
+  operator-sdk generate csv --csv-version "$VERSION" --default-channel --operator-name "$TYPE" --csv-channel=rhmi --update-crds
 }
 
 set_version() {
-  "${SED_INLINE[@]}" "s/$PREVIOUS_VERSION/$VERSION/g" Makefile
-  "${SED_INLINE[@]}" "s/$PREVIOUS_VERSION/$VERSION/g" version/version.go
+  if [[ -z "$PREVIOUS_VERSION" ]]
```
laurafitzgerald, line 31: Looks like we need `$PREVIOUS_VERSION` to be set earlier in the script too. Wdyt about having that as an earlier check and not continuing unless it's set?

scripts/prepare-release.sh
```diff
@@ -2,30 +2,50 @@
...
+  ORG="$ORG"
+fi
 
 create_new_csv() {
-  operator-sdk generate csv --csv-version "$VERSION" --default-channel --operator-name integreatly-operator --csv-channel=rhmi --update-crds --from-version "$PREVIOUS_VERSION"
+  if [[ -z "$PREVIOUS_VERSION" ]]
+    then
+      operator-sdk generate csv --csv-version "$VERSION" --default-channel --operator-name "$TYPE" --csv-channel=rhmi --update-crds
+    else
+      operator-sdk generate csv --csv-version "$VERSION" --default-channel --operator-name "$TYPE" --csv-channel=rhmi --update-crds --from-version "$PREVIOUS_VERSION"
```
laurafitzgerald, line 22: I'm not sure but I think we are checking if there isn't an entry in $PREVIOUS_VERSION but then using it?

scripts/prepare-release.sh
```diff
@@ -2,30 +2,50 @@
 set -e
 set -o pipefail
 
-PREVIOUS_VERSION=$(grep integreatly-operator deploy/olm-catalog/integreatly-operator/integreatly-operator.package.yaml | awk -F v '{print $2}')
+if [[ -z "$TYPE" ]]; then
+  TYPE="integreatly-operator"
```
laurafitzgerald, line 6: Should we be really specific here? i.e. if there's no entry default to integreatly-operator but if there is entry check that it is either integreatly-operator _or_ managed-api-service?

deploy/olm-catalog/managed-api-service/managed-api.package-service.yaml
```diff
@@ -0,0 +1,5 @@
+channels:
```
laurafitzgerald, line 1: Should the file name here be `managed-api-service.package.yaml`?

laurafitzgerald: nvm. I see the comments now. I see that the directories etc should be renamed.

scripts/README.md
obrienrobert: I would reword this section: `The org to where any images maybe stored on quay.io`. Maybe something like `The quay.io org to where the images will be pushed` (or something to that effect).

deploy/olm-catalog/managed-api-service/managed-api.package-service.yaml
Boomatang: The plan was to have v0.0.0 there as a place holder till we were ready to do a release. Sense you are the second person to ask this maybe we should readdress this plan.

scripts/prepare-release.sh
Boomatang: This is just an update to the initial script to allow for parameters and didn't see the need to rewrite the whole thing.

On a personal I don't like the use of `yq` as its not a stand tool on most systems. But that is my personal opinion and would be will to change.

scripts/prepare-release.sh
Boomatang: The `-z` is checking if a string has a length of zero, so we are only using the `PREVIOUS_VERSION` in the `else` which means it has been set.

scripts/prepare-release.sh
Boomatang: If we set this check to exit the script earlier it would remove the possibility of creating the initial CSV files. These initial CSV file would be require been created in a manual process.

scripts/prepare-release.sh
Boomatang: I am not sure I fully understand the question been asked but here is what I think is the answer. 

The integreatly-operator is set as the default for now for two reasons.
1. The integreatly-operator still takes priority at least till the 2.7.1 has been released
2. There has been no release of the managed-api-service to date. I feel it would be hard set the default for something not existing yet

The question about checking for  integreatly-operator or managed-api-service, I feel where it would be a good idea there would be more work required which might be pushing the scope of the ticket a bit. It might be better to have sane defaults than trying to check every possibility.

scripts/README.md
Boomatang: I have done these changes locally, there maybe be other changes need to the script part. But yes `maybe` might not have been the right word to use.

Boomatang:
I know this is after getting a bit messy. I have push changes which should allow for the work done in #1210. 

The Makefile has a list of TODO marks. These markers are for function which I feel need to be changed to allow us to have different release types. If some one who has dealt with delorean could give a once over and let me know that yes the makred function need new targets.

There is also a question should the new targets be prefixed with rhoam or try place them in existing sections with a different naming format.?

@philbrookes @laurafitzgerald @pmccarthy  calling ye out as we spoke about this earlier. 
When this is ready for a full review again I will tidy up the commits :)

Makefile
```diff
@@ -1,20 +1,22 @@
 include ./make/*.mk
 
 ORG ?= integreatly
-NAMESPACE_PREFIX=redhat-rhmi-
+NAMESPACE_PREFIX ?= redhat-rhmi-
```
laurafitzgerald, line 4: is it too early to move this namespace prefix into the block below that manages the envs for each installation. @CathalOConnorRH what did we land on in the end for the managed api prefix?

Makefile
```diff
@@ -1,20 +1,22 @@
...
 NAMESPACE=$(NAMESPACE_PREFIX)operator
 PROJECT=integreatly-operator
+RHOAM_PROJECT=managed-api-service
```
laurafitzgerald, line 7: I guess this is the project in quay? can we use one PROJECT env and move it to the block below that manages the envs?

Makefile
```diff
@@ -1,20 +1,22 @@
...
 REG=quay.io
 SHELL=/bin/bash
-TAG ?= 2.7.0
+RHMI_TAG ?= 2.7.0
```
laurafitzgerald, line 10: Same with the tags?

Makefile
```diff
@@ -331,6 +355,7 @@ deploy/integreatly-rhmi-cr.yml:
 	sed "s/USE_CLUSTER_STORAGE/$(USE_CLUSTER_STORAGE)/g" > deploy/integreatly-rhmi-cr.yml
 	@-oc create -f deploy/integreatly-rhmi-cr.yml
 
+#TODO: new target required for rhoam
```
laurafitzgerald, line 358: I think this todo is out of date. Am I right in saying we wont need to add a new target but we'll update delorean to have a new flag which we can set the installatin type by?

Makefile
Boomatang: I will check with @CathalOConnorRH  again but from what I understood we are not ready to make this a full time change just. I think there is one thing missing but I will follow up again.

Makefile
Boomatang: Yes we can move it

Makefile
Boomatang: The tags can be moved but the naming would want to be different. The prepare-release.sh script changes the tag values when making a new release.

Makefile
Boomatang: I was nopeing to push this work off to a different ticket. Currently progressing the changes needed. So yes this todo is now out of date and is been worked on.

Makefile
CathalOConnorRH: I think it's too early for moving this yet, I think we were waiting on RC2 but I might have missed if thats been released. I think we should wait for this release to be in prod before we change the default

Boomatang: Ok @CathalOConnorRH I may have miss understood you. You were waiting to for RC2 before making the change for the default. I have added in a block which sets you the defaults depending on the install type. We can set the `NAMESPACE_PREFIX` in there. line 27 to 70 of this file. Not sure now to get a url link to work correctly. If the operator can run with the new name spaces we can do that change in the defaults block that I added.

Makefile
Boomatang: Waiting on this PR https://github.com/integr8ly/delorean/pull/148 to be merged, bu the change has been put in place

Makefile
```diff
@@ -334,20 +360,12 @@ deploy/integreatly-rhmi-cr.yml:
...
 prepare-patch-release:
 	$(CONTAINER_ENGINE) pull quay.io/integreatly/delorean-cli:master
-	$(CONTAINER_ENGINE) run --rm -e KUBECONFIG=/kube.config -v "${HOME}/.kube/config":/kube.config:z -v "${HOME}/.delorean.yaml:/.delorean.yaml" quay.io/integreatly/delorean-cli:master delorean release openshift-ci-release --config /.delorean.yaml --version $(TAG)
+	$(CONTAINER_ENGINE) run --rm -e KUBECONFIG=/kube.config -v "${HOME}/.kube/config":/kube.config:z -v "${HOME}/.delorean.yaml:/.delorean.yaml" quay.io/integreatly/delorean-cli:master delorean release openshift-ci-release --config /.delorean.yaml --olmType $(olmType) --version $(TAG)
 
 .PHONY: release/prepare
 release/prepare:
 	@./scripts/prepare-release.sh
 
-.PHONY: push/csv
```
laurafitzgerald, line 343: I've been using this recently. I think there was some discussion that we don't use it anymore. Is there an alternative way? If not can we leave this in?

scripts/prepare-release.sh
```diff
@@ -2,30 +2,75 @@
...
+    then
+      operator-sdk generate csv --csv-version "$VERSION" --default-channel --operator-name "$OLM_TYPE" --csv-channel=rhmi --update-crds --make-manifests=false
+    else
+      operator-sdk generate csv --csv-version "$VERSION" --default-channel --operator-name "$OLM_TYPE" --csv-channel=rhmi --update-crds --from-version "$PREVIOUS_VERSION" --make-manifests=false
+  fi
 }
 
 update_csv() {
-  operator-sdk generate csv --csv-version "$VERSION" --default-channel --operator-name integreatly-operator --csv-channel=rhmi --update-crds --make-manifests=false
+  operator-sdk generate csv --csv-version "$VERSION" --default-channel --operator-name "$OLM_TYPE" --csv-channel=rhmi --update-crds --make-manifests=false
```
laurafitzgerald, line 40: Do you know if this is run using your local operator-sdk or does it pick up the OPERATOR_SDK in the same way as Makefile?

Makefile
```diff
@@ -334,20 +360,12 @@ deploy/integreatly-rhmi-cr.yml:
 .PHONY: prepare-patch-release
 prepare-patch-release:
 	$(CONTAINER_ENGINE) pull quay.io/integreatly/delorean-cli:master
-	$(CONTAINER_ENGINE) run --rm -e KUBECONFIG=/kube.config -v "${HOME}/.kube/config":/kube.config:z -v "${HOME}/.delorean.yaml:/.delorean.yaml" quay.io/integreatly/delorean-cli:master delorean release openshift-ci-release --config /.delorean.yaml --version $(TAG)
+	$(CONTAINER_ENGINE) run --rm -e KUBECONFIG=/kube.config -v "${HOME}/.kube/config":/kube.config:z -v "${HOME}/.delorean.yaml:/.delorean.yaml" quay.io/integreatly/delorean-cli:master delorean release openshift-ci-release --config /.delorean.yaml --olmType $(olmType) --version $(TAG)
```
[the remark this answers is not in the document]
Boomatang: Good catch

scripts/prepare-release.sh
Boomatang: I'm going to say it uses the local operator-sdk , if the script is directly ran. I am not even sure if you were to use the `make release/prepare` command would it pick up the correct version.

Makefile
Boomatang: With the time constrain I have added this back in. Not sure if it was updated correctly to work with but rhmi and rhoam. It's some thing I don't use so not sure the best way to test.

scripts/prepare-release.sh
laurafitzgerald: It picks it up here: https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L35-L39

Makefile
```diff
@@ -1,27 +1,26 @@
...
-INTEGREATLY_OPERATOR_IMAGE ?= $(REG)/$(ORG)/$(PROJECT):v$(TAG)
-MANAGED_API_OPERATOR_IMAGE ?= $(REG)/$(ORG)/managed-api-service:v$(MGDAPI_TAG)
+
+
 CONTAINER_ENGINE ?= docker
 TEST_RESULTS_DIR ?= "test-results"
 TEMP_SERVICEACCOUNT_NAME="rhmi-operator"
 CLUSTER_URL:=$(shell sh -c "oc cluster-info | grep -Eo 'https?://[-a-zA-Z0-9\.:]*'")
 
+RHMI_TAG ?= 2.7.0
```
laurafitzgerald, line 21: I see that these get assigned to TAG in the if statement below. Could we just move them down there?

Makefile
```diff
@@ -48,18 +47,42 @@ endif
...
+	export OLM_TYPE ?= integreatly-operator
+endif
+
+ifeq ($(INSTALLATION_TYPE), managed-api)
+	PROJECT=managed-api-service
+	TAG ?= RHOAM_TAG
+	OPERATOR_IMAGE=$(REG)/$(ORG)/$(PROJECT):v$(TAG)
+	NAMESPACE_PREFIX ?= redhat-rhoam-
+	APPLICATION_REPO ?= managed-api-service
+	export INSTALLATION_PREFIX ?= redhat-rhoam
```
laurafitzgerald, line 72: This is used in the path of file to set up the service account running locally. https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L73
In this case I believe we need it to be `redhat-managed-api` unless there are plans to rename the folder.

Makefile
```diff
@@ -48,18 +47,42 @@ endif
...
 export OPERATORS_IN_PRODUCT_NAMESPACE ?= false # e2e tests and createInstallationCR() need to be updated when default is changed
 export DELOREAN_PULL_SECRET_NAME ?= integreatly-delorean-pull-secret
 export ALERTING_EMAIL_ADDRESS = noreply-test@rhmi-redhat.com
 
+ifeq ($(INSTALLATION_TYPE), managed)
+	PROJECT=integreatly-operator
+	TAG ?= RHMI_TAG
+	OPERATOR_IMAGE=$(REG)/$(ORG)/$(PROJECT):v$(TAG)
+	NAMESPACE_PREFIX ?= redhat-rhmi-
+	APPLICATION_REPO ?= integreatly
```
laurafitzgerald, line 61: APPLICATION_REPO confuses me. In the case of managed we are saying it's `integreatly` but in the case of managed-api` we are saying `managed-api-service`. I'm assuming this relates to application in QUAY? Do we use the application `integreatly` for rhmi and the application `managed-api-service`. I think it's probably fine but just want to be sure I'm understanding it correctly. Seems it's only used for operator-courier anyways right?

Makefile
laurafitzgerald: By the way, it's possible I only used this because of some legacy knowledge I have about installing RHMI. If there is another way that this is to be used, I'd like to learn about it. 😄

README.md
```diff
@@ -1,5 +1,15 @@
 # Integreatly Operator
 
+---
+**NOTE**
```
laurafitzgerald, line 4: Could we also leave a similar note above the INSTALLATION_TYPE variable in the Makefile. I see that it's being left as managed (for now?). Just thinking that it should be clear that if you want to do anything in the makefile.

laurafitzgerald (commented):
Hi @Boomatang just doing a short review here. Some questions inline.

Makefile
Boomatang: These tags are modified by the prepare-release script. In the `if` blocks the TAG is assigned  with the correct value. This value is used in the image url which is also set at this stage. Assigned the TAG value also meant not having to change any existing make targets that maybe using the TAG. Possible I should put in a comment about why the tags are done this way, Phil also asked the question.

README.md
Boomatang: sure

Makefile
Boomatang: I think this is fine. These values have to do with the update_csv target which you commented about how you still use. This will set the defaults that should be used if its ran for the ORG of integreatly. Which I don't think it every will be.

Makefile
Boomatang: Yes this was something I was miss understanding and it needs to be changed. I will also check the file for the any other cases where this is been used.

scripts/prepare-release.sh
Boomatang: I think we can mark this as resolved. Call the script through the make target will use the correct operator-sdk while calling the script directly will used the default system version. From what I have seen this is the same for other scripts in the that scripts folder. The issue I can see is. Delorean calls that script directly, would delorean use it's version of the operator-sdk or in the pipelines does the correct version of the operator-sdk get installed?

PR #1157, opened 2020-09-01
New test case for rhmidoc-241

bhardesty:
# Description
Doc test case to verify content no longer contains any references to config changes through Solution Explorer.

## Type of change

## Checklist


test-cases/tests/documentation/i16-verify-partial-ag-rhmi-config.md
```diff
@@ -0,0 +1,28 @@
+---
+environments:
+  - osd-fresh-install
+tags:
+  - manual-selection
+---
```
b1zzu, line 6:
```suggestion
  - manual-selection
targets:
  - 2.7.0
---
```
Just need to add also the 2.7.0 target otherwise it will not be executed during the 2.7.0 release

PR #1174, opened 2020-09-10
update the test case F04 change the output of the command to json

austincunningham:
# Description

Update the steps in test case F04 as the AWS command returns json

jira: https://issues.redhat.com/browse/INTLY-10019

## Type of change
update test cast markdown

## Checklist


psturc:
need to update the PR after running `npm run prettier`

PR #1175, opened 2020-09-10
rename from managed-3scale to managed-api

briangallagher:
# Description
rename from managed-3scale to managed-api

## Verify
**Install managed-api**
1. Install the managed-api operator and verify 
1.1 `export INSTALLATION_TYPE=managed-api`
3. Run the functional tests `make test/functional`
4. Verify all the tests in the managed api install type test suite ran

**Install rhmi**
1. Install the rhmi operator and verify 
1.1 `export INSTALLATION_TYPE=rhmi`
3. Run the functional tests `make test/functional`
4. Verify all the tests in the rhmi install type test suite ran

### DIFF
--- a/pkg/controller/installation/installation_controller.go
+++ b/pkg/controller/installation/installation_controller.go
@@ -47,6 +47,7 @@ import (
 const (
 	deletionFinalizer                = "finalizer/configmaps"
 	DefaultInstallationName          = "rhmi"
+	ManagedApiInstallationName       = "managed-api"
 	DefaultInstallationConfigMapName = "installation-config"
 	DefaultInstallationPrefix        = "redhat-rhmi-"
 	DefaultCloudResourceConfigName   = "cloud-resource-config"
@@ -159,7 +160,7 @@ func createInstallationCR(ctx context.Context, serverClient k8sclient.Client) er
 
 		installation = &integreatlyv1alpha1.RHMI{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      DefaultInstallationName,
+				Name:      getCrName(installType),
 				Namespace: namespace,
 			},
 			Spec: integreatlyv1alpha1.RHMISpec{
@@ -186,6 +187,14 @@ func createInstallationCR(ctx context.Context, serverClient k8sclient.Client) er
 	}
 
 	return nil
+}
+
+func getCrName(installType string) string {
+	if installType == string(integreatlyv1alpha1.InstallationTypeManagedApi) {
+		return ManagedApiInstallationName
+	} else {
+		return DefaultInstallationName
+	}
 }
 
 var _ reconcile.Reconciler = &ReconcileInstallation{}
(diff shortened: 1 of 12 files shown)


trepel (changes requested):
What about this occurrence, should not it be changed as well?

https://github.com/integr8ly/integreatly-operator/blob/master/deploy/cro-configmaps.yaml#L18

briangallagher:
> @briangallagher What about this occurrence, should not it be changed as well?
> 
> https://github.com/integr8ly/integreatly-operator/blob/master/deploy/cro-configmaps.yaml#L18

Yes, thanks!

Makefile
```diff
@@ -43,8 +43,8 @@ else
 endif
 
 export SELF_SIGNED_CERTS   ?= true
-export INSTALLATION_TYPE   ?= managed
-export INSTALLATION_NAME   ?= rhmi
+export INSTALLATION_TYPE   ?= managed-api
+export INSTALLATION_NAME   ?= managed-api
```
trepel, line 47: I think we should stick with defaulting to 'rhmi' at this point. Defaulting to managed-api would require some README.md updates, changing the INSTALLATION_PREFIX default value and (likely) more.

trepel (changes requested):

Makefile
briangallagher: @trepel Is that in relation to the type, name or both?

briangallagher: @philbrookes You specifically mentioned changing the installation name, do you have a strong opinion on this?

trepel: Both. I definitely agree that Manged API should be default in the end, I am just not sure if everything has to be addressed in one big PR. I leave that up to you guys :)

Just make sure installation of RHMI is also part of verification steps.

philbrookes: Sorry, I missed this ping. I think the default behaviour of the operator should be RHMI, it should only do managed-api when we pass in the env var for installation type.

pkg/controller/installation/installation_controller.go
```diff
@@ -154,12 +155,12 @@ func createInstallationCR(ctx context.Context, serverClient k8sclient.Client) er
 		logrus.Infof("Creating a %s rhmi CR with USC %s, as no CR rhmis were found in %s namespace", installType, useClusterStorage, namespace)
 
 		if installType == "" {
-			installType = string(integreatlyv1alpha1.InstallationTypeManaged)
+			installType = string(integreatlyv1alpha1.InstallationTypeManagedApi)
```
philbrookes, line 158: Same here, this should stay on managed

philbrookes:
I think once the default behaviour is switched back to managed RHMI, this PR is good to go.

pkg/controller/installation/installation_controller_test.go
```diff
@@ -206,7 +206,7 @@ func TestCreateInstallationCR_installationTypeDefault(t *testing.T) {
 	mockClient := fake.NewFakeClientWithScheme(buildScheme())
 	ctx := context.TODO()
 
-	installationType := "managed"
+	installationType := "managed-api"
```
trepel, line 209: Should this stay on managed too then?

philbrookes: I think it's ok for the tests to default to managed-api, after 2.7.1 we should not be doing many releases for RHMI.

philbrookes: Hmm, though actually, do the tests pass like this? I feel like a bunch will fail due to missing namespaces, etc.

briangallagher: @philbrookes They've been split based on type and further refined here:
https://issues.redhat.com/browse/MGDAPI-46

trepel: We have the pipeline in place so we will monitor the tests daily. There was just one failure last night.

I just think it is a little strange for the tests defaulting to 'managed-api' whereas the operator itself defaulting to 'managed-api'. But I am ok with that.

PR #1190, opened 2020-09-17
MGDAPI-89: Update ConsoleLink reconciliation

sergioifg94:
# Description

Update reconciliation of ConsoleLink CRs to:

1. Add link to SSO as "API Management SSO"
2. Rename link to 3scale as "API Management"
3. Group these links into "OpenShift Managed Services" section

## Verification steps

In order to fully verify this PR:

1. Install the operator from master
2. Once the operator is installed, check that there's a 3scale link in the Applications menu
3. Change the operator to this PR
4. Wait for it to start and fully reconcile
5. Verify that the Applications menu contains:
    ```
    OpenShift Managed Services

    API Management
    API Management SSO
    ```
6. Uninstall the operator by deleting the RHMI CR
7. Verify that the links have been removed

## Type of change

## Checklist

### DIFF
--- a/pkg/products/rhsso/reconciler.go
+++ b/pkg/products/rhsso/reconciler.go
@@ -10,6 +10,7 @@ import (
 	"github.com/integr8ly/integreatly-operator/pkg/resources/events"
 	userHelper "github.com/integr8ly/integreatly-operator/pkg/resources/user"
 
+	consolev1 "github.com/openshift/api/console/v1"
 	corev1 "k8s.io/api/core/v1"
 
 	keycloakCommon "github.com/integr8ly/keycloak-client/pkg/common"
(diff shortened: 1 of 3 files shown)


pkg/products/rhsso/reconciler.go
```diff
@@ -195,6 +202,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
+	if err := r.reconcileConsoleLink(ctx, serverClient); err != nil {
```
philbrookes, line 205: I think we should only be doing this on managed-api install type.

sergioifg94: Good point. Should we also be keeping the previous text for 3scale ("3scale API Management") when it's not managed-api?

philbrookes: I think calling it API Management in RHMI 2.x is fine.

philbrookes: I don't think it's used much in RHMI 2.x anyway, I'd say it's mostly accessed from the solution explorer.

sergioifg94: Cool, I just updated the PR with the condition to ensure the ConsoleLink is not created if the type isn't managed-api:
https://github.com/integr8ly/integreatly-operator/pull/1190/files#diff-74bcf7e72cc996549e6ed196e756ab37R410-R414

PR #1193, opened 2020-09-18
disabling AZ script

austincunningham:
# Description

Jira: https://issues.redhat.com/browse/MGDAPI-53
Need a method to disable an aws availability zone. This script creates dummy ACL(deny all traffic) on aws and points all subnets on a particular AZ to these dummy ACL. 

## Type of change
Script to disable and AZ and re-enable again

### Setup
- Deploy a multi AZ byoc
- Change the rhmi cr to deploy `type: managed-api` and `useClusterStorage: 'false'` 
- run `make cluster/prepare/local`
- run ` oc create -f deploy/crds/examples/rhmi.cr.yaml`
- run `make code/run`
### Test the script
Before running the script check the ACL for the account 
should look something like
image

Disable AZ
```bash
./disableAz.sh
No option set expect params true/false, the default option is disable which is true
true - disable the AvailabilityZone
false - restore the original configuration (run only after a successful disable)
Disabling AvailabilityZone
acl-0e6064637db103dbd
aclassoc-0ae2f313cd55928f3
{
    "NewAssociationId": "aclassoc-04218d21fb9162c71"
}
acl-0230cb9b94f332966
aclassoc-013c5e171044eb352
{
    "NewAssociationId": "aclassoc-0867d96a9057fe41c"
}
acl-0344cf512e041926c
aclassoc-0966c189592afdd8a
{
    "NewAssociationId": "aclassoc-00cec7884f9acc1e5"
}
```
Check the ACL after running the command and you should see the new associations
image

You can also check the machineSet although these can take a while to change
```bash
 oc get machineset -n openshift-machine-api
NAME                                      DESIRED   CURRENT   READY   AVAILABLE   AGE
mw-collab-multi-zmvv6-infra-eu-west-1a    1         1         1       1           3d
mw-collab-multi-zmvv6-infra-eu-west-1b    1         1         1       1           3d
mw-collab-multi-zmvv6-infra-eu-west-1c    1         1                             3d
mw-collab-multi-zmvv6-worker-eu-west-1a   3         3         3       3           3d1h
mw-collab-multi-zmvv6-worker-eu-west-1b   3         3         3       3           3d1h
mw-collab-multi-zmvv6-worker-eu-west-1c   3         3                             3d1h

```

Re-enable the AZ
```bash
./disableAz.sh false
Re-enable AvailabilityZone
acl-0e6064637db103dbd
aclassoc-04218d21fb9162c71
{
    "NewAssociationId": "aclassoc-091984c33350a6634"
}
acl-0230cb9b94f332966
aclassoc-0867d96a9057fe41c
{
    "NewAssociationId": "aclassoc-0d876901f17655855"
}
acl-0344cf512e041926c
aclassoc-00cec7884f9acc1e5
{
    "NewAssociationId": "aclassoc-05d59a48b57dd5ba8"
}
```
recheck the ACL after running should see them return to normal
image

you can also check the MachineSet but these take a while to reconcile
```bash
oc get machineset -n openshift-machine-api
NAME                                      DESIRED   CURRENT   READY   AVAILABLE   AGE
mw-collab-multi-zmvv6-infra-eu-west-1a    1         1         1       1           3d
mw-collab-multi-zmvv6-infra-eu-west-1b    1         1         1       1           3d
mw-collab-multi-zmvv6-infra-eu-west-1c    1         1         1       1           3d
mw-collab-multi-zmvv6-worker-eu-west-1a   3         3         3       3           3d1h
mw-collab-multi-zmvv6-worker-eu-west-1b   3         3         3       3           3d1h
mw-collab-multi-zmvv6-worker-eu-west-1c   3         3         3       3           3d1h
```

## Checklist

### DIFF
--- a/scripts/disableAz.sh
+++ b/scripts/disableAz.sh
@@ -7,8 +7,7 @@
 #  - aws-cli
 
 # check the correct command line args are passed two expected
-if (( $# < 2 ))
-then
+if (( $# < 2 )); then
   echo "Incorrect amount of arguments passed in"
  	echo "usage ./disableAz.sh <true/false> <aws az>"
  	exit 1
@@ -16,8 +15,7 @@ fi
 
 # Checks the command line args $1 true to disable and false to re-enable
 DISABLE=$1
-if [[ "$DISABLE" == "true" ]] || [[ "$DISABLE" == "false" ]]
-then
+if [[ "$DISABLE" == "true" ]] || [[ "$DISABLE" == "false" ]]; then
   echo "Arg $1 accepted"
 else
  	echo "Incorrect option for first arg expect true/false"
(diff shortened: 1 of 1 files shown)


psturc:
👀

scripts/disableAz.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+# Usage: disable and renable an AZ
+#
+# prereq
+#  - jq
+#  - aws-cli
+
+# AZ to disable
+AZ=eu-west-1a
```
psturc, line 10: I was trying it out with a different region than I have stored in `~/.aws/config`

i.e. content of my `~/.aws/config`:
```
[default]
region = eu-central-1
```

and the script didn't work, because all aws commands are automatically targeting `eu-central-1` region

I think if all aws commands in this script would contain additional parameter `--region ${AZ:0:-1}`, it should work regardless what user has stored in `~/.aws/config`. Wdyt?

Also, do you think it would make sense to add the AZ as an optional parameter? (i.e. to use it like this: `./scripts/disableAz.sh --az eu-central-1c`)

psturc:
LGTM, just a few comments.

scripts/disableAz.sh
austincunningham: Thanks @psturc , yea makes sense.

psturc (commented):
changes look good, going to verify them in a sec...

scripts/disableAz.sh
```diff
@@ -0,0 +1,100 @@
...
+# Function changeAcl takes two arguments for disable or enable
+# $1 should be NetworkAclAssociationId filename
+# $2 should be NetworkAclId filename
+function ChangeAcl() {
+	count=1
+	cat $1 | while read NetworkAclAssociationId
+	do
+		echo $(sed -n "${count}p" < NetworkAclId.tmp)
+		echo $NetworkAclAssociationId
+		aws ec2 replace-network-acl-association --region ${AZ:0:-1} --association-id $NetworkAclAssociationId --network-acl-id $(sed -n "${count}p" < $2)
```
psturc, line 58: Sorry, just found out that when using `bash` for running the script, the expression `${AZ:0:-1}` is not compatible (it works when using zsh)

I created a branch with the suggested change that is compatible with bash and zsh as well
https://github.com/austincunningham/integreatly-operator/compare/MGDAPI-53-script-disable-az...psturc:MGDAPI-53-script-disable-az-bash-compatible?expand=1

scripts/disableAz.sh
```diff
@@ -0,0 +1,100 @@
...
+	    ((count=count+1))
+	done
+}
+
+if $DISABLE
+then
+  echo "Disabling AvailabilityZone"
+  # use the subnetId to get the NetworkAclAssociationId to create the new acl association and get the NetworkAclId so can revert the change
```
psturc, line 66: do you think it would make sense to remove following files

	NetworkAclAssociationId.tmp
	NetworkAclId-restore.tmp
	NetworkAclId.tmp

to make sure there are no such files from previous run of this script?

scripts/disableAz.sh
austincunningham: :+1:

scripts/disableAz.sh
austincunningham: Yea it would :100:

PR #1194, opened 2020-09-21
Adding initial marin3r code requiring bump to operator sdk v0.17.1

laurafitzgerald:
# Description
JIRA: https://issues.redhat.com/browse/MGDAPI-93

Changes included in this pr:
- [x] Add marin3r dependency
- [x] Bump operator-sdk version dependency
- [x] Bump operator-sdk version in Makefile and openshift-ci iamge
- [x] Add marin3r manifest
- [x] Add marin3r reconciler

## Verification

We probably want to verify both paths but the dependency change here is triggered by the addition of marin3r so the managed-api install type is more important.

To install that - update the file `deploy/crds/examples/rhmi.cr.yaml` spec.type to `managed-api`
 and/or update the INSTALLATION_TYPE value in the Makefile to `managed-api`

Against an Openshift cluster:
Run `make cluster/prepare`
Run `make code/run`

Verify that the `redhat-rhmi-marin3r-operator` and `redhat-rhmi-marin3r` namespaces are installed and that the finalizer `finalizer.marin3r.integreatly.org` is added to the rhmi cr.

Delete the cr.

Verify that the two namespaces are deleted.

## Type of change

## Checklist


pkg/controller/installation/types.go
```diff
@@ -45,6 +45,7 @@ var (
 				Products: map[integreatlyv1alpha1.ProductName]integreatlyv1alpha1.RHMIProductStatus{
 					integreatlyv1alpha1.Product3Scale:    {Name: integreatlyv1alpha1.Product3Scale},
 					integreatlyv1alpha1.ProductRHSSOUser: {Name: integreatlyv1alpha1.ProductRHSSOUser},
+					integreatlyv1alpha1.ProductMarin3r:   {Name: integreatlyv1alpha1.ProductMarin3r},
```
CathalOConnorRH, line 48: We should probably add this to the products uninstall stage too ?

laurafitzgerald: Yes. I will add it. 👍

Makefile
```diff
@@ -108,7 +108,7 @@ pkg/apis/integreatly/v1alpha1/zz_generated.deepcopy.go:	pkg/apis/integreatly/v1a
 
 .PHONY: code/gen
 code/gen: setup/moq deploy/crds/integreatly.org_rhmis_crd.yaml pkg/apis/integreatly/v1alpha1/zz_generated.deepcopy.go pkg/apis/integreatly/v1alpha1/zz_generated.openapi.go
-	find ./ -name *_moq.go -type f -not -path "./vendor/*" -delete
+	# find ./ -name *_moq.go -type f -not -path "./vendor/*" -delete
```
philbrookes, line 111: any reason not to delete this? No need to hold up the PR or anything, just curious

laurafitzgerald: yes I think it can be deleted. @grdryn do you recall what problem this one was causing that we removed it?

pkg/resources/reconciler_test.go
```diff
@@ -206,28 +206,6 @@ func TestReconciler_reconcilePullSecret(t *testing.T) {
 		Config       *config.ConfigReadWriterMock
 		Validate     func(c k8sclient.Client) error
 	}{
-		{
```
laurafitzgerald, line 209: Note: this test case is being removed because the function logic has changed. In the past it used to provide a default but that's not the case any more after some refactoring. Not sure how it wasn't failing before but it will fail now because there's no default provided. Also the function being tested would never run if there is no Secret Object in the RHMI.

CathalOConnorRH:
I have validated the SDK changes on cluster. This should be fine to merge. 
/lgtm

PR #1199, opened 2020-09-23
update manifests to follow correct format for rhmi

laurafitzgerald:
# Description

https://issues.redhat.com/browse/MGDAPI-178

- [x] update manifests
- [x] add deployment of the operator
- [x] add creation of the discovery service
- [x] add creation of the required secrets

## Verification

- update the file `deploy/crds/examples/rhmi.cr.yaml` spec.type to `managed-api`
 and/or update the INSTALLATION_TYPE value in the Makefile to `managed-api`

Against an Openshift cluster:
Run `make cluster/prepare`
Run `make code/run`

Verify that the following is in place
-  Inside Namespace - redhat-rhmi-marin3r
- marin3r instance service with related pod
- Secrets titled `marin3r-server-cert-instance` and `marin3r-ca-cert-instance` 

## Type of change

## Checklist


sergioifg94:
Code changes look good, and verified in the cluster that the service and secrets are created, as well as the marin3r operator is deployed and reconciled correctly. I'm just noticing an error in the marin3r operator logs. Seems like it's trying to create the `marin3r-ca-cert-instance` `DiscoveryServiceCertificate` and it already exists. Not sure if it's a bug in the operator itself:

```
2020-09-30T09:05:30.245Z	DEBUG	controller_dicoveryservice	Reconciling CA certificate	{"Request.Name": "instance"}
2020-09-30T09:05:30.272Z	ERROR	controller-runtime.controller	Reconciler error	{"controller": "discoveryservice-controller", "request": "/instance", "error": "discoveryservicecertificates.operator.marin3r.3scale.net \"marin3r-ca-cert-instance\" already exists"}
```

sergioifg94:
Verified that the operator is now installed and initialized correctly. It still logs a couple of errors as it fails to find the secrets for the certificates, but it stabilizes soon after and I can see that the secrets and the DiscoveryServiceCertificate CRs are reconciled successfully.
There's only one thing I noticed and it's that when uninstalling, the DiscoveryService is not deleted (as it's not bound to the `redhat-rhmi-marin3r(-operator)` namespace), maybe we should add some explicit logic to delete it in the finalizer?

laurafitzgerald:
@sergioifg94 awesome stuff. I agree that the Discover Service should be cleaned up. Can that be done by making the integreatly cr the owner?

Something like? 

```_, err = controllerutil.CreateOrUpdate(ctx, client, discoveryservice, func() error {
		owner.AddIntegreatlyOwnerAnnotations(discoveryservice, r.installation)
		return nil
	})```

sergioifg94:
@laurafitzgerald I thought about it but I think this wouldn't work as it's a cluster-scoped CR:
From [the k8s documentation](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents):

> Note:
Cross-namespace owner references are disallowed by design. This means:
> 
> 1. Namespace-scoped dependents can only specify owners in the same namespace, and owners that are cluster-scoped.
> 2. **Cluster-scoped dependents can only specify cluster-scoped owners, but not namespace-scoped owners.**

laurafitzgerald:
Cool cool cool, clean up in the finalizer it is so. Would you like me to make that change or do you want it? 😄

sergioifg94:
I'm happy to do it myself, and I've the cluster perfectly set up for testing it atm :stuck_out_tongue:

sergioifg94:
Just pushed the change to delete the DiscoveryService. All looking good!
/lgtm

PR #1202, opened 2020-09-24
Managed-api via addon e2e test flakes

MStokluska:
# Description

Skipping h07/h11/h03 failure message after failing to log in to 3scale. Issue reported in https://issues.redhat.com/browse/INTLY-10087

Skipping B04 due to flaky behavior on managed-api installation via addon. Issue reported in https://issues.redhat.com/browse/INTLY-10156

Editing A14 to accept equal or greater than the expected number of replicas. During managed-api installation via an addon, at some stage, system-app replicas go up to 3 for a brief moment, while the expected value is 2. Issue reported in https://issues.redhat.com/browse/INTLY-10157

Increasing timout on alerting mechanism test from 25mins to 40mins as the test requires more time on osde2e to complete

### DIFF
--- a/test/common/deployment_types.go
+++ b/test/common/deployment_types.go
@@ -258,7 +258,7 @@ func TestDeploymentExpectedReplicas(t *testing.T, ctx *TestingContext) {
 				continue
 			}
 
-			if deployment.Status.Replicas != product.ExpectedReplicas {
+			if deployment.Status.Replicas < product.ExpectedReplicas {
 				t.Errorf("Deployment %s in namespace %s doesn't match the number of expected replicas. Replicas: %v / Expected Replicas: %v",
 					product.Name,
 					namespace.Name,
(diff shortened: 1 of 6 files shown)


test/common/deployment_types.go
```diff
@@ -258,7 +258,7 @@ func TestDeploymentExpectedReplicas(t *testing.T, ctx *TestingContext) {
 				continue
 			}
 
-			if deployment.Status.Replicas != product.ExpectedReplicas {
+			if deployment.Status.Replicas != product.ExpectedReplicas && deployment.Status.Replicas < product.ExpectedReplicas {
```
KevFan, line 261: Just wondering is the first condition,`deployment.Status.Replicas != product.ExpectedReplicas`, of the `if` blocks still necessary ? 

I think `deployment.Status.Replicas < product.ExpectedReplicas` would cover the check for replicas to be greater or equal to the expected number of replicas :thinking:

KevFan: @MStokluska

MStokluska: Thanks @KevFan that makes sense! updated the PR

PR #1203, opened 2020-09-24
MGDAPI-90 Add customer facing Grafana to managed-api

CathalOConnorRH:
# Description
Added grafana operator and instance to managed-api install type

To verify: 
1, Run managed-api install. 
2, Confirm redhat-rhmi-customer-monitoring-operator product namespace is created.
3. Confirm grafana-operator pod exists and is ready
4. Confirm grafana-deployment pod exists and is ready
5. Confirm grafana route exists and works. 
6. Login to grafana instance using kubeadmin or dedicated-admin user. 

## Type of change


pkg/products/grafana/reconciler.go
```diff
@@ -0,0 +1,320 @@
...
+			Name:      "grafana",
+			Namespace: r.Config.GetOperatorNamespace(),
+		}, Spec: grafanav1alpha1.GrafanaSpec{
+			Config: grafanav1alpha1.GrafanaConfig{
+				Log: &grafanav1alpha1.GrafanaConfigLog{
+					Mode:  "console",
+					Level: "warn",
+				},
+				Auth: &grafanav1alpha1.GrafanaConfigAuth{
+					DisableLoginForm:   &[]bool{false}[0],
```
philbrookes, line 185: this is nuts, is this because it has to be a pointer to a bool?

CathalOConnorRH: I think so, the error I get when using just true/false is 
cannot convert false (untyped bool constant) to *bool
Is there a better way to do it ?

philbrookes: no, this is fine, just looks odd! :)

PR #1204, opened 2020-09-25
MGDAPI-193: Only reconcile backups blob storage on Managed type

sergioifg94:
# Description

Add condition in CRO reconciler to only reconcile the backups blob storage when
the InstallationType is `managed`.

Modify e2e test to not check for this blob storage when the installation type
is `managed-api`

## Verification steps

1. Install the Managed API Operator
2. Check that the installation completes successfully
3. Check that only the 3scale blobstorage has been created:
    ```sh
    $ oc get blobstorages -n redhat-rhmi-operator

    NAME                                 AGE
    threescale-blobstorage-managed-api   17h
    ```

## Type of change

## Checklist


sergioifg94:
Rebased after #1199 was merged due to conflict

PR #1205, opened 2020-09-25
MGDAPI-90_Install_Grafana

CathalOConnorRH:
# Description
Added grafana operator and instance to managed-api install type

To verify: 
1, Run managed-api install. 
2, Confirm redhat-rhmi-customer-monitoring-operator product namespace is created.
3. Confirm grafana-operator pod exists and is ready
4. Confirm grafana-deployment pod exists and is ready
5. Confirm grafana route exists and works. 
6. Login to grafana instance using kubeadmin or dedicated-admin user. 

## Type of change


pkg/products/grafana/reconciler.go
```diff
@@ -0,0 +1,320 @@
...
+	})
+
+	if err != nil {
+		logrus.Errorf("failed to create grafana proxy secret, %s: %w", secret, err)
+	}
+
+	return nil
+}
+
+func (r *Reconciler) reconcileComponents(ctx context.Context, client k8sclient.Client, installation *integreatlyv1alpha1.RHMI, cr *grafanav1alpha1.Grafana) (integreatlyv1alpha1.StatusPhase, error) {
```
laurafitzgerald, line 165: Is it required to pass the cr here? I don't see it used in the function.

pkg/products/grafana/reconciler.go
```diff
@@ -0,0 +1,320 @@
...
+		catalogSourceReconciler,
+	)
+}
+
+func (r *Reconciler) preUpgradeBackupExecutor() backup.BackupExecutor {
+	return backup.NewNoopBackupExecutor()
+}
+
+// PopulateSessionProxySecret generates a session secret
+func PopulateSessionProxySecret() string {
```
laurafitzgerald, line 303: Does this need to be exported. I think it's used somewhere else as well. Is it exported there too? Could we combine or just have it internal.

pkg/products/grafana/reconciler.go
```diff
@@ -0,0 +1,320 @@
...
+	p, err := GeneratePassword(43)
+	if err != nil {
+		logrus.Info("Error executing PopulateSessionProxySecret")
+	}
+	return p
+}
+
+// GeneratePassword returns a base64 encoded securely random bytes.
+func GeneratePassword(n int) (string, error) {
```
laurafitzgerald, line 312: Same here.

pkg/products/grafana/reconciler.go
```diff
@@ -0,0 +1,320 @@
...
+
+func (r *Reconciler) preUpgradeBackupExecutor() backup.BackupExecutor {
+	return backup.NewNoopBackupExecutor()
+}
+
+// PopulateSessionProxySecret generates a session secret
+func PopulateSessionProxySecret() string {
+	p, err := GeneratePassword(43)
+	if err != nil {
+		logrus.Info("Error executing PopulateSessionProxySecret")
```
laurafitzgerald, line 306: Do we need to fail the reconcile if the secret isn't generated or is it fine to log the error? Could p be nil on the return in this case and cause some problems with the install?

pkg/products/grafana/reconciler.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	phase, err = r.ReconcileNamespace(ctx, operatorNamespace, installation, client)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, fmt.Sprintf("Failed to reconcile %s ns", operatorNamespace), err)
+		return phase, err
+	}
+
+	err = r.createSecrets(ctx, client, installation, &grafanav1alpha1.Grafana{})
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
```
laurafitzgerald, line 123: I think we should either remove the phase check here and return a new phase or have the `createSecrets` function return a phase failure rather than a standard err. if it fails to create the secret it's going to rerun the previous phase from above as it's not updated here.

laurafitzgerald (commented):
@CathalOConnorRH I've left a few comments inline. Overall this change looks great. I haven't done the verification yet but happy to do that in a while.

pkg/products/grafana/reconciler.go
CathalOConnorRH: Good spot, Have removed it.

pkg/products/grafana/reconciler.go
CathalOConnorRH: No real need but the other exported version is in application monitoring which I dont want to rely on always being there. I've stopped the export but left the function in.

pkg/products/grafana/reconciler.go
CathalOConnorRH: No real need but the other exported version is in application monitoring which I dont want to rely on always being there. I've stopped the export but left the function in.

pkg/products/grafana/reconciler.go
CathalOConnorRH: I have updated this to reconcile and report phase failed.

pkg/products/grafana/reconciler.go
CathalOConnorRH: I have updated this to reconcile and report phase failed.

laurafitzgerald:
Worked through the verification steps. All working except for one permission thing solved by `oc adm policy add-cluster-role-to-user system:auth-delegator -z grafana-serviceaccount -n redhat-rhmi-customer-monitoring-operator` Once that's in place I think this is in really good shape. After that was applied everything worked as described.

CathalOConnorRH:
@laurafitzgerald  I've updated the roles now and grafana should install and be accessible without running the oc adm command.

laurafitzgerald:
👀

pkg/products/grafana/reconciler.go
```diff
@@ -0,0 +1,314 @@
...
+				AuthBasic: &grafanav1alpha1.GrafanaConfigAuthBasic{
+					Enabled: &[]bool{true}[0],
+				},
+				AuthAnonymous: &grafanav1alpha1.GrafanaConfigAuthAnonymous{
+					Enabled: &[]bool{true}[0],
+				},
+			},
+			Containers: []v1.Container{
+				{Name: "grafana-proxy",
+					Image: "quay.io/openshift/origin-oauth-proxy:4.2",
```
laurafitzgerald, line 191: Do we hardcode the oauth proxy image elsewhere? I'm wondering if this is something that could be more configured.

CathalOConnorRH: I took that from the Kafka project but I think we could set it as a parameter at least. Or we could set it to the latest tag but I dont know what the impact might be on changes.

laurafitzgerald:
@CathalOConnorRH I left one question above. I've ran the branch and everything is as expected in the description 👍 
/approve

laurafitzgerald:
I don't have approvers permission. I have verified this pr but happy for other folks to approve
/lgtm

PR #1208, opened 2020-09-28
MGDAPI-92 Reconcile rate limit service

sergioifg94:
# Description

Create RateLimitServiceReconciler component to reconcile the resources for the rate limit service. It expects a secret that points to the Redis instance to be used by the service, and creates a ConfigMap, Deployment and Service to configure, run and expose this service.

Create unit tests for reconciler.

### 1. Include the marin3r reconciler

Uncomment the line with the `integreatlyv1alpha1.ProductMarin3r` reconciler to include the product in the ManagedAPI installation:
https://github.com/integr8ly/integreatly-operator/blob/master/pkg/controller/installation/types.go#L48

### 2. Install the Managed API Operator

```sh
make cluster/prepare/local
make code/run
```

### 3. Create Redis instance

> This step will be automated by [MGDAPI-88](https://issues.redhat.com/browse/MGDAPI-88). Meanwhile, create the in-cluster Redis instance and a Service to expose it

```sh
cat << EOF | oc create -f - -n redhat-rhmi-operator
kind: Redis
metadata:  
  name: rate-limit
spec:
  secretRef:
    name: rate-limit-redis
  tier: production
  type: managed-api
EOF
```

Wait for the CRO to reconcile the Redis CR and set it to complete. Then create a Secret in the `redhat-rhmi-marin3r` namespace referencing the Redis Service:

```sh
REDIS_URL=$(oc get secret/rate-limit-redis -n redhat-rhmi-operator -o json | jq -r '.data.uri' | base64 --decode):$(oc get secret/rate-limit-redis -n redhat-rhmi-operator -o json | jq -r '.data.port' | base64 --decode)
oc create secret generic redis -n redhat-rhmi-marin3r --from-literal=URL=$REDIS_URL
```

### 4. Check resources are reconciled

Once the secret with the Redis URL is created, the operator will finish reconciling the resources. Check that the following resources exist and are in good state:

* Rate limit configuration ConfigMap
    ```sh
    oc get configmaps/ratelimit-config -n redhat-rhmi-marin3r
    ```
* Rate limit service Deployment
    ```sh
    oc get deployments/ratelimit -n redhat-rhmi-marin3r
    ```
* Exposed service
    ```sh
    oc get services/ratelimit -n redhat-rhmi-marin3r
    ```
## Type of change

## Checklist

### DIFF
--- a/pkg/products/marin3r/rateLimitService.go
+++ b/pkg/products/marin3r/rateLimitService.go
@@ -18,6 +18,12 @@ import (
 type RateLimitServiceReconciler struct {
 	Namespace       string
 	RedisSecretName string
+	StatsdConfig    *StatsdConfig
+}
+
+type StatsdConfig struct {
+	Host string
+	Port string
 }
 
 func NewRateLimitServiceReconciler(namespace, redisSecretName string) *RateLimitServiceReconciler {
@@ -62,6 +68,12 @@ func (r *RateLimitServiceReconciler) ReconcileRateLimitService(ctx context.Conte
 	}
 
 	return r.reconcileService(ctx, client)
+}
+
+// WithStatsdConfig mutates r setting r.StatsdConfig to the value of config
+func (r *RateLimitServiceReconciler) WithStatsdConfig(config StatsdConfig) *RateLimitServiceReconciler {
+	r.StatsdConfig = &config
+	return r
 }
 
 func (r *RateLimitServiceReconciler) reconcileConfigMap(ctx context.Context, client k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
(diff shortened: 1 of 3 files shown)


CathalOConnorRH:
Verified on cluster 
/lgtm

pkg/products/marin3r/rateLimitService.go
```diff
@@ -0,0 +1,322 @@
...
+// exposes it as a Service
+func (r *RateLimitServiceReconciler) ReconcileRateLimitService(ctx context.Context, client k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
+	phase, err := r.reconcileConfigMap(ctx, client)
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, err
+	}
+
+	if phase == integreatlyv1alpha1.PhaseInProgress {
+		return phase, nil
+	}
```
KevFan, line 67: Is this needed as `r.reconcileConfigMap` doesn't look to return a `PhaseInProgress` ? :thinking:

pkg/products/marin3r/rateLimitService.go
```diff
@@ -0,0 +1,322 @@
...
+
+	phase, err = r.reconcileDeployment(ctx, client)
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, err
+	}
+	if phase == integreatlyv1alpha1.PhaseAwaitingComponents {
+		return phase, nil
+	}
```
KevFan, line 75:
```suggestion
	phase, err = r.reconcileDeployment(ctx, client)
	if phase != integreatlyv1alpha1.PhaseCompleted {
		return phase, err
	}
```
What do you think on just returning the `phase` and `err` if the `phase` isnt `PhaseCompleted` from the function calls in general as they each return `PhaseCompleted` if all goes well ?

Think it might be a bit cleaner instead of additional if blocks :thinking:

KevFan (commented):
Looks good :+1: 

Just have a few comments

pkg/products/marin3r/rateLimitService.go
sergioifg94: It is not. I wrote it in advance thinking that it would be needed but forgot to remove it. I'll get rid of that block :+1:

pkg/products/marin3r/rateLimitService.go
sergioifg94: True! That's much more elegant :smile:

sergioifg94:
@CathalOConnorRH @KevFan Thanks for the reviews! I updated the PR with Kevin's suggestions

KevFan:
\lgtm

sergioifg94:
Rebased after #1199 was merged due to conflict

laurafitzgerald:
/lgtm
rebase.

PR #1214, opened 2020-10-02
test cases: add support for more products

psturc:
# Description
https://issues.redhat.com/browse/MGDAPI-176

## Type of change

## Checklist

## TODO
- [ ] Apply a new test case style to all test cases
- [ ] Update linter

## Migration of existing test cases/creating a new test case

Old format:
```
---
automation:
  - INTLY-7748
components:
  - product-sso
environments:
  - osd-post-upgrade
  - osd-fresh-install
targets:
  - 2.7.0
estimate: 1h
---
```

New format
```
---
automation:
  - INTLY-7748
estimate: 1h
components:
  - product-sso
products:
   - name: rhmi
     environments:
       - osd-post-upgrade
       - osd-fresh-install
     targets:
       - 2.7.0
   - name: rhoam
     environments:
       - osd-fresh-install
     targets:
       - 0.1.0
---
```


test-cases/tools/lib/test-case.ts
```diff
@@ -111,14 +117,24 @@ function extractCategory(file: string): string {
...
+        .map((f) => loadTestCase(f, productName))
+        .filter((t) => t != null);
 }
 
-function loadTestCase(file: string): TestCase {
+function loadTestCase(file: string, productName: string): TestCase | null {
     const m = matter.read(file);
     const data = m.data as Metadata;
 
+    const product = data.products.find((p) => p.name === productName);
```
jhellar, line 133:
```suggestion
    const product = data.products && data.products.find((p) => p.name === productName);
```

jhellar:
`lint` command won't work, because it does not pass `product` to `loadTestCases`, so it won't load any test cases to lint. Also README should be updated to capture new metadata format.

test-cases/tools/lib/test-case.ts
psturc: thanks, will add it

psturc:
@jhellar thanks for a review.

Updating a lint is in TODO, I'll add that later.

test-cases/README.md
```diff
@@ -187,33 +187,33 @@ It is also possible to set the Jira username and password in environment variabl
 Use the `export csv` cmd to list and export the test cases in csv:
 
 ```
-./tools.sh export csv
+./tools.sh export csv --product PRODUCT_NAME
```
b1zzu, line 190: Not sure if we should always filter by-product, the product should be a column in the CSV not a mandatory filter in my opinion

test-cases/tools/lib/test-case.ts
```diff
@@ -111,14 +117,24 @@ function extractCategory(file: string): string {
     return path.basename(path.dirname(file));
 }
 
-function loadTestCases(testDirectory?: string): TestCase[] {
-    return walk(testDirectory || TEST_DIR, TEST_FILTER).map(loadTestCase);
+function loadTestCases(
+    productName?: string,
```
b1zzu, line 121: The productName can not be optional here and mandatory in the loadTestCase method
```suggestion
    productName: string,
```

psturc: @b1zzu I was thinking to make `productName` parameter optional for both `loadTestCases` and `loadTestCase` methods. I.e. if `productName` is not specified, return all test cases (for all products). 

This would allow to not filter out test cases based on the product for `tools.sh export` command, as you suggested above.

wdyt?

b1zzu:
@psturc the introduction of the `product` key will break more things that I could predict, as it is the `lint`, `rename` and `plan` cmds are not working. As this is a major change we should test all cmds. It would be so easier if the rhmi and rhoam will be on two separate branches or repository :facepalm:

test-cases/tools/lib/test-case.ts
b1zzu: yes and no, the problem is that if the `test.enviroments` and `test.targets` will be undefined if the product is not specified, also right now I've changed again opinion, and proabbly keeping the mandatory product filter for the `export` cmd is better :sweat_smile:

psturc:
@b1zzu do you mind giving this PR a final look? Feel free to ignore the changes to test cases (.md files), they will be updated in the following sprint anyway.

test-cases/README.md
```diff
@@ -247,6 +247,7 @@ Tags are used to define specific characteristics of a test case.
 - `per-release` Indicates that this test case should be executed on each release but can be skip if passed in the previous build
 - `manual-selection` Indicates that this test case is excluded from the standard test recycle and therefore it will not be retested every tree releases
 - `destructive` Indicates that the test case may conflict with other tests and therefore can't be executed in parallel with other tests, all destructive tests must be executed sequentially
+- `rhmi`
```
b1zzu, line 250: I think we can remove this part

test-cases/tools/cmd/lint.ts
```diff
@@ -26,6 +26,9 @@ type error = string | null;
 
 const AUTOMATION = /^[A-Z]+-[0-9]+$/;
 
+// Update the README.md too
```
b1zzu, line 29: Have you updated the README.md? :)

test-cases/tools/cmd/lint.ts
```diff
@@ -96,7 +99,7 @@ function lintDuplicateIDs(): Linter {
     const parsed: { [id: string]: TestCase } = {};
 
     return (test: TestCase): error => {
-        if (test.id in parsed) {
```
b1zzu, line 99: I don't understand what `parsed[test.id].title !== test.title` does

test-cases/tools/lib/test-case.ts
```diff
@@ -71,8 +79,8 @@ function extractTitle(content: string): { title: string; content: string } {
 }
 
 function extractId(title: string): { id: string; title: string } {
-    // A01 - Title
-    const match = /^(?<id>[A-Z][0-9]{2})\s-\s(?<title>.*)$/.exec(title);
+    // A01a - Title
+    const match = /^(?<id>[A-Z][0-9]{2}[a-z]?)\s-\s(?<title>.*)$/.exec(title);
```
b1zzu, line 83: Because we have only `a` or `b` as post id letter we could updated the regex to include only them, also it would be nice to have the id all uppercase , like this:
```suggestion
    const match = /^(?<id>[A-Z][0-9]{2}[AB]?)\s-\s(?<title>.*)$/.exec(title);
```

b1zzu:
@psturc let me know what do you think about this approach: https://github.com/psturc/integreatly-operator/pull/5

test-cases/tools/cmd/lint.ts
psturc: solved by your PR https://github.com/psturc/integreatly-operator/pull/5

test-cases/tools/cmd/lint.ts
psturc: who reads the documentation? :P

will fix it in next commit

psturc:
hey @b1zzu thanks so much for all the suggestions.
I hope I addressed all of them and we're finally ready to merge this tiny PR ;)

b1zzu:
@psturc to me is ready, if you need I can approve it :)

psturc:
I hope prow will allow me to approve it as well :)
/approve

PR #1215, opened 2020-10-02
[WIP] Mgdapi 88 install redis

briangallagher:
# Description

## Type of change

## Checklist


pkg/products/marin3r/reconciler.go
```diff
@@ -129,6 +151,74 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+	redisName := fmt.Sprintf("%s%s", constants.RateLimitRedisPrefix, r.installation.Name)
+	rateLimitRedis, err := croUtil.ReconcileRedis(ctx, client, defaultInstallationNamespace, r.installation.Spec.Type, croUtil.TierProduction, redisName, ns, redisName, ns, func(cr metav1.Object) error {
+		owner.AddIntegreatlyOwnerAnnotations(cr, r.installation)
+		return nil
+	})
+	if err != nil {
+		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to reconcile backend redis request: %w", err)
+	}
+
+	r.redisSecretName = rateLimitRedis.Spec.SecretRef.Name
```
sergioifg94, line 176: The way the RateLimitReconciler is implemented expects the Redis secret to have a URL field [\[1\]](https://github.com/integr8ly/integreatly-operator/pull/1215/commits/e8d6598690a4ccba24a70d03f598e4d226eb066b#diff-898ea6af7f0284619a8a48f611898689R183) with the host and port of the redis instance. This is done following the existing implementation for other Redis resources [\[2\]](https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/threescale/reconciler.go#L796-L819), where the operator retrieves the secret created from the CRO and creates a new secret with `URL` concatenating the `uri` and `host` fields from the secret created by CRO.
There's two possible approaches to make it work:
1. To create a new secret as it's done in the 3scale reconciler, setting the `URL` field
2. To modify the RateLimitReconciler to use the `uri` and `port` fields instead of the `URL` 

I don't see a reason not to go for option 2 and save us from copying the secret, maybe someone who knows better than me can tell

pkg/products/marin3r/reconciler.go
```diff
@@ -263,6 +263,12 @@ func (r *Reconciler) preUpgradeBackupExecutor() backup.BackupExecutor {
...
-	//todo add backup for redis once it's added to the reconciler
-	return backup.NewNoopBackupExecutor()
+
+	return backup.NewConcurrentBackupExecutor(
+		backup.NewAWSBackupExecutor(
+			r.installation.Namespace,
+			"ratelimit-service-redis-rhmi",
+			backup.RedisSnapshotType,
+		),
+	)
```
sergioifg94, line 273:
```suggestion
	return backup.NewAWSBackupExecutor(
			r.installation.Namespace,
			"ratelimit-service-redis-rhmi",
			backup.RedisSnapshotType,
	)
```

I would only use the `ConcurrentBackupExecutor` for one or more resources that need backup. Otherwise we'll create a single new goroutine and wait for it, which has no benefit

sergioifg94 (changes requested):
Looking good so far. Just a couple things

pkg/products/marin3r/reconciler.go
briangallagher: Thanks @sergioifg94 I'm closing this PR and opening https://github.com/integr8ly/integreatly-operator/pull/1220 I've taken the 1. approach as there was a precedent in the code but I can change to 2. if you think there is a need?

pkg/products/marin3r/reconciler.go
briangallagher: Good call

briangallagher:
Covered by: https://github.com/integr8ly/integreatly-operator/pull/1220

pkg/products/marin3r/reconciler.go
sergioifg94: Yeah, I think it's safer to go with 1. I don't know exactly what's the reasoning behind but if it was done that way I'd assume it's for a good reason. I'll have a look at the new PR

PR #1220, opened 2020-10-04
add redis to rate limiting service

briangallagher:
# Description

## How to Verify

- Install manged-api and verify it's as expected 

- Verify that the rate-limiting Deployment exists
`oc get deployments/ratelimit -n redhat-rhmi-marin3r`

- Verify redis installed
`oc get redis ratelimit-service-redis-managed-api`

- Check the Pod logs of rate-limiting service and confirm no error when it connects to redis.

### DIFF
--- a/pkg/products/marin3r/reconciler.go
+++ b/pkg/products/marin3r/reconciler.go
@@ -298,7 +298,7 @@ func (r *Reconciler) preUpgradeBackupExecutor() backup.BackupExecutor {
 
 	return backup.NewAWSBackupExecutor(
 		r.installation.Namespace,
-		externalRedisSecretName,
+		fmt.Sprintf("%s%s", constants.RateLimitRedisPrefix, r.installation.Name),
 		backup.RedisSnapshotType,
 	)
 }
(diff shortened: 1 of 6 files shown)


pkg/products/marin3r/reconciler.go
```diff
@@ -220,6 +295,10 @@ func (r *Reconciler) preUpgradeBackupExecutor() backup.BackupExecutor {
 	if r.installation.Spec.UseClusterStorage != "false" {
 		return backup.NewNoopBackupExecutor()
 	}
-	//todo add backup for redis once it's added to the reconciler
-	return backup.NewNoopBackupExecutor()
+
+	return backup.NewAWSBackupExecutor(
+		r.installation.Namespace,
+		externalRedisSecretName,
```
sergioifg94, line 301:
```suggestion
		fmt.Sprintf("%s%s", constants.RateLimitRedisPrefix, r.installation.Name),
```

This parameter is the name of the Redis CR, which is referenced by the RedisSnapshot created when running the backups

sergioifg94 (changes requested):
Verified on the cluster, and the code looks good. Only one small suggestion

pkg/products/marin3r/reconciler.go
briangallagher: Good catch, thanks

sergioifg94:
I'm not sure if it's ready to merge given that it's marked as WIP but it looks good to me
/lgtm

PR #1221, opened 2020-10-04
add prometheus resources to scrape rate limiting service

briangallagher:
# Description
Add Prometheus to the Rate Limiting Solution

## How to Verify
- Install managed-api and confirm complete
- Confirm that the RateLimit metrics are being scraped 

metrics

### DIFF
--- a/pkg/products/marin3r/reconciler.go
+++ b/pkg/products/marin3r/reconciler.go
@@ -3,11 +3,16 @@ package marin3r
 import (
 	"context"
 	"fmt"
+
+	prometheus "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1/types"
 	croUtil "github.com/integr8ly/cloud-resource-operator/pkg/client"
 	"github.com/integr8ly/integreatly-operator/pkg/resources/owner"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"strconv"
 
 	marin3r "github.com/3scale/marin3r/pkg/apis/operator/v1alpha1"
 	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
@@ -38,6 +43,10 @@ const (
 	discoveryServiceName = "instance"
 
 	externalRedisSecretName = "redis"
+
+	statsdHost  = "prom-statsd-exporter"
+	statsdPort  = 9125
+	metricsPort = 9102
 )
 
 type Reconciler struct {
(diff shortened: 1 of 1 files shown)


pkg/products/marin3r/reconciler.go
```diff
@@ -223,3 +254,148 @@ func (r *Reconciler) preUpgradeBackupExecutor() backup.BackupExecutor {
...
+
+	return integreatlyv1alpha1.PhaseCompleted, nil
+}
+
+func (r *Reconciler) reconcilePromStatsdExporterService(ctx context.Context, client k8sclient.Client, namespace string) (integreatlyv1alpha1.StatusPhase, error) {
+	logrus.Infof("Start reconcilePromStatsdExporterService for marin3r")
+
+	service := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "prom-statsd-exporter",
```
KevFan, line 325: What do you think on using `statsdHost` constant here and other places that use `prom-statsd-exporter` ? 
Might be unlikely this will change but would be good to have one place to change the name since this is referenced quite a bit as a label `selector` :thinking:

pkg/products/marin3r/reconciler.go
```diff
@@ -223,3 +254,148 @@ func (r *Reconciler) preUpgradeBackupExecutor() backup.BackupExecutor {
...
+						Name:  "prom-statsd-exporter",
+						Image: "prom/statsd-exporter:v0.18.0",
+						Ports: []corev1.ContainerPort{
+							{
+								Name:          "prom-statsd",
+								ContainerPort: 9125,
+							},
+							{
+								Name:          "metrics",
+								ContainerPort: 9102,
```
KevFan, line 302: What do you think on making these constants also as it's also references a few times in other deployments ? :thinking:

KevFan (commented):
Looks good :+1: 

Verified with a manual deploy of redis (https://github.com/integr8ly/integreatly-operator/pull/1208 Step 3) as https://github.com/integr8ly/integreatly-operator/pull/1220 is not merged yet

Have just a few comments

pkg/products/marin3r/reconciler.go
```diff
@@ -140,8 +148,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+
+	phase, err = r.reconcilePromStatsdExporterService(ctx, client, productNamespace)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, fmt.Sprintf("Failed to reconcile Prometheus StatsD exporter service"), err)
+		return phase, err
+	}
+
+	rlReconciler := NewRateLimitServiceReconciler(productNamespace, externalRedisSecretName)
+	rlReconciler.WithStatsdConfig(statsdConfig)
+	phase, err = rlReconciler.ReconcileRateLimitService(ctx, client)
```
sergioifg94, line 169:
```suggestion
	phase, err = NewRateLimitServiceReconciler(productNamespace, externalRedisSecretName).
	    WithStatsdConfig(statsdConfig).
	    ReconcileRateLimitService(ctx, client)
```

Just a tiny note. The `WithStatsdConfig` function returns the `RateLimitServiceReconciler` instance, so you can instantiate it, add the config, and reconcile it in the same statement, without allocating a variable for it

pkg/products/marin3r/reconciler.go
briangallagher: Yeah, it's cleaner

pkg/products/marin3r/reconciler.go
briangallagher: yep, sure thing

pkg/products/marin3r/reconciler.go
briangallagher: Yep, its cleaner

KevFan:
@briangallagher PR needs a rebase

PR #1223, opened 2020-10-05
Adding Rate Limiting Configuration

laurafitzgerald:
# Description
https://issues.redhat.com/browse/MGDAPI-95

### Installation
On a clean cluster, run:

- `make cluster/prepare/local`
- Change INSTALLATION_TYPE in Makefile to `managed-api`
- `make cluster/run`

Verify that

- There are two pods running for each `apicast-staging` and `apicast-production` with 2 containers in each include an evoy sidecar container
- There are two EnvoyConfig cr's created titled `apicast-staging` and `apicast-production`.
- The ratelimit-config ConfigMap in the marin3r namespace has the data entrys:
```
apicast-ratelimiting.yaml: |
    domain: apicast-ratelimiting
    descriptors:
    - key: generic_key
      value: slowpath
      rate_limit:
        requests_per_unit: 20
        unit: minute
      descriptors: []
```

### Functionality

Verify that Rate Limiting is working as expected

The Rate Limiting is currently set to Rate Limit 20 per minute so on the second run of a curl command inside 1 minute the request should return a 429.

1. Navigate to `https://3scale-admin.[your_admin_domain].devshift.org/p/admin/onboarding/wizard/intro` and follow the steps outlined in the wizard to create a 3Scale service. You can get your admin domain from `Routes` under `redhat-rhmi-3scale` namespace, use the admin user credentials from the `system-seed` secret in the 3scale namespace
2. Navigate to `https://3scale-admin.[your_admin_domain].devshift.org/apiconfig/services/2/integration`
3. Copy curl command from API Cast staging.

## Type of change

## Checklist


pkg/products/threescale/reconciler.go
```diff
@@ -342,6 +383,260 @@ func (r *Reconciler) backupSystemSecrets(ctx context.Context, serverClient k8scl
...
+		for i, port := range ports {
+			if port.Name == "gateway" {
+				logrus.Info("updating the gateway port")
+				service.Spec.Ports[i].Port = 8443
+				service.Spec.Ports[i].TargetPort = intstr.FromInt(8443)
+				break
+			}
+		}
+		return nil
+	})
```
sergioifg94, line 438:
```suggestion
if service.Annotations == nil {
	service.Annotations = map[string]string{}
}
// add annotation in order to trigger openshift to create a required tls secret
service.Annotations["service.beta.openshift.io/serving-cert-secret-name"] = secretName

// update the gateway port to point to 8443 so that the request is via the envoy sidecar container

ports := service.Spec.Ports
for i, port := range ports {
	if port.Name == "gateway" {
		logrus.Info("updating the gateway port")
		service.Spec.Ports[i].Port = 8443
		service.Spec.Ports[i].TargetPort = intstr.FromInt(8443)
		break
	}
}

if err := client.Update(ctx, service); err != nil {
    return fmt.Errorf("failed to update service: %v", err)
}
```

There's no need to use the `CreateOrUpdate` function as this code won't be reached if the service doesn't exist (it will have returned `AwaitingComponents`). Same thing can be changed for the DeploymentConfig

pkg/products/threescale/reconciler.go
```diff
@@ -342,6 +383,260 @@ func (r *Reconciler) backupSystemSecrets(ctx context.Context, serverClient k8scl
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
 
+func getDeploymentConfig(ctx context.Context, client k8sclient.Client, dcName string) (*appsv1.DeploymentConfig, integreatlyv1alpha1.StatusPhase, error) {
+	apiCastDeploymentConfig := &appsv1.DeploymentConfig{}
+
+	err := client.Get(ctx, k8sTypes.NamespacedName{Name: dcName, Namespace: "redhat-rhmi-3scale"}, apiCastDeploymentConfig)
```
sergioifg94, line 389:
```suggestion
func (r *Reconciler) getDeploymentConfig(ctx context.Context, client k8sclient.Client, dcName string) (*appsv1.DeploymentConfig, integreatlyv1alpha1.StatusPhase, error) {
	apiCastDeploymentConfig := &appsv1.DeploymentConfig{}

	err := client.Get(ctx, k8sTypes.NamespacedName{Name: dcName, Namespace: r.Config.GetProductNamespace()}, apiCastDeploymentConfig)
```

Change to avoid hardcoding the namespace. I'd suggest the same in the `getService` function

pkg/products/threescale/reconciler.go
```diff
@@ -342,6 +383,260 @@ func (r *Reconciler) backupSystemSecrets(ctx context.Context, serverClient k8scl
...
+	rateLimitClusterYamlMarshalled, err := k8syaml.JSONToYAML(ratelimitClusterJsonMarshalled)
+	if err != nil {
+		return fmt.Errorf("failed to create the %s envoy config: %w", deploymentConfig.Name, err)
+
+	}
+	listenerMarshalledYaml, err := k8syaml.JSONToYAML(listenerJsonMarshalled)
+	if err != nil {
+		return fmt.Errorf("failed to create the %s envoy config: %w", deploymentConfig.Name, err)
+
+	}
```
sergioifg94, line 592: Maybe we could create a utility function `JSONObjectToYAML` that'd look like this

```go
func JSONObjectToYAML(o interface{}) ([]byte, error) {
  jsonMarshelled, err := json.Marshal(o)
  if err := nil {
    return nil, err
  }

  return k8syaml.JSONToYAML(jsonMarshalled)
}
```

It'd make this section shorter and more readable

sergioifg94 (commented):
Code looks good so far, I just added a few comments

laurafitzgerald:
@sergioifg94 thanks for the review. comments are addressed. 👍

pkg/products/threescale/reconciler.go
```diff
@@ -342,6 +392,349 @@ func (r *Reconciler) backupSystemSecrets(ctx context.Context, serverClient k8scl
...
+	return nil
+}
+
+func (r *Reconciler) createEnvoyRateLimitingConfig(ctx context.Context, client k8sclient.Client) error {
+	var podIp string
+
+	pods := &corev1.PodList{}
+
+	listOpts := []k8sclient.ListOption{
+		k8sclient.InNamespace("redhat-rhmi-marin3r"),
```
laurafitzgerald, line 455: @MStokluska I think we should probably parameterise this one in case the prefix changes. Although I think we are going to update the endpoint address to point to the service at some point or something similar so if this is going away it might be fine for now. wdyt?

pkg/products/threescale/reconciler.go
```diff
@@ -342,6 +392,349 @@ func (r *Reconciler) backupSystemSecrets(ctx context.Context, serverClient k8scl
...
+			},
+			{
+				Name: "envoy.router",
+			},
+		},
+	}
+
+	pbst, err := ptypes.MarshalAny(manager)
+	if err != nil {
+		panic(err)
```
laurafitzgerald, line 640: Do we definitely want to panic here or is it sufficient to return a phase failed?

pkg/products/threescale/reconciler.go
```diff
@@ -342,6 +392,349 @@ func (r *Reconciler) backupSystemSecrets(ctx context.Context, serverClient k8scl
...
+	yamlCluster, err := yaml.JSONToYAML(clusterJson)
+
+	listenerJson, err := ResourcesToJSON(listener)
+
+	yamlListener, err := yaml.JSONToYAML(listenerJson)
+
+	envoyconfig := &marin3rv1alpha.EnvoyConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      apicastRatelimiting,
+			Namespace: "redhat-rhmi-3scale",
```
laurafitzgerald, line 680: We should probably use the parmaterised prefix or pass the productNamespace into the function rather than hardcoding.

R-Lawton:
i think adding more comments of the functions are doing would be a good idea

pkg/products/threescale/reconciler.go
```diff
@@ -342,6 +392,349 @@ func (r *Reconciler) backupSystemSecrets(ctx context.Context, serverClient k8scl
...
+	}
+
+	err := client.List(ctx, pods, listOpts...)
+
+	for _, pod := range pods.Items {
+		if pod.Status.Phase == "Running" {
+			podIp = pod.Status.PodIP
+			break
+		}
+	}
```
sergioifg94, line 466: I'm probably missing something but asking just in case: Wouldn't it be better to point to the Service that's serving the ratelimit deployment? [1] This would also solve the issue with replication as the service could load balance the rate limit pods

1. https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/marin3r/rateLimitService.go#L259

pkg/products/threescale/reconciler.go
MStokluska: Good idea, thanks Laura

pkg/products/threescale/reconciler.go
MStokluska: Thanks :)

pkg/products/threescale/reconciler.go
MStokluska: Good idea, thank you

pkg/products/threescale/reconciler.go
MStokluska: Thanks @sergioifg94, that idea works well!

R-Lawton:
installation and functionality worked as expected

pkg/products/threescale/reconciler.go
```diff
@@ -342,6 +396,343 @@ func (r *Reconciler) backupSystemSecrets(ctx context.Context, serverClient k8scl
...
+	yamlRateLimitCluster, err := yaml.JSONToYAML(rateLimitClusterJson)
+
+	clusterJson, err := ResourcesToJSON(&cluster)
+
+	yamlCluster, err := yaml.JSONToYAML(clusterJson)
+
+	listenerJson, err := ResourcesToJSON(listener)
+
+	yamlListener, err := yaml.JSONToYAML(listenerJson)
+
```
laurafitzgerald, line 674: Could we handle the above errors here also.

MStokluska: Thanks @laurafitzgerald, updated the PR now

MStokluska:
> installation and functionality worked as expected
@R-Lawton thanks! Updated the PR now

laurafitzgerald:
thanks for addressing the comments, code changes
/lgtm

R-Lawton:
both code and functionality works as expected /lgtm

PR #1225, opened 2020-10-06
fix: backup-restore script typo

psturc:
# Description
Seen when working on https://issues.redhat.com/browse/MGDAPI-267

### DIFF
--- a/test/scripts/backup-restore/postgres.sh
+++ b/test/scripts/backup-restore/postgres.sh
@@ -188,7 +188,7 @@ EOF
   echo "Test finished successfully."
 }
 
-dump_databse() {
+dump_database() {
   DUMP_FILE=$1
 
   echo "Creating throwaway Postgres container..."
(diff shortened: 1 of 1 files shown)


sergioifg94 (changes requested):
I was aware of this typo, but I thought it had already been fixed.

Looks good, just needs to be updated in the first function invocation: https://github.com/integr8ly/integreatly-operator/pull/1225/files#diff-6d48514803dd9bb39c85d864718b7311R46

psturc:
thanks for review, it should be fixed now

sergioifg94:
There's another error that @MStokluska brought to my attention. In [this line](https://github.com/psturc/integreatly-operator/blob/patch-6/test/scripts/backup-restore/postgres.sh#L171), it's `dump_after.sql` instead of `database_after.sql`

Seems like this test might still be a bit flaky anyways, so I'll try to verify it and ensure it works well

PR #1231, opened 2020-10-07
MGDAPI-234 Add priorityClass and configure RHSSO, USERSSO and threescale to use the priorityClass

CathalOConnorRH:
Add priorityClass and configure RHSSO, USERSSO and threescale to use the priorityClass

Verification steps. 
1. Install managed-api type. 
2. Confirm a new priorityclass CR called `managed-service-priority`
3. Confirm managed-api CR has the field `priorityClassName` and it's value is `managed-service-priority`
4. Confirm RHSSO and USERSSO `keycloak` statefulsets have the field `priorityClassName` with the value of `managed-service-priority` 
5. Confirm threescale deployment configs have the field `priorityClassName` with the value of `managed-service-priority` 

Extra verification step if time allows, Install managed type and confirm the above doesn't exist.

### DIFF
--- a/pkg/controller/installation/installation_controller.go
+++ b/pkg/controller/installation/installation_controller.go
@@ -56,6 +56,8 @@ const (
 	alertingEmailAddressEnvName      = "ALERTING_EMAIL_ADDRESS"
 	buAlertingEmailAddressEnvName    = "BU_ALERTING_EMAIL_ADDRESS"
 	installTypeEnvName               = "INSTALLATION_TYPE"
+	priorityClassNameEnvName         = "PRIORITY_CLASS_NAME"
+	managedServicePriorityClassName  = "managed-service-priority"
 )
 
 var (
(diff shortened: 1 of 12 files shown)


pkg/controller/installation/installation_controller.go
```diff
@@ -154,13 +156,18 @@ func createInstallationCR(ctx context.Context, serverClient k8sclient.Client) er
...
 
 		logrus.Infof("Creating a %s rhmi CR with USC %s, as no CR rhmis were found in %s namespace", installType, useClusterStorage, namespace)
 
 		if installType == "" {
 			installType = string(integreatlyv1alpha1.InstallationTypeManaged)
 		}
 
+		if installType == string(integreatlyv1alpha1.InstallationTypeManagedApi) && priorityClassName == "" {
+			priorityClassName = managedServicePriorityClassName
+		}
```
sergioifg94, line 169: I see this block is reached only when the CR does not exist. When upgrading the operator from a previous version, as the installation is already complete and the CR exists, the `.spec.priorityClassName` field won't be set automatically, so the PriorityClass won't be created unless the field is set manually. If it's something that we want to support maybe we could move this logic to the Reconcile loop

sergioifg94 (commented):
Verified it on a cluster, just that one comment

pmccarthy:
@CathalOConnorRH looks like this needs rebasing

philbrookes:
Verified locally.

PR #1234, opened 2020-10-12
MGDAPI-242: Support rate limit configuration via SKU ConfigMap

sergioifg94:
# Description

Add package `marin3r/config` with the functionality to retrieve the rate limit configuration from a ConfigMap that's expected to be already defined with the rate limits for multiple SKUs

Add make target to create this ConfigMap for development purposes. This ConfigMap is expected to be reconciled by hive from managed-tenants

## Verification steps

1. Begin an installation of Managed API type from this PR
3. Create the ConfigMap
    ```sh
    make cluster/prepare/ratelimits
    ```
4. Verify that the installation progresses and completes successfully
5. Open a terminal in the Rate Limit pod in the marin3r namespace, and check that the config matches the ConfigMap
    ```sh
    cat /srv/runtime_data/current/config
    ```
6. Modify the SKU Rate Limits ConfigMap to a different value
7. Wait a minute and repeat step 4 to verify that the config is updated accordingly

## Type of change

## Checklist

### DIFF
--- a/deploy/sku-limits-configmap.yaml
+++ b/deploy/sku-limits-configmap.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: sku-limits-managed-api-service
+data:
+  rate_limit: |-
+    {
+      "RHOAM SERVICE SKU": {
+        "unit": "minute",
+        "requests_per_unit": 20
+      }
+    }
--- a/pkg/controller/installation/bootstrapReconciler.go
+++ b/pkg/controller/installation/bootstrapReconciler.go
@@ -2,6 +2,7 @@ package installation
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"math/rand"
 	"strings"
@@ -29,6 +30,8 @@ import (
 	"k8s.io/client-go/tools/record"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	marin3rconfig "github.com/integr8ly/integreatly-operator/pkg/products/marin3r/config"
 )
 
 func NewBootstrapReconciler(configManager config.ConfigReadWriter, installation *integreatlyv1alpha1.RHMI, mpm marketplace.MarketplaceInterface, recorder record.EventRecorder) (*Reconciler, error) {
@@ -92,6 +95,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		events.HandleError(r.recorder, installation, phase, "Failed to check cloud resources config settings", err)
 		return phase, err
 	}
+
+	phase, err = r.checkRateLimitsConfig(ctx, serverClient)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, "Failed to check rate limits config settings", err)
+		return phase, err
+	}
+
 	events.HandleStageComplete(r.recorder, installation, integreatlyv1alpha1.BootstrapStage)
 
 	metrics.SetRHMIInfo(installation)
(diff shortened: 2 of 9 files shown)


pkg/controller/installation/installation_controller.go
```diff
@@ -589,6 +591,32 @@ func (r *ReconcileInstallation) preflightChecks(installation *integreatlyv1alpha
...
 	}
 
+	if installation.Spec.Type == string(integreatlyv1alpha1.InstallationTypeManagedApi) {
+		configMap := &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      marin3rconfig.RateLimitConfigMapName,
+				Namespace: installation.Namespace,
+			},
+		}
+		if exists, err := resources.Exists(context.TODO(), r.client, configMap); err != nil {
```
laurafitzgerald, line 601: Could we do something like we do with the clourd-resource-stragies config here: https://github.com/integr8ly/integreatly-operator/blob/master/pkg/controller/installation/bootstrapReconciler.go#L104-L133 where we have a default "development" config unless there is a sku provided rather than the installation returning in cases where the config map is not there?

sergioifg94: Good idea. I'll add that to the PR

deploy/sku-limits-configmap.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: sku-limits-managed-api-service
+data:
+  rate_limit: |-
+    {
+      "RHOAM SERVICE SKU": {
```
pmccarthy, line 8: @sergioifg94 this SKU configmap template would only be used for dev correct? If so we should probably look at reducing the reqPerUnit to something that can be easily tested.

It has yet to be decided what unit we'll be limiting on in production: `second` or `minute` but for dev minute makes most sense.

sergioifg94: Yes it would. For now I simply copied the example values from the ticket, but the operator won't reconcile back manual changes that are made to this ConfigMap (hive will, in the case of addon installations), so it's possible for us to tweak this values in development regardless of the initial value

pkg/controller/installation/bootstrapReconciler.go
```diff
@@ -132,6 +142,44 @@ func (r *Reconciler) checkCloudResourcesConfig(ctx context.Context, serverClient
...
+		}
+
+		if _, ok := rlConfig.Data["rate_limit"]; ok {
+			return nil
+		}
+
+		defaultConfig := map[string]*marin3rconfig.RateLimitConfig{
+			marin3rconfig.ManagedApiServiceSKU: {
+				Unit:            "minute",
+				RequestsPerUnit: 231,
```
pmccarthy, line 165: These values will most likely change in the near future depending on load testing results and BU discussions but the default should try to closely match what we want to set in production. The current goal is 20 million requests per day which equals 231 requests per second or 13,860 requests per minute.

To stay consistent across the board we should probably stick with `minute` as the unit and `13860` as the `RequestsPerUnit` value.

deploy/sku-limits-configmap.yaml
pmccarthy: Also, this is my fault for specifying an incorrect name in the configmap example in the Jira but the parameter is `requests_per_unit` rather than `reqPerUnit`. See official docs: https://github.com/envoyproxy/ratelimit#rate-limit-definition

pmccarthy: > Yes it would. For now I simply copied the example values from the ticket, but the operator won't reconcile back manual changes that are made to this ConfigMap (hive will, in the case of addon installations), so it's possible for us to tweak this values in development regardless of the initial value

Sure, that makes sense. I don't feel particularly strongly about it but I think having spoke with @MStokluska on a related Jira we decided to go with 20 requests per minute for now

pkg/controller/installation/bootstrapReconciler.go
sergioifg94: I would only expect the configuration to be set here in development scenarios, as it won't override an existing configuration (which should have been created already with the add-on), shouldn't we then have low values as you suggested in the earlier comment?
I will update the value in the MR in managed-tenants to `13860` though

pmccarthy: My understanding was that should the configmap not exist on cluster this would be the default that would get set? Not sure if it could ever happen but if Hive didn't apply the configmap for some reason it might be safer to have these defaults set to production values? The dev configmap map added as part of this PR should also override these right?

deploy/sku-limits-configmap.yaml
sergioifg94: > the parameter is `requests_per_unit` rather than `reqPerUnit`

That's handled anyways, as the operator will set the correct field in the ConfigMap that's mounted to the Rate Limit deployment:
https://github.com/integr8ly/integreatly-operator/blob/0cdf5dac9b9fcbcf6c0b5f0b442b11cccf7de348/pkg/products/marin3r/rateLimitService.go#L43
https://github.com/integr8ly/integreatly-operator/blob/0cdf5dac9b9fcbcf6c0b5f0b442b11cccf7de348/pkg/products/marin3r/rateLimitService.go#L99
Unless we want the fields to match in both ConfigMaps. In that case I could update that field

> we decided to go with 20 requests per minute for now

Sounds good, I'll update the development ConfigMap for this value

pkg/controller/installation/bootstrapReconciler.go
sergioifg94: That's a good point. The dev configmap won't be created on an add-on flow, so if hive doesn't reconcile it there wouldn't be a config map at all. I'll update this value for the production-like one

deploy/sku-limits-configmap.yaml
pmccarthy: > Unless we want the fields to match in both ConfigMaps. In that case I could update that field
> > we decided to go with 20 requests per minute for now

For consistency it might make sense to have them matching unless there is a valid reason not to

sergioifg94: Cool, I just updated this PR and [the managed-tenants MR](https://gitlab.cee.redhat.com/service/managed-tenants/-/blob/2e0bde81b7764d8bae4a697dc10d4c74245cfa93/addons/managed-api-service/metadata/stage/sku-limits.yaml.j2) to use the `requests_per_unit` field

pmccarthy: Thanks @sergioifg94 changes look good. In the process of validating now

pmccarthy:
@sergioifg94 I ran through the validation steps listed above and all is working as expected. However, it seems that the rate limit pod needs to be restarted before the new rate limit configs take effect. We'll need to figure out how the pod can be restarted when the configmap changes

sergioifg94:
Thanks @pmccarthy, looking at the rate limit service it appears like there's a couple alternatives to trigger a configuration reload, so I'm going to see if I get that to work and then automate it in the operator
https://github.com/envoyproxy/ratelimit#loading-configuration

sergioifg94:
@pmccarthy I updated the PR with a solution to restart the pods when the configuration changes (see new commit). Unfortunately I couldn't make the service hot-reload the configuration, due to the nature of the config map volume mounts and the permissions in these volumes. This approach will force to restart the pods only if the configuration values are different

pmccarthy:
@sergioifg94 thanks for the update, I'm in the process of validating these new changes now

pmccarthy:
@sergioifg94 successfully validated though changes and confirmed that the rate limit pods restarts each time. Also checked the rate limiting itself by curling a 3scale application endpoint. Looks good 👍 

Probably need to re-run the prow tests before I approve?

PR #1238, opened 2020-10-12
create/update grafana datasource for customer grafana

laurafitzgerald:
# Description

## Verification

`make cluster/prepare/local`
`export INSTALLATION_TYPE=managed-api`
`make code/run`

Verify there is a GrafanaDatasource Cr created in the `redhat-rhmi-customer-monitoring-operator` namespace.
Navigate to the customer grafana 
Log in using the credentials from the secret `grafana-admin-credentials`
On left nav select the cog -> Datasources -> prometheus -> Test

You should see `Data source is working`

## Type of change

## Checklist

### DIFF
--- a/pkg/products/grafana/reconciler.go
+++ b/pkg/products/grafana/reconciler.go
@@ -5,6 +5,8 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"fmt"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/global"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
 
 	grafanav1alpha1 "github.com/integr8ly/grafana-operator/v3/pkg/apis/integreatly/v1alpha1"
 	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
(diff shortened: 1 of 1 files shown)


pkg/products/grafana/reconciler.go
```diff
@@ -281,6 +282,49 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, client k8sclient.C
 
 	logrus.Info("Grafana Status: ", status)
 
+	prometheusNamespace := fmt.Sprintf("%smiddleware-monitoring-operator", global.NamespacePrefix)
+
+	url := fmt.Sprintf("http://%s.%s.svc:%s", "prometheus-service", prometheusNamespace, "9090")
```
pmccarthy, line 287: Would there ever be a need to make this URL configurable? Can't think of a reason off the top of my head but seeing hardcoded entries always makes me wary

laurafitzgerald: Yes, if the service name or port changed we'd end up creating the datasource without error but grafana wouldn't be able to access it. An alternative here would be to get the service, the namespace would still be hardcoded, at least the `middleware-monitoring-operator` and `service name` piece but it would be a little bit more robust. We could error if no service is found.

pmccarthy: It seems that `prometheus-service` is hardcoded in our monitoring checks so if it ever changed the alert should fire as part of release testing: https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/monitoring/prometheusRules.go#L270

I think we can leave this as is for now

pmccarthy:
Validated latest changes on OSD based on steps outlined above. All working as expected 👍

laurafitzgerald:
@pmccarthy would you mind adding the `lgtm` label here again. Looks like the small final change I pushed knocked it off.

PR #1239, opened 2020-10-12
MGDAPI-215 removed the command to delete pipelineKey.pem as it is not…

tonyxrmdavidson:
removal of command from tearDown.sh.  It is not needed.


tonyxrmdavidson:
@trepel

trepel:
lgtm

PR #1242, opened 2020-10-14
MGDAPI-227

martinhesko:
# Description
Added a new script that will be used to deploy hyperfoil on an EC2 instance

## Type of change

## Checklist


trepel:
reviewing. I haven't verified fully yet. I tried only running the command from the script in sequence, I did not try to run the script itself, but it seems it is working.
Waiting with lgtm to see if this script is indeed the way to go. Other investigations around running Hyperfoil controller on Jenkins slave directly are in progress.

philbrookes:
If this is still relevant, please re-open.

PR #1252, opened 2020-10-19
add support for BU contact via envVar and RHMI CR

jjaferson:
# Description

We need to contact BU to notify customer API usage, so this PR adds the ability to include BU email address to an envVar or via  RHMI CR 

https://issues.redhat.com/browse/MGDAPI-200

## Verification steps

1) Export an envVar variable with the BU email 
```bash
BU_ALERTING_EMAIL_ADDRESS=noreply-bu-test@rhmi-redhat.com
```
2) Install operator locally from this branch
3) Verify if the bu email is present in the RHMI CR
```bash
oc get rhmi rhmi -o json -n redhat-rhmi-operator | jq .spec.alertingEmailAddresses

{
  "businessUnit": "noreply-bu-test@rhmi-redhat.com",
  "cssre": ""
}

```

## Type of change


Makefile
```diff
@@ -53,6 +53,8 @@ export USE_CLUSTER_STORAGE ?= true
 export OPERATORS_IN_PRODUCT_NAMESPACE ?= false # e2e tests and createInstallationCR() need to be updated when default is changed
 export DELOREAN_PULL_SECRET_NAME ?= integreatly-delorean-pull-secret
 export ALERTING_EMAIL_ADDRESS = noreply-test@rhmi-redhat.com
```
MStokluska, line 55: Hey @jjaferson, I might be missing something but should it not be `?=` - currently when specifing the email addresses my envvars are being overwritten by the values in makefile.

Makefile
```diff
@@ -53,6 +53,8 @@ export USE_CLUSTER_STORAGE ?= true
...
+export BU_ALERTING_EMAIL_ADDRESS = noreply-test@rhmi-redhat.com
```
MStokluska, line 56: This seems to be the same issue as above

MStokluska:
Hi @jjaferson this looks good! Left a couple of comments which I think could be good to address. Thanks for answering all the other questions I had. 

/lgtm

philbrookes:
verified locally

PR #1255, opened 2020-10-20
MGDAPI-404 Deleting enabled ns CRs

MStokluska:
# Description
This PR addresses a bug where during the uninstallation of managed-api there was a race condition between 3scale and the marin3r reconciler, leading to uninstallation being stuck in 3scale.

# Verification
Because the bug has a chance to trigger there is no way of recreating it every time. Best way to verify it would be to install managed-api, uninstall it and expect it to be removed from the cluster.

## Prerequisites
- Cluster provisioned
- Logged in to the cluster

## Steps
- run `export INSTALLATION_TYPE=managed-api`
- run `make cluster/prepare/local`
- run `make code/run`
- wait for the installation to complete
- once completed remove managed-api CR from `redhat-rhmi-operator` namespace
- expect uninstallation to be completed

### DIFF
--- a/pkg/products/marin3r/reconciler.go
+++ b/pkg/products/marin3r/reconciler.go
@@ -9,12 +9,14 @@ import (
 	prometheus "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1/types"
 	croUtil "github.com/integr8ly/cloud-resource-operator/pkg/client"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/global"
 	"github.com/integr8ly/integreatly-operator/pkg/resources/owner"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	marin3rv1alpha "github.com/3scale/marin3r/pkg/apis/marin3r/v1alpha1"
 	marin3r "github.com/3scale/marin3r/pkg/apis/operator/v1alpha1"
 	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
 	"github.com/integr8ly/integreatly-operator/pkg/config"
@@ -42,6 +44,12 @@ const (
 	metricsPort                  = 9102
 	discoveryServiceName         = "instance"
 	externalRedisSecretName      = "redis"
+)
+
+var (
+	enabledNamespaces = []string{
+		global.NamespacePrefix + "3scale",
+	}
 )
 
 type Reconciler struct {
@@ -104,6 +112,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 	productNamespace := r.Config.GetNamespace()
 
 	phase, err := r.ReconcileFinalizer(ctx, client, installation, string(r.Config.GetProductName()), func() (integreatlyv1alpha1.StatusPhase, error) {
+		if err := r.deleteEnabledNamespacesFinalizers(ctx, client); err != nil {
+			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to delete enabled namespaces finalizers: %v", err)
+		}
+
+		if err := r.deleteDiscoveryService(ctx, client); err != nil {
+			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to delete discovery service: %v", err)
+		}
+
 		phase, err := resources.RemoveNamespace(ctx, installation, client, productNamespace)
 		if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
 			return phase, err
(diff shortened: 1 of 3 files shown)


pkg/resources/ratelimit/envoyconfig.go
```diff
@@ -0,0 +1,50 @@
...
+
+		// Only change the status phase if it was Completed, to ensure that
+		// as long as one of the namespaces returns InProgress, the phase is
+		// set to InProgress
+		if phase == integreatlyv1alpha1.PhaseCompleted {
+			phase = nsPhase
+		}
+	}
+
+	return phase, nil
```
KevFan, line 28: Not quite sure I understand this. `phase` is already preset to `PhaseComplete` so this will always be true for the first run of the for loop. So is the thinking, `phase` gets set to whatever `nsPhase` is and be unblocking for each namespace. Then on the next reconcile, the `phase` will only be complete if all the `EnvoyConfigs` are deleted for each namespace  :thinking: 

Is there any problems with just a return if `nsPhase` is `PhaseInProgress` instead? Might be a bit easier to understand :thinking: 

```suggestion
	phase := integreatlyv1alpha1.PhaseCompleted

	for _, namespace := range namespaces {
		nsPhase, err := DeleteEnvoyConfigsInNamespace(ctx, client, namespace)
		if err != nil {
			return integreatlyv1alpha1.PhaseFailed, err
		}

		if phase == integreatlyv1alpha1.InProgress {
			return nsPhase, err 
		}
	}

	return phase, nil
```

KevFan (commented):
Looks good :+1: 
Just have some comments / questions

pkg/products/marin3r/reconciler.go
```diff
@@ -104,7 +112,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
 
 	phase, err := r.ReconcileFinalizer(ctx, client, installation, string(r.Config.GetProductName()), func() (integreatlyv1alpha1.StatusPhase, error) {
-		phase, err := resources.RemoveNamespace(ctx, installation, client, productNamespace)
+		phase, err := ratelimit.DeleteEnvoyConfigsInNamespaces(ctx, client, enabledNamespaces...)
+		if err != nil {
+			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to clear envoy config CRs: %v", err)
+		}
+		if phase != integreatlyv1alpha1.PhaseCompleted {
+			return phase, nil
+		}
```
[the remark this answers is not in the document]
MStokluska: @KevFan thank you, I agree, @sergioifg94 WDYT?

pkg/products/threescale/reconciler.go
```diff
@@ -171,7 +172,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
 
 	phase, err := r.ReconcileFinalizer(ctx, serverClient, installation, string(r.Config.GetProductName()), func() (integreatlyv1alpha1.StatusPhase, error) {
-		phase, err := resources.RemoveNamespace(ctx, installation, serverClient, productNamespace)
+		phase, err := ratelimit.DeleteEnvoyConfigsInNamespaces(ctx, serverClient, productNamespace)
+		if err != nil {
+			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to delete envoy config: %v", err)
+		}
+		if phase != integreatlyv1alpha1.PhaseCompleted {
+			return phase, nil
+		}
```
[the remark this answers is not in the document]
MStokluska: @KevFan thank you, I agree, @sergioifg94 WDYT?

pkg/resources/ratelimit/envoyconfig.go
MStokluska: This seems like a good point to me, thanks @KevFan, wdyt @sergioifg94

KevFan: Actually might be able to use the same 2 checks from the previous comments here instead also :thinking: 
```suggestion
		for _, namespace := range namespaces {
		nsPhase, err := DeleteEnvoyConfigsInNamespace(ctx, client, namespace)
		if err != nil || nsPhase != integreatlyv1alpha1.PhaseCompleted {
			return nsPhase, err
		}
	}
	return integreatlyv1alpha1.PhaseCompleted, nil
```

sergioifg94: @KevFan `phase` is set to whatever `nsPhase` is but only if it currently is PhaseComplete. If it is InProgress it will always stay in progress. Your suggestion would fulfill that purpose, but in the scenario that the first iteration of the loop returns `InProgress`, the whole function would be shortcut and it wouldn't reach the other namespaces. This approach ensures that at least we send the Delete request for all namespaces no matter the order

KevFan: @sergioifg94 Thanks for the explaination. Yeah fair, I'm good either or anyway, more of a question than anything :smiley: . I think the main fix is there and currently only 1 namespace is passed to this function anyhow :+1:

KevFan:
Verified :+1: 
/lgtm

pkg/resources/ratelimit/envoyconfig.go
sergioifg94: Oh yes I forgot to mention that. For now it's only one namespace so it doesn't make a difference either way :stuck_out_tongue:

PR #1241, opened 2020-10-13
MGDAPI-378: Support rate limit alerts configuration via ConfigMap

sergioifg94:
# Description

Modify AlertReconciler instantiation for the marin3r product to create the AlertConfiguration items from the marin3rconfig.AlertConfig elements that are obtained from a ConfigMap.
Modify bootstrap phase to create this ConfigMap with default values, as well as owner annotations to trigger a reconciliation when the ConfigMap is updated.

## Verification steps

1. Begin an installation from this PR
2. Once the bootstrap phase is complete, check that the alerts ConfigMap has been created
    ```sh
    oc get cm rate-limit-alerts -n redhat-rhmi-operator
    ```
3. Wait for the installation to complete
4. Check that the alerts have been reconciled
   ```sh
    oc get prometheusrules -n redhat-rhmi-marin3r

        NAME                     AGE
    api-usage-alert-level1   123m
    api-usage-alert-level2   123m
    api-usage-alert-level3   123m
    ```
5. Modify the alerts ConfigMap
6. A reconciliation should be triggered
7. Verify that the PrometheusRules CRs are updated with the changes

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist
- [ ] This change requires a documentation update 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added a test case that will be used to verify my changes 
- [ ] Verified independently on a cluster by reviewer

PR #1262, opened 2020-10-22
Adding customer grafana dashboard

laurafitzgerald:
# Description

## Type of change
Adding customer grafana dashboard
Adding tests
E06 - verify customer dashboard exists
E07 - verify customer route is accessible

## Verification.

Run the operator against an openshift cluster

Create an application in 3scale using the following steps

Navigate to `https://3scale-admin.[your_admin_domain].devshift.org/p/admin/onboarding/wizard/intro` and follow the steps outlined in the wizard to create a 3Scale service. You can get your admin domain from Routes under `redhat-rhmi-3scale` namespace, use the admin user credentials from the `system-seed` secret in the 3scale namespace
Navigate to `https://3scale-admin.[your_admin_domain].devshift.org/apiconfig/services/2/integration`
Copy curl command from API Cast staging.

```
wrk -t6 -c400 -d5m <copied curl command from above>
```

- Navigate to the customer grafana route and verify the dashboard is created.
 
Based on the state of the current data set. e.g. what load is being inflicted on the cluster the dashboard should look something like the below images.

Things to verify are
- the per minute dropped requests goes orange if there has been dropped requests in the last minute
- the count for the daily requests accurately reflects the amount of requests hitting the cluster

Screenshot 2020-10-29 at 17 07 22

Screenshot 2020-10-29 at 14 12 11
Screenshot 2020-10-29 at 12 44 07
Screenshot 2020-10-29 at 12 37 17

Alert firing are based on alerts being added as part of [this pr](https://github.com/integr8ly/integreatly-operator/pull/1241)

## Checklist

### DIFF
--- a/pkg/products/grafana/customerDashboard.go
+++ b/pkg/products/grafana/customerDashboard.go
@@ -1,0 +1,606 @@
+package grafana
+
+const CustomerMonitoringGrafanaRateLimitingJSON = `{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "id": 2,
+  "iteration": 1603968007503,
+  "links": [],
+  "panels": [
+    {
+      "collapsed": false,
+      "datasource": null,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 0
+      },
+      "id": 8,
+      "panels": [],
+      "title": "RHOAM API Rate Limiting",
+      "type": "row"
+    },
+    {
+      "cacheTimeout": null,
+      "colorBackground": true,
+      "colorPostfix": false,
+      "colorPrefix": false,
+      "colorValue": false,
+      "colors": [
+        "#299c46",
+        "rgba(237, 129, 40, 0.89)",
+        "#d44a3a"
+      ],
+      "datasource": null,
+      "description": "",
+      "format": "none",
+      "gauge": {
+        "maxValue": 100,
+        "minValue": 0,
+        "show": false,
+        "thresholdLabels": false,
+        "thresholdMarkers": true
+      },
+      "gridPos": {
+        "h": 5,
+        "w": 3,
+        "x": 0,
+        "y": 1
+      },
+      "id": 4,
+      "interval": null,
+      "links": [],
+      "mappingType": 1,
+      "mappingTypes": [
+        {
+          "name": "value to text",
+          "value": 1
+        },
+        {
+          "name": "range to text",
+          "value": 2
+        }
+      ],
+      "maxDataPoints": 100,
+      "nullPointMode": "connected",
+      "nullText": null,
+      "options": {},
+      "postfix": "",
+      "postfixFontSize": "50%",
+      "prefix": "",
+      "prefixFontSize": "50%",
+      "rangeMaps": [
+        {
+          "from": "null",
+          "text": "N/A",
+          "to": "null"
+        }
+      ],
+      "sparkline": {
+        "fillColor": "rgba(31, 118, 189, 0.18)",
+        "full": false,
+        "lineColor": "rgb(31, 120, 193)",
+        "show": false,
+        "ymax": null,
+        "ymin": null
+      },
+      "tableColumn": "",
+      "targets": [
+        {
+          "expr": "increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits[1m])",
+          "instant": true,
+          "refId": "A"
+        }
+      ],
+      "thresholds": "$perMinuteTwentyMillion",
+      "timeFrom": null,
+      "timeShift": null,
+      "title": "Last Minute - No. Requests",
+      "transparent": true,
+      "type": "singlestat",
+      "valueFontSize": "80%",
+      "valueMaps": [
+        {
+          "op": "=",
+          "text": "N/A",
+          "value": "null"
+        }
+      ],
+      "valueName": "avg"
+    },
+    {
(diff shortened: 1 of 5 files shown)


pkg/products/grafana/customerDashboard.go
```diff
@@ -0,0 +1,546 @@
...
+    {
+      "datasource": null,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 0
+      },
+      "id": 8,
+      "title": "3Scale Api Rate Limiting",
```
obrienrobert, line 33: Should `Api` be capitalized?

laurafitzgerald: yes. 👍

test/common/dashboards_exist.go
```diff
@@ -84,7 +84,57 @@ var rhmi2ExpectedDashboards = []dashboardsTestRule{
 	},
 }
 
-func TestIntegreatlyDashboardsExist(t *testing.T, ctx *TestingContext) {
+var customerRHOAMDashboards = []dashboardsTestRule{
+	{
+		Title: "3Scale Api Rate Limiting",
```
obrienrobert, line 89: @laurafitzgerald one more - does this have to be updated now as well?

laurafitzgerald: Yeah. 🙈

laurafitzgerald: Done. 👍

test/common/dashboards_exist.go
```diff
@@ -151,7 +205,14 @@ func TestIntegreatlyDashboardsExist(t *testing.T, ctx *TestingContext) {
...
+	if installType == string(integreatlyv1alpha1.InstallationTypeManagedApi) {
+		return customerRHOAMDashboards
+	}
+	return nil
+}
+
+func getExpectedMiddlewareDashboard(installType string) []dashboardsTestRule {
 	if installType == string(integreatlyv1alpha1.InstallationTypeManagedApi) {
 		return commonExpectedDashboards
 	} else {
```
obrienrobert, line 218: @laurafitzgerald would having a single function for this work, as we're passing in the same `installType` into both functions? If we add any additional install types in the future we might be able to just add an extra case statement. Maybe this isn't the best approach, but it's just an idea. Something like: 
```
func getExpectedDashboards(installType string) []dashboardsTestRule {
	switch installType {
	case string(integreatlyv1alpha1.InstallationTypeManagedApi):
		return append(customerRHOAMDashboards, commonExpectedDashboards...)
	case string(integreatlyv1alpha1.InstallationTypeManaged), string(integreatlyv1alpha1.InstallationTypeSelfManaged:
		return append(commonExpectedDashboards, rhmi2ExpectedDashboards...)
	}
	return append(commonExpectedDashboards, rhmi2ExpectedDashboards...)
}
```

laurafitzgerald: The returns are slightly different there. The first one is returning expected dashboards for the customer grafana and the second one for the middleware grafana.

obrienrobert: Yep I see. Thanks for explaining.

pkg/products/grafana/reconciler.go
```diff
@@ -143,6 +143,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
+	phase, err = r.reconcileGrafanaDashboards(ctx, client, "rate-limiting")
```
MStokluska, line 146: Hey @laurafitzgerald do you think it would be better to make the name of dashboard a variable, just in case we would want to change it or add more dashboards later on?

pkg/products/grafana/reconciler.go
```diff
@@ -179,6 +185,34 @@ func (r *Reconciler) reconcileSecrets(ctx context.Context, client k8sclient.Clie
...
+		},
+	}
+
+	opRes, err := controllerutil.CreateOrUpdate(ctx, serverClient, grafanaDB, func() error {
+		grafanaDB.Labels = map[string]string{
+			"monitoring-key": "customer",
+		}
+		grafanaDB.Spec = grafanav1alpha1.GrafanaDashboardSpec{
+			Json: CustomerMonitoringGrafanaRateLimitingJSON,
+			Name: "rate-limiting",
```
MStokluska, line 203: Same as the previous comment, do you think it would be better to use a const instead?

MStokluska:
@laurafitzgerald I've verified it on a cluster and the Grafana dashboard is created, I've also left a couple of comments. Great job!

test/common/tests.go
```diff
@@ -49,10 +49,12 @@ var (
...
 				{"C01 - Verify Alerts are not pending or firing apart from DeadMansSwitch", TestIntegreatlyAlertsPendingOrFiring},
 				{"C04 - Verify Alerts exist", TestIntegreatlyAlertsExist},
-				{"E01 - Verify Grafana Route is accessible", TestGrafanaExternalRouteAccessible},
+				{"E01 - Verify Middleware Grafana Route is accessible", TestGrafanaExternalRouteAccessible},
 				{"E02 - Verify that all dashboards are installed and all the graphs are filled with data", TestDashboardsData},
-				{"E03 - Verify dashboards exist", TestIntegreatlyDashboardsExist},
+				{"E03 - Verify middleware dashboards exist", TestIntegreatlyMiddelewareDashboardsExist},
 				{"E05 - Verify Grafana Route returns dashboards", TestGrafanaExternalRouteDashboardExist},
+				{"E09 - Verify customer dashboards exist", TestIntegreatlyCustomerDashboardsExist},
+				{"E10 - Verify Customer Grafana Route is accessible", TestCustomerGrafanaExternalRouteAccessible},
```
KevFan, line 57: @laurafitzgerald e2e tests looks to be failing due to these tests because they are `managed-api` specific so might be better to put these tests as a seperate tests suite that is specific to `managed-api` only. What do you think ? :thinking: 
* https://github.com/integr8ly/integreatly-operator/blob/master/test/common/tests.go#L17

laurafitzgerald: Nicely spotted. thanks @KevFan I will make that change. :)

KevFan:
Verified :+1: 
/lgtm
/test e2e

PR #1267, opened 2020-10-22
fix: test A25

psturc:
# Description
https://issues.redhat.com/browse/MGDAPI-394


psturc:
will be fixed as a part of https://github.com/integr8ly/integreatly-operator/pull/1279

PR #1274, opened 2020-10-23
Expose Grafana URL in API usage notifications

jjaferson:
# Description

Expose the Grafana URL in the alert notifications for the API usage

https://issues.redhat.com/browse/MGDAPI-228

## Verification steps

1) Install the operator locally from this branch
2) Get the Prometheus Rules CRs
```oc get prometheusrules -n redhat-rhmi-marin3r``` 
3)  Verify if the `grafanaConsole` is present in the annotation block of the Prometheus Rules CRs
```oc get prometheusrules [prometheus_rule_cr_name]  -o json -n redhat-rhmi-marin3r | jq .spec.groups[0].rules[0].annotations[0].grafanaConsole```

## Type of change

### DIFF
--- a/pkg/products/grafana/reconciler.go
+++ b/pkg/products/grafana/reconciler.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"fmt"
+
 	"github.com/integr8ly/integreatly-operator/pkg/resources/global"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 
@@ -416,4 +417,22 @@ func (r *Reconciler) reconcileHost(ctx context.Context, serverClient k8sclient.C
 	}
 
 	return integreatlyv1alpha1.PhaseCompleted, nil
+}
+
+func GetGrafanaConsoleURL(ctx context.Context, serverClient k8sclient.Client, installation *integreatlyv1alpha1.RHMI) (string, error) {
+
+	grafanaConsoleURL := installation.Status.Stages[integreatlyv1alpha1.ProductsStage].Products[integreatlyv1alpha1.ProductGrafana].Host
+	if grafanaConsoleURL != "" {
+		return grafanaConsoleURL, nil
+	}
+
+	ns := installation.Spec.NamespacePrefix + defaultInstallationNamespace
+	grafanaRoute := &routev1.Route{}
+
+	err := serverClient.Get(ctx, k8sclient.ObjectKey{Name: defaultRoutename, Namespace: ns}, grafanaRoute)
+	if err != nil {
+		return "", fmt.Errorf("Failed to get route for Grafana: %w", err)
+	}
+
+	return "https://" + grafanaRoute.Spec.Host, nil
 }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pkg/products/grafana/reconciler.go [hunk 1]
```diff
@@ -417,3 +418,21 @@ func (r *Reconciler) reconcileHost(ctx context.Context, serverClient k8sclient.C
...
 }
+
+func GetGrafanaConsoleURL(ctx context.Context, serverClient k8sclient.Client, installation *integreatlyv1alpha1.RHMI) (string, error) {
+
+	grafanaConsoleURL := installation.Status.Stages[integreatlyv1alpha1.ProductsStage].Products[integreatlyv1alpha1.ProductGrafana].Host
+	if grafanaConsoleURL != "" {
+		return grafanaConsoleURL, nil
+	}
+
+	ns := installation.Spec.NamespacePrefix + defaultInstallationNamespace
```
laurafitzgerald, line 429: I know that you are using the one in the file here which was already there but we have these centralised here: https://github.com/integr8ly/integreatly-operator/blob/c7c20a4d49150d304257816756787040c593f4a3/test/common/types.go#L45 I think we should try to use that instead.

pkg/products/marin3r/reconciler.go
```diff
@@ -229,7 +230,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 }
 
 func (r *Reconciler) reconcileAlerts(ctx context.Context, client k8sclient.Client, installation *integreatlyv1alpha1.RHMI) (integreatlyv1alpha1.StatusPhase, error) {
-	alertReconciler, err := r.newAlertsReconciler()
+
+	granafaConsoleURL, err := grafana.GetGrafanaConsoleURL(ctx, client, installation)
+	if err != nil {
+		logrus.Errorf("failed to get Grafana console URL %w", err)
```
laurafitzgerald, line 236: Should we return a phaseawaitingcomponents here rather than just log? There could be a case where the url is `""`? Or are we okay to pass an empty string to the newAlertsReconciler until it's found? A case where we don't want to stop the reconcile perhaps.

laurafitzgerald (commented):
@jjaferson just a couple of comments inline. Overall the change looks good. I haven't verified yet.

pkg/products/marin3r/reconciler.go
jjaferson: Yeah, I didn't want to stop the reconcile just because there was an issue getting the Granafa URL, it could prevent the operator from completing an upgrade for example.

pkg/products/grafana/reconciler.go
jjaferson: I agree, although I thought these variables were just used for the automated tests. If we are going to use them I'd suggest we  move it somewhere that is not under the test package, wdyt?

pkg/products/marin3r/reconciler.go
laurafitzgerald: Cool cool cool. I'm fine with that. We do it elsewhere. 👍

MStokluska:
Verified it on the cluster and it works as expected. I think @laurafitzgerald verified the code.
/lgtm

pkg/products/grafana/reconciler.go
laurafitzgerald: I think we'll be addressing namespace stuff in the next sprint/release so for me this is fine as we are using this as you have it here.

PR #1289, opened 2020-10-27
add priorityClassName: rhoam-pod-priority to rhmi.cr.yaml

austincunningham:
# Description

add priorityClassName: rhoam-pod-priority to rhmi.cr.yaml

## Type of change

## Checklist

### DIFF
--- a/deploy/crds/examples/rhmi.cr.yaml
+++ b/deploy/crds/examples/rhmi.cr.yaml
@@ -10,3 +10,4 @@ spec:
   smtpSecret: redhat-rhmi-smtp
   deadMansSnitchSecret: redhat-rhmi-deadmanssnitch
   pagerDutySecret: redhat-rhmi-pagerduty
+  priorityClassName: rhoam-pod-priority
--- a/pkg/controller/installation/installation_controller.go
+++ b/pkg/controller/installation/installation_controller.go
@@ -57,7 +57,7 @@ const (
 	buAlertingEmailAddressEnvName    = "BU_ALERTING_EMAIL_ADDRESS"
 	installTypeEnvName               = "INSTALLATION_TYPE"
 	priorityClassNameEnvName         = "PRIORITY_CLASS_NAME"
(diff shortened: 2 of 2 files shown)


matskiv:
LGTM, but I would like to hear @CathalOConnorRH opinion, since he knows most about the new field.

deploy/crds/examples/rhmi.cr.yaml
```diff
@@ -10,3 +10,4 @@ spec:
   smtpSecret: redhat-rhmi-smtp
   deadMansSnitchSecret: redhat-rhmi-deadmanssnitch
   pagerDutySecret: redhat-rhmi-pagerduty
+  priorityClassName: rhoam-pod-priority
```
CathalOConnorRH, line 13: In theory this should be fine, I just wonder should we also change https://github.com/integr8ly/integreatly-operator/blob/master/pkg/controller/installation/installation_controller.go#L60 as part of this to keep the name consistent?

austincunningham: no objections from me with that change

KevFan:
Not too sure but `managed-service-priority` is also reference here, which might also need to be updated if the name is changed (or better passed in as a variable to this function from the RHMI CR) :thinking: 
* https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/podPriority.go#L41

matskiv:
@KevFan oh, that's interesting.
If the class is configurable via CR, it shouldn't be hard coded here: https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/podPriority.go#L41
right?

KevFan:
> @KevFan oh, that's interesting.
> If the class is configurable via CR, it shouldn't be hard coded here: https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/podPriority.go#L41
> right?

Yeah, that's my understanding, otherwise it's always updated to the hardcoded value :thinking: This might have been missed in the original PR or I'm misunderstanding something :confused:

PR #1296, opened 2020-10-27
update rhsso to 10.0.1

briangallagher:
# Description
Bump from 10.0.0 to 10.0.1 to overcome production issue:
https://issues.redhat.com/browse/OHSS-1620

## Verify - Regression test RHMI / Managed API

- Install RHMI and confirm CR reconciles 
- Verify RHSSO image is 10.0.1
- Run through the sso walkthroughs 

## Verify - OHSS-1620
Add the following block to the Keycloak CR:

```
  keycloakDeploymentSpec:
    experimental:
      args:
        - "bash"
        - "-c"
        - "echo Running SED; sed -i \"s/standalone-openshift.xml/standalone-openshift.xml -Djboss.as.management.blocking.timeout=600/g\" /opt/eap/bin/openshift-launch.sh; echo STARTING; /opt/eap/bin/openshift-launch.sh"
```

- Wait for the Keycloak Pods to be restarted. 

- Check the Keycloak Logs. Check the JBOSS Configured system properties: Confirm the following property is set to be value specified above in `Djboss.as.management.blocking.timeout`

`jboss.as.management.blocking.timeout`

- Using the above example we should see the following in the logs:

`jboss.as.management.blocking.timeout = 600`


R-Lawton:
/lgtm 
verified SSO walkthroughs worked and right version is showing as well as verified the OHSS verification steps with Brian on a call

PR #1301, opened 2020-10-29
MGDAPI-521 - Ensure 3scale deployment config rollout is ready before returning complete

KevFan:
# Description
Previously 3scale phase can report complete before being truly ready. This is due to when the deployment configs changes (from adding topologySpreadConstraints, podPriority etc), there is no check that the rollout has completed before returing complete. 

This causes the products stage to become `PhaseComplete`, then back to `InProgress` before returning to `PhaseComplete` again.

Due to this, tests can run early, where 3scale is not fully ready, causing some flaky failure in tests such as product version, deployment replicas etc
  * OSDE2E - https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/osde2e-stage-aws-addon-integreatly-operator/1321421556915638272

To prevent this, we should check that the deployment config rollout has completed before returning phase complete in the 3scale reconciler

Jira:
* https://issues.redhat.com/browse/MGDAPI-521

## Verification
* Checkout this branch
* Install managed-api
```
make cluster/prepare/local
```
* Monitor products stage on RHMI CR and ensure it does not go back to in progress once it reported complete 
```
while true; do
	echo "products stage: $(oc get rhmi managed-api -n redhat-rhmi-operator -o template --template {{.status.stages.products.phase}})"
done
```
## Type of change

## Checklist

### DIFF
--- a/pkg/products/threescale/objects_test.go
+++ b/pkg/products/threescale/objects_test.go
@@ -678,6 +678,20 @@ func getSuccessfullTestPreReqs(integreatlyOperatorNamespace, threeScaleInstallat
 	systemEnvConfigMap.Namespace = threeScaleInstallationNamepsace
 	oauthClientSecrets.Namespace = integreatlyOperatorNamespace
 	installation.Namespace = integreatlyOperatorNamespace
+	apicastProduction.Namespace = threeScaleInstallationNamepsace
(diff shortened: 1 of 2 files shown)


MStokluska:
Good job! Verified it on the cluster and it is working as expected. Checked code and it looks good to me.
/lgtm

PR #1302, opened 2020-10-29
[WIP]Update prepare-release script for RHOAM

CathalOConnorRH:
# Description
Update prepare-release script to display RHOAM to users rather than RHMI

### DIFF
--- a/scripts/prepare-release.sh
+++ b/scripts/prepare-release.sh
@@ -82,6 +82,24 @@ set_installation_type() {
   fi
 }
 
+set_descriptions() {
+  case $OLM_TYPE in
+   "integreatly-operator")
+      echo "using default descriptions"
+      ;;
+    "managed-api-service")
+      echo "Updating descriptions"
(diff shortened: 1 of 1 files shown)


scripts/prepare-release.sh
```diff
@@ -30,14 +30,14 @@ create_new_csv() {
 
   if [[ -z "$PREVIOUS_VERSION" ]]
     then
-      operator-sdk generate csv --csv-version "$VERSION" --default-channel --operator-name "$OLM_TYPE" --csv-channel=rhmi --update-crds
+      operator-sdk generate csv --csv-version "$VERSION" --default-channel --operator-name "$OLM_TYPE" --csv-channel=rhmi --update-crds --make-manifests=false
```
philbrookes, line 33: have we updated the version of operator-sdk in the pipelines now? Last time we tried to add this flag the operator-sdk version in the pipelines was too old to understand this parameter.

philbrookes: @laurafitzgerald

philbrookes: Ah, @CathalOConnorRH confirmed this is just for the sake of verifying the PR locally, and will not be merged.

PR #1319, opened 2020-11-03
[WIP] MGDAPI-386: Alerts for tiered SKUs soft limits

sergioifg94:
# Description

Add option in rate limit configuration for soft limits in each SKU.
Create alerts that are sent to the BU when the the daily soft limit is reached

## Type of change

## Checklist


pmccarthy:
@sergioifg94 we'll also need to add/update our API usage test case to include these alerts

sergioifg94:
@pmccarthy Yup, I'm adding these and renaming the alerts to `RHOAMApiUsageSoftLimitReachedTierN` as previously discussed

laurafitzgerald:
@sergioifg94 I've installed and things are running okay. Feel free to give me a shout when you have verification steps. Also if you want another set of eyes on the code feel free to request that from others too but the code looks great to me. 👍

sergioifg94:
@laurafitzgerald thanks for the review, and sorry for the delay with the verification steps. I just added them to the description

pkg/products/marin3r/softLimitPrometheusRules.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func mapSoftLimitRules(softDailyLimits []uint32) []monitoringv1.Rule {
+	result := make([]monitoringv1.Rule, len(softDailyLimits))
+
+	for i, softDailyLimit := range softDailyLimits {
+		result[i] = monitoringv1.Rule{
+			Alert: fmt.Sprintf("RHOAMApiUsageSoftLimitReachedTier%d", i+1),
+			Expr: intstr.FromString(
+				fmt.Sprintf("(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits - ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits offset 1d) > %d", softDailyLimit),
```
KevFan, line 39: Just a question on this, as I'm not sure am I doing this right. I've been playing around with this `offset` modifier.
For me it looks to return no data if offset is in a range where there was no data present:

image

But it will return data if there is data thoughout this range: 

image

So because of this, I'm thinking these alerts wont fire on the first day on install due to the above. The alert would fire on the second day once the expression can resolve correctly :thinking: 

image

Wondering is this a concern or am I missing something ? :sweat_smile:

sergioifg94: No, you're right, I noticed this as well, that's why I suggest to modify it to 5 minutes in order to test the query. I'm not sure how much of a concern it'd be, as it means that if any of the limits is reached in the first day after the cluster is created it wouldn't trigger an alert. @pmccarthy wdyt?

laurafitzgerald: Would it work with `increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits[24h]) > <value>`? We are using a similar query in the customer grafana dashboard.

KevFan: Yeah, i think using `increase` would work if this is a concern :thinking:

laurafitzgerald: If we have an alternative which would allow for the metric to be in place rather than not having it available on the first day, Í think we should go with that.

sergioifg94: @laurafitzgerald @KevFan I experimented with the `increase` function but for some reason the values I see in the graph don't make sense (they keep fluctuating even after the requests stop). I managed to come up with a query that defaults to 0 if there's no data. It's a bit complicated but seems to do the job. I'm not an expert in promql so there might be a better way. Wdyt? 

I broke down the query for readability, hope it makes sense:
```sql
# Current number of requests
scalar(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits)
-
scalar(
  # Number of requests a day ago
  (ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits offset 1d) 
  or
  # Current number of requests - Current number of requests => 0 (Using vector(0) doesn't work as the vector types are different)
  (ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits - 
   ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits)
)
```

sergioifg94: Now this one looks a bit nicer
```sql
scalar(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits) - 
scalar(
  (ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits offset 1d)
  or
  on() vector(0)
)
```

laurafitzgerald:
👀

laurafitzgerald:
Hi @sergioifg94 I had verified this change but then was checking alert and noticed that all the alerts were firing.
Screenshot 2020-11-13 at 18 00 40
Could we chat about it on Monday.

laurafitzgerald:
/override e2e
/override rhoam-e2e
/hold cancel

tests already passed. final push is a small text change to the severity level.

PR #1338, opened 2020-11-04
MGDAPI-532: Include anti affinity rules for Multi-AZ support

sergioifg94:
# Description

Add logic to include anti affinity rules in the products pod templates (SSO, User SSO, Rate limit and 3scale)
Refactor logic that updates pod template of product components by unifying into a `UpdatePodTemplateIfExists` function that receives composable mutations.
Add anti affinity rules to 3scale APIManager CR to delegate this logic in the 3scale operator

## Verification steps

The behaviour of the operator in regards to the anti affinity rules is different depending on whether the `FORCE_ZONE_DISTRIBUTIO` env var is set. Install the operator and run it with both combinations

1. Install RHOAM from this PR
    1. Preferred
        > By default the pod affinity rule will be set to preferred
        ```sh
        ```
    2. Required
        ```sh
        FORCE_ZONE_DISTRIBUTION=true INSTALLATION_TYPE=managed-api make code/run
        ```

2. For each of the following components, check that the `podAntiAffinity` field is added to the pod template:
    > 1.  Required
    >     ```yaml
    >     affinity:
    >      podAntiAffinity:
    >        requiredDuringSchedulingIgnoredDuringExecution:
    >          - labelSelector:
    >              matchLabels: # set of labels that identify the set of pod replicas
    >            topologyKey: topology.kubernetes.io/zone
    >     ```
    >
    > 2. Preferred
    >     ```yaml
    >     affinity:
    >       podAntiAffinity:
    >         preferredDuringSchedulingIgnoredDuringExecution:
    >           - weight: 100
    >             podAffinityTerm:
    >               labelSelector:
    >                 matchLabels: # set of labels that identify the set of pod replicas
    >               topologyKey: topology.kubernetes.io/zone
    >     ```

    | Namespace | Kind | Components |
    | --------------- | ------- | ---------------- |
    | `redhat-rhmi-rhsso` | `StatefulSet` | `keycloak` |
    | `redhat-rhmi-usersso` | `StatefulSet` | `keycloak` |
    | `redhat-rhmi-marin3r` | `Deployment` | `ratelimit` |
    | `redhat-rhmi-3scale` | `DeploymentConfig` | `apicast-production`, `apicast-staging`, `backend-cron`, `backend-listener`, `backend-worker`, `system-app`, `system-sidekiq`, `zync`, `zync-que` |

## Type of change

## Checklist

### DIFF
--- a/pkg/controller/installation/installation_controller.go
+++ b/pkg/controller/installation/installation_controller.go
@@ -2,9 +2,11 @@ package installation
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
@@ -593,6 +595,16 @@ func (r *ReconcileInstallation) preflightChecks(installation *integreatlyv1alpha
 
 	eventRecorder := r.mgr.GetEventRecorderFor("Preflight Checks")
 
+	// Validate the env vars used by the operator
+	if err := checkEnvVars(map[string]func(string, bool) error{
+		resources.AntiAffinityRequiredEnvVar: optionalEnvVar(func(s string) error {
+			_, err := strconv.ParseBool(s)
+			return err
+		}),
+	}); err != nil {
+		return result, err
+	}
+
 	if strings.ToLower(installation.Spec.UseClusterStorage) != "true" && strings.ToLower(installation.Spec.UseClusterStorage) != "false" {
 		installation.Status.PreflightStatus = integreatlyv1alpha1.PreflightFail
 		installation.Status.PreflightMessage = "Spec.useClusterStorage must be set to either 'true' or 'false' to continue"
@@ -851,6 +863,37 @@ func (r *ReconcileInstallation) addCustomInformer(crd runtime.Object, namespace 
 
 	logrus.Infof("Cache synced. A %s watch in %s namespace successfully initialized.", gvk, namespace)
 	return nil
+}
+
+func checkEnvVars(checks map[string]func(string, bool) error) error {
+	for env, check := range checks {
+		value, exists := os.LookupEnv(env)
+		if err := check(value, exists); err != nil {
+			return fmt.Errorf("validation failure for env var %s: %w", env, err)
+		}
+	}
+
+	return nil
+}
+
+func optionalEnvVar(check func(string) error) func(string, bool) error {
+	return func(value string, ok bool) error {
+		if !ok {
+			return nil
+		}
+
+		return check(value)
+	}
+}
+
+func requiredEnvVar(check func(string) error) func(string, bool) error {
+	return func(value string, ok bool) error {
+		if !ok {
+			return errors.New("required env var not present")
+		}
+
+		return check(value)
+	}
 }
 
 type multiErr struct {
--- a/pkg/resources/podAntiAffinity.go
+++ b/pkg/resources/podAntiAffinity.go
@@ -1,0 +1,143 @@
+package resources
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strconv"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// ZoneLabel is the label that specifies the zone where a node is
+	ZoneLabel = "topology.kubernetes.io/zone"
+	// AntiAffinityRequiredEnvVar is an environment variable that, when set to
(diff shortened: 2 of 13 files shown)


pkg/controller/installation/installation_controller.go
```diff
@@ -853,6 +865,37 @@ func (r *ReconcileInstallation) addCustomInformer(crd runtime.Object, namespace
...
+	return func(value string, ok bool) error {
+		if !ok {
+			return nil
+		}
+
+		return check(value)
+	}
+}
+
+func requiredEnvVar(check func(string) error) func(string, bool) error {
```
MStokluska, line 889: Hey @sergioifg94 I am little bit confused by the syntax of the checkEnvVars call at line 599, but can't see the requiredEnvVar function being used anywhere? Could you explain what am I missing please?

sergioifg94: `checkEnvVars` is a little function to validate the env vars during the preflight checks, so the installation doesn't progress if an invalid variable is passed. It accepts a map in the form _Env var_ -> _Validating function_, so for now I'm passing the `FORCE_ZONE_DISTRIBUTION`, and a function to check that the value can be parsed as `bool`.

The `optionalEnvVar` and `requiredEnvVar` functions are utility closures for the sake of re utilization. I left `requiredEnvVar` there even though it's unused in case it's needed at some point

MStokluska: Thanks Sergio

MStokluska:
hey @sergioifg94, thanks for the call today with all the explanations. I've verified it on the cluster and it works as expected with the envvar passed in and not. Looking at the code I am a bit confused about how the functions work, so got a bit of upskilling to be done. Initially, it looks good to merge with me but I'd like someone with a bit more experience to look at the code side of things.

pkg/controller/installation/installation_controller.go
sergioifg94: @MStokluska @laurafitzgerald fyi this all could be implemented in a more OOP fashion that might be easier to understand (but more boilerplate code imo):

```go
type EnvVarValidation interface {
   CheckEnvVar(value string, ok bool) error
}

func checkEnvVars(validations map[string]EnvVarValidation) error {
   for env, check := range checks {
    value, exists := os.LookupEnv(env)
    if err := check.CheckEnvVar(value, exists); err != nil {
      return fmt.Errorf("validation failure for env var %s: %w", env, err)
    }
  }

  return nil
}
```

Instead of a function an interface, which would require to implement with structs

```go

type OptionalEnvVar struct {
  InnerCheck EnvVarValidation
}

func (c *OptionalEnvVar) CheckEnvVar(value string, ok bool) error {
  if !ok {
    return nil
  }

  return c.InnerCheck.CheckEnvVar(value, ok)
}

type BoolEnvVar struct {}

func (c *BoolEnvVar) CheckEnvVar(value string, ok bool) error {
  _, err := strconv.ParseBool(value)
  return err
}
```

So the map would look like

```go
map[string]EnvVarValidation {
  resources.AntiAffinityRequiredEnvVar: &OptionalEnvVar{
    InnerCheck: &BoolEnvVar{},
  } 
}
```

pkg/resources/podAntiAffinity.go
```diff
@@ -0,0 +1,143 @@
...
+	if err := client.List(ctx, nodeList); err != nil {
+		return false, err
+	}
+
+	// If there's no nodes, fail
+	if len(nodeList.Items) == 0 {
+		return false, fmt.Errorf("no nodes found")
+	}
+
+	// If there's only one node, directly return true
```
obrienrobert, line 124: Should this be `directly return false`?

sergioifg94: Oops yes, good catch. I'll fix it up with another typo that I found

pkg/resources/podAntiAffinity.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	return MultiAZAntiAffinityPreferred(matchLabels)
+}
+
+// IsAntiAffinityRequired checks whether the anti affinity rule must be set
+// to required or preferred.
+//
+// It currently checks the value of the FORCE_ZONE_DISTRIBUTION bool env var
+func IsAntiAffinityRequired(_ context.Context, _ k8sclient.Client) (bool, error) {
```
obrienrobert, line 100: Are the function parameters needed here? I don't see them being used.

sergioifg94: Not at the moment. For now the result is based on an environment variable, but we might externalize it to a ConfigMap at some point, so I left the parameters that will be needed in the function signature

obrienrobert:
Completed a code review. I didn't verify this on my cluster, but this was already [completed](https://github.com/integr8ly/integreatly-operator/pull/1338#issuecomment-723028250) by Michal.

/lgtm

PR #1344, opened 2020-11-04
MGDAPI-419

CathalOConnorRH:
# Description
Removal of RHMI from customer facing parts of RHOAM installation 

To Validate: Do an install of managed-api and confirm no mentions of RHMI in customer accessible areas such as namespace
secrets etc. Also, check for mentions of managed-api as these should be RHOAM


philbrookes:
/lgtm
/approve

verified on my local cluster.

philbrookes:
/override ci/prow/e2e
/override ci/prow/rhoam-e2e

already passed

PR #1345, opened 2020-11-04
updates to grafana dashboard based on feedback

laurafitzgerald:
# Description

## Type of change

Updating text - SKU -> Requests
Adding two additional panels which are percentages of rejected/total per last 24 hours and last 1 minute.
Making timescale of last 1 minute and last 24 hours clearer

## Verification
Run the operator against an openshift cluster with the installType set to `managed-api`

Create an application in 3scale using the following steps

Navigate to `https://3scale-admin.[your_admin_domain].devshift.org/p/admin/onboarding/wizard/intro` and follow the steps outlined in the wizard to create a 3Scale service. You can get your admin domain from Routes under `redhat-rhmi-3scale` namespace, use the admin user credentials from the `system-seed` secret in the 3scale namespace
Navigate to `https://3scale-admin.[your_admin_domain].devshift.org/apiconfig/services/2/integration`
Copy curl command from API Cast staging.

```
wrk -t6 -c400 -d5m <copied curl command from above>
```

- Navigate to the customer grafana route and verify the dashboard is created.
 
Based on the state of the current data set. e.g. what load is being inflicted on the cluster the dashboard should look something like the below images.

Verify specifically that the SKU entry has changed to Requests for each line on the graph and that the text matches on each of the single stats to what's in the below image.

Screenshot 2020-11-04 at 18 22 49

## Checklist

### DIFF
--- a/pkg/products/grafana/customerDashboard.go
+++ b/pkg/products/grafana/customerDashboard.go
@@ -18,7 +18,7 @@ const CustomerMonitoringGrafanaRateLimitingJSON = `{
   "gnetId": null,
   "graphTooltip": 0,
   "id": 2,
-  "iteration": 1603968007503,
+  "iteration": 1604491512771,
   "links": [],
   "panels": [
     {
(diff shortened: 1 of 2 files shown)


Boomatang (commented):
Still trying to run this in a cluster

PR #1347, opened 2020-11-05
Manifest generation

MStokluska:
# Description
Initial manifests for integreatly-operator. Addresses ticket https://issues.redhat.com/browse/MGDAPI-224
# Verify
- Delete the file in rhoam-manifests called `0.5.0-manifest.txt` 
- Run `make manifest` to generate a manifest file.
- Confirm that the format of the manifest matches the following format: 
```services-rhoam:<VERSION>/<PACKAGE_URL>/<PACKAGE_NAME>-v<VERSION>``` 
- Alternatively, download the following repo: https://gitlab.cee.redhat.com/dmisharo/mangen and follow the instructions to generate a manifest for the integreatly operator via the CLI tool and compare the results.

### CODE REVIEW

File: rhoam-manifests/0.5.0-manifest.txt [hunk 1]
```diff
@@ -0,0 +1,1966 @@
...
+services-rhoam:0.5.0/google.golang.org/genproto-v0.0.0-20200108215221-bd8f9a0ef82f
+services-rhoam:0.5.0/google.golang.org/genproto-v0.0.0-20200115191322-ca5a22157cba
+services-rhoam:0.5.0/google.golang.org/genproto-v0.0.0-20200122232147-0452cf42e150
+services-rhoam:0.5.0/google.golang.org/genproto-v0.0.0-20200204135345-fa8e72b47b90
+services-rhoam:0.5.0/google.golang.org/genproto-v0.0.0-20200212174721-66ed5ce911ce
+services-rhoam:0.5.0/google.golang.org/genproto-v0.0.0-20200224152610-e50cd9704f63
+services-rhoam:0.5.0/google.golang.org/genproto-v0.0.0-20200331122359-1ee6d9798940
+services-rhoam:0.5.0/google.golang.org/genproto-v0.0.0-20200420144010-e5e8543f8aeb
+services-rhoam:0.5.0/google.golang.org/grpc-v0.0.0-20160317175043-d3ddb4469d5a
+services-rhoam:0.5.0/google.golang.org/grpc-v1.14.0
```
Boomatang, line 1706: When validating this file I used the `google.golang.org/grpc` as a random example. The grpc-v1.14.0 is not in my installation.  I am wondering what we would need to do in this case. To help resolve this issue my installed go version is `go1.13.8`

Makefile
```diff
@@ -410,3 +410,7 @@ vendor/check: vendor/fix
 vendor/fix:
 	go mod tidy
 	go mod vendor
+
+.PHONY: manifest
+manifest:
```
Boomatang, line 415: Is there a better name that we could use for this make target. `manifest` is very generic and we use the term manifest when talking about the CSV also. If this is for security checks maybe `security/manifest` maybe a better target name.

scripts/rhoam-manifest-generator.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
Boomatang, line 1: Can you add some detail about this script to the readme in the scripts folder. Details like how to use the script, any variables it may use or what the expected out come maybe.
I also have a question about the name of the script being misleading having the manifest in the name.

scripts/rhoam-manifest-generator.sh
```diff
@@ -0,0 +1,24 @@
...
+file="${RHOAM_VERSION}-manifest.txt"
+
+# Dependencies used
+go mod graph | cut -d " " -f 2 | tr @ - | while read x; do echo "${SERVICE_NAME}:${RHOAM_VERSION}/$x" >> "rhoam-manifests/$pre_sorted_file"; done
+
+# Remove repeating dependencies
+sort -u "rhoam-manifests/$pre_sorted_file" >> "rhoam-manifests/$file"
+
+# Delete pre-sorted file
+rm -rf "rhoam-manifests/$pre_sorted_file"
```
Boomatang, line 24:
```suggestion
rm -f "rhoam-manifests/$pre_sorted_file"
```
if the `--recursive` flag is required then I feel you `pre_sorted_file` variable needs to be renamed.

rhoam-manifests/0.5.0-manifest.txt
```diff
@@ -0,0 +1,1966 @@
+services-rhoam:0.5.0/bazil.org/fuse-v0.0.0-20160811212531-371fbbdaa898
```
Boomatang, line 1: Should this be rebased off master as there is now a 0.7.0 version.

scripts/rhoam-manifest-generator.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+# Set the service name and version
+SERVICE_NAME="services-rhoam"
+
+RHOAM_VERSION=$(grep managed-api-service.v deploy/olm-catalog/managed-api-service/managed-api-service.package.yaml | tail -c 6)
+
+# Manifest folder search
+MW_MANIFESTS=( 3scale cloud-resources grafana marin3r rhsso )
```
Boomatang, line 9: Can this be removed or am I missing something.

Boomatang (commented):
I have left a few small things. The one big problem is your 0.5.0-manifest.txt has entries which I don't have locally. If you rebase off master I will run a larger diff if required.

rhoam-manifests/0.5.0-manifest.txt
MStokluska: I think as the manifest aims to run vulnerability checks on dependencies used, we are safe with whichever official Go version we are using.

Makefile
MStokluska: Use of `/` makes me feel like there should be more than one item available but I agree that the `make manifest` isn't best. Maybe `make manifest/release` as later on, there will also be a manifest for current master? WDYT?

scripts/rhoam-manifest-generator.sh
MStokluska: Thanks @Boomatang - I'll add a description. How else would you like to see this script to be called?

scripts/rhoam-manifest-generator.sh
MStokluska: Yes, this is part of old script, thanks!

rhoam-manifests/0.5.0-manifest.txt
Boomatang: Looking into this a bit more I feel the first time I tried this I may have done something wrong and this should be regraded as an none issue.

scripts/rhoam-manifest-generator.sh
```diff
@@ -0,0 +1,25 @@
...
+RHOAM_VERSION=$(grep managed-api-service.v deploy/olm-catalog/managed-api-service/managed-api-service.package.yaml | tail -c 6)
+
+# Pre-filetered manifest file
+pre_sorted_file="pre-sorted-file.txt"
+
+# Output file name
+file="${RHOAM_VERSION}-manifest.txt"
+
+# Dependencies used
+go mod graph | cut -d " " -f 2 | tr @ - | while read x; do echo "${SERVICE_NAME}:${RHOAM_VERSION}/$x" >> "rhoam-manifests/$pre_sorted_file"; done
```
Boomatang, line 19: I have found a bug with here. If the script is ran twice it wall append the new values to the existing file making it twice as long. It might be best to check if the file already exists and if it does to exist then exit the script telling the user why it exited.

MStokluska:
> I have left a few small things. The one big problem is your 0.5.0-manifest.txt has entries that I don't have locally. If you rebase off master I will run a larger diff if required.

Thanks for the review @Boomatang, I've rebased so the most recent manifest is 0.7.0.

scripts/rhoam-manifest-generator.sh
Boomatang: On calling the script I know you intended people use the make target but someone could call the script directly. 

Also a second point I wan to added in here. When I ran the script using the make target there was no output to tell me what happened. I didn't know what file or where it was created or if the scripted finished successfully. You may want to add some log statements.

scripts/rhoam-manifest-generator.sh
MStokluska: Good spot! thanks Jim. I've pushed an update

PR #1349, opened 2020-11-05
Improvements to scripts, mainly podsAz.sh

matskiv:
# Description
I optimized scripts/podsAz.sh a bit, now it does only 2 API requests.
This made feasible support for all namespaces.

Also small improvement to scripts/disableAz.sh - it will print timestamp(UTC) when it finishes disabling an AZ.

## Type of change


CathalOConnorRH:
verified against Multi-AZ cluster and it's much improved. Great work. 
/lgtm

PR #1357, opened 2020-11-10
Update namespace delete label

CathalOConnor:
# Description
Minor refactor of namespace deletion label


psturc:
/lgtm
verified on my cluster

philbrookes:
We are matching the correct label for OCM atm, so I think this PR is no longer required?

PR #1358, opened 2020-11-10
script to gather hyperfoil stats

austincunningham:
# Description

script to gather hyperfoil stats

## Type of change
independent script for load testing

## Verification
- Run a load test on hyperfoil and get the run number
```bash
export HYPERFOIL_URL=http://hyperfoil-hyperfoil.apps.mwcollabaz-byoc.2mv4.s1.devshift.org"
./hyperfoil-request-stats.sh <RUN>
```
Should get the following results for a run
```bash
./hyperfoil-request-stats.sh 0007
failed =  839
total =  31075
one percent = 310
===============================================================================================
 
Percentage failed =  2%
Percentage passed =  98%
 
===============================================================================================
90% Percentile
 
Mean =  176125478  Max =  1002438655

```

### DIFF
--- a/scripts/hyperfoil-request-stats.sh
+++ b/scripts/hyperfoil-request-stats.sh
@@ -33,11 +33,11 @@ echo 'Percentage failed = ' $percent_failed"%"
 echo 'Percentage passed = ' $percent_passed"%"
 echo " "
 echo '==============================================================================================='
-echo '90% Percential'
+echo '90% Percentile'
 echo " "
-curl -s ${HYPERFOIL_URL}/run/${RUN}/stats/total | jq '.statistics[].summary.percentileResponseTime."90.0"' > percential.txt
-NoOfEndpoints=$(wc -l percential.txt | awk '{print $1}')
-SumPercential=$(awk '{s+=$1} END {print s}' percential.txt)
+curl -s ${HYPERFOIL_URL}/run/${RUN}/stats/total | jq '.statistics[].summary.percentileResponseTime."90.0"' > percentile.txt
+NoOfEndpoints=$(wc -l percentile.txt | awk '{print $1}')
+SumPercential=$(awk '{s+=$1} END {print s}' percentile.txt)
 mean=$(expr $SumPercential / $NoOfEndpoints)
-max=$(grep -Eo '[0-9]+' percential.txt | sort -rn | head -n 1)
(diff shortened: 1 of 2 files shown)


scripts/hyperfoil-request-stats.sh
```diff
@@ -0,0 +1,43 @@
...
+echo " "
+echo '==============================================================================================='
+echo '90% Percentile'
+echo " "
+curl -s ${HYPERFOIL_URL}/run/${RUN}/stats/total | jq '.statistics[].summary.percentileResponseTime."90.0"' > percentile.txt
+NoOfEndpoints=$(wc -l percentile.txt | awk '{print $1}')
+SumPercential=$(awk '{s+=$1} END {print s}' percentile.txt)
+mean=$(expr $SumPercential / $NoOfEndpoints)
+max=$(grep -Eo '[0-9]+' percentile.txt | sort -rn | head -n 1)
+echo "Mean = " $mean " Max = " $max
```
trepel, line 43: It would be good to mention units here. When tried with my run, the max value was `158.33 ms` (seen via stats <run>) and the script printed out  `158334975`. Is that nonoseconds?

trepel (approved):
It think it is good to go. Except for that unit output I would consider excluding rampUp phase from computing the stats (not sure if that's a good idea though and if it is something simple to do) and maybe think of exporting the data in a format that would allow us just to copy/paste to the spreadsheet.

scripts/hyperfoil-request-stats.sh
austincunningham: your right the cli returns ms, but the rest returns ns have amended to ns.

austincunningham:
@trepel should we exclude rampUp? I can add an additional select to the jq to select only steadyLoad e.g. 
```bash
curl -s http://hyperfoil-hyperfoil.apps.mwcollabaz-byoc.3tq7.s1.devshift.org/run/0003/stats/total | jq '.statistics[] | select(.phase | test("steady.")) | select(.metric | test("post.")) | .summary.percentileResponseTime."90.0"'
```

trepel:
@austincunningham I'd say if it is simple enough let's exclude the rampUp from the stats.

philbrookes:
Please re-open this PR if it is still valid.

austincunningham:
@trepel can you put /approve on this

PR #1360, opened 2020-11-10
Bump Grafana version and enable pod Priority

CathalOConnorRH:
# Description
Update granfana operator to v3.6.0 and enable pod priority class name
To validate install this branch as a managed api install type and confirm version 3.6.0 is installed and check the `grafana-deployment` for the pod priority class name of `rhoam-pod-priority`


MStokluska:
Verified on the cluster and working as expected. Good job!
/lgtm

PR #1362, opened 2020-11-10
Updating C18 API usage test case to validate email notifications

pmccarthy:
# Description
Updating C18 API usage test case to validate email notifications for the BU, SRE and Customer respectively.

Also replaced references of `RHMI` with `RHOAM`

Related Jira: https://issues.redhat.com/browse/MGDAPI-537

## Type of change

- [ ] Test Case Enhancement

## Checklist

## Verification
- [ ] Provision an OSD cluster and install RHOAM from master locally (not via the add-on)
- [ ] Complete the SOP end to end

### DIFF
--- a/test-cases/tests/alerts/c18-validate-api-usage-alerts.md
+++ b/test-cases/tests/alerts/c18-validate-api-usage-alerts.md
@@ -14,148 +14,170 @@ tags:
 
 1. Login via `oc` as a user with **cluster-admin** role (kubeadmin):
 
-```
-oc login --token=<TOKEN> --server=https://api.<CLUSTER_NAME>.s1.devshift.org:6443
-```
+   ```
+   oc login --token=<TOKEN> --server=https://api.<CLUSTER_NAME>.s1.devshift.org:6443
+   ```
 
-2. RHMI is installed using the [Jenkins pipeline](https://master-jenkins-csb-intly.cloud.paas.psi.redhat.com/job/ManagedAPI/job/managed-api-install-master/build?delay=0sec) with following parameters:
+2. RHOAM is installed using the [Jenkins pipeline](https://master-jenkins-csb-intly.cloud.paas.psi.redhat.com/job/ManagedAPI/job/managed-api-install-master/build?delay=0sec) with following parameters:
 
-- integreatlyOperatorBranchName: (Go to integreatly-operator github repo and select latest RHMI release branch)
-- ocmClusterLifespan: 36
-- openshiftVersion: (search for the version in the test plan document)
-- useByoc: false
-- multiAZ: false
-- patchCloudResAwsStrCM: false
-- clusterID: api-usage-alerts
-- emailRecipients: <your@email.address>
-- stepsToDo: provision + install
+   - integreatlyOperatorBranchName: (Go to integreatly-operator github repo and select latest RHOAM release branch)
+   - ocmClusterLifespan: 36
+   - openshiftVersion: (search for the version in the test plan document)
+   - useByoc: false
+   - multiAZ: false
+   - patchCloudResAwsStrCM: false
+   - clusterID: api-usage-alerts
+   - emailRecipients: <your@email.address>
+   - stepsToDo: provision + install
+
+3. Valid SMTP credentials have been added to the `redhat-rhoam-smtp` secret
+
+   _NOTE:_ Please reach out to Paul McCarthy <pamccart@redhat.com> for a valid SendGrid API Key
+
+   Details on how to create a valid smtp secret can be found in this [SOP](https://github.com/RHCloudServices/integreatly-help/blob/master/sops/2.x/install/create_cluster_smtp_configuration.md). The verification section of this SOP regarding DMS and PagerDuty configs can be skipped. Also make sure to specify the `redhat-rhoam-operator` namespace rather than `redhat-rhmi-operator`.
 
 ## Steps
 
 1. Validate that rate-limiting alerts ConfigMap has been created
 
-```sh
-oc get cm rate-limit-alerts -n redhat-rhmi-operator
-```
+   ```sh
+   oc get cm rate-limit-alerts -n redhat-rhoam-operator
+   ```
 
 2. Verify that level1, level2 and level3 rate-limiting alerts are present
 
-```sh
-oc get prometheusrules -n redhat-rhmi-marin3r
-```
+   ```sh
+   oc get prometheusrules -n redhat-rhoam-marin3r
+   ```
 
-3. Modify the `sku-limits-managed-api-service` configmap to set the rate limit to 20 requests per minute.
+3. Modify the `sku-limits-managed-api-service` configmap to set the rate limit to 100 requests per minute.
    Run
 
-```sh
-oc edit configmap sku-limits-managed-api-service -n redhat-rhmi-operator
-```
+   ```sh
+   oc patch configmap sku-limits-managed-api-service -n redhat-rhoam-operator -p '"data": {        "rate_limit": "{\n  \"RHOAM SERVICE SKU\": {\n    \"unit\": \"minute\",\n    \"requests_per_unit\": 100\n  }\n}"    }'
+   ```
 
-And insert the following data:
+4. Modify the `rate-limit-alerts` to allow alerts to fire on a per minute basis:
 
-```yaml
-data:
-  rate_limit: |-
-    {
-      "RHMI SERVICE SKU": {
-        "unit": "minute",
-        "requests_per_unit": 100
-      }
-    }
-```
+   ```sh
+   oc patch configmap rate-limit-alerts -n redhat-rhoam-operator -p '"data": {
+       "alerts": "{\n  \"api-usage-alert-level1\": {\n    \"ruleName\": \"RHOAMApiUsageLevel1ThresholdExceeded\",\n    \"level\": \"warning\",\n    \"minRate\": \"80%\",\n    \"maxRate\": \"90%\",\n    \"period\": \"1m\"\n  },\n  \"api-usage-alert-level2\": {\n    \"ruleName\": \"RHOAMApiUsageLevel2ThresholdExceeded\",\n    \"level\": \"warning\",\n    \"minRate\": \"90%\",\n    \"maxRate\": \"95%\",\n    \"period\": \"1m\"\n  },\n  \"api-usage-alert-level3\": {\n    \"ruleName\": \"RHOAMApiUsageLevel3ThresholdExceeded\",\n    \"level\": \"warning\",\n    \"minRate\": \"95%\",\n    \"period\": \"1m\"\n  }\n}"
+   }'
+   ```
 
-4. Modify the `rate-limit-alerts` to allow alerts to fire in a reasonable time from the testing perspective:
+5. Patch the `rhoam` CR to specify BU, SRE and Customer email addresses:
 
-```sh
-oc edit configmap rate-limit-alerts -n redhat-rhmi-operator
-```
+   _NOTE:_ Replace `<rh_username>` references in the below commands with a valid Red Hat username. For example: `pamccart+BU@redhat.com`
 
-and insert the following data:
+   Patch BU and SRE email addresses:
 
-```json
-{
-  "api-usage-alert-level1": {
-    "ruleName": "RHOAMApiUsageLevel1ThresholdExceeded",
-    "level": "warning",
-    "minRate": "80%",
-    "maxRate": "90%",
-    "period": "1m"
-  },
-  "api-usage-alert-level2": {
-    "ruleName": "RHOAMApiUsageLevel2ThresholdExceeded",
-    "level": "warning",
-    "minRate": "90%",
-    "maxRate": "95%",
-    "period": "1m"
-  },
-  "api-usage-alert-level3": {
-    "ruleName": "RHOAMApiUsageLevel3ThresholdExceeded",
-    "level": "warning",
-    "minRate": "95%",
-    "period": "1m"
-  }
-}
-```
+   ```sh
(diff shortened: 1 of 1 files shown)


psturc:
👀

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
```diff
@@ -14,148 +14,126 @@ tags:
...
+   Patch BU and SRE email addresses:
+
+   ```sh
+   oc patch rhmi rhoam -n redhat-rhoam-operator --type merge -p '{"spec":{"alertingEmailAddresses":{"businessUnit":"<rh_username>+BU@redhat.com", "cssre":"<rh_username>+SRE@redhat.com"}}}'
+   ```
+
+   Patch Customer email addresses with two emails to validate multiple addresses:
+
+   ```sh
+   oc patch rhmi rhoam -n redhat-rhoam-operator --type merge -p '{"spec":{"alertingEmailAddress":"<rh_username>+CUSTOMER1@redhat.com <rh_username>+CUSTOMER2@redhat.com"}}'
```
laurafitzgerald, line 81: The format of `alertingEmailAddresses` in the cr is ```  alertingEmailAddresses:
    businessUnit: noreply-test@rhmi-redhat.com
    cssre: noreply-test@rhmi-redhat.com```
In the previous command the `businessUnit` entry is specified. Is there some key for the customer emails that should be in place? or does it sit under `alertingEmailAddress`

psturc: it's a different key:
`alertingEmailAddress` vs `alertingEmailAddresses`

It's actually not easy to distinguish between them actually :D 

I think it would be clearer if we had a different key name for the customer email address...

laurafitzgerald: Yes I agree we should rename that one if we can, I thought it was a typo to be honest.

pmccarthy: Agreed, it's very confusing. I logged this ticket on the backlog yesteday: https://issues.redhat.com/browse/MGDAPI-553

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
```diff
@@ -14,148 +14,126 @@ tags:
...
+   open "https://$(oc get route -n redhat-rhoam-3scale | grep 3scale-admin | awk {'print $2'})"
+   ```
+
+9. Click on `Ok, how does 3scale work?` and follow the 3Scale wizard to create an API
+10. Once on your API overview page, click on `Integration` on the left, then on `Configuration`
+11. Take note of the `example curl for testing` for `Staging-APIcast`
+12. Trigger API Usage alerts
+
+    | Alert to fire                        | `numRequests` | `interval` | `recipients`          |
+    | ------------------------------------ | ------------- | ---------- | --------------------- |
```
laurafitzgerald, line 108: Seein as all requests are now 200 can we move that into the command rather than having to paste it?

laurafitzgerald:
# Verification

## Level 1
### Command
Screenshot 2020-11-12 at 20 33 51
### Prometheus
Screenshot 2020-11-12 at 20 34 08
### Alert
Screenshot 2020-11-12 at 20 32 55
### Email
Screenshot 2020-11-12 at 20 33 27

laurafitzgerald:
## Level 2
### Command
Screenshot 2020-11-12 at 21 04 29



### Prometheus
Screenshot 2020-11-12 at 21 03 47



### Alert
Screenshot 2020-11-12 at 21 03 37


### Email
Screenshot 2020-11-12 at 21 03 58

laurafitzgerald:
## Level 3
### Command
```
for i in {1..500}; do curl -i "route"; sleep 0.0009; done
```

### Prometheus
Screenshot 2020-11-12 at 21 18 10




### Alert
Screenshot 2020-11-12 at 21 15 15



### Email
Screenshot 2020-11-12 at 21 17 55

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
```diff
@@ -14,148 +14,126 @@ tags:
...
+    | RHOAMApiUsageLevel1ThresholdExceeded | 200           | 0.48       | BU and Customers      |
+    | RHOAMApiUsageLevel2ThresholdExceeded | 200           | 0.43       | BU and Customers      |
+    | RHOAMApiUsageLevel3ThresholdExceeded | 200           | 0.4        | BU, Customers and SRE |
```
psturc, line 111: @pmccarthy what's the motivation behind changing these intervals and the script? They worked in previous release testing, but I cannot make the alerts firing with the new script and new values.

laurafitzgerald: @psturc I was able to get the alerts to fire yesterday with the intervals that I included in the comments. I think that these intervals are going to be different in each case so we probably need to update the test-case to tell the executor that they will need to try different intervals. The way that I did it was start with 0.60 as the interval and then check in prometheus what number it was at. In my case it was aroun 60 so I kept reducing the interval and running the command until I got to 82 and then to 86 and then to 92. I think we'd need to write steps for the verify to do that rather than a specific set of intervals. cc @pmccarthy

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
```diff
@@ -14,148 +14,126 @@ tags:
...
-    | Alert to fire                        | `numRequests` | `interval` |
-    | ------------------------------------ | ------------- | ---------- |
-    | RHOAMApiUsageLevel1ThresholdExceeded | 170           | 0.71       |
-    | RHOAMApiUsageLevel2ThresholdExceeded | 186           | 0.645      |
-    | RHOAMApiUsageLevel3ThresholdExceeded | 240           | 0.5        |
+    ```sh
+    open "https://$(oc get routes prometheus-route -n redhat-rhoam-middleware-monitoring-operator -o jsonpath='{.spec.host}')"
+    ```
 
-    > Only the requests for level 3 should be eventually failing with a `429 Too Many Requests` status code
```
psturc, line 141: Maybe we could leave this check here?

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
```diff
@@ -14,148 +14,126 @@ tags:
...
     >
     > ```
-    > increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits[1m]))
+    > increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits[1m])
     > ```
     >
     > The alert should fire when the value is in the expected range
-18. Repeat the same process for the remaining alerts and verify that they are firing as expected. Make sure to wait at least a minute before running each script.
+
+13. Once each of the above `for loops` have completed, verify that a `FIRING` and associated `RESOLVED` email notification is sent. Check the `to` field of the email to ensure that it matches the `recipients` listed in the table above.
```
psturc, line 139: @pmccarthy I was chatting to @laurafitzgerald about the customer grafana dashboard - that we don't have it covered properly with a test case and I suggested to include the verification steps to this test case, since it perfectly connects to it.

the verification steps would be:
1. Open customer grafana dashboard
```
open "https://$(oc get routes grafana-route -n redhat-rhoam-customer-monitoring-operator -o jsonpath='{.spec.host}')"
```
2. Select the **Rate Limiting** dashboard
3. Run the script above with the following configuration:`export numRequests=10000 interval=0.1`
4. Check the graphs are updating and showing the correct data (you should see more than 90% of requests being rejected)

Screenshot 2020-11-13 at 12 46 44

wdyt?

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
pmccarthy: Sure, can do 👍

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
laurafitzgerald: After discussing it with @psturc I'm also in favour of reverting to the old script which was working before. I found the above to be a very lengthy process and probably will be each time.

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
pmccarthy: Table has changed again along with the `numrequests` based on latest changes. Can chat about it on Monday

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
pmccarthy: I've made a number of updates and am trying a new approach for triggering the alerts that doesn't require intervals. Maybe we can see how it goes and if it's still not working we can revert back to the original script.

Personally I'm not a fan of creating a script as I'm lazy. Copying a one liner seems easier to me :)

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
pmccarthy: Done 👍

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
pmccarthy: Section has been added 👍

pmccarthy:
@laurafitzgerald @psturc hey guys, I've made more changes to this PR. Let me know what you think. Can sync up about it next week. Thanks

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
laurafitzgerald: Hey @pmccarthy I had a cluster up so I checked if the commands trigger the alert and I was able to reliably trigger the alerts following the steps here. I'm cool with these new steps. 👍  will come back to the other verification Monday.

laurafitzgerald:
hey @pmccarthy I thought the problem was with prow initially so reran the test but looks like there is a lint problem in the file. You can run `make test-cases/fix` to fix it.

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
```diff
@@ -14,148 +14,172 @@ tags:
...
+      ```sh
+      curl -i https://<DUMMY_URL>//?user_key=<DUMMY_KEY>
+      ```
+
+      Click the `Execute` button in the Prometheus console again until a count other than `0` is returned
+
+      Now, every few seconds click the `Execute` button. As soon as the count goes back to `0` run the command prepared earlier in this section. This ensures that we send our requests in alignment with the Prometheus metrics scrape interval:
+
+      ```sh
+      for i in {1..$numRequests}; do curl -i https://<DUMMY_URL>//?user_key=<DUMMY_KEY>; done
```
psturc, line 154: With the script above, I wasn't able to reach the alert thresholds, because those curl commands took too long to finish. If I ran the commands in background, they were working fine.
```suggestion
      for i in {1..$numRequests}; do curl -i https://<DUMMY_URL>//?user_key=<DUMMY_KEY>& done
```

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
```diff
@@ -14,148 +14,172 @@ tags:
...
-    > ```
-    >
-    > The alert should fire when the value is in the expected range
-18. Repeat the same process for the remaining alerts and verify that they are firing as expected. Make sure to wait at least a minute before running each script.
+    Select the Rate Limiting dashboard
+
+    Run the following command:
+
+    ```sh
+    for i in {1..150}; do curl -i https://<DUMMY_URL>//?user_key=<DUMMY_KEY>; done
```
psturc, line 176:
```suggestion
    for i in {1..1000}; do curl -i https://<DUMMY_URL>//?user_key=<DUMMY_KEY>& done
```

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
```diff
@@ -14,148 +14,172 @@ tags:
...
+    ```
+
+    _NOTE:_ The above command should eventually result in failing `429 Too Many Requests` status codes. This is to be expected. If no requests have been rejected make sure to check the current Rate Limit configuration in the `sku-limits-managed-api-service` configmap of the `redhat-rhoam-operator` namespace
+
+    While the command is running, check that the graphs in the Grafana dashboard are updating every minute. For example, the "Last 1 Minute - Rejected" percentage should be above 90%
+
+    See example:
+
+    ![](https://user-images.githubusercontent.com/4881144/99069295-5e34fc00-25ae-11eb-874b-3cf5c4a03736.png)
```
psturc, line 185:
```suggestion
    See [this example](https://user-images.githubusercontent.com/4881144/99288530-07dced00-283c-11eb-9cba-906151dd7dfb.png)
```
^ I would add a web link to the file, since the test cases will accessed from JIRA

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
```diff
@@ -14,148 +14,172 @@ tags:
...
+
+    Run the following command:
+
+    ```sh
+    for i in {1..150}; do curl -i https://<DUMMY_URL>//?user_key=<DUMMY_KEY>; done
+    ```
+
+    _NOTE:_ The above command should eventually result in failing `429 Too Many Requests` status codes. This is to be expected. If no requests have been rejected make sure to check the current Rate Limit configuration in the `sku-limits-managed-api-service` configmap of the `redhat-rhoam-operator` namespace
+
+    While the command is running, check that the graphs in the Grafana dashboard are updating every minute. For example, the "Last 1 Minute - Rejected" percentage should be above 90%
```
psturc, line 181:
```suggestion
    While the command is running, check that the graphs in the Grafana dashboard are updating every minute. For example, the "Last 1 Minute - Rejected" percentage should be around 90%
```

psturc (commented):
@pmccarthy  just a few suggest suggestions, otherwise LGTM

pmccarthy:
Thanks @psturc - all suggestions have now been added

PR #1365, opened 2020-11-11
Add rate limit pod priority

MStokluska:
# Description
This PR addresses ticket https://issues.redhat.com/browse/MGDAPI-551. 

#Verification
Install from this branch and verify that the pod priority class name of `rhoam-pod-priority` is set in `rate-limit` deployment under marin3r namespace.

### DIFF
--- a/pkg/products/marin3r/rateLimitService_test.go
+++ b/pkg/products/marin3r/rateLimitService_test.go
@@ -31,7 +31,8 @@ func TestRateLimitService(t *testing.T) {
 			Reconciler: NewRateLimitServiceReconciler(&marin3rconfig.RateLimitConfig{
 				Unit:            "minute",
 				RequestsPerUnit: 1,
-			}, "redhat-test-marin3r", "ratelimit-redis"),
+			},
+				&integreatlyv1alpha1.RHMI{}, "redhat-test-marin3r", "ratelimit-redis"),
 			InitObjs: []runtime.Object{
 				&corev1.Secret{
 					ObjectMeta: v1.ObjectMeta{
@@ -109,6 +110,7 @@ func TestRateLimitService(t *testing.T) {
 					Unit:            "minute",
 					RequestsPerUnit: 1,
 				},
+				&integreatlyv1alpha1.RHMI{},
 				"redhat-test-marin3r",
 				"ratelimit-redis",
 			).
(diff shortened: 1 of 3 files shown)


pkg/products/marin3r/rateLimitService_test.go
```diff
@@ -31,7 +31,12 @@ func TestRateLimitService(t *testing.T) {
 			Reconciler: NewRateLimitServiceReconciler(&marin3rconfig.RateLimitConfig{
 				Unit:            "minute",
 				RequestsPerUnit: 1,
-			}, "redhat-test-marin3r", "ratelimit-redis"),
+			},
+				&integreatlyv1alpha1.RHMI{
+					Spec: integreatlyv1alpha1.RHMISpec{
+						PriorityClassName: "rhoam-pod-priority",
```
obrienrobert, line 37: Would it make sense to make for this value to be a const instead of hard-coding it? If the value ever needs to be changed, we only need to make a change to the const value.

MStokluska: Hey @obrienrobert , thanks for the review/question! 
I think it should not matter, reason why is that it is a test so we are mocking the value in the RHOAM CR and later on, comparing if the expected result matches the value above. So essentially it could be anything at all. I think that we could even live without these values being set at all in the first 3 tests, as it is only the last one that cares about this value.

obrienrobert: @MStokluska understood, thanks for the detailed reply!

obrienrobert: > I think that we could even live without these values being set at all in the first 3 tests
Just reading over this again - If these values are not needed for the first 3 tests, could they be removed?

obrienrobert: > I think that we could even live without these values being set at all in the first 3 tests

If we don't need this value for the first 3 tests, could it be removed?

obrienrobert:
Installed and verified that the `priorityClassName` gets set in the deployment, and also the the pods:

rate_limit_priority_class

pkg/products/marin3r/rateLimitService_test.go
MStokluska: @obrienrobert 
Sure! Done! Sorry was meant to do it yesterday!

PR #1368, opened 2020-11-11
Api 519

laurafitzgerald:
# Description

## Type of change

- Small updates to fix typo and some functions inside the customer grafana dashboard.
- Adding a new alert which evaluates every 30 minutes that checks if the total requests has gone over the total allowed per minute in any minute in the last 30 minutes

## Verification.

This verification contains some maths. Sorry. ➕ ➖  

Steps

1. Install RHOAM
2. Create 3scale Application 

- Navigate to https://3scale-admin.[your_admin_domain].devshift.org/p/admin/onboarding/wizard/intro and follow the steps outlined in the wizard to create a 3Scale service. You can get your admin domain from Routes under redhat-rhmi-3scale namespace, use the admin user credentials from the system-seed secret in the 3scale namespace
- Navigate to https://3scale-admin.[your_admin_domain].devshift.org/apiconfig/services/2/integration
- Copy the route from the curl command from API Cast staging.

3. Verify Grafana Changes 

Run 21 curl commands , heres a command for your convenience.
```for i in {1..21}; do curl "<your app route>" -v; done```

Navigate to the grafana instance in the `redhat-rhmi-customer-monitoring` namespace
Navigate to the rate limiting dashboard

Verify that the last 1 Minute single stat entires have 21 for the total and 1 for rejected. Image below as example, The reason for this verification is that the query has been updated to account for discrepencies in the increase function when using [1m] time period, you can see the change in the code.
Screenshot 2020-11-11 at 15 34 29

While you're there check that the text change has taken effect. `Reqests` -> `Requests` in the Last 1 minute No of Requests single stat. 😉 

4 Verify the Alert is fires/doesn't fire as expected

4.1 Verify that the Alert is created with the inital 20 limit on install
Screenshot 2020-11-17 at 18 24 24

4.2 Update data entry `"requests_per_unit": 13389,` in  config map `sku-limits-managed-api-service`

4.3 Navigate to the Prometheus Alerts Dashboard in the `redhat-rhoam-middleware-operator` namespace

Verify that the Alert is in place with the updated request_per_unit entry. It may take a minute or so for the update to propogate through to prometheus

The alert should appear the same as above but the query and the message should be updated with the new limit.

4.4 Load the api

```
wrk -t6 -c100 -d5m <copied url from above>
```

4.5. Check that the load has been sufficient to push the requests up 
Click on the expr in the alert. If you haven't run the command above yet there may be no data. Click on the Graph and you can see historical data. If the total amount is over 13888 * 2 check the alert.
It shouldn't fire for 30 minutes. You'll need to wait to verify that. the 30 minute interval is from when the alert is created.
After 30 minutes verify the alert is firing.
Screenshot 2020-11-11 at 11 31 07
Wait 30 minutes
Verify that the alert isn't firing anymore.

`If you're reading this, you've gotten out. And if you've come this far, maybe you're willing to come a little further.`

5. In order to verify the expression is working as close to accurately as possible.

5.1 Add additional graphs to your Prometheus view.

the following 3 queries are required as graphs - set up the three as graphs together.
1. `ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits`
2. `increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits[1m]) *  (60 - 30) / 60`
3. `max_over_time(increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits[1m])[30m:]) `

5.3 Wait a couple of minutes after you have done any previous loading then run the same load request as previous. 
Run the load script from above
Wait 1 minute for the data to be scraped.

5.4 Run the execute command for each of the graphs you've set up.

5.5 From graph 1 take the starting and ending value of the first increase you see related to the latest load script.
Subtract the lower from the greater. Let's call that the `new increase`.

5.6 Check the relative increase on Graph 2. the amount should match the `new increase`.

5.7 Then go to graph number 3 and the amount for the related time should be 2*`new increase`.

The reason to verify this is to show the need to have ` * 2` at the end of the alert expression and to show that all the relevant queries are providing us with the most accurate data possible.

6. Verify Grafana Route in Usage Alerts
6.1 Navigate to Prometheus
6.2 Verify that the Grafana Route in the below three alerts is update to `https://<your-clusters-customer-grafana-router>/d/66ab72e0d012aacf34f907be9d81cd9e/rate-limiting`

```
RHOAMApiUsageLevel1ThresholdExceeded (0 active)
RHOAMApiUsageLevel2ThresholdExceeded (0 active)
RHOAMApiUsageLevel3ThresholdExceeded

```
6.3 Verify text in alerts has been updated to `"Total API usage in your API Management service is between %s%% and %s%% percent of the allowable threshold, %d requests per %s, during the last %s",`

In particular verify that the percentage symbols are parse correctly. e.g. the entry should be Total API usage in your API Management service is between 80% and 90% percent of the allowable threshold, X requests per minute, during the last 4 hours"

7. There are now some changes to the existing config map for rate limit alerts. The following steps would be good to verify the new logic works as expected and that the older config isn't broken.

7.1 Verify the `rate-limit` config map is in and has the following data
```{
  "api-usage-alert-level1": {
    "type": "Threshold",
    "level": "info",
    "ruleName": "RHOAMApiUsageLevel1ThresholdExceeded",
    "period": "4h",
    "threshold": {
      "minRate": "80%",
      "maxRate": "90%"
    }
  },
  "api-usage-alert-level2": {
    "type": "Threshold",
    "level": "info",
    "ruleName": "RHOAMApiUsageLevel2ThresholdExceeded",
    "period": "2h",
    "threshold": {
      "minRate": "90%",
      "maxRate": "95%"
    }
  },
  "api-usage-alert-level3": {
    "type": "Threshold",
    "level": "info",
    "ruleName": "RHOAMApiUsageLevel3ThresholdExceeded",
    "period": "30m",
    "threshold": {
      "minRate": "95%"
    }
  },
  "rate-limit-spike": {
    "type": "Spike",
    "level": "warning",
    "ruleName": "RHOAMApiUsageOverLimit",
    "period": "30m"
  }
}
```

7.2 Verify that the four alerts, 1 new and 3 rafactored are in place

```
oc get prometheusrules -n redhat-rhoam-marin3r
NAME                     AGE
api-usage-alert-level1   3s
api-usage-alert-level2   3s
api-usage-alert-level3   3s

rate-limit-spike         3s
```

`Do you know what the Mexicans say about the Pacific? They say it has no memory. That's where I want to live the rest of my life; A warm place with no memory.`  of this verification.


pkg/products/marin3r/apiUsagePrometheusRules.go
```diff
@@ -80,10 +80,10 @@ func mapAlertsConfiguration(namespace, rateLimitUnit string, rateLimitRequestsPe
 					Alert: alertConfig.RuleName,
 					Annotations: map[string]string{
 						"message": fmt.Sprintf(
-							"RHOAM API usage is between %s and %s of the allowable threshold, %d requests per %s, during the last %s",
+							"Total API usage in your API Management service is between %s%% and %s%% percent of the allowable threshold, %d requests per %s, during the last %s",
```
pmccarthy, line 83: Is this correct syntax? `%s%%`? Should it not be `%s`?

laurafitzgerald: @pmccarthy `%s%%` would result in `90%`. The double % is the syntax required by fmr.Sprintf to produce a % symbol in the output. You can see that running here: https://play.golang.org/p/efuSxUz4vRl.
Although I see now I've got the word `percent` in there as well. I'll remove that.

pmccarthy: Ok cool, just said I'd ask in any case 👍

pmccarthy:
@laurafitzgerald It looks like this new alert is not configurable via the `rate-limit-alerts` configmap in the `redhat-rhoam-operator` namespace like the other API usage alerts? We'll need to do this here so that SRE can tweak in production if needed

pkg/products/marin3r/apiUsagePrometheusRules.go
```diff
@@ -41,59 +41,84 @@ func (r *Reconciler) newAlertsReconciler(grafanaConsoleURL string) (resources.Al
...
+						Alert: alertConfig.RuleName,
+						Annotations: map[string]string{
+							"message": fmt.Sprintf(
+								"Total API usage in your API Management service is between %s%% and %s%% of the allowable threshold, %d requests per %s, during the last %s",
+								alertConfig.MinRate, upperMessage, rateLimitRequestsPerUnit, rateLimitUnit, alertConfig.Period,
+							),
+							"grafanaConsole": grafanaDashboardURL,
+						},
+						Expr:   intstr.FromString(expr),
+						Labels: map[string]string{"severity": alertConfig.Level},
```
sergioifg94, line 115:
```suggestion
                var alert resources.AlertConfiguration
		switch alertConfig.Type {
		case marin3rconfig.AlertTypeSpike:
		        alert = mapSpikeAlert(alertConfig /*, other parameters */)
                case marin3rconfig.AlertTypeThreshold:
                        alert = mapThresholdAlert(alertConfig /*, other parameters */)
                default:
                        alert = mapThresholdAlert(alertConfig /*, other parameters */)
```

Maybe we could externalize the creation of each type of alert in functions? It'd make the switch shorter, and we could reuse the threshold alert function to default in case the type doesn't match (existing config maps that don't have this field or invalid values)

laurafitzgerald: Yes, think it would be easier to read also. 👍

laurafitzgerald: Do we definitely want to do a default where it's the ThresholdAlert? I would expect if it's not either of those types we don't want to do anything as it may be the case that a new type is added and the switch is maybe just not updated?

sergioifg94: Makes sense, another option would be to `continue` on default, maybe log that an unsupported type was found. I was thinking of defaulting  to this in case the operator is run on a cluster that already has the ConfigMap (which wouldn't include the new field), but will probably occur just in development so not really a concern

laurafitzgerald: I've added an info log the default. Agree it would be good to at least log so that the it's obvious in the future if there is a case where the type is unsupported. 👍

laurafitzgerald:
Hi @pmccarthy that change is implemented now. The pr is ready for review again if you have some time. Thanks @sergioifg94 for the review and feedback. 👍

laurafitzgerald:
I know I cried wolf earlier 🐺  but it's ready for verification now. 👍

pmccarthy:
@laurafitzgerald the new alert needs to be added to the alert manager template: https://github.com/integr8ly/integreatly-operator/blob/master/templates/monitoring/alertmanager/alertmanager-application-monitoring.yaml#L20-L31

With the recipient set to `BUandCustomer `

pmccarthy:
We also need to update the API usage test case to include this new alert: https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/alerts/c18-validate-api-usage-alerts.md

It looks like the `RHOAMApiUsageSoftLimitReachedTier` alerts are missing from the test case too but this can be addressed separately. I'll log a new Jira

pkg/products/marin3r/apiUsagePrometheusRules.go
```diff
@@ -41,60 +41,97 @@ func (r *Reconciler) newAlertsReconciler(grafanaConsoleURL string) (resources.Al
...
+			// upper limit is set, if not, assign the lower one
+			expr := *lowerExpr
+			upperMessage := "*"
+			if upperExpr != nil {
+				expr = fmt.Sprintf("%s and %s", expr, *upperExpr)
+				upperMessage = *alertConfig.Threshold.MaxRate
+			}
+			annotations := map[string]string{
+				"message": fmt.Sprintf(
+					"Total API usage in your API Management service is between %s%% and %s%% of the allowable threshold, %d requests per %s, during the last %s",
```
pmccarthy, line 88: Message is showing two `%` symbols:

Screenshot 2020-11-19 at 15 55 25

pmccarthy:
Ran through verification and everything worked as expected aside from some minor tweaks that I've commented on above.

Customer dashboard displaying requests and rejected requests:

Screenshot 2020-11-19 at 14 52 08

Alert triggering:

Screenshot 2020-11-19 at 14 52 56

Prometheus queries returning expected values:

Screenshot 2020-11-19 at 15 54 05

Changes to alert configurations in the  rate limit configmap are reflected in prometheus:

Screenshot 2020-11-19 at 16 01 53

pmccarthy:
Test case update will be included as part of https://issues.redhat.com/browse/MGDAPI-586

Alert Manager configs and `%%` typo will be addressed as part of https://issues.redhat.com/browse/MGDAPI-585

/lgtm
/approve

laurafitzgerald:
As per image below the e2e tests are passed. just waiting for prow to report it back, could we go ahead and override? cc @philbrookes @pmccarthy 

Screenshot 2020-11-20 at 12 37 58

PR #1370, opened 2020-11-12
Remove RHMI reference in user sso console link

Boomatang:
# Description

Remove RHMI reference from the user sso console link which is only used in managed-api installs.

## Type of change

## Checklist


KevFan:
Verified :+1: 

image

/lgtm

philbrookes:
is this going to upset RHMI installations?

KevFan:
@philbrookes No, should not, this is exclusive to RHOAM, and ensure's it's deleted for RHMI also
* https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/rhssouser/reconciler.go#L1158-L1164
* https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/rhssouser/reconciler.go#L1193-L1205

PR #1371, opened 2020-11-12
add test case for authenticate RHOAM with LDAP IDP

jjaferson:
# Description

Test case to validate the integration of RHOAM with LDAP IDP

https://issues.redhat.com/browse/MGDAPI-520

## Verification Steps

Go through the steps in the test case and make sure it works as expected

### DIFF
--- a/test-cases/tests/authorization/b09-validate-authentication-with-ldap-idp.md
+++ b/test-cases/tests/authorization/b09-validate-authentication-with-ldap-idp.md
@@ -1,0 +1,179 @@
+---
+# See the metatadata section in the README.md for details on the
+# allowed fields and values
+components:
+  - product-3scale
+  - product-sso
+estimate: 1h
+products:
+  - name: rhoam
+    environments:
+      - osd-fresh-install
+    targets:
+      - 1.0.0
+---
+
+# B09 - Validate Authentication with LDAP IDP
+
+## Description
+
+We want to validate that RHOAM is able to authenticate to an LDAP server via IDP.
+
+## Steps
+
+First lets start the AWS EC2 instance that have the LDAP service installed.
+
+1. Access the AWS EC2 console, you can find the credentials in https://gitlab.cee.redhat.com/integreatly-qe/vault/-/blob/master/SECRETS.md
+
+2. On the AWS EC2 console, right click on the LDAP Server EC2 instance and select start instance.
+
+   > Instance should change to running state
+
+3. Once the instance is in a running state verify if the LDAP service is up and running.
+
+   ```bash
+   // you need to change the <EC2 instance IP> to the EC2 instance IP
+
+   curl -vv "ldap://<EC2 instance IP>:389/dc=ec2-3-133-150-27,dc=us-east-2,dc=compute,dc=amazonaws,dc=com?uid?sub?(uid=rhoam)"
+
+   DN: uid=rhoam,ou=people,dc=ec2-3-133-150-27,dc=us-east-2,dc=compute,dc=amazonaws,dc=com
+    uid: rhoam
+   ```
+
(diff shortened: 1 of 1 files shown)


psturc:
👀

test-cases/tests/authorization/b09-validate-authentication-with-ldap-idp.md
```diff
@@ -0,0 +1,117 @@
...
+
+   > You should be redirected to the 3scale login page
+
+5. Click on the `Authenticate through Red Hat Single Sign-On` link
+
+   > You should be redirected to Red Hat single sign on page
+
+6. Click on the LDAP option.
+
+7. Enter _rhoam_ for `username` and `password` and click the log in button.
```
psturc, line 107: do you think it would make sense to have 2 users, that we would use for `customer-admin-user` and `regular-user` scenarios? e.g. `rhoam-test-user` and `rhoam-customer-admin`?

test-cases/tests/authorization/b09-validate-authentication-with-ldap-idp.md
```diff
@@ -0,0 +1,117 @@
+---
+# See the metatadata section in the README.md for details on the
+# allowed fields and values
+automation:
+  - INTLY-0000
```
psturc, line 5: should this be updated?

test-cases/tests/authorization/b09-validate-authentication-with-ldap-idp.md
```diff
@@ -0,0 +1,117 @@
...
+components:
+  - product-3scale
+  - product-sso
+estimate: 1h
+products:
+  - name: rhoam
+    environments:
+      - osd-fresh-install
+    targets:
+      - 0.2.0
```
psturc, line 15: I think we should add a next release version here, i.e. `1.0.0`

psturc (changes requested):
just a few comments, otherwise `test-cases-lint` job has failed. need to rerun `npm run prettier && npm run lint`

test-cases/tests/authorization/b09-validate-authentication-with-ldap-idp.md
jjaferson: hmm, not sure it would make much difference, all users when they are imported they will be created as `test-user` afaik unless we can add some flag in the ldap server that the operator can detect and distinguish which group they belong. 

When I was testing I manually changed the user between `dedicated-admin` and `test-user` and it worked as expected.

test-cases/tests/authorization/b09-validate-authentication-with-ldap-idp.md
jjaferson: I'll remove field as we don't have any ticket to automate this test case

test-cases/tests/authorization/b09-validate-authentication-with-ldap-idp.md
psturc: We're going to perform a full test run against a cluster with LDAP IDP (i.e. all relevant manual tests from test case folder). Some tests require logging in as regular user, some require customer-admin permissions, so I think having 2 users with dedicated permissions could be helpful for us.

Also I think that adding a step for changing the permissions for the customer(dedicated) admin in ocm should be sufficient.

Wdyt?

jjaferson: in this case that makes sense to have two users, one dedicated-admin and regular user, I'll create them a update the test case

psturc:
@jjaferson 
running `npm run prettier && npm run lint` and pushing the new changes should fix that failing test

PR #1372, opened 2020-11-12
E09 e2e flaky test fix

MStokluska:
# Description
Addresses https://issues.redhat.com/browse/MGDAPI-528, fixes the E09 e2e test.

# How
Currently, customer monitoring grafana pod does not support "curl" command. I am re-using monitoring grafana pod to curl customer monitoring grafana in order to get customer monitoring dashboards. Also, as the rate limit dashboard is within a top-level dashboard, the top-level dashboard must be skipped. 

# Verification
PR e2e for RHMI/verified manually
rhoam e2e needs manual verification

### DIFF
--- a/test/common/dashboards_exist.go
+++ b/test/common/dashboards_exist.go
@@ -3,7 +3,7 @@ package common
 import (
 	goctx "context"
 	"encoding/json"
-	"os"
+	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


KevFan:
Verified :+1: 

/lgtm

test/common/dashboards_exist.go
```diff
@@ -131,33 +121,20 @@ func TestIntegreatlyCustomerDashboardsExist(t *testing.T, ctx *TestingContext) {
...
-	if err != nil {
-		t.Fatal("failed to list pods", err)
-	}
-
-	if len(pods.Items) != 1 {
-		t.Fatal("grafana pod not found")
-	}
+	monitoringGrafanaPods := getGrafanaPods(t, ctx, MonitoringOperatorNamespace)
 
 	output, err := execToPod("curl localhost:3000/api/search",
```
philbrookes, line 136: Does this work? I got errors saying curl didn't exist when I tried this command

philbrookes:
@KevFan Can you confirm this test was executed locally during verification? The ci/prow/rhoam/e2e does not currently execute the e2e tests so is not a valid verification of this PR and this test also does not run in the RHMI e2e tests.

test/common/dashboards_exist.go
KevFan: :thinking: the functional test itself passed locally, but yeah when the curl command is run in the grafana pod terminal manually, there is an error, not sure whats happening here :man_shrugging: 

Maybe the solution here would be to get any pod **but** the grafana pod in the monitoring namespace and curl to the grafana pod ip instead ? What do you think @MStokluska ?

MStokluska: @philbrookes @KevFan monitoring grafana pods do have curl, customer monitoring grafana pods don't. 
This is why I am using monitoring grafana pods to curl the customer monitoring pods. Alternative to it would be to either create a throwaway pod to connect to customer grafana or, install curl on customer grafana which was not possible due to pod security blocking it.

KevFan: Ah I see, makes sense now :+1:

KevFan:
> @KevFan Can you confirm this test was executed locally during verification? The ci/prow/rhoam/e2e does not currently execute the e2e tests so is not a valid verification of this PR and this test also does not run in the RHMI e2e tests.

@philbrookes Yeah I've ran this test (E09 and E10) locally for RHOAM installation and both passed

PR #1374, opened 2020-11-13
Isolate IDP tests

CathalOConnorRH:
# Description
Isolate IDP tests so they can be run nightly rather than against each PR


test/common/tests.go
```diff
@@ -77,6 +67,29 @@ var (
 		},
 	}
 
+	IDP_BASED_TESTS = []TestSuite{
```
psturc, line 70: should we update the [functional tests runner](https://github.com/integr8ly/integreatly-operator/blob/1023e85c9ecdfdc975a94b5175d567aaee70d8af/test/functional/integreatly_test.go#L13) to include these IDP based tests?

CathalOConnorRH: I've added it now. Thanks

PR #1377, opened 2020-11-13
MGDAPI-564 RHOAM/RHMI Installation/upgrade/reconciler alerts

MStokluska:
# Description
Addresses https://issues.redhat.com/browse/MGDAPI-564.
Currently, when RHOAM is installed and/or running the following alerts could trigger:
```
RHMIInstallationControllerIsNotReconciling
RHMIInstallationControllerStoppedReconciling
```
and the following alerts triggers and don't go away after installation completes:
```
RHMIOperatorInstallDelayed
RHMIUpgradeExpectedDurationExceeded
```
The reason for it is because the alerts expect rhmi_status and rhmi_version present, while we are having rhoam installed. 

# What
Both `RHMIOperatorInstallDelayed` and `RHOAMUpgradeExpectedDurationExceeded` fired during installation and remained in firing state after installation reports completed. The reason for it is because we were checking RHMI installation stage status instead of RHOAM.

# How
Added metrics for RHOAM status and edited the alerts to be flexible so that alerts names and queries are built based on the install type

# Verification
Both RHMI and RHOAM must be verified.

## RHOAM

- install RHOAM from this branch via OLM ( you can follow steps in https://github.com/RHCloudServices/integreatly-help/blob/master/guides/olm/installing-rhmi-bundle-format.md )
- during installation following alert should be pending:
```
RHOAMOperatorInstallDelayed
```
- and the following alerts will fire:
```
RHOAMUpgradeExpectedDurationExceeded
```
The reason for the RHOAMUpgradeExpectedDurationExceeded alert firing is that this alert aims to monitor upgrade time, so the `version` field during initial installation is not present, and thus, the alert fires. 
- when the installation completes, confirm that none of the following alerts are triggered:
```
1. RHOAMOperatorInstallDelayed
2. RHOAMInstallationControllerIsNotReconciling
3. RHOAMInstallationControllerStoppedReconciling
4. RHOAMUpgradeExpectedDurationExceeded
```

## RHMI

- install RHMI from this branch via OLM ( you can follow steps in https://github.com/RHCloudServices/integreatly-help/blob/master/guides/olm/installing-rhmi-bundle-format.md )
- during installation following alert should be pending:
```
RHMIOperatorInstallDelayed
```
- and the following alerts will fire:
```
RHMIUpgradeExpectedDurationExceeded
```
- when the installation completes, confirm that none of the following alerts are triggered:
```
1. RHMIOperatorInstallDelayed
2. RHMIInstallationControllerIsNotReconciling
3. RHMIInstallationControllerStoppedReconciling
4. RHMIUpgradeExpectedDurationExceeded
```

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -80,6 +80,16 @@ var (
 			"to_version",
 		},
 	)
+
+	RHOAMStatus = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "rhoam_status",
+			Help: "RHOAM status of an installation",
+		},
+		[]string{
+			"stage",
+		},
+	)
 )
 
 // SetRHMIInfo exposes rhmi info metrics with labels from the installation CR
@@ -101,6 +111,11 @@ func SetRHMIStatus(installation *integreatlyv1alpha1.RHMI) {
 	RHMIStatus.Reset()
 	if string(installation.Status.Stage) != "" {
 		RHMIStatus.With(prometheus.Labels{"stage": string(installation.Status.Stage)}).Set(float64(1))
+	}
+
+	RHOAMStatus.Reset()
+	if string(installation.Status.Stage) != "" {
+		RHOAMStatus.With(prometheus.Labels{"stage": string(installation.Status.Stage)}).Set(float64(1))
 	}
 }
 
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -65,6 +65,7 @@ func init() {
 	customMetrics.Registry.MustRegister(integreatlymetrics.RHMIVersion)
 	customMetrics.Registry.MustRegister(integreatlymetrics.RHMIStatus)
 	customMetrics.Registry.MustRegister(integreatlymetrics.RHOAMVersion)
+	customMetrics.Registry.MustRegister(integreatlymetrics.RHOAMStatus)
 	integreatlymetrics.OperatorVersion.Add(1)
 }
 
--- a/pkg/controller/installation/prometheusRules.go
+++ b/pkg/controller/installation/prometheusRules.go
@@ -2,6 +2,7 @@ package installation
 
 import (
 	"fmt"
+	"strings"
 
 	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/sirupsen/logrus"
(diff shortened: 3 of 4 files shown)


laurafitzgerald:
👀

laurafitzgerald:
Hey @MStokluska I think that we maybe don't check these alerts in the `alertsexist` test but if we plan to do that I think we need to update this value https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L202 This get's used in the RHOAM e2e tests so will set the name of the object which you are using to build the alert name. I think probably it's not failing because we are not checking the alert is in place. Alternatively I guess we could use the install name in the test also. If we don't need to check that then it's maybe not an issue.

pkg/metrics/metrics.go
```diff
@@ -80,6 +80,16 @@ var (
 			"to_version",
 		},
 	)
+
+	RHOAMStatus = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "rhoam_status",
```
laurafitzgerald, line 86: we're using the name of the cr to reference this query in the alerts. if the name of the cr changes for any reason, this would break. if we expect that not to change we are good but just wanted to call it out.

MStokluska:
> Hey @MStokluska I think that we maybe don't check these alerts in the `alerts_exist` test but if we plan to do that I think we need to update this value https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L202 This get's used in the RHOAM e2e tests so will set the name of the object which you are using to build the alert name. I think probably it's not failing because we are not checking the alert is in place. Alternatively I guess we could use the install name in the test also. If we don't need to check that then it's maybe not an issue.

Thanks @laurafitzgerald -I've updated the alerts_exists.go with 2 alerts, other 2 are under openshift ns and we are not checking for them in RHMI too so I am going to leave them out too.

pkg/metrics/metrics.go
MStokluska: Thanks, as agreed during our call, we don't need to worry about it

laurafitzgerald:
Steps followed so far:
build image from branch by updating org in Makefile to your own quay org
`INSTALLATION_TYPE=managed-api make image/build/push`
install via add on ocm cluster
update the image tags in the rhoam clusterserviceversion as soon as it's created.

verify the pod pullls your image.
wait for the install and continue with the verification.

jjaferson:
/lgtm

Verified the changes on an OSD cluster and the alerts worked as expected.

PR #1380, opened 2020-11-15
webhook selector is always rhmi-operator

philbrookes:
# Description
Fixing a few more namespace prefix bugs


philbrookes:
/override ci/prow/e2e

failed on a flake

PR #1387, opened 2020-11-16
add default cidr value

philbrookes:
# Description
add a default CIDR range value

### DIFF
--- a/pkg/products/cloudresources/reconciler.go
+++ b/pkg/products/cloudresources/reconciler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -357,12 +358,10 @@ func (r *Reconciler) reconcileCIDRValue(ctx context.Context, client k8sclient.Cl
 		return err
 	}
 
-	if cidrValue == "" {
+	//!ok means the param wasn't found so we want to default rather than return
+	//but don't do it until the installation object is more than a minute old in case the secret is slow to create
+	if !ok || cidrValue == "" && r.installation.ObjectMeta.CreationTimestamp.Time.Before(time.Now().Add(-(1*time.Minute))) {
 		cidrValue = "10.1.0.0/16"
-	}
-
-	if !ok {
-		return nil
 	}
 
 	cfgMap := &corev1.ConfigMap{}
--- a/pkg/products/cloudresources/reconciler.go
+++ b/pkg/products/cloudresources/reconciler.go
@@ -357,6 +357,10 @@ func (r *Reconciler) reconcileCIDRValue(ctx context.Context, client k8sclient.Cl
 		return err
 	}
 
+	if cidrValue == "" {
+		cidrValue = "10.1.0.0/16"
+	}
+
 	if !ok {
 		return nil
 	}


laurafitzgerald:
Steps carried out to verify this change.
Built image from branch to `quay.io/laurafitzgerald/managed-api-service:v0.8.0-cidr`
Deployed via olm without passing cidr block
Patched image in csv to the above.
Screenshot 2020-11-16 at 20 53 41
Unfortunately the configmap did not get patched with the expected value 
And the operator faced the same blocking await message.

laurafitzgerald:
After looking into it a little further, the function where the default is being added is not hit when there is no `cidr` label added to the namespace. Is it possible to add a default value of `""` to the cidr add on param so that the label is created with an emtpy value so that it triggers the function and add the default value? Or alternatively provide a default via the addon param if the user doesn't enter one in the wizard?

pkg/products/cloudresources/reconciler.go
```diff
@@ -357,6 +357,10 @@ func (r *Reconciler) reconcileCIDRValue(ctx context.Context, client k8sclient.Cl
 		return err
 	}
 
+	if cidrValue == "" {
+		cidrValue = "10.1.0.0/16"
+	}
+
 	if !ok {
 		return nil
 	}
```
laurafitzgerald, line 366:
```suggestion
		//!ok means the param wasn't found so we want to default rather than return
	if !ok || cidrValue == "" {
		cidrValue = "10.1.0.0/16"
	}
```

pkg/products/cloudresources/reconciler.go
```diff
@@ -357,8 +357,9 @@ func (r *Reconciler) reconcileCIDRValue(ctx context.Context, client k8sclient.Cl
 		return err
 	}
 
-	if !ok {
-		return nil
+	//!ok means the param wasn't found so we want to default rather than return
+	if !ok || cidrValue == "" {
```
maleck13, line 361: is there any possibility the secret has not arrived to the cluster yet? I can't recall what Mike said, is the secret always created? Wondering do we want to allow a little time for the secret to get there from hive. Seems like it should be there as it will be created at the same time as the label to apply the SSS but just wondering if there is a chance we set the default then the secret comes in

laurafitzgerald: We should check with Mike. As soon as the `_network` entry is added cro will create the vpc and progress with creating the resources and it cannot be changed after that. I'll ping him in the chat.

pkg/products/cloudresources/reconciler.go
```diff
@@ -357,8 +357,9 @@ func (r *Reconciler) reconcileCIDRValue(ctx context.Context, client k8sclient.Cl
...
+		cidrValue = "10.1.0.0/16"
```
ciaranRoche, line 362: Defaulting to a static range could run into issues, as this CIDR can not overlap with Cluster CIDR. There is logic in CRO to get the Cluster CIDR, we should set the default there or export the logic so we can insure there is no overlap and the CIDR is valid.

laurafitzgerald: This could happen in the case of a customer provided CIDR also though right?

ciaranRoche: yep, that could also happen, the validation on the form is only regex and ensures it is a valid CIDR. There is no other validation provided via CS.

laurafitzgerald: CRO will report back if the CIDR range is not suitable so I think we're covered there. We just need to find a way to provide some sensible default via rhoam.

laurafitzgerald:
👀

laurafitzgerald:
Built and deployed this branch using the following steps.

- Install RHOAM via add on in ocm.
- Didn't pass an entry for the CIDR value
- Patch the csv to use the new image - doing it manually but working on a command. Something like the below but it's not working yet. think I'm doing something wrong with the patch command. the json seems fine.
```
 oc patch clusterserviceversion managed-api-service.v0.8.0 --type json -p '{"spec": {"install": { "spec": { "deployments": [{"name": "rhmi-operator", "spec": {"template": {"spec": {"containers": [{"name": "rhmi-operator", "image": "quay.io/laurafitzgerald/managed-api-service:v0.8.0-cidr"}]}}}}]}}}}'
```

- If you've already installed an add on you'll need to add the dms secret
`INSTALLATION_TYPE=managed-api make cluster/prepare/dms`

- Patch the cr to useClusterStorage false 
- Wait a minute or so and check the `cloud-resources-aws-strategies` config map
` oc patch rhmi rhoam -n redhat-rhoam-operator --type merge -p '{"spec":{"useClusterStorage": "false"}}'  `
Screenshot 2020-11-17 at 21 32 18


Followed the above steps again but passing a cidr value 10.10.0.0/16 and it was added correctly to the _network 
Screenshot 2020-11-17 at 21 51 39

philbrookes:
/override ci/prow/e2e 
manually verified by Laura

philbrookes:
/override ci/prow/rhoam-e2e
manually verified by Laura

PR #1388, opened 2020-11-17
Add alert for 3scale ApiCast worker process restart

jjaferson:
# Description

Add an alert to detect when 3scale apicast worker restarts

Jira: https://issues.redhat.com/browse/MGDAPI-322

PS.: this PR is in WIP because the existing SOP to troubleshoot the alert but it needs to be improved with more details on how to fix the issue in case the alert start firing and be approved by SRE. 

## Verification steps

1) Install the operator from this branch
2) Check if the alert is created in prometheus
3) Go to one of the production APICast pod and run the command below to restart nginx worker
```for k in /proc/[0-9]*; do grep -i "^nginx: worker process" $k/cmdline 2>&1 >/dev/null; if [[ $? == 0 ]]; then echo ${k##*/}; fi; done | xargs kill -9```
4) Verify if the alert is in a pending state

### DIFF
--- a/pkg/products/threescale/prometheusRules.go
+++ b/pkg/products/threescale/prometheusRules.go
@@ -275,6 +275,19 @@ func (r *Reconciler) newAlertReconciler() resources.AlertReconciler {
 						For:    "15m",
 						Labels: map[string]string{"severity": "warning"},
 					},
+					{
+						Alert: "ThreescaleApicastWorkerRestart",
+						Annotations: map[string]string{
+							"summary":     "A new worker process in Nginx has been started",
+							"description": "A new thread has been started. This could indicate that a worker process has died due to the memory limits being exceeded. Please investigate the memory pressure on pod (instance {{ $labels.instance }})",
+							"sop_url":     "https://github.com/3scale/3scale-Operations/blob/master/sops/alerts/apicast_worker_restart.adoc",
+						},
+						Expr: intstr.FromString(fmt.Sprintf(`changes(worker_process{kubernetes_namespace='%s', kubernetes_pod_name=~'apicast-production.*'}[5m]) > 0`, r.Config.GetNamespace())),
+						For:  "5m",
+						Labels: map[string]string{
+							"severity": "critical",
+						},
+					},
 				},
 			},
 		},
--- a/test/common/alerts_exist.go
+++ b/test/common/alerts_exist.go
@@ -400,6 +400,7 @@ func commonExpectedRules() []alertsTestRule {
 				"ThreeScaleContainerHighCPU",
 				"ThreeScaleZyncPodAvailability",
 				"ThreeScaleZyncDatabasePodAvailability",
+				"ThreescaleApicastWorkerRestart",
 			},
 		},
 		{

### CODE REVIEW

File: pkg/products/threescale/prometheusRules.go [hunk 1]
```diff
@@ -275,6 +275,19 @@ func (r *Reconciler) newAlertReconciler() resources.AlertReconciler {
 						For:    "15m",
 						Labels: map[string]string{"severity": "warning"},
 					},
+					{
+						Alert: "ThreescaleApicastWorkerRestart",
+						Annotations: map[string]string{
+							"summary":     "A new worker process in Nginx has been started",
```
Nanyte25, line 281: @pepedocs  and @jjaferson  I notice the annotation for the alert mentions this `Annotations: map[string]string{
							"summary":     "A new worker process in Nginx has been started",` should this not review and describe the particular component in RHMI or RHOAM?

The link SOP would need improving also @jjaferson  for example: **https://github.com/3scale/3scale-Operations/blob/master/sops/alerts/apicast_worker_restart.adoc**

```
Pod hits the memory limits and one of the worker processes was killed. Check why the limits were reached, and if it’s needed, increase them

Node is out of memory, so try to kill containers to schedule to another node

```
We should have an example of the pod hitting it memory limit and what that would look like. What action can we take to patch, scale the update the pods memory via request limits or budget disruption budget.

`oc events` how to up the memory of the pod using a `oc patch` command etc.

What action can an CS-SRE perform on this component for this managed service?

pkg/products/threescale/prometheusRules.go
```diff
@@ -275,6 +275,19 @@ func (r *Reconciler) newAlertReconciler() resources.AlertReconciler {
...
+							"description": "A new thread has been started. This could indicate that a worker process has died due to the memory limits being exceeded. Please investigate the memory pressure on pod (instance {{ $labels.instance }})",
+							"sop_url":     "https://github.com/3scale/3scale-Operations/blob/master/sops/alerts/apicast_worker_restart.adoc",
+						},
+						Expr: intstr.FromString(fmt.Sprintf(`changes(worker_process{kubernetes_namespace='%s', kubernetes_pod_name=~'apicast-production.*'}[5m]) > 0`, r.Config.GetNamespace())),
+						For:  "5m",
+						Labels: map[string]string{
+							"severity": "warning",
```
pmccarthy, line 288: @jjaferson confirmed with CS-SRE earlier today that this alert severity can be bumped to `Critical`

pkg/products/threescale/prometheusRules.go
pmccarthy: @jjaferson @Nanyte25 as discussed on the sync call with @brizrobbo, CS-SRE will take an action to work with 3Scale to improve the content of the SOP itself. That work doesn't block this PR from being merged as the SOP URL won't change.

pkg/products/threescale/prometheusRules.go
```diff
@@ -275,6 +275,19 @@ func (r *Reconciler) newAlertReconciler() resources.AlertReconciler {
 						For:    "15m",
 						Labels: map[string]string{"severity": "warning"},
 					},
+					{
+						Alert: "ThreescaleApicastWorkerRestart",
+						Annotations: map[string]string{
+							"summary":     "A new worker process in Nginx has been started",
+							"description": "A new thread has been started. This could indicate that a worker process has died due to the memory limits being exceeded. Please investigate the memory pressure on pod (instance {{ $labels.instance }})",
```
pepedocs, line 282: Not a biggie, just that we can be more consistent with using the word "process" instead of "thread" here.

pepedocs:
Looks good to me assuming the SOP is not a blocker and alert set to critical.

Nanyte25:
Looks good to me now, I think there is a higherlevel conversation need around how we have RHOAM SOP in three-scale repo verus keep the base SOP generate for three-scale SAAS to be had. 

/lgtm

pkg/products/threescale/prometheusRules.go
jjaferson: @pepedocs this alert is a copy from an alert created in the 3scale operator https://github.com/3scale/3scale-operator/pull/484/files 

When we upgrade the 3scale operator to the version that hasthis alert, we'll remove the alert created by RHOAM.

pkg/products/threescale/prometheusRules.go
Nanyte25: Sounds good to me @pmccarthy

PR #1389, opened 2020-11-17
MGDAPI-565: Modify API Usage queries for statsd pod restart support

sergioifg94:
# Description

> Link to Jira: https://issues.redhat.com/browse/MGDAPI-565

The metrics exposed by the rate limiting service include in their labels information about the statsd Pod such as it's host and Pod name. When the statsd Pod is restarted, a different vector is exposed with the counter restarted to 0, which affects the Grafana dashboard as well as the alerts that rely on these metrics.
Wrap the `increase()` function calls to `sum()`. When the metric is not exposed anymore the value of `increase()` will remain and, by using `sum()`, the values of the different vectors will be added up.

## Verification steps

This PR modifies the promql queries for the Grafana dashboard, the API Usage alerts, and the soft limit alerts. In order to verify this, verify that the alerts still work and the Grafana dashboard works after restarting the statsd Pod

1. ### Restart the statsd Pod
    1. Create a 3scale application.
    2. Produce some traffic
    3. In Prometheus, check that there is some data present for the query
        ```python
        increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits[24h])
        ```
    4. Restart the prom-stats-d-exporter pod in the redhat-rhoam-marin3r namespace. 
    5. Scale down the rate-limit Deployment, scale it up again and wait a minute for it to be deployed
2. ### Verify API Usage alerts
    Follow the verification steps from https://github.com/integr8ly/integreatly-operator/pull/1319
    > The queries use the `increase()` function now. Verify that they work the same with the following format
    > ```python
    > floor(sum(increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits[24h]))) > <soft limit>
    > ```

3. ### Verify soft limits alerts
    Follow the steps in the manual test case to verify it still works:
    https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/alerts/c18-validate-api-usage-alerts.md#steps
    
4. ### Verify Grafana dashboard
    Navigate to the Grafana dashboard and verify that the figures make sense and no error is shown.
    > A value of `"NaN%"` for "Last 1 Minute Rejected/Requests" is acceptable, as if there's no requests it's a division by 0
## Type of change

## Checklist

### DIFF
--- a/pkg/products/grafana/customerDashboard.go
+++ b/pkg/products/grafana/customerDashboard.go
@@ -102,7 +102,7 @@ const CustomerMonitoringGrafanaRateLimitingJSON = `{
       "tableColumn": "",
       "targets": [
         {
-          "expr": "increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits[1m])",
+          "expr": "sum(increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits[1m]))",
           "instant": true,
           "refId": "A"
         }
@@ -188,7 +188,7 @@ const CustomerMonitoringGrafanaRateLimitingJSON = `{
       "tableColumn": "",
       "targets": [
         {
-          "expr": "increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_over_limit[1m])",
+          "expr": "sum(increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_over_limit[1m]))",
           "instant": true,
           "refId": "A"
         }
@@ -273,7 +273,7 @@ const CustomerMonitoringGrafanaRateLimitingJSON = `{
       "tableColumn": "",
       "targets": [
         {
-          "expr": "(increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_over_limit[1m])/increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits[1m]))*100",
+          "expr": "sum(increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_over_limit[1m]))/sum(increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits[1m]))*100",
           "instant": true,
           "refId": "A"
         }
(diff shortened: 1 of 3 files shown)


pmccarthy:
@sergioifg94 While following the steps in the "Verify API Usage alerts" section above I seem to be getting a slightly higher count with the updated query. Is this to be expected? See screenshot below:

Screenshot 2020-11-17 at 19 32 52

pmccarthy:
@sergioifg94 Also noticing strange behaviour in the Grafana dashboard where `No. of Requests` is showing twice? The "Reqest" typo is being fixed in a separate PR so we can ignore that :)

Screenshot 2020-11-17 at 19 39 27

sergioifg94:
@pmccarthy It would only make sense if those 2 requests of difference are the requests that were made before restarting the statsd Pod. In that case it would prove that the new query is better, as the count isn't lost despite the restart

sergioifg94:
@pmccarthy I'm not sure what's the cause of the No of Requests being showed twice. The only change made on Grafana was to the queries, so I'd say this was already present 🤔

pmccarthy:
> @pmccarthy I'm not sure what's the cause of the No of Requests being showed twice. The only change made on Grafana was to the queries, so I'd say this was already present 🤔

Any thoughts on this one @laurafitzgerald ?

laurafitzgerald:
The dashboard also needs the update. 
It maps similarly to the prometheus graphs.
So in prometheus you see two different colours until you add the `sum()` which you've done on the others.
I've tested adding that and it aggregates them into one amount. 

Existing (see the two entries)
Screenshot 2020-11-17 at 20 06 53

Updating the query
Screenshot 2020-11-17 at 20 06 38


@sergioifg94 would you mind also updating that query here?

sergioifg94:
Thanks @laurafitzgerald I had missed a couple queries that weren't wrapped by `sum()` indeed. I just updated the PR to include it

pmccarthy:
Validated latest changes based on outline steps and confirmed the following:

- API usage alerts are still firing as expected
- Grafana Dashboard is showing relevant count based on usage
- `floor` query is displaying a count which equals the `sum` of all vectors counts returned in:

```
increase(ratelimit_service_rate_limit_apicast_ratelimit_generic_key_slowpath_total_hits[24h])
```

/lgtm
/approve

PR #1392, opened 2020-11-17
skip flaky tests

philbrookes:
# Description
Skip flaky tests


test/common/authdelay_first_broker_login.go
```diff
@@ -32,7 +32,7 @@ const (
 )
 
 func TestAuthDelayFirstBrokerLogin(t *testing.T, ctx *TestingContext) {
-
+	t.SkipNow()
```
MStokluska, line 35:
```suggestion
        t.Log("skipping due to flaky behaviour")
	t.SkipNow()
```

PR #1393, opened 2020-11-17
Bump AMO to v1.4.0 and added pod anti-affinity rules

obrienrobert:
## Description
[MGDAPI-410](https://issues.redhat.com/browse/MGDAPI-410)
- Bumps AMO to `v1.4.0`
- Implements Anti-Affinity within the AMO spec for Prometheus and Alertmanager pods
- Updates the number of expected pods in the `MiddlewareMonitoringPodCount` alert depending on if the cluster if multi-AZ or not

#### Single AZ cluster
- Install via this branch
- Ensure the the correct AMO version is being used (`v.1.4.0`)
- Within the `middleware-monitoring` namespace, ensure both Prometheus and Alertmanager have 1 pod each
- Login to Prometheus and ensure that the `MiddlewareMonitoringPodCount` alert expression expects 7 pods, and the alert isn't firing

#### Multi-AZ cluster
- Install via this branch
- Ensure the the correct AMO version is being used (`v.1.4.0`)
- Within the `middleware-monitoring` namespace, ensure both Prometheus and Alertmanager have 3 pods each
- Login to Prometheus and ensure that the `MiddlewareMonitoringPodCount` alert expression expects 11 pods, and the alert isn't firing
- Ensure the the Prometheus and Alertmanager pods are spread across different AZ's

## Type of change

## Checklist


pkg/products/monitoring/reconciler.go
```diff
@@ -569,6 +574,14 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, serverClient k8scl
 			PrometheusInstanceNamespaces:     r.Config.GetOperatorNamespace(),
 			SelfSignedCerts:                  r.installation.Spec.SelfSignedCerts,
 		}
+
+		if isClusterMultiAZ {
+			m.Spec.Affinity = resources.SelectAntiAffinityForCluster(isClusterMultiAZ, map[string]string{
+				"prometheus":   "application-monitoring",
+				"alertmanager": "application-monitoring",
+			})
+		}
```
philbrookes, line 583: Could you add an:
```
} else {
    m.Spec.Affinity = nil
}
```

pkg/products/monitoring/reconciler.go
```diff
@@ -203,6 +203,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
+	isMultiAZCluster, err := resources.IsMultiAZCluster(ctx, serverClient)
```
philbrookes, line 206: Did you test that this IsMultiAzCluster works then? Or did you check with @sergioifg94 ?

pkg/products/monitoring/reconciler.go
```diff
@@ -569,6 +574,16 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, serverClient k8scl
 			PrometheusInstanceNamespaces:     r.Config.GetOperatorNamespace(),
 			SelfSignedCerts:                  r.installation.Spec.SelfSignedCerts,
 		}
+
+		if isMultiAZCluster {
+			m.Spec.Affinity = resources.SelectAntiAffinityForCluster(isMultiAZCluster, map[string]string{
```
sergioifg94, line 579:
```suggestion
			m.Spec.Affinity = resources.SelectAntiAffinityForCluster(antiAffinityRequired, map[string]string{
```

We still need to use the `antiAffinityRequired` value (from the env var) for this parameter. Otherwise the affinity rules will be set to required for every Multi AZ cluster

obrienrobert: I've made the update as per your suggestion. Thanks Sergio.

pkg/products/monitoring/reconciler.go
obrienrobert: I've tested this on both a single and multi-AZ cluster.

sergioifg94:
The code changes look good and I verified on both single and mutli AZ clusters
/lgtm

PR #1425, opened 2020-11-23
Test cases updates

psturc:
# Description
* simple namespace prefix fix: https://github.com/integr8ly/integreatly-operator/pull/1425/commits/1a0fdc3b653258401d7d08fea67fe2d91f313107
* update p01 test case: add steps to verify pod redistribution after AZ recovery: https://github.com/integr8ly/integreatly-operator/pull/1425/commits/fa96c0854443fedffb2627d372a7a8f17ab8a1de

### DIFF
--- a/test-cases/tests/multiAZ/p01-measure-downtime-during-down-az.md
+++ b/test-cases/tests/multiAZ/p01-measure-downtime-during-down-az.md
@@ -175,6 +175,21 @@ Measure the downtime of the RHOAM components during a AWS Availability Zone fail
     # e.g.
     ./scripts/disableAz.sh false eu-west-1a
     ```
-
-16. Consult the results with engineering (especially in case some components have a longer downtime than 30min
+16. Wait until the AZ is restored and OpenShift starts using it
+  ```bash
+  # You should see a similar output to the one below when running the following oc command:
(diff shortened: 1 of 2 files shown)


test-cases/tests/multiAZ/p01-measure-downtime-during-down-az.md
```diff
@@ -176,5 +176,25 @@ Measure the downtime of the RHOAM components during a AWS Availability Zone fail
...
+mw-collab-multi-45qpn-worker-eu-west-1c   3         3         3       3           94m
+```
+
+17. Run the automated test for checking the correct pod distribution across all AZs and make sure it passes (it might take a while until all pods are correctly redistributed)
+
+```bash
+$ go clean -testcache && MULTIAZ=true WATCH_NAMESPACE=redhat-rhoam-operator go test -v ./test/functional -run="//^F09"
+```
+
+18. Consult the results with engineering (especially in case some components have a longer downtime than 30min
```
obrienrobert, line 199: Do we need to consult the results with engineering if the tests pass successfully?

psturc: Good point, we only need to consult the measured downtimes of components.

I updated the test case now

PR #1429, opened 2020-11-23
e2e failing tests fix

MStokluska:
This is a temporary fix for F05 and F08 to check only if the minimum required amount of replicas is present following approach from A13 test.


MStokluska:
closing as it seems that the issue has been resolved in another PRs.

PR #1441, opened 2020-11-25
MGDAPI-373: Add alert for rate of rejected requests

sergioifg94:
# Description

> Link to Jira: https://issues.redhat.com/browse/MGDAPI-373

Create an alert that will fire when the rate of rejected requests doesn't match the expected rate based on the configuration

Example: The rate limit service is configured with a rate limit of `200 req/minute`, and the API is receiving `240 req/minute`. We'd expect ~16% of the requests to be rate limited. If this doesn't match, the alert will be triggered

## Verification steps

1. Install RHOAM
1. Create a 3scale application
2. Set the rate limit configuration to 200 requests per minute
3. Wait for the configuration to be propagated
4. Verify that the alert was created
    ```sh
    oc get prometheusrules rejected-requests -n redhat-rhoam-marin3r
    ```
5. Run the following script to force requests to be rate limited, and verify that the alert doesn't fire
    > It will perform requests for 5 minutes at 240 requests per minute
    ```sh
    for i in {1..1200}; do
        curl -i "<3scale API URL>" &
        sleep 0.25
    done
    ```

## Type of change

## Checklist

### DIFF
--- a/pkg/products/marin3r/config/unit_conversion.go
+++ b/pkg/products/marin3r/config/unit_conversion.go
@@ -1,0 +1,73 @@
+package config
+
+import "fmt"
+
+var (
+	Second = "second"
+	Minute = "minute"
+	Hour   = "hour"
+	Day    = "day"
+)
+
+type unitConversion func(int) float64
+
+func divideBy(factor int) unitConversion {
+	return func(n int) float64 {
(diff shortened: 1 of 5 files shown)


jjaferson:
/lgtm 

Verified on an OSD cluster

PR #1445, opened 2020-11-25
INTLY-10307 ThreeScaleSystemAdminUIBBT alert update

CathalOConnorRH:
# Description
This alert fired in production while 3scale admin ui was still responding correctly. Route cause is that the blackbox exporter didnt scrap metrics for a period of time. This update will check for when the exporter returns 0 rather than nothing

# Validation
To validate 1. remove the url from the blackbox target and the alert should not fire. Once confirmed. replace the url and move on to step 2.
2. stop 3scale operator and scale down system-app replicas until the alert fires.


sergioifg94:
Verified it on cluster
/lgtm

laurafitzgerald:
@CathalOConnorRH @philbrookes Should we be addressing the 3 BBT Alerts, I think they could all suffer from the same issue?

philbrookes:
As Cathal is on the success team currently, let's merge this and make a Jira to fix the others in the backlog.

PR #1463, opened 2020-11-30
Reconcile webhooks before preflight checks

sergioifg94:
# Description

Webhooks were only being reconciled after the preflight checks had passed. A webhook was set up to delete the operator
Subscription when the RHMI CR is deleted, however; in the chance that the CR is deleted before the preflight checks have passed (example: the `useClusterStorage` field hasn't been set), the webhook won't be in place and the Subscription won't be deleted.
This change ensures that the webhooks are created before the preflight checks

## Verification steps

In order to verify this, install the add-on and use an image from this PR, then delete the add-on and check that the subscription is deleted

1. Create an image from this PR
2. Install the add-on
3. Update the operator CSV to reference the custom image
4. Wait for the operator Pod to be restarted
5. Uninstall the add-on
6. Verify that the RHMI CR is deleted and the Subscription is deleted as well

## Type of change

## Checklist

### DIFF
--- a/pkg/controller/installation/installation_controller.go
+++ b/pkg/controller/installation/installation_controller.go
@@ -341,6 +341,11 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, err
 	}
 
+	// Reconcile the webhooks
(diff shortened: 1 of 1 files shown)


jjaferson:
/lgtm 

Verified, the RHMI CR and subscription are deleted

PR #1465, opened 2020-11-30
replace in_prow with scale_down for use in local dev

philbrookes:
# Description
replace `in_prow` with `scale_down` so it can be used more widely without being confusing.

Allow value to come from an env var to facilitate local development.


R-Lawton:
/lgtm 
verified on a osd cluster that 1 replica gets created for rhsso, rhssouser a 3scale also verified that the tests pass locally.

PR #1470, opened 2020-12-01
Updates grafana image from dockerhub to quay

jjaferson:
# Description

One of the requirements for RHOAM GA is that all images used in the product should be hosted in quay, so this PR updates the granafa image from dockerhub to quay

https://issues.redhat.com/browse/MGDAPI-809

## Verification steps

1) install the operator from this branch
2) check if the granafa image used in AMO and `redhat-rhoam-customer-monitoring-operator` namespaces use the image from quay `quay.io/integreatly/grafana:7.1.1`

PR #1474, opened 2020-12-02
E02 and e03 flaky tests

philbrookes:
# Description
e02 and e03 have become unstable tests, apparently caused by this PR: https://github.com/integr8ly/integreatly-operator/pull/1470 marking them as flaky until we can address https://issues.redhat.com/browse/MGDAPI-815

Also slipping in a small change to the functional Dockerfile that was noticed in the rc3 release.


jjaferson:
/lgtm 

Happy for these tests to be skipped for now, the dashboard are being created in Grafana as expected.

PR #1478, opened 2020-12-03
Increases ThreeScaleZyncPodAvailability alert firing time

jjaferson:
# Description

With the increase of replicas for [3scale zync pods to 3](https://github.com/integr8ly/integreatly-operator/blob/master/pkg/config/threescale.go#L128), the time for all the pods to become ready has increased which is making the alert `ThreeScaleZyncPodAvailability` fire. We need to extend the length of time for this alert to avoid it from fire prior the installation is completed

https://issues.redhat.com/browse/MGDAPI-913

## Verification steps

1) install the operator from this branch in a multi AZ cluster
2) verify that `ThreeScaleZyncPodAvailability` is not firing during the installation of 3scale.

### DIFF
--- a/pkg/products/threescale/prometheusRules.go
+++ b/pkg/products/threescale/prometheusRules.go
@@ -239,10 +239,10 @@ func (r *Reconciler) newAlertReconciler() resources.AlertReconciler {
 						Alert: "ThreeScaleZyncPodAvailability",
 						Annotations: map[string]string{
 							"sop_url": resources.SopUrlAlertsAndTroubleshooting,
-							"message": "3Scale Zync has {{   $value  }} pods in a ready state. Expected a minimum of 2 pods.",
+							"message": fmt.Sprintf("3Scale Zync has {{   $value  }} pods in a ready state. Expected %d of pods.", r.Config.GetReplicasConfig(r.installation)["zyncApp"]),
 						},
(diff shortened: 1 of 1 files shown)


MStokluska:
hey @jjaferson , I've verified it on the cluster and it works as expected, however, I have 2 questions and wonder if you can help me out here:
1. Looking at the query - we are checking if the number of zync pods != 3  ( or whatever is currently set in threescale.go ). Do you think it is ever possible that zync pods will scale up - to 4/5 for a customer and then, this alert will trigger? I see this is not a change done by you but by whoever wrote the initial query. If so, would it be better to change the `!=` to `<` ? Please correct me if I'm wrong on this!
2. So the alert did not trigger - however, it was at 8minute mark where it resolved from warning to normal, and I wonder if we could maybe increase it a bit more? Do you think we are safe with just 2 mins extra?

jjaferson:
> hey @jjaferson , I've verified it on the cluster and it works as expected, however, I have 2 questions and wonder if you can help me out here:
> 
> 1. Looking at the query - we are checking if the number of zync pods != 3  ( or whatever is currently set in threescale.go ). Do you think it is ever possible that zync pods will scale up - to 4/5 for a customer and then, this alert will trigger? I see this is not a change done by you but by whoever wrote the initial query. If so, would it be better to change the `!=` to `<` ? Please correct me if I'm wrong on this! If there is not a chance that there will be more zync pods present - if it cannot be less or more, I would update the message from `Expected a minimum of <AMOUNT>` to something like `Expected <AMOUNT> of pods`
> 2. So the alert did not trigger - however, it was at 8minute mark where it resolved from pending to normal, and I wonder if we could maybe increase it a bit more? Do you think we are safe with just 2 mins extra?

Hey @MStokluska, 

1) the number zync pods will change according to the number configuration in threescale.go https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/threescale/prometheusRules.go#L244 

2) Good point, I will increase the time to 15m

PR #1479, opened 2020-12-04
Add pod distribution check and balance

briangallagher:
## Description
Add pod distribution check and balance. On multiAZ clusters, the rhoam operator should reconcile pod distribution on deployments (and RS, SS) with > 1 replica count.  First check if the distribution meets the requirements i.e. Minimum pod distribution states that a replica count > 1 on clusters with > 1 zone should not all be on the same zone. If the minimum pod distribution is not satisfied for a given pod restart <=50% of the replica count. Limit this check and redistribution attempt to 3. If it fails after 3 attempts there may be another cause for the failing distribution, for example, resource usage.

More details on [JIRA](https://issues.redhat.com/browse/MGDAPI-680): 

## Verify

- Clone this repo
- Modify the [replica numbers](https://github.com/integr8ly/integreatly-operator/blob/95d6b50918c6937f5ae9a1eb3ae39550acc9d51a/pkg/config/threescale.go#L120) for 3Scale to 2. This simply increases the chances of pods not fully distributing i.e. facilitate with testing
- Install RHOAM with `USE_CLUSTER_STORAGE = 'true'`
- Stop the operator
- Run `./scripts/disableAZ.sh true us-east-1c` to disable the us-east-1c zone. Change details if needed depending on where the cluster is installed. Disable a second zone, i.e. two out of three. 
- Wait for all pods to be recreated in the remaining zone. This can take some time and the console may go offline for a period. Once pods are recreated, re-enable the zones you disabled `./scripts/disableAZ.sh false us-east-1c` 
- Check the  `MultiAZPodDistribution` alert in prometheus, it should now be triggered. 
- Run the `MultiAZPodDistribution` query in prometheus to identify the pods not distributed
- Restart the rhoam operator. Pods not balanced in the below namespace should be restarted by the operator to attempt to force a rebalance. 
- Check the logs for indications of this. See below. 
- Verify ReplicaSet, DeploymentConfig, StatefulSet associated with the pods have a new annotation  ``` pod-balance-attempts: '1' ``` indicating the number of times a pod on the controller was restarted. 
- Most pods restarted will schedule on a different zone to its peer pods, thus alter the alert. This is not guaranteed.
- Verify that some, if not all Pods, from `MultiAZPodDistribution` have resolved. 

## Verify disabling rebalancing 
Setting env var `export REBALANCE_PODS="false"` will disable any attempt to rebalance. 
Restart the operator, after a full install, and verify the rebalance logic has not executed. The attempt to rebalance will happen almost immediately after a restart and when all products are fully installed. Simply check for non existence of log entry `"Reconciling Pod Balance in ns"`

### Namespaces
- redhat-rhoam-3scale
- redhat-rhoam-rhsso
- redhat-rhoam-userrhsso
- redhat-rhoam-marin3r

### Logs to grep:
- "Reconciling Pod Balance in ns"
- "Requires pod rebalance"
- "Attempting to delete pod"
- "Updating pod balance annotation"

### DIFF
--- a/pkg/controller/installation/installation_controller.go
+++ b/pkg/controller/installation/installation_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/poddistribution"
 	"os"
 	"reflect"
 	"strconv"
@@ -171,6 +172,7 @@ func createInstallationCR(ctx context.Context, serverClient k8sclient.Client) er
 	// Creates installation CR in case there is none
 	if len(installationList.Items) == 0 {
 		useClusterStorage, _ := os.LookupEnv("USE_CLUSTER_STORAGE")
+		rebalancePods := getRebalancePods()
 		cssreAlertingEmailAddress, _ := os.LookupEnv(alertingEmailAddressEnvName)
 		buAlertingEmailAddress, _ := os.LookupEnv(buAlertingEmailAddressEnvName)
 
@@ -209,6 +211,7 @@ func createInstallationCR(ctx context.Context, serverClient k8sclient.Client) er
 			Spec: integreatlyv1alpha1.RHMISpec{
 				Type:                 installType,
 				NamespacePrefix:      namespacePrefix,
+				RebalancePods:        rebalancePods,
 				SelfSignedCerts:      false,
 				SMTPSecret:           namespacePrefix + "smtp",
 				DeadMansSnitchSecret: namespacePrefix + "deadmanssnitch",
@@ -235,6 +238,13 @@ func createInstallationCR(ctx context.Context, serverClient k8sclient.Client) er
 	}
 
 	return nil
+}
+func getRebalancePods() bool {
+	rebalance, exists := os.LookupEnv("REBALANCE_PODS")
+	if !exists || rebalance == "true" {
+		return true
+	}
+	return false
 }
 
 func getCrName(installType string) string {
@@ -435,11 +445,29 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		installation.Status.Stage = integreatlyv1alpha1.StageName("complete")
 		metrics.RHMIStatusAvailable.Set(1)
 		retryRequeue.RequeueAfter = 5 * time.Minute
+		if installation.Spec.RebalancePods {
+			r.reconcilePodDistribution(installation)
+		}
 	}
 	metrics.SetRHMIStatus(installation)
 
 	err = r.updateStatusAndObject(originalInstallation, installation)
 	return retryRequeue, err
+}
+
+func (r *ReconcileInstallation) reconcilePodDistribution(installation *integreatlyv1alpha1.RHMI) {
+
+	serverClient, err := k8sclient.New(r.restConfig, k8sclient.Options{})
+	if err != nil {
+		logrus.Errorf("Error getting server client for pod distribution %v", err.Error())
+		installation.Status.LastError = err.Error()
+		return
+	}
+	err = poddistribution.ReconcilePodDistribution(context.TODO(), serverClient, installation.Spec.NamespacePrefix, installation.Spec.Type)
+	if err != nil {
+		logrus.Errorf("Error reconciling pod distributions %v", err.Error())
+		installation.Status.LastError = err.Error()
+	}
 }
 
 func (r *ReconcileInstallation) updateStatusAndObject(original, installation *integreatlyv1alpha1.RHMI) error {
@@ -512,7 +540,7 @@ func (r *ReconcileInstallation) handleUninstall(installation *integreatlyv1alpha
 	metrics.SetRHMIStatus(installation)
 
 	// Clean up the products which have finalizers associated to them
-	merr := &multiErr{}
+	merr := &resources.MultiErr{}
 	finalizers := []string{}
 	for _, finalizer := range installation.Finalizers {
 		finalizers = append(finalizers, finalizer)
@@ -549,7 +577,7 @@ func (r *ReconcileInstallation) handleUninstall(installation *integreatlyv1alpha
 		//don't move to next stage until all products in this stage are removed
 		//update CR and return
 		if pendingUninstalls {
-			if len(merr.errors) > 0 {
+			if len(merr.Errors) > 0 {
 				installation.Status.LastError = merr.Error()
 				r.client.Status().Update(context.TODO(), installation)
 			}
@@ -790,9 +818,9 @@ func (r *ReconcileInstallation) processStage(installation *integreatlyv1alpha1.R
 
 		if err != nil {
 			if mErr == nil {
-				mErr = &multiErr{}
+				mErr = &resources.MultiErr{}
 			}
-			mErr.(*multiErr).Add(fmt.Errorf("failed installation of %s: %w", product.Name, err))
+			mErr.(*resources.MultiErr).Add(fmt.Errorf("failed installation of %s: %w", product.Name, err))
 		}
 
 		// Verify that watches for this product CRDs have been created
@@ -929,20 +957,4 @@ func requiredEnvVar(check func(string) error) func(string, bool) error {
 
 		return check(value)
 	}
-}
-
-type multiErr struct {
-	errors []string
-}
-
-func (mer *multiErr) Error() string {
-	return "product installation errors : " + strings.Join(mer.errors, ":")
-}
-
-//Add an error to the collection
-func (mer *multiErr) Add(err error) {
-	if mer.errors == nil {
-		mer.errors = []string{}
-	}
-	mer.errors = append(mer.errors, err.Error())
 }
--- a/pkg/controller/installation/installation_controller.go
+++ b/pkg/controller/installation/installation_controller.go
@@ -463,10 +463,10 @@ func (r *ReconcileInstallation) reconcilePodDistribution(installation *integreat
 		installation.Status.LastError = err.Error()
 		return
 	}
-	err = poddistribution.ReconcilePodDistribution(context.TODO(), serverClient, installation.Spec.NamespacePrefix, installation.Spec.Type)
-	if err != nil {
-		logrus.Errorf("Error reconciling pod distributions %v", err.Error())
-		installation.Status.LastError = err.Error()
+	mErr := poddistribution.ReconcilePodDistribution(context.TODO(), serverClient, installation.Spec.NamespacePrefix, installation.Spec.Type)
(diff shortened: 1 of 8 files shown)


pkg/resources/poddistribution/poddistribution.go
```diff
@@ -0,0 +1,386 @@
...
+	}
+	logrus.Errorf("Unexpected obj kind %s", objDtls[kind])
+	return nil, nil
+}
+
+func getDeploymentConfig(ctx context.Context, client k8sclient.Client, objDtls map[string]string) (metav1.Object, error) {
+	deploymentConfig := &appsv1.DeploymentConfig{}
+	err := client.Get(ctx, k8sTypes.NamespacedName{Name: objDtls[name], Namespace: objDtls[ns]}, deploymentConfig)
+	if err != nil {
+		logrus.Errorf("Error getting DeploymentConfig %v", err)
```
maleck13, line 82: does it makes sense to log here? Or better to return err, deployment config and let the caller figure out what to do with the error?

pkg/resources/poddistribution/poddistribution.go
```diff
@@ -0,0 +1,386 @@
...
+		logrus.Errorf("Error getting StatefulSet %v", err)
+		return nil, err
+	}
+	return ss, nil
+}
+
+func parseKnn(knn string) map[string]string {
+	objDtls := map[string]string{}
+	s := strings.Split(knn, "/")
+	objDtls[kind] = s[0]
```
maleck13, line 111: is it possible this will be malformed and we end up with an indexOutOfBound error?

pkg/resources/poddistribution/poddistribution.go
```diff
@@ -0,0 +1,386 @@
...
+		nsPrefix + "user-sso",
+	}
+
+	if installType == string(integreatlyv1alpha1.InstallationTypeManagedApi) {
+		namespaces = append(namespaces, nsPrefix+"marin3r")
+	}
+	return namespaces
+}
+
+func ReconcilePodDistribution(ctx context.Context, client k8sclient.Client, nsPrefix string, installType string) {
```
maleck13, line 130: should this return an error?

pkg/resources/poddistribution/poddistribution.go
```diff
@@ -0,0 +1,386 @@
...
+	for _, ns := range getNamespaces(nsPrefix, installType) {
+		logrus.Infof("Reconciling Pod Balance in ns %s", ns)
+		unbalanced, err := findUnbalanced(ctx, ns, client)
+		if err != nil {
```
maleck13, line 134: I guess we would want to use a multi error here add a ``` continue ``` and remove the else block

pkg/resources/poddistribution/poddistribution.go
```diff
@@ -0,0 +1,386 @@
...
+			logrus.Errorf("Error getting pods to balance on namespace %s. %v", ns, err)
+		} else {
+			for knn, pods := range unbalanced {
+				objDtls := parseKnn(knn)
+				obj, err := getObject(ctx, client, objDtls)
+				if err != nil || obj == nil {
+					logrus.Errorf("Failed to get object from string %s", knn)
+					continue
+				}
+				balance := verifyBalanceCount(ctx, client, obj)
```
maleck13, line 144: maybe rename to ```rebalance``` so it becomes ``` if rebalance ```

pkg/resources/poddistribution/poddistribution.go
```diff
@@ -0,0 +1,386 @@
...
+	nodesToZone := map[string]string{}
+	for _, n := range nodes.Items {
+		for _, a := range n.Status.Addresses {
+			if a.Type == corev1.NodeInternalIP {
+				nodesToZone[a.Address] = n.Labels[ZoneLabel]
+				break
+			}
+		}
+	}
+	logrus.Infof("nodes to zone %v", nodesToZone)
```
maleck13, line 169: prob can remove this log not sure it is useful outside of development

pkg/resources/poddistribution/poddistribution.go
```diff
@@ -0,0 +1,386 @@
...
+	balance := map[string][]string{}
+	objPods := map[string][]string{}
+	l := &corev1.PodList{}
+	listOpts := []k8sclient.ListOption{
+		k8sclient.InNamespace(nameSpace),
+	}
+	if err := client.List(ctx, l, listOpts...); err != nil {
+		logrus.Errorf("Error getting pod lists %v", err)
+		return unBalanced, err
```
maleck13, line 178: I think we can remove the log and instead do ```  return fmt.Errorf("error getting pod list %v", err) ```

pkg/resources/poddistribution/poddistribution.go
```diff
@@ -0,0 +1,386 @@
...
+	}
+
+	// need to check if there is more than 1 pod of a kind
+	podCount := map[string]int{}
+	logrus.Infof("total pods in ns %s: %d", nameSpace, len(l.Items))
```
maleck13, line 183: Can remove this I think it is just noise outside development

pkg/resources/poddistribution/poddistribution.go
```diff
@@ -0,0 +1,386 @@
...
+		if sameZone(z) && podCount[knn] > 1 {
+			logrus.Warningf("Requires pod rebalance %s", knn)
+			unBalanced[knn] = objPods[knn]
+		}
+	}
+
+	return unBalanced, nil
+}
+
+func sameZone(zones []string) bool {
```
maleck13, line 233: not sure I get this? If all the pods are in 1 zone will there not just be 1 entry in the array?

pkg/resources/poddistribution/poddistribution.go
briangallagher: @maleck13  I took the general approach of treating nothing really as an error because nothing is "broke" as such. This is all optimisation. Also, I  took the approach of continuing when there was an issue, for example, to the next Controller or namespace. However, I can change the code to return an error or maybe array of errors if you like? 

On returning to the installation controller, we can simply log the errors or write to the status block. If we write to the status block we can use the existing "last_error" field, which may get immediately over wrote, by or product error for example, or create a new field. wdyt?

pkg/resources/poddistribution/poddistribution.go
briangallagher: Will add check here

pkg/resources/poddistribution/poddistribution.go
briangallagher: In general I'd like to keep the logs but can change to Debug level? You ok with that?

pkg/resources/poddistribution/poddistribution.go
briangallagher: @maleck13 
No, multiple entries in the array.  example:
DeploymentConfig/redhat-rhoam-3scale/dc1 : ["zone1", "zone1"]

pkg/resources/poddistribution/poddistribution.go
briangallagher: done

pkg/resources/poddistribution/poddistribution.go
briangallagher: updated

pkg/resources/poddistribution/poddistribution.go
briangallagher: done

pkg/resources/poddistribution/poddistribution.go
briangallagher: done

pkg/resources/poddistribution/poddistribution.go
briangallagher: done

pkg/resources/poddistribution/poddistribution.go
briangallagher: done

pkg/resources/poddistribution/poddistribution.go
briangallagher: changed to debug

pkg/resources/poddistribution/poddistribution.go
briangallagher: removed

deploy/olm-catalog/managed-api-service/1.1.0/integreatly.org_rhmiconfigs_crd.yaml
```diff
@@ -0,0 +1,113 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
philbrookes, line 1: why is this 1.1.0 stuff in here?

pkg/controller/installation/installation_controller.go
```diff
@@ -435,13 +445,28 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
...
 	}
 	metrics.SetRHMIStatus(installation)
 
 	err = r.updateStatusAndObject(originalInstallation, installation)
 	return retryRequeue, err
 }
 
+func (r *ReconcileInstallation) reconcilePodDistribution(installation *integreatlyv1alpha1.RHMI) {
+
+	serverClient, _ := k8sclient.New(r.restConfig, k8sclient.Options{})
```
philbrookes, line 460: why are we ignoring this error? Should it be handled?

deploy/olm-catalog/managed-api-service/1.1.0/integreatly.org_rhmiconfigs_crd.yaml
briangallagher: @philbrookes  Added 'rebalancePods' field to the csv, as requested by Craig, to turn this functionality on and off. Wasn't able to catch you at the time, but spoke to Paul and Craig and agreed on 1.1.0 as an appropriate version

pkg/controller/installation/installation_controller.go
briangallagher: I'm not sure if we should do anything beyond logging the error. The pod distribution functionality is an attempt to optimize the distribution but if it fails the operator can continue to work as normal. Failed distribution will then be highlighted to SRE through the PodDistributionAlert. wdyt?

philbrookes:
/approve

code looks good but ran out of time trying to verify this locally,, @CathalOConnorRH or @MStokluska can you help @briangallagher wrap this up tomorrow please? Not much left in it!

CathalOConnorRH:
Will have a look in the morning,  enjoy the break @philbrookes

briangallagher:
> Will have a look in the morning, enjoy the break @philbrookes

@CathalOConnorRH Ping me in morning when ready. WIll have cluster set up.

PR #1494, opened 2020-12-09
feat: test cases tool: update affects version field

psturc:
# Description
* updates "affects version" field in JIRA test cases with the version that's set in "fix version" field

Verified by generating test cases for this epic https://issues.redhat.com/browse/MGDAPI-951


psturc:
@pmccarthy fyi

PR #1503, opened 2020-12-11
Skipping ho3,h11, ho7 ,a16and c03 as they are flakey

R-Lawton:
## Description
test flake removal for [ho3](https://issues.redhat.com/browse/MGDAPI-935) and [ho7](https://issues.redhat.com/browse/MGDAPI-936)


MStokluska:
hey @R-Lawton, I wonder if skipping it either with a different message for each failure or identifying at which point the test fails and skip it there would be better? The reason why is because, with the current approach, we are going to skip the test with the same message at any point if it fails, later on, when looking through logs it will be hard to track down why these tests failed. I also think the issue we have with IDP tests is the `loginToThreeScale` function.

PR #1505, opened 2020-12-12
Add abstracted logger and log context

briangallagher:
# Description
Add abstracted logger and log context
JIRA: [MGDAPI-961](https://issues.redhat.com/browse/MGDAPI-961)

# Verify

- Verify successful install and uninstall of RHMI and RHOAM
- Scan the logs and verify all looks reasonable.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -162,6 +162,7 @@ github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751/go.mod h1:LOuy
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
 github.com/alecthomas/units v0.0.0-20190717042225-c3de453c63f4 h1:Hs82Z41s6SdL1CELW+XaDYmOH4hkBN4/N9og/AsOv7E=
 github.com/alecthomas/units v0.0.0-20190717042225-c3de453c63f4/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
+github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d h1:UQZhZ2O0vMHr2cI+DC1Mbh0TJxzA3RcLoMsFw+aXw7E=
 github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d/go.mod h1:rBZYJk541a8SKzHPHnH3zbiI+7dagKZ0cgpgrD7Fyho=
 github.com/aliyun/aliyun-oss-go-sdk v2.0.4+incompatible/go.mod h1:T/Aws4fEfogEE9v+HPhhw+CntffsBHJ8nXQCwKr0/g8=
 github.com/andreyvit/diff v0.0.0-20170406064948-c7f18ee00883/go.mod h1:rCTlJbsFo29Kk6CurOXKm700vrz8f0KW0JNfpkRJY/8=
--- a/pkg/controller/installation/installation_controller.go
+++ b/pkg/controller/installation/installation_controller.go
@@ -5,12 +5,14 @@ import (
 	"errors"
 	"fmt"
 	"github.com/integr8ly/integreatly-operator/pkg/resources/poddistribution"
+	"github.com/sirupsen/logrus"
 	"os"
 	"reflect"
 	"strconv"
 	"strings"
 	"time"
 
+	l "github.com/integr8ly/integreatly-operator/pkg/resources/logger"
 	"github.com/integr8ly/integreatly-operator/version"
 
 	marin3rconfig "github.com/integr8ly/integreatly-operator/pkg/products/marin3r/config"
@@ -20,8 +22,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	usersv1 "github.com/openshift/api/user/v1"
-
-	"github.com/sirupsen/logrus"
 
 	"github.com/integr8ly/integreatly-operator/pkg/addon"
 	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
@@ -65,6 +65,7 @@ const (
 
 var (
 	productVersionMismatchFound bool
+	log                         = l.NewLoggerWithContext(l.Fields{l.ControllerLogContext: "installation_controller"})
 )
 
 // Add creates a new Installation Controller and adds it to the Manager. The Manager will set fields on the Controller
@@ -135,7 +136,7 @@ func add(mgr manager.Manager, r ReconcileInstallation) error {
 	// Watch for changes to rate limit alerts config
 	err = c.Watch(
 		&source.Kind{Type: &corev1.ConfigMap{}},
-		&EnqueueIntegreatlyOwner{},
+		&EnqueueIntegreatlyOwner{log: log},
 	)
 	if err != nil {
 		return err
@@ -157,7 +158,7 @@ func createInstallationCR(ctx context.Context, serverClient k8sclient.Client) er
 		return err
 	}
 
-	logrus.Infof("Looking for rhmi CR in %s namespace", namespace)
+	log.Infof("Looking for rhmi CR", l.Fields{"namespace": namespace})
 
 	installationList := &integreatlyv1alpha1.RHMIList{}
 	listOpts := []k8sclient.ListOption{
@@ -179,7 +180,7 @@ func createInstallationCR(ctx context.Context, serverClient k8sclient.Client) er
 		installType, _ := os.LookupEnv(installTypeEnvName)
 		priorityClassName, _ := os.LookupEnv(priorityClassNameEnvName)
 
-		logrus.Infof("Creating a %s rhmi CR with USC %s, as no CR rhmis were found in %s namespace", installType, useClusterStorage, namespace)
+		log.Infof("No rhmi CRs found, creating one", l.Fields{"installType": installType, "USC": useClusterStorage, "ns": namespace})
 
 		if installType == "" {
 			installType = string(integreatlyv1alpha1.InstallationTypeManaged)
@@ -272,6 +273,7 @@ type ReconcileInstallation struct {
 // Reconcile reads that state of the cluster for a Installation object and makes changes based on the state read
 // and what is in the Installation.Spec
 func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+
 	installInProgress := false
 	installation := &integreatlyv1alpha1.RHMI{}
 	err := r.client.Get(context.TODO(), request.NamespacedName, installation)
@@ -300,21 +302,21 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 
 	cssreAlertingEmailAddress := os.Getenv(alertingEmailAddressEnvName)
 	if installation.Spec.AlertingEmailAddresses.CSSRE == "" && cssreAlertingEmailAddress != "" {
-		logrus.Infof("Adding CS-SRE alerting email address to RHMI CR")
+		log.Info("Adding CS-SRE alerting email address to RHMI CR")
 		installation.Spec.AlertingEmailAddresses.CSSRE = cssreAlertingEmailAddress
 		err = r.client.Update(context.TODO(), installation)
 		if err != nil {
-			logrus.Errorf("Error while copying alerting email addresses to RHMI CR: %v", err)
+			log.Error("Error while copying alerting email addresses to RHMI CR", err)
 		}
 	}
 
 	buAlertingEmailAddress := os.Getenv(buAlertingEmailAddressEnvName)
 	if installation.Spec.AlertingEmailAddresses.BusinessUnit == "" && buAlertingEmailAddress != "" {
-		logrus.Infof("Adding BU alerting email address to RHMI CR")
+		log.Info("Adding BU alerting email address to RHMI CR")
 		installation.Spec.AlertingEmailAddresses.BusinessUnit = buAlertingEmailAddress
 		err = r.client.Update(context.TODO(), installation)
 		if err != nil {
-			logrus.Errorf("Error while copying alerting email addresses to RHMI CR: %v", err)
+			log.Error("Error while copying alerting email addresses to RHMI CR", err)
 		}
 	}
 
@@ -326,12 +328,12 @@ func (r *ReconcileInstallation) Reconcile(request reconcile.Request) (reconcile.
 		"notification-email",
 	)
 	if err != nil {
-		logrus.Errorf("failed while retrieving addon parameter: %v", err)
+		log.Error("failed while retrieving addon parameter", err)
 	} else if ok && customerAlertingEmailAddress != "" && installation.Spec.AlertingEmailAddress != customerAlertingEmailAddress {
-		logrus.Info("Updating customer email address from parameter")
+		log.Info("Updating customer email address from parameter")
 		installation.Spec.AlertingEmailAddress = customerAlertingEmailAddress
 		if err := r.client.Update(context.TODO(), installation); err != nil {
-			logrus.Errorf("Error while updating customer email address to RHMI CR: %v", err)
+			log.Error("Error while updating customer email address to RHMI CR", err)
 		}
 	}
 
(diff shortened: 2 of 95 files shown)


MStokluska:
So far I've reviewed the logs and they look good, examples:
Error:
error
Warning:
warning
Normal:
normal

pkg/controller/installation/installation_controller.go
```diff
@@ -598,11 +601,11 @@ func (r *ReconcileInstallation) handleUninstall(installation *integreatlyv1alpha
 			return retryRequeue, merr
 		}
 
-		logrus.Infof("uninstall completed")
+		log.Info("uninstall completed")
```
MStokluska, line 604: Hey Brian, just a question more than anything else, is there a reason why we sometimes log an info message starting with capital letters and sometimes we don't?

pkg/controller/namespacelabel/namespacelabel_controller.go
```diff
@@ -252,20 +255,20 @@ func CheckCidrValueAndUpdate(value string, request reconcile.Request, r *Reconci
...
-		logrus.Infof("Cidr value is already set to : %v , not updating", cidr)
+		r.log.Infof("Cidr value is already set, not updating", l.Fields{"value": cidr})
 		return nil
 	}
 
 	// replace - character from label with / so that the cidr value is set correctly.
 	// / is not a valid character in namespace label values.
 	newCidr := strings.Replace(value, "-", "/", -1)
-	logrus.Infof("No cidr has been set in configmap yet, Setting cidr from namespace label : %v", newCidr)
+	r.log.Infof("No cidr has been set in configmap yet, Setting cidr from namespace label", l.Fields{"newCidr": newCidr})
```
MStokluska, line 271:
```suggestion
	r.log.Infof("No cidr has been set in configmap yet, setting cidr from namespace label", l.Fields{"newCidr": newCidr})
```

pkg/controller/subscription/subscription_controller.go
```diff
@@ -139,7 +140,7 @@ type ReconcileSubscription struct {
 func (r *ReconcileSubscription) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	// skip any Subscriptions that are not integreatly operator
 	if !r.shouldReconcileSubscription(request) {
-		logrus.Infof("not our subscription: %+v, %s", request, r.operatorNamespace)
+		log.Infof("Not our subscription", l.Fields{"request": request, "opNS": r.operatorNamespace})
```
MStokluska, line 143: Another question, why do we sometimes, for example in RHMI config controller have the logger returned in `newReconciller` and use the `r.log.xxxx` and as in the example above we use it directly via `log.xxxx` - I am probably missing something and wonder if you could help me understand the difference?

pkg/products/grafana/reconciler.go
```diff
@@ -79,43 +79,42 @@ func NewReconciler(configManager config.ConfigReadWriter, installation *integrea
...
-		logger:        logger,
+		log:           logger,
 		Reconciler:    resources.NewReconciler(mpm),
 		recorder:      recorder,
 	}, nil
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1alpha1.RHMI, product *integreatlyv1alpha1.RHMIProductStatus, client k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
-	logrus.Infof("Start Grafana reconcile")
+	r.log.Info("Start Grafana reconcile")
```
MStokluska, line 94: I like the idea of having "START" and "FINISH" of reconciling highlighted, wonder if you think we could do the same for other products? It seems there's a bit of inconsistency, for example, fuse is missing the "start" of the reconciling log. To me, logs are much easier to read when the information they provide is consistent and I think either removing the above the line or having it done the same way for other products would be great!

pkg/products/marin3r/apiUsagePrometheusRules.go
```diff
@@ -157,8 +157,9 @@ func getRateLimitInSeconds(rateLimitUnit string, rateLimitRequestsPerUnit uint32
 	} else if rateLimitUnit == "day" {
 		return float64(rateLimitRequestsPerUnit) / (60 * 60 * 24), nil
 	} else {
-		logrus.Errorf("Unexpected Rate Limit Unit %v, while creating 3scale api usage alerts", rateLimitUnit)
-		return 0, errors.New(fmt.Sprintf("Unexpected Rate Limit Unit %v, while creating 3scale api usage alerts", rateLimitUnit))
+		err := errors.New(fmt.Sprintf("Unexpected Rate Limit Unit %v, while creating 3scale api usage alerts", rateLimitUnit))
+		r.log.Error("", err)
```
MStokluska, line 161: Is the empty string, err intended? I see that the error string got created @160 but the empty string is a bit confusing.

pkg/controller/installation/installation_controller.go
```diff
@@ -644,7 +647,7 @@ func (r *ReconcileInstallation) preflightChecks(installation *integreatlyv1alpha
 		installation.Status.PreflightStatus = integreatlyv1alpha1.PreflightFail
 		installation.Status.PreflightMessage = "Spec.useClusterStorage must be set to either 'true' or 'false' to continue"
 		_ = r.client.Status().Update(context.TODO(), installation)
-		logrus.Infof("preflight checks failed on useClusterStorage value")
+		log.Info("preflight checks failed on useClusterStorage value")
```
MStokluska, line 650: Wonder if this could be marked as an error as the installation won't continue without it?

pkg/controller/installation/installation_controller.go
```diff
@@ -671,33 +674,33 @@ func (r *ReconcileInstallation) preflightChecks(installation *integreatlyv1alpha
...
 	}
 
-	logrus.Infof("getting namespaces")
+	log.Info("getting namespaces")
 	namespaces := &corev1.NamespaceList{}
 	err := r.client.List(context.TODO(), namespaces)
 	if err != nil {
 		// could not list namespaces, keep trying
-		logrus.Infof("error listing namespaces, will retry")
+		log.Info("error listing namespaces, will retry")
```
MStokluska, line 688: Not sure if the `will retry` bit is needed, just seems like everything in our reconcilers is going to be `retried`, maybe an error or stay consistent with other logs (`going to retry` or not for all logs that might have trigger on failure)?

pkg/controller/installation/installation_controller.go
MStokluska: ah, nevermind, I see there's no error available.

pkg/products/rhssocommon/reconciler.go
```diff
@@ -127,7 +127,7 @@ func (r *Reconciler) CleanupKeycloakResources(ctx context.Context, inst *integre
 	if err != nil {
 		return integreatlyv1alpha1.PhaseFailed, err
 	} else if len(users.Items) > 0 {
-		logrus.Println("rhsso deletion of users in progress")
+		r.Log.Info("rhsso deletion of users in progress")
```
MStokluska, line 130: Would it make sense to split it between user-sso and rhsso?

pkg/products/rhssocommon/reconciler.go
```diff
@@ -136,7 +136,7 @@ func (r *Reconciler) CleanupKeycloakResources(ctx context.Context, inst *integre
 	if err != nil {
 		return integreatlyv1alpha1.PhaseFailed, err
 	} else if len(clients.Items) > 0 {
-		logrus.Println("rhsso deletion of clients in progress")
+		r.Log.Info("rhsso deletion of clients in progress")
```
MStokluska, line 139: Same as above

pkg/controller/installation/installation_controller.go
MStokluska: hmm, maybe use of r.Log.Warning would be better?

pkg/products/ups/reconciler.go
```diff
@@ -107,34 +104,34 @@ func (r *Reconciler) VerifyVersion(installation *integreatlyv1alpha1.RHMI) bool
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1alpha1.RHMI, product *integreatlyv1alpha1.RHMIProductStatus, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
-	logrus.Infof("Reconciling %s", defaultUpsName)
+	r.log.Info("Start Reconcile")
```
MStokluska, line 107: Feels like this could be a bit more descriptive. "Starting UPS Reconcile" ?

pkg/resources/logger/logger_test.go
```diff
@@ -0,0 +1,30 @@
...
+	log.Warningf("This is a Warningf log", Fields{"agr1": "agr1"})
+
+	err := errors.New("This is an error")
+	log.Error("This is a Error log", err)
+	log.Errorf("This is a Errorf log", Fields{"agr1": "agr1"}, err)
+
+	log.Error("This is a Error log with nil err object", nil)
+	log.Errorf("This is a Errorf log with nil err object", nil, nil)
+
+	//log.Fatal("This is a Fatal log: ", err)
```
MStokluska, line 27: Are the commented out lines required?

pkg/resources/marketplace/manager.go
```diff
@@ -75,14 +76,14 @@ func (m *Manager) InstallOperator(ctx context.Context, serverClient k8sclient.Cl
 	}
 	err = serverClient.Create(ctx, og)
 	if err != nil && !k8serr.IsAlreadyExists(err) {
-		logrus.Infof("error creating operator group")
+		log.Info("error creating operator group")
```
MStokluska, line 79: could this be made into an log.Error message?

pkg/resources/marketplace/manager.go
```diff
@@ -75,14 +76,14 @@ func (m *Manager) InstallOperator(ctx context.Context, serverClient k8sclient.Cl
...
 		return err
 	}
 
-	logrus.Infof("creating subscription in ns if it doesn't already exist: %s", t.Namespace)
+	log.Infof("Creating subscription in ns if it doesn't already exist", l.Fields{"ns": t.Namespace})
 	err = serverClient.Create(ctx, sub)
 	if err != nil && !k8serr.IsAlreadyExists(err) {
-		logrus.Infof("error creating sub")
+		log.Info("error creating sub")
```
MStokluska, line 86: same as above

pkg/resources/marketplace/manifestHelper.go
```diff
@@ -100,7 +100,7 @@ func GetCRDsFromManifestAsStringList(dir string, regex string, packageYaml strin
 			currentVersion, err = GetCurrentCSVFromManifest(packageYaml)
 		}
 		if err != nil {
-			logrus.Fatal("Error")
+			log.Fatal("Error ", err)
```
MStokluska, line 103: I'm confused with this err log, we are logging an error at line 93, wonder if I'm missing something but can't see why are we returning it again, and the "Error " isn't very descriptive to me.

MStokluska:
I've reviewed it and it looks fine, there's an improvement in logging to me, I like the new way we are logging messages. 
Left some comments, however, I am not they are really a part of the scope of this ticket, just feels like we would be better with some of the logs removed/changed and have a bit more consistency, for example, when a reconcile loop starts and completes, have a message stating which reconcile starts. I see this is done for some but not for all products.  The new logger logic works perfectly fine. Well done!

pkg/controller/installation/installation_controller.go
briangallagher: We don't have a standard for that

pkg/controller/subscription/subscription_controller.go
briangallagher: Thats a bit inconsistent. I've changed the RHMI config controller to just log.

In some cases, the product reconcilers for example, we use the logger from the installation controller so it makes sense to add it to the reconciler. But not for RHMI config controller for you're right to point that out.

pkg/products/grafana/reconciler.go
briangallagher: We don't yet have standards around this. The main purpose of this JIRA was to get the contextual and structure logging in place. It wasn't to fix all logging issues or set the standards. It's a good suggestion but I don't think it should be in scope of this JIRA.

pkg/products/marin3r/apiUsagePrometheusRules.go
briangallagher: In most cases it's useful to pass context to the error log. In this case because we're creating the error there's no need. Anything here would just be superfluous.

pkg/controller/installation/installation_controller.go
briangallagher: Changed to warning

pkg/controller/installation/installation_controller.go
briangallagher: Updated to: `log.Warningf("error listing namespaces", l.Fields{"error": err.Error()})`

pkg/products/rhssocommon/reconciler.go
briangallagher: The log will output with the context of user-sso or rhsso

pkg/products/ups/reconciler.go
briangallagher: Log will output with context product=UPS

pkg/resources/logger/logger_test.go
briangallagher: removed

pkg/resources/marketplace/manager.go
briangallagher: done

pkg/resources/marketplace/manager.go
briangallagher: done

pkg/products/grafana/reconciler.go
MStokluska: Thanks Brian, kinda felt that way :D sorry for pointing the actual contents of the logs in your PR then!

pkg/products/marin3r/apiUsagePrometheusRules.go
MStokluska: Makes sense, thank you.

pkg/resources/marketplace/manifestHelper.go
```diff
@@ -90,7 +90,7 @@ func GetCRDsFromManifestAsStringList(dir string, regex string, packageYaml strin
 
 		err := ReverseSlice(vs)
 		if err != nil {
-			logrus.Error("ReverseSlice erorr %w", err)
+			log.Error("ReverseSlice erorr %w", err)
```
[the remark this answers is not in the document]
briangallagher: done 👍

pkg/resources/marketplace/manifestHelper.go
briangallagher: I've updated the msg to "Error getting current csv from manifest" 
The log is necessary because of the statement on L100

MStokluska: Makes sense, thanks!

briangallagher:
> n a reconcile loop

@MStokluska Thanks. With regards to Improving the log statements I agree. I think we can do this as we touch on different aspects of the code base in other JIRAs. Jim raised a related question on a retro recently and the conclusion was that we should make small quick improvements to the code when working in a particular file(s). I think logging improvements is a perfect example. Might be no harm setting some standards also.

go.sum
```diff
@@ -162,6 +162,7 @@ github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751/go.mod h1:LOuy
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
 github.com/alecthomas/units v0.0.0-20190717042225-c3de453c63f4 h1:Hs82Z41s6SdL1CELW+XaDYmOH4hkBN4/N9og/AsOv7E=
 github.com/alecthomas/units v0.0.0-20190717042225-c3de453c63f4/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
+github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d h1:UQZhZ2O0vMHr2cI+DC1Mbh0TJxzA3RcLoMsFw+aXw7E=
```
philbrookes, line 165: @MStokluska does this change mean that we need to update the manifest file?

PR #1506, opened 2020-12-15
MGDAPI-498 - Marin3r Critical SLO Panel

KevFan:
# Description

Branched from https://github.com/integr8ly/integreatly-operator/pull/1502 with an additional commit to fix dashboard panel were not loaded when Critical SLO Dashboard page is initially visited

Jira:
* https://issues.redhat.com/browse/MGDAPI-498
## Type of change

## Verification

- Install RHOAM from this branch
- Run though E04B test case https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/dashboards/e04b-verify-slo-dashboard.md
- Access Grafana and verify two alerts are firing in the Critical SLO dashboard in the Marin3r Panel
Screenshot from 2020-12-15 16-54-15
Screenshot from 2020-12-15 17-15-11

## Checklist

### DIFF
--- a/pkg/products/monitoring/dashboards/criticalSLOAlerts.go
+++ b/pkg/products/monitoring/dashboards/criticalSLOAlerts.go
@@ -4096,7 +4096,7 @@ func GetMonitoringGrafanaDBCriticalSLORHMIAlertsJSON(nsPrefix string) string {
 				"current": {
 					"selected": true,
 					"text": "All",
-					"value": "$__all"
+					"value": ["$__all"]
 				},
 				"hide": 0,
 				"includeAll": true,
@@ -4148,7 +4148,7 @@ func GetMonitoringGrafanaDBCriticalSLORHMIAlertsJSON(nsPrefix string) string {
 				"current": {
 					"selected": true,
 					"text": "All",
-					"value": "$__all"
+					"value": ["$__all"]
 				},
 				"hide": 0,
 				"includeAll": true,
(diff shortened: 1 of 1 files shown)


obrienrobert:
Verified that 2 Mari3r alerts fired when executing the test case:
mari3r_alerts

@KevFan the E04B may have to be updated once this is merged to scale down the Mari3r deployments and trigger the alerts.

KevFan:
> Verified that 2 Mari3r alerts fired when executing the test case:
> mari3r_alerts
> 
> @KevFan the E04B may have to be updated once this is merged to scale down the Mari3r deployments and trigger the alerts.

Thanks @obrienrobert, good point :+1: E04B probably should also include scaling down customer-grafana also in this case, though I think currently the alerts from scalling down are all `warning` severity so won't be reflected in the dashboard for now. 
* https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/marin3r/prometheusRules.go
* https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/grafana/prometheusRules.go
I can update the test case to include scaling down these deployments as part of this PR, what do you think ? :thinking:

obrienrobert:
@KevFan If using the scripts in the test case to scale the 3scale deployment down also causes the marin3r alerts to trigger in the SLO dashboard then I think everything looks good to me!

KevFan:
> @KevFan If using the scripts in the test case to scale the 3scale deployment down also causes the marin3r alerts to trigger in the SLO dashboard then I think everything looks good to me!

Cool, thanks @obrienrobert. Yeah the current test case will cause the Marin3r critical alerts to fire due to the envoy side car container is down once the 3scale pods are scaled down  :+1: 
* https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/threescale/envoy_prometheusRules.go#L22-L41

PR #1507, opened 2020-12-16
MGDAPI-562 User-SSO ratelimiting

MStokluska:
# Description
Addresses https://issues.redhat.com/browse/MGDAPI-562

# What
Currently, we are not rate-limiting user-sso login requests. This PR aims to rate-limit login requests under the same rate-limiting count as we have for apicasts. 

# How
In order to be able to select paths which we want to rate-limit in user sso, I had to add the following:
- Edit keycloak-edge route to point to a newly created service called "ratelimit" under `redhat-rhoam-user-sso` namespace
- Edit keycloak-edge route to use "edge" `tlsTermination`
- From the edge route, traffic is directed into rate limit service which forwards it to envoy sidecar container that is on the keycloak pod under port `8444`
- Evoy checks the route that it is being accessed from
a) if route ends in `/auth/realms/master/protocol/openid-connect/token` we are not rate-limiting as it means an SSO admin console is being accessed
b) if the route ends in `/token` - we are rate-limiting
c) any other route is not rate limited
- Traffic then is redirected back to rate limit service under user-sso ns and pushed to keycloak pod on port 8080

## Envoy configuration file
We are creating two envoy configuration files, one for 3scale apicasts and one for user-sso logins.
Configuration CR consists of 
- Clusters - clusters define where to push traffic to, we have 2 clusters in both of the products
a) rate limit cluster - this is a description of the rate limit service under marin3r ns, this cluster checks the rate limit count has been reached or not
b) outbound traffic cluster - where to direct the traffic after a successful response from rate limit cluster (a response is considered successful when the rate limit hasn't been reached)
- Listener - a listener describes what incoming traffic to listen for, in this example, we are listening for any traffic coming to sidecar on port 8443 for 3scale, and on 8444 for keycloak. The listener also defines the route filtering and which cluster push the traffic to. 
For example, we can specify in the listener what traffic to rate limit, for apicasts, it's everything (`/` route), for keycloak, we first specify that we don't want to rate limit admin route, then, we want to rate-limit `/token` route, and finally, we don't want to rate limit anything else. The rate-limit path filtering are a REGEX rules. 

# How to verify
In order to verify, we are going to create a new realm, client, user and integrate it with 3scale API, after which we are going to connect to 3Scale and verify that the rate limit works on login and, it still for reaching the apicasts.

## Prerequisites
- Postman - unless you are happy with authentication via curl!

## Installation
- Run `INSTALLATION_TYPE=managed-api make cluster/prepare/local`
- Install RHOAM from this branch via `INSTALLATION_TYPE=managed-api make code/run` 

NOTE: If you are using a 4 node cluster please update the replica values in https://github.com/integr8ly/integreatly-operator/blob/master/pkg/config/threescale.go#L120 to 1 or 2 for each of the replicas. Or, alternatively increase the node count to 6.

- Confirm that the installation has been successful 

## Rate limit control 
In order to verify this PR, it is good to keep an eye on the current rate limit count. In order to do that, please create and run the following script:
```
#!/bin/sh
echo "Creating throwaway redis container..."
  cat << EOF | oc create -f - -n redhat-rhoam-operator
  kind: Redis
  metadata:
    name: throw-away-redis-pod
    labels:
      productName: productName
  spec:
    secretRef:
      name: throw-away-redis-sec
    tier: development
    type: workshop
EOF

  while true
  do
      PHASE=`oc get redis/throw-away-redis-pod -n redhat-rhoam-operator -o template --template={{.status.phase}}`
      if [ "$PHASE" = 'complete' ]; then
        break
      fi

      echo "Waiting for throwaway Redis container to complete. Current phase: $PHASE..."
      sleep 10s
  done

echo "Container ready..."
echo ""
echo "Running Redis connections count for 30 minutes..."
echo ""

POD_NAME=$(oc get pod -n redhat-rhoam-operator | grep -v "deploy" | grep throw-away-redis-pod | awk '{print $1}')
REDIS_HOST=$(oc get secrets/ratelimit-service-redis-rhoam -n redhat-rhoam-operator -o template --template={{.data.uri}} | base64 --decode)

OS=$(uname)
if [[ $OS = Linux ]]; then
    RUNTIME="30 minute"
    ENDTIME=$(date -ud "$RUNTIME" +%s)
elif [[ $OS = Darwin ]]; then
    RUNTIME="+30m"
    ENDTIME=$(date -v $RUNTIME +%s)
fi

while [[ $(date -u +%s) -le $ENDTIME ]]
do
RESULT=$(kubectl exec $POD_NAME \
    -n redhat-rhoam-operator \
    -- /opt/rh/rh-redis32/root/usr/bin/redis-cli -c -h $REDIS_HOST -p 6379 KEYS '*' | grep -v "liveness-probe")

    if (( $(grep -c . <<<"$RESULT") > 1 )); then
        echo "Flushing REDIS, time slot for allowed hits per unit has been reached. Ping APICAST again to create a new key value pair"
        kubectl exec $POD_NAME \
            -n redhat-rhoam-operator \
            -- /opt/rh/rh-redis32/root/usr/bin/redis-cli -c -h $REDIS_HOST -p 6379 FLUSHALL
    else
        HITCOUNT=$(kubectl exec $POD_NAME \
            -n redhat-rhoam-operator \
            -- /opt/rh/rh-redis32/root/usr/bin/redis-cli -c -h $REDIS_HOST -p 6379 GET $RESULT)

        if echo $HITCOUNT | grep -q "ERR";
        then
            echo ""
            echo "Curl Apicast or keycloak login endpoint to create a REDIS key value pair"
            echo ""
        else
            echo ""
            echo "REDIS key in rate limit redis: $RESULT"
            echo "Current hit count: $HITCOUNT"
            echo ""
        fi
fi
    echo "-------------------------------------------------------------------------------"
    sleep 5s
done
```
NOTE: The script needs to reset the rate limit count every minute (Redis FLUSH). It will run for 30 minutes, and output the current count of rate-limited requests (whether it is apicasts or keycloak token requests). After 30 minutes, please re-run the script. The best approach here would be to have the script opened in another window and keep an eye on it. We do not want the counter to increase unless specified by the verification steps.

## Keycloak Setup
In order to integrate 3scale application with user-sso we need to setup keycloak realm and clients first.
- Create cluster IDP, you can use either `bash scripts/setup-sso-idp.sh`
NOTE: Please take note of the customer-admin id and password
- Once IDP is created, login to your cluster via the customer-admin role in a private window
- Navigate to redhat-rhoam-user-sso ns and open the keycloak-edge route in the tab
- Log in to keycloak via customer admin
- Create a new realm (top left of the screen under realm section (hover over "master"))
- Name of the realm `3scale`
- Navigate to `Clients` and click on `Create` with client ID `3scale` and click save
- Set access type to `confidential`
- Enable flows:
``` 
standard flow
implicit flow
direct access grant
service accounts
authorization flow
```
- `*` for `Valid Redirect URIs`
- Click save
- `Service accounts roles` tab should now appear 
- Open the service accounts tab and click on client roles and type `realm management`.
- choose manage-clients and click add selected
- navigate to the credentials tab and take note of the secret
- next, click on the Users tab on the left side and add a user
- set user name to `test-user` and click save
- navigate to credentials and set a password for the user to `test-user` also, but before applying your changes, please make sure that the `temporary password` field is set to `off`. 

You should now have a client ID - which is `3scale` and a secret associated with this client, and a single user with `test-user` id and `test-user` password. The purpose of it was that we are going to integrate a 3scale product with this realm and we want 3scale to be able to create clients for us, and also, we want to use our created test-user to be able to request token to access apicast.

Keep user sso console opened.

NOTE: The rate limit count should not have been increased at any point while following the above settings

## 3Scale application integration
Next, we are going to create apicast
- Navigate to routes under the 3scale namespace
- Open the admin route 
- Click on login via keycloak and select testing idp
- Once logged in, follow the wizard to create a sample api (just keep following wizard steps after `ok, how does 3scale work` until you are brought to the 3scale console)
NOTE: The rate limit count should have increased by 1 at this stage ( however, it might have happened that the wizard tried to access the apicast endpoint at the time of redis rate-limiting count reset, if such happened, you would not see an increase in rate limit count but that's ok, we are going to confirm that apicasts are ratelimited later on)
- Once in 3scale console click on Integration (left side of the screen) and Settings
- Under authentication click on OpenID Connect
- In the `OpenID Connect Issuer Type` select Red Hat Single Sign On
NOTE: `OpenID Connect Issuer` is as follows:
https://<client_id_from_rhsso_realm>:<client_secret_from_rhhso_realm>@<KEYCLOAK ROUTE WITHOUT HTTPS://>/auth/realms/<REALM_NAME>
Example:
`https://3scale:12345678912345@keycloak-edge-redhat-rhoam-user-sso.apps.mstoklus.1234567.s1.devshift.org/auth/realms/3scale`
- Select all OIDC authorization flows
- Choose HTTP Headers for Credentials Location
- Click update product at the bottom of settings screen
- Navigate to `Audience` (scroll down tab on top beside Red Hat 3scale API Management logo)
- Click on `Listing` on the left of the screen
- Click on `Create` on the right of the screen
- Username `testUser`, Emal `testUser@something.com`, Passwod `testUser`, organization `testOrg` and click `create`
NOTE: The rate limit should be bumped by 2 now. This is happening as there are requests sent to Keycloak in order to register a new application.
- Navigate back to User SSO console > Clients section and notice that a new client has been created 
- Click on that client (usually the name looks like a bunch of random numbers) and confirm that the client's org is testOrg
- Select `Authorization Enabled` and `*` for valid URIs and save changes
- Take note of client ID and secret (from credentials tab)
- Go back to 3scale console and select from the top scroll down menu `Product: API`
- Click on integration > configuration
- Click on `Promote v. 2 to Staging APIcast` and `Promote v. 1 to Production APIcast`
- Copy the staging URL

## Postman
In postman first, we need to get the ./wellknown configuration of OPENID
- GET the following route `https://<USER_SSO_URL>/auth/realms/3scale/.well-known/openid-configuration`
- Keep the results opened
- Open another GET with the apicast url, for example: `https://api-3scale-apicast-staging.apps.mstokluska.y9c8.s1.devshift.org:443`
- Click on Authorization section under above GET url command
- Select OAuth 2.0
- Click on `Get New Access Token`
- Token Name -  anything you want
- Auth URL - the `/auth` path from the result of the ./wellknown GET
- Access Token URL - the `/token` path from the result of the ./wellknown GET
- Client ID - the 3scale generated client (not the 3scale client but the autogenerated one)
- Client Secret - the 3scale generated client secret
- scope - openid
- Grant Type - Authorization Code
- Tick `Request access token locally`
- Hit the `Request Token` button
- In the user sso login screen login as test user - if you have followed above instructions, it will be `test-user` `test-user`
- Once you login - a new token will be obtained and a rate limit count should bump by 1.
- If it happened that you have logged in during REDIS flush (happens every 1 minute), delete the token by clicking on it in postman, and clicking the `Delete` button
- Click on the token and select to to be added as a Header
- Send a request to the apicast url
- Rate limit count should increase by 1.

You can navigate to user-sso, go into users, select test-user > sessions and terminate ongoing session to trigger the login process again.

@pb82 - I wonder if you could go through the flow of the above rate-limiting and let me know what you think about rate-limiting the /token endpoint.


pkg/resources/ratelimit/envoyconfig.go
```diff
@@ -50,3 +80,398 @@ func DeleteEnvoyConfigsInNamespace(ctx context.Context, client k8sclient.Client,
...
+						},
+					},
+				},
+			},
+			{
+				Match: &v2route.RouteMatch{
+					PathSpecifier: &v2route.RouteMatch_SafeRegex{
+						SafeRegex: &envoy_type_matcher.RegexMatcher{
+							EngineType: &envoy_type_matcher.RegexMatcher_GoogleRe2{},
+							Regex:      ".*/token",
```
philbrookes, line 198: doesn't this also match line 181? So can we remove lines 176-192?

philbrookes: nvm, I see the actions are different now.

matskiv:
We will need to disable the route created by Keycloak operator, because that one wouldn't be rate limited.
https://github.com/integr8ly/integreatly-operator/blob/037655d0afda718b19def325591eacda5b5c7d50/pkg/products/rhssouser/reconciler.go#L395
`kc.Spec.ExternalAccess.Enabled = false`

PR #1508, opened 2020-12-17
MGDAPI-1002 - improve logging in marin3r reconciler

matskiv:
# Description

JIRA: https://issues.redhat.com/browse/MGDAPI-1002

Marin3r and customer monitoring [are in the same stage](https://github.com/integr8ly/integreatly-operator/blob/a78b3aaad51a74f32355b60b524bb7e12d4514aa/pkg/controller/installation/types.go#L48-L49),
and marin3r reconciler [tries to retrieve grafana URL](https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/marin3r/reconciler.go#L225-L263) before "redhat-rhoam-customer-monitoring" ns is created.
This results into an error message in logs which can in fact be ignored under specific circumstances (during installation, before grafana is installed). Error message:

> time="2020-12-17T00:29:28Z" level=error msg="failed to get Grafana console URL Failed to get route for Grafana: routes.route.openshift.io \"grafana-route\" is forbidden: User \"system:serviceaccount:redhat-rhoam-operator:rhmi-operator\" cannot get resource \"routes\" in API group \"route.openshift.io\" in the namespace \"redhat-rhoam-customer-monitoring\""

This change should ensure that this error is not logged when Garafana isn't fully installed.

## Type of change

## Checklist

### DIFF
--- a/pkg/products/grafana/reconciler.go
+++ b/pkg/products/grafana/reconciler.go
@@ -492,7 +492,7 @@ func GetGrafanaConsoleURL(ctx context.Context, serverClient k8sclient.Client, in
 
 	err := serverClient.Get(ctx, k8sclient.ObjectKey{Name: defaultRoutename, Namespace: ns}, grafanaRoute)
 	if err != nil {
-		return "", fmt.Errorf("Failed to get route for Grafana: %w", err)
+		return "", err
 	}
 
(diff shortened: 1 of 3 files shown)


grdryn (approved):
/lgtm

PR #1511, opened 2020-12-18
Add support for a from alert email address

maleck13:
# Description
$subject

As part of this change we will need to add a new parameter to the managed tenants repo see PR 471

**Verification**

Spin up a cluster and run the operator locally overriding ALERT_SMTP_FROM and setting it to a value of your choice.

Once the install is complete, check the smtp config in alert manager

```
oc get secret alertmanager-application-monitoring -n redhat-rhoam-middleware-monitoring-operator -o=yaml
```

it should have the ```smtp_from``` set to your chosen email address.

## Type of change

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,7 @@ export SELF_SIGNED_CERTS   ?= true
 # Setting the INSTALLATION_TYPE to managed-api will configure the values required for RHOAM installs
 export INSTALLATION_TYPE   ?= managed
 
+export ALERT_SMTP_FROM ?=no-reply-alert@devshift.org
 export USE_CLUSTER_STORAGE ?= true
 export OPERATORS_IN_PRODUCT_NAMESPACE ?= false # e2e tests and createInstallationCR() need to be updated when default is changed
 export DELOREAN_PULL_SECRET_NAME ?= integreatly-delorean-pull-secret
--- a/deploy/crds/integreatly.org_rhmis_crd.yaml
+++ b/deploy/crds/integreatly.org_rhmis_crd.yaml
@@ -17,20 +17,18 @@ spec:
       description: RHMI is the Schema for the RHMI API
       properties:
         apiVersion:
-          description: 'APIVersion defines the versioned schema of this representation
-            of an object. Servers should convert recognized schemas to the latest
-            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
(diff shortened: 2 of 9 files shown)


deploy/crds/integreatly.org_rhmis_crd.yaml
```diff
@@ -76,26 +66,22 @@ spec:
...
-                modifying this file Add custom validation using kubebuilder tags:
-                https://book.kubebuilder.io/beyond_basics/generating_crd.html'
+              description: 'INSERT ADDITIONAL SPEC FIELDS - desired state of cluster Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html'
               type: string
             useClusterStorage:
               type: string
-            rebalancePods:
-              type: boolean
           required:
+          - alertFromAddress
```
matskiv, line 84: I don't think this field is intended to be mandatory.
This is currently causing failure in e2e test in CI:
```
"error":"Could not create rhmi CR in redhat-rhmi-operator namespace: RHMI.integreatly.org \"rhmi\" is invalid: spec.alertFromAddress: Required value"
```

matskiv:
This worked on the existing cluster (with and without `alertFromAddress` CR field).
New installation is however blocked by the error mentioned in my other comment^.

deploy/operator.yaml
```diff
@@ -54,3 +54,6 @@ spec:
               value: "managed"
             - name: REBALANCE_PODS
               value: "true"
+            # this should be set for production and development via MT repo
+            - name: ALERT_SMTP_FROM
+              value: "default@test.com"
```
Boomatang, line 59: If this is been set by the MT repo should the `ALERT_SMTP_FROM` be also added to the delorean configuration?

https://github.com/integr8ly/delorean/blob/master/configurations/managed-tenants-addons-config-rhoam.yaml

matskiv: Yes, I think so.
Good catch 👍

Makefile
```diff
@@ -52,6 +52,7 @@ export SELF_SIGNED_CERTS   ?= true
 # Setting the INSTALLATION_TYPE to managed-api will configure the values required for RHOAM installs
 export INSTALLATION_TYPE   ?= managed
 
+export ALERT_SMTP_FROM ?=no-reply-alert@devshift.org
```
MStokluska, line 55: Shouldn't the address be `noreply-alert@devshift.org`? That's what it is in MTR.

PR #1515, opened 2021-01-06
Mgdapi 995 cherry picked

Boomatang:
This PR has made changes to allow the work in #1511 pass the e2e test.
# Description

As part of this change we will need to add a new parameter to the managed tenants repo see PR 471

Verification

Spin up a cluster and run the operator locally overriding ALERT_SMTP_FROM and setting it to a value of your choice.

Once the install is complete, check the smtp config in alert manager

oc get secret alertmanager-application-monitoring -n redhat-rhoam-middleware-monitoring-operator -o=yaml

it should have the smtp_from set to your chosen email address.

## Type of change

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ export SELF_SIGNED_CERTS   ?= true
 # Setting the INSTALLATION_TYPE to managed-api will configure the values required for RHOAM installs
 export INSTALLATION_TYPE   ?= managed
 
-export ALERT_SMTP_FROM ?=no-reply-alert@devshift.org
+export ALERT_SMTP_FROM ?= noreply-alert@devshift.org
 export USE_CLUSTER_STORAGE ?= true
 export OPERATORS_IN_PRODUCT_NAMESPACE ?= false # e2e tests and createInstallationCR() need to be updated when default is changed
 export DELOREAN_PULL_SECRET_NAME ?= integreatly-delorean-pull-secret
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,7 @@ export SELF_SIGNED_CERTS   ?= true
 # Setting the INSTALLATION_TYPE to managed-api will configure the values required for RHOAM installs
 export INSTALLATION_TYPE   ?= managed
 
+export ALERT_SMTP_FROM ?=no-reply-alert@devshift.org
 export USE_CLUSTER_STORAGE ?= true
 export OPERATORS_IN_PRODUCT_NAMESPACE ?= false # e2e tests and createInstallationCR() need to be updated when default is changed
 export DELOREAN_PULL_SECRET_NAME ?= integreatly-delorean-pull-secret
--- a/pkg/controller/installation/installation_controller.go
+++ b/pkg/controller/installation/installation_controller.go
@@ -664,6 +664,12 @@ func (r *ReconcileInstallation) preflightChecks(installation *integreatlyv1alpha
 			_, err := strconv.ParseBool(s)
 			return err
 		}),
+		integreatlyv1alpha1.EnvKeyAlertSMTPFrom: requiredEnvVar(func(s string) error {
+			if s == "" {
+				return fmt.Errorf(" env var %s is required ", integreatlyv1alpha1.EnvKeyAlertSMTPFrom)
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -52,6 +52,7 @@ export SELF_SIGNED_CERTS   ?= true
 # Setting the INSTALLATION_TYPE to managed-api will configure the values required for RHOAM installs
 export INSTALLATION_TYPE   ?= managed
 
+export ALERT_SMTP_FROM ?=no-reply-alert@devshift.org
```
MStokluska, line 55: I think this should be `noreply-alert@devshift.org` - that's what it is in MTR

Boomatang: Yap I agree, I will make any change for that.

Makefile
```diff
@@ -52,6 +52,7 @@ export SELF_SIGNED_CERTS   ?= true
 # Setting the INSTALLATION_TYPE to managed-api will configure the values required for RHOAM installs
 export INSTALLATION_TYPE   ?= managed
 
+export ALERT_SMTP_FROM ?=noreply-alert@devshift.org
```
MStokluska, line 55:
```suggestion
export ALERT_SMTP_FROM ?= noreply-alert@devshift.org
```
Seems like it's passing the formatting anyway but maybe if we want to keep it in line with other envs?

Boomatang: Done, but just saying it would nicer if you had pointed that out in the last comment. Now the builds have been retriggered :(

MStokluska: Agree, didn't notice it at the first time

pkg/controller/installation/installation_controller.go
```diff
@@ -664,6 +664,12 @@ func (r *ReconcileInstallation) preflightChecks(installation *integreatlyv1alpha
 			_, err := strconv.ParseBool(s)
 			return err
 		}),
+		integreatlyv1alpha1.EnvKeyAlertSMTPFrom: requiredEnvVar(func(s string) error {
```
MStokluska, line 667: @Boomatang just more of a question than anything else, the requiredEnvVar returns an error also, and I wonder would it make sense to return the error from the requiredEnvVar instead, to me, and I might be missing something, we are creating a new error message here with just the addiotion of which envvar is missing - maybe editing the requiredEnvVar so that it would print out also which env var is missing, so that in future, the `checkEnvVars` function could be cleaner with more structured? way of error logging?

MStokluska: also, I think we have agreed recently that we could be making these slight improvement to the code as we go, on lines @174 - about 180 we are also checking for envvars, wonder if it would make sense to improve these as well to use the checkEnvVar function? WDYT?

matskiv: Because we are close to release cut date, and this PR is critical item for the release, I would rather put code improvements into a separate PR.

philbrookes: yes, I agree with @MStokluska in general, but in this case we should move that to a separate PR - or a Jira.

MStokluska:
Verified it on cluster, works as exptect, my SMTPFrom address could be found in alertmanager.yaml encrypted part. 
/lgtm

pkg/controller/installation/installation_controller.go
Boomatang: Ok I will create a follow Jira for this

philbrookes:
Do we have any idea how to verify this will function in production? This was never an issue in staging, right?

Boomatang:
@philbrookes While speaking with @alanmoran for verifying this on production we need to try it on production. Even using staging the domain for stage is the same a email domain we are going to be using.

alanmoran:
We could always stop the operator in production to stop reconcile, apply changes to the alertmanager template manually to include the new smtp_from address, restart alertmanager and test? 

While it doesn't test this PR, it will confirm that the new from address will work with the production sendgrid account. 

Please reach out to CS SRE to test this. We can use the Eng long lived cluster to validate this

philbrookes:
well, we need this merged either way, the issue with testing in prod is a separate issue to this

/approve

PR #1517, opened 2021-01-06
Remove keycloak route MGDAPI-1026

MStokluska:
JIRA: https://issues.redhat.com/browse/MGDAPI-1026

# Description
Currently, we have 2 routes for user-sso for RHOAM, one called `keycloak` is KCO generated route, and another one, `keycloak-edge`, is managed by RHOAM operator. Both routes use re-encrypt TLS termination policy and it makes no point in keeping two as only the edge route is being used.

# How
- KCO route has been disabled leaving only the `keycloak-edge` route available. 
- The name of the `keycloak-edge` route has been updated to `keycloak` since it is no longer using edge termination policy, and also since this work is preparing us for enabling user-sso rate-limiting in the future if it happens that we decide not to enable rate-limiting on user-sso, we can switch back to KCO managed routing without changing the host address.
- Unit test has been updated to test if the RHOAM full reconcile works as expected ( with the new route )
- E2E test has been updated to reflect changes
- Some test cases have been updated to match the new name for the route

# Verify
## RHOAM
- Install Managed-API from this branch
- Navigate to user sso ns and confirm that there's only 1 route present. 
- Check if user-sso can be accessed
- Confirm that RHOAM e2e test passed

## RHMI
- Install RHMI from this branch
- Navigate to user sso ns and confirm that there's 2 routes present (edge and keycloak route)
- Confirm that user-sso can be accessed via the edge route
- Confirm that RHMI e2e test passed on this PR

### DIFF
--- a/pkg/products/rhssocommon/reconciler.go
+++ b/pkg/products/rhssocommon/reconciler.go
@@ -175,18 +175,17 @@ func (r *Reconciler) CleanupKeycloakResources(ctx context.Context, inst *integre
 // this should use the same valid certs as the cluster itself but for some reason the
 // signing operator gives out self signed certs
 // to circumvent this we create another keycloak route with edge termination
-func (r *Reconciler) CreateKeycloakRoute(ctx context.Context, serverClient k8sclient.Client, config config.ConfigReadable, ssoCommon *config.RHSSOCommon) (integreatlyv1alpha1.StatusPhase, error) {
-	// We need a route with edge termination to serve the correct cluster certificate
-	edgeRoute := &routev1.Route{
+func (r *Reconciler) CreateKeycloakRoute(ctx context.Context, serverClient k8sclient.Client, config config.ConfigReadable, ssoCommon *config.RHSSOCommon, routeName string) (integreatlyv1alpha1.StatusPhase, error) {
+	keycloakRoute := &routev1.Route{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "keycloak-edge",
+			Name:      routeName,
 			Namespace: ssoCommon.GetNamespace(),
 		},
 	}
 
-	or, err := controllerutil.CreateOrUpdate(ctx, serverClient, edgeRoute, func() error {
-		host := edgeRoute.Spec.Host
-		edgeRoute.Spec = routev1.RouteSpec{
+	or, err := controllerutil.CreateOrUpdate(ctx, serverClient, keycloakRoute, func() error {
+		host := keycloakRoute.Spec.Host
+		keycloakRoute.Spec = routev1.RouteSpec{
 			Host: host,
 			To: routev1.RouteTargetReference{
 				Kind: "Service",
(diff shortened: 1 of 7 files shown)


pkg/products/rhssocommon/reconciler.go
```diff
@@ -175,11 +176,22 @@ func (r *Reconciler) CleanupKeycloakResources(ctx context.Context, inst *integre
...
-	// We need a route with edge termination to serve the correct cluster certificate
+func (r *Reconciler) CreateKeycloakRoute(ctx context.Context, serverClient k8sclient.Client, config config.ConfigReadable, ssoCommon *config.RHSSOCommon, installation *integreatlyv1alpha1.RHMI) (integreatlyv1alpha1.StatusPhase, error) {
+
+	// For now, edge termination is used in order to support future plans for implementing user-sso rate limiting for RHOAM.
+	var (
+		terminationType = routev1.TLSTerminationReencrypt
+		hostName        = "keycloak-edge"
+	)
+
+	if config.GetNamespace() == common.RHSSOUserProductOperatorNamespace && r.Installation.Spec.Type == string(integreatlyv1alpha1.InstallationTypeManagedApi) {
```
matskiv, line 187: Checking the 
`config.GetNamespace() == common.RHSSOUserProductOperatorNamespace`
seems like a bad pattern, maybe just make terminationType and hostName parameters of this function.

pkg/products/rhssocommon/reconciler.go
```diff
@@ -175,11 +176,22 @@ func (r *Reconciler) CleanupKeycloakResources(ctx context.Context, inst *integre
...
+		terminationType = routev1.TLSTerminationEdge
+		hostName = "user-sso"
```
matskiv, line 189: If we decide to rename it, I would remove `user-` prefix from the name,
as I heard from somebody(maybe docs), that we shouldn't refer to it as "User SSO" officially.

pkg/products/rhssocommon/reconciler.go
MStokluska: Thanks @matskiv, let me know if the recent change is better.

pkg/products/rhssocommon/reconciler.go
MStokluska: I've update it to `sso`, thanks @matskiv

pkg/products/rhssouser/reconciler.go
```diff
@@ -186,7 +187,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
-	phase, err = r.CreateKeycloakRoute(ctx, serverClient, r.Config, r.Config.RHSSOCommon)
+	// Setting a name for keycloak-edge to "sso" for managed-api install type
```
matskiv, line 190: Comment doesn't match code (`"sso"` part :) )
Also, describing _why_ that is being set might be useful.

MStokluska: Ah, good spot! Thanks a lot @matskiv

PR #1518, opened 2021-01-06
Added test case for 3scale custom policy

trepel:
# Description

Added a new test case, see https://issues.redhat.com/browse/MGDAPI-370. Hopefully we won't need to do this too often.

# Verification Steps
eye review only I guess. I had a few calls with 3scale QE and the comments in MGDAPI-370 were verified against RHOAM with 3scale 2.9.1.
But you can try yourself for sure.

### DIFF
--- a/test-cases/tests/products/h24-verify-selfmanaged-apicast-and-custom-policy.md
+++ b/test-cases/tests/products/h24-verify-selfmanaged-apicast-and-custom-policy.md
@@ -1,0 +1,198 @@
+---
+products:
+  - name: rhoam
+    environments:
+      - osd-fresh-install
+estimate: 2h
+tags:
+  - manual-selection
+---
+
+# H24 - Verify self-managed APIcast and custom policy
+
+## Description
+
+This test case should prove that it is possible for customers to deploy self-managed APIcast and use custom policies on it. The 3scale QE team will perform this test case in RHOAM each time there is an upgrade of 3scale. We (RHOAM QE) should only perform this if there are modifications on our end that might break the functionality - typically changes in permissions in RHOAM and/or OSD.
+Additional context can be found in [MGDAPI-370](https://issues.redhat.com/browse/MGDAPI-370)
+
+## Steps
+
+_1. This test case must be done using `customer-admin` user (or any other user from `dedicated-admin` group). Do not use `kubeadmin`!_
+_2. Create a new namespace (e.g. selfmanaged-apicast) for self-managed APIcast_
+
+- `oc project selfmanaged-apicast`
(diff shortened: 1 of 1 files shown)


test-cases/tests/products/h24-verify-selfmanaged-apicast-and-custom-policy.md
```diff
@@ -0,0 +1,196 @@
...
+with
+
+```
+    strategy:
+      type: Source
+      sourceStrategy:
+        from:
+          kind: ImageStreamTag
+          namespace: selfmanaged-apicast
+          name: apicast-gateway-rhel8:3scale2.9.1
```
psturc, line 66:
```suggestion
          name: apicast-gateway-rhel8:3scale<version>
```
would it be also worth mentioning how to find out the 3scale version?

test-cases/tests/products/h24-verify-selfmanaged-apicast-and-custom-policy.md
```diff
@@ -0,0 +1,196 @@
...
+
+_9. Start custom policy builds_
+
+- `oc start-build apicast-example-policy`
+- `oc start-build apicast-custom-policies`
+- once finished, the `image-registry.openshift-image-registry.svc:5000/selfmanaged-apicast/apicast-policy:example` should be available to be used as image for self-managed APIcast
+
+_10. Install "Red Hat Integration - 3scale APIcast gateway" operator via Operator Hub in OSD web console for the namespace_
+
+- do not sue community version!
```
psturc, line 141:
```suggestion
- do not use community version!
```

psturc (commented):
just a few comments/suggestions, otherwise LGTM

test-cases/tests/products/h24-verify-selfmanaged-apicast-and-custom-policy.md
trepel: good catch, will fix it

test-cases/tests/products/h24-verify-selfmanaged-apicast-and-custom-policy.md
trepel: apparently, this is not clear enough. that whole `3scale<version>` is an image tag. So you need to look in the catalog and see what tags are available and pick one - likely the one that corresponds with the 3scale version in RHOAM. Will improve the wording.

trepel:
@psturc Your comments have been addressed, would you mind taking another look?

I added two bullet points to the step 5 about the tag so it is more clear hopefully.

PR #1519, opened 2021-01-06
Update prometheus reconcile rules so not always true

austincunningham:
# Description

issue:https://issues.redhat.com/browse/MGDAPI-1017
remove RHOAMInstallationControllerIsNotReconciling and RHOAMInstallationControllerStoppedReconciling as these alerts constantly trigger and replace with RHOAMInstallationControllerIsInReconcilingErrorState

## Type of change

### Verification
- change the operator.yaml image to point at this quay.io/austincunningham/integreatly-operator:v2.7.0 and deploy to redhat-rhmi-operator namespace, use a managed-api cr to deploy RHOAM

or you can look at it deployed here https://console-openshift-console.apps.mw-collab-byoc.sskm.s1.devshift.org/
Verify if  RHOAMInstallationControllerIsNotReconciling and RHOAMInstallationControllerStoppedReconciling are removed and replace with RHOAMInstallationControllerIsInReconcilingErrorState and is present and not firing:

- Find route for Prometheus in redhat-rhmi-middleware-monitoring-operator namespace
- Open its URL
- Go to the Alerts tab and look for RHOAMInstallationControllerIsInReconcilingErrorState
- Click on the expr
- In the query page on the execute button 
- Verify if the rhoam_status metric no data is returned

## Checklist


test-cases/tests/alerts/c13a-verify-rhmi-reconciliation-alerts-are-not-firing.md
```diff
@@ -0,0 +1,27 @@
+---
+products:
+  - name: rhmi
+    environments:
+      - osd-post-upgrade
+    targets:
+      - 2.5.0
```
laurafitzgerald, line 7: should it be all releases?

test-cases/tests/alerts/c13b-verify-rhoam-reconciliation-alerts-are-not-firing.md
```diff
@@ -0,0 +1,30 @@
+---
+products:
+  - name: rhoam
+    environments:
+      - osd-post-upgrade
+      - osd-fresh-install
+    targets:
```
laurafitzgerald, line 7: @psturc do you think this one should be per-release?

psturc: This test case now says that it should `Verify RHOAMnstallationControllerIsInReconcilingErrorState alerts are present and are not firing`

I think that the first part (check that alert is present) should be covered by [the automated test](https://github.com/integr8ly/integreatly-operator/pull/1519/files#diff-91e53f30685bc5065b0c3501fc4c08dc7a2ed0778429cdea3c77d04d9bbd330bR248)
The second part (alert is not firing) is already covered by [this test case](https://github.com/integr8ly/integreatly-operator/blob/fcc80cc5080db8a8a9e3d022f755e7ae66ab5ace/test-cases/tests/alerts/c05b-verify-alerts-are-exposed-and-no-critical-alerts-have-fired.md)

So maybe c13b test case is not needed anymore in the end?

laurafitzgerald: Even better! The alert is also changed for RHMI so perhaps we could remove both test cases?

psturc: I believe so

austincunningham: this is a [warning level alert](https://github.com/integr8ly/integreatly-operator/pull/1519/files#diff-7d10ed51e49fd1fef8a5c0a2f1074fd700809f789237a1dd39cfe136e5655d4fL45) so don't think it will be covered by [this test case](https://github.com/integr8ly/integreatly-operator/blob/fcc80cc5080db8a8a9e3d022f755e7ae66ab5ace/test-cases/tests/alerts/c05b-verify-alerts-are-exposed-and-no-critical-alerts-have-fired.md)

pkg/controller/installation/prometheusRules.go
```diff
@@ -35,25 +35,15 @@ func (r *ReconcileInstallation) newAlertsReconciler(logger *logrus.Entry, instal
...
 					{
-						Alert: fmt.Sprintf("%sInstallationControllerIsNotReconciling", strings.ToUpper(installationName)),
+						Alert: fmt.Sprintf("%sInstallationControllerIsInReconcilingErrorState", strings.ToUpper(installationName)),
 						Annotations: map[string]string{
 							"sop_url": resources.SopUrlAlertsAndTroubleshooting,
-							"message": fmt.Sprintf("%s operator has not reconciled successfully in the interval of 15m over the past 1 hour", strings.ToUpper(installationName)),
+							"message": fmt.Sprintf("%s operator has finished installing, but has been in a error state while reconciling in the interval of 5m over the past 1 hour", strings.ToUpper(installationName)),
 						},
-						Expr:   intstr.FromString(fmt.Sprintf("%s_status{stage='complete'} AND on(namespace) rate(controller_runtime_reconcile_total{controller='installation-controller', result='success'}[15m]) == 0", installationName)),
+						Expr:   intstr.FromString(fmt.Sprintf("%s_status{stage='complete'} AND on(namespace) rate(controller_runtime_reconcile_total{controller='installation-controller', result='error'}[5m]) >= 1", installationName)),
```
laurafitzgerald, line 43: Hey @austincunningham @pepedocs I was thinking about this query a little more. The check is whether in the last hour the expression is satisfied over 5 minutes. If it happened in the last hour then it could be the case that it's resolved itself and the alert notification wouldn't be of much value at that stage. 
The alert could be simplified to `rhoam_status{stage='complete'} AND on(namespace) controller_runtime_reconcile_total{controller='installation-controller', result='error'} >=1` and the for entry changed to `For: 5m`, that way the alert would fire at the time the problem is occuring. 
The message would then be `operator has finished installing, but has been in a error state for the last 5 minutes` 
wdyt?

austincunningham: Agree [changing](https://github.com/integr8ly/integreatly-operator/pull/1519/files#diff-7d10ed51e49fd1fef8a5c0a2f1074fd700809f789237a1dd39cfe136e5655d4fL44) to 5m and removing the rate change on the metrics expression

test-cases/tests/alerts/c13b-verify-rhoam-reconciliation-alerts-are-not-firing.md
austincunningham: actually having read the test case , only the title mentions critical alerts, the test case checks for any alert firing , happy to remove these

pkg/controller/installation/prometheusRules.go
austincunningham: When removing the rate not sure if >=1 makes sense as these metrics `controller_runtime_reconcile_total{controller="installation-controller"}` continuously rise when we don't use rate
image making it difficult to trigger these alerts on a specific value

think we should reintroduce rate 5m and check the alert every 10m WDYT @laurafitzgerald ?

austincunningham: Also think `>0` might be better to trigger on, having looked at the rates changes above they don't exceed 1 ever.

laurafitzgerald:
/approve
@pepedocs if you could take a look and `/lgtm` if you are good with it now.

Nanyte25:
@austincunningham and myself have verified this alert today 
Screenshot 2021-01-20 at 13 49 40


/lgtm

PR #1521, opened 2021-01-08
MGDAPI-582

laurafitzgerald:
# Description

https://issues.redhat.com/browse/MGDAPI-582

bump version of cro to include defaulting of cidr block if no param i…s provided to the config map

remove logic that was defaulting in integreatly

## Verification

Scenarios:
1) Verify fresh install by add on with cidr block passed -> the passed cidr block should be used by cro
2) Verify fresh install by add on with cidr block not passed -> cro should search for a default and find a suitable one and create a vpc
3) Verify upgrade -> the upgrade should succeed and no change to an existing vpc setup should occur.

**Steps taken to verify 1**

- install the add on passing a cidr value of 10.3.0.0/26
- update the image to `quay.io/laurafitzgerald/managed-api-service:v1.0.0` before rhoam pod is running
- Verify that secret `addon-managed-api-service-parameters` contains the cidr value
Screenshot 2021-01-08 at 15 14 19
- Verify that that is populated to the `cloud-resources-aws-strategies` config map
Screenshot 2021-01-08 at 15 14 03
- Verify that this entry is used to create the vpc in aws.
Screenshot 2021-01-08 at 15 15 29

**Steps taken to verify 2** 

`INSTALLATION_TYPE=managed-api USE_CLUSTER_STORAGE=false make cluster/prepare/local`
`INSTALLATION_TYPE=managed-api USE_CLUSTER_STORAGE=false make code/run/service_account`

Verify `cloud-resources-aws-strategies` config map does not contain a default entry `{ "production": { "createStrategy": {} } }`
Verify that the vpc in aws has been created with a sensible default. 
Verify that the rds is created inside that vpc.

**Steps to verify 3**

- Install 1.0.1
- Do an upgrade -> As cro doesn't recreate any networking resources there should be no change to aws networking

## Type of change

## Checklist


deploy/cro-aws-config.yml
```diff
@@ -4,6 +4,6 @@ metadata:
   name: cloud-resources-aws-strategies
 data:
   _network: >-
-    { "production": { "createStrategy": { "CidrBlock": "10.1.0.0/23" } } }
+    { "production": { "createStrategy": {} } }
```
laurafitzgerald, line 7: i'm removing this because cro should now default. this was included previously to allow for development in a non add-on scenario. For dev purposes, we should allow cro to use it's default logic. A dev can specify an entry here if they wish.

pkg/products/cloudresources/reconciler.go
```diff
@@ -408,10 +406,8 @@ func (r *Reconciler) reconcileCIDRValue(ctx context.Context, client k8sclient.Cl
 		return err
 	}
 
-	//!ok means the param wasn't found so we want to default rather than return
-	//but don't do it until the installation object is more than a minute old in case the secret is slow to create
-	if !ok || cidrValue == "" && r.installation.ObjectMeta.CreationTimestamp.Time.Before(time.Now().Add(-(1*time.Minute))) {
-		cidrValue = "10.1.0.0/16"
+	if !ok {
```
laurafitzgerald, line 409: This change reverts https://github.com/integr8ly/integreatly-operator/pull/1387

R-Lawton:
/lgtm
verified with Laura the following scenarios 
Verify fresh install by add on with cidr block passed -> the passed cidr block should be used by cro
Verify fresh install by add on with cidr block not passed -> cro should search for a default and find a suitable one and create a vpc 
upgrade will be tested during testing

PR #1524, opened 2021-01-08
Add test case for cidr range

psturc:
# Description
Built on top of https://github.com/integr8ly/integreatly-operator/pull/1504 (that is waiting to be merged)

Commit to review: https://github.com/integr8ly/integreatly-operator/commit/bb22482ac811e4396a74f8173d02fd462e6d5a4c

### DIFF
--- a/test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
+++ b/test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
@@ -9,6 +9,7 @@ products:
       - osd-fresh-install
     targets:
       - 1.0.0
+      - 1.0.1
 ---
 
 # A30 - Validate installation of RHOAM addon and integration with LDAP IDP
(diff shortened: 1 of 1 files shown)


test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
```diff
@@ -57,7 +59,7 @@ Host prefix: /26
 3. Fill in the following parameters and click on `Install`
 
 ```
-CIDR range: "10.1.0.0/24"
+CIDR range: "10.1.0.0/24" (note this down to use it later for another verification step)
```
laurafitzgerald, line 62: Since this was written we've decided to use a /26 as default. It doesn't really change the verification but if it doesn't delay things could we update that also here?

test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
```diff
@@ -86,6 +88,24 @@ watch "oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq .status.stage"
...
+AWS_REGION=$(oc get infrastructure cluster -o jsonpath='{.status.platformStatus.aws.region}')
+```
+
+2. Run following AWS command
+
+```bash
+aws ec2 describe-vpcs --filters "Name=tag-key,Values=integreatly.org/clusterID" --region $AWS_REGION | jq -r '.Vpcs[0].CidrBlockAssociationSet[0].CidrBlock'
+```
+
+> Verify that the CIDR block you get on the output matches with the one that was specified in the installation form (iva OCM UI) ^
```
laurafitzgerald, line 107: nit
```suggestion
> Verify that the CIDR block you get on the output matches with the one that was specified in the installation form (via OCM UI) ^
```

test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
```diff
@@ -9,6 +9,7 @@ products:
       - osd-fresh-install
     targets:
       - 1.0.0
+      - 1.0.1
```
laurafitzgerald, line 12: Should this also include 1.1.0?

test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
```diff
@@ -236,7 +256,39 @@ Verify the integration with the LDAP server for a regular user
 
    > You should be redirected to the 3scale main page.
 
-As the test is finished lets stop the EC2 instance.
+**Verify RHOAM uninstall**
```
laurafitzgerald, line 259: Is this to verify that on a cluster that's had rhoam uninstalled that if there is a reinstall later the logic for the cidr block still works as if it were a new one?

laurafitzgerald (commented):
@psturc thanks for adding this. I have just a couple of small comments/questions inline.

test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
psturc: We're not checking if CIDR block logic works correctly after reinstallation. At the moment, the test case for reinstallation is there to check that after reinstall, all required secrets (dms, smtp, pagerduty) are created (there was an issue with it before) and we're not waiting for installation to complete.

Do you think it would be worth to add the CIDR check to the reinstallation test case as well?

test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
psturc: sure, will do

test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
psturc: good spot 👍

test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
psturc: thanks for pointing that out, I will actually update this test case to be automatically included every release

test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
laurafitzgerald: No i think we should be fine there. Just thought this was connected to the cidr change as it was added in the same pr.

test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
```diff
@@ -86,6 +87,24 @@ watch "oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq .status.stage"
 7. Go to `Monitoring` section of your cluster in OCM UI
    > Make sure no alerts are firing
 
+**Verify custom RHOAM CIDR range was applied correctly**
```
laurafitzgerald, line 90: Could we also write a case where there is no cidr value passed by the add on params that cro does provide a sensible default and the installation was successful?
It would be the same steps as this one except for not passing the cidr param and when checking in aws the cidr block may vary but should be within certain ranges listed here: https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Subnets.html#vpc-sizing-ipv4 and should not overlap with the cluster machine cidr. If it does overlap then cro would error out. If it doesn't find an option then cro would error out. so validation would be that there are no errors from cro.

psturc: This scenario should be covered by installation pipeline - it [doesn't provide any addon params](https://github.com/integr8ly/delorean/blob/0cd8e05a49540c0c505c3c291629dd737d7cc818/scripts/ocm/ocm.sh#L144), so CIDR block has to be created by CRO. If the pipeline finishes successfully, it means that the CIDR block was correct. So I will add a note to this test case that this scenario is already covered by pipeline.

PR #1543, opened 2021-01-18
MGDAPI-1004 Update AddOn Delete Watch

CathalOConnorRH:
# Description
Add the ability for hive to remove a cluster using a labeled configmap. 

To Verify:
1. Install RHOAM/RHMI from this branch
2. Once install is complete, add a config map in the integreatly operator namespace called `managed-service-api` / `rhmi` 
3. Label the config map with `api.openshift.com/addon-managed-api-service-delete=true`/ `api.openshift.com/addon-rhmi-operator-delete=true`
4. once the configmap is read the rhmi cr should have a deletion timestamp and the uninstall process should start. 

## Type of change

## Checklist

### DIFF
--- a/pkg/controller/namespacelabel/namespacelabel_controller.go
+++ b/pkg/controller/namespacelabel/namespacelabel_controller.go
@@ -50,6 +50,8 @@ import (
 var (
 	log           = l.NewLoggerWithContext(l.Fields{l.ControllerLogContext: "namespacelabel_controller"})
 	configMapName = "cloud-resources-aws-strategies"
+	//configMap name derived from the addons metadata id
+	deletionConfigMap = "managed-api-service"
 )
 
 //  patchStringValue specifies a patch operation for a string.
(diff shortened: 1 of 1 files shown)


Boomatang:
I have tried this with both fully installed version of RHMI & RHOAM and while both version were still in the install process. Each time it worked as expected.
/lgtm

pkg/controller/namespacelabel/namespacelabel_controller.go
```diff
@@ -130,6 +143,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch the SKU rate limits config map
```
philbrookes, line 146: Why is this watch on the SKU configmap added?

CathalOConnorRH: That was a copy paste mistake on the comment. I've update it now.

PR #1545, opened 2021-01-19
MGDAPI-700 Add Request and Limit resources to the statsd-exporter pod.

Boomatang:
# Description

Add Request and Limit resources to the statsd-exporter pod
Resources are higher than required to allow for initial spike in resources usages during start up.
*JIRA* https://issues.redhat.com/browse/MGDAPI-700

*Verify*
This is hard to Verify.
- On install the `prom-statsd-exporter` pod in the `redhat-rhoam-main3r` namespace should start correctly.
- Destroying and recreating the pod should not go into any crash off loops
- Metrics should show in grafana after about 3 minutes from pod creation
- Cluster metrics from a load test run shows the pod is does not increase under system load. https://snapshot.raintank.io/dashboard/snapshot/zPx50Zdzpn3tl0TrF1bGWHl27abgdChB?orgId=2 You are wanting to look at the charts for the `prom-statsd-exporter-6bf4478bdf-c58lq` pod. Note these charts do not have the initial spike in resource usage.

## Type of change

## Checklist

### DIFF
--- a/pkg/products/marin3r/reconciler.go
+++ b/pkg/products/marin3r/reconciler.go
@@ -4,9 +4,7 @@ import (
 	"context"
 	"fmt"
 	l "github.com/integr8ly/integreatly-operator/pkg/resources/logger"
-
 	"github.com/pkg/errors"
-
 	"strconv"
 
 	prometheus "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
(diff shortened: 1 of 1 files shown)


laurafitzgerald:
Verification 

- [ ] On install the prom-statsd-exporter pod in the redhat-rhoam-main3r namespace should start correctly.
- [ ] Destroying and recreating the pod should not go into any crash off loops
- [ ] Metrics should show in grafana after about 3 minutes from pod creation
- [ ] Cluster metrics from a load test run shows the pod is does not increase under system load. https://snapshot.raintank.io/dashboard/snapshot/zPx50Zdzpn3tl0TrF1bGWHl27abgdChB?orgId=2 You are wanting to look at the charts for the prom-statsd-exporter-6bf4478bdf-c58lq pod. Note these charts do not have the initial spike in resource usage.

matskiv:
When the pod is (re)created, we should also check that there is no restart - the container comes up successfully on the first try.

As for requests x limits, having them the same is better because we get `Guaranteed` QoS class. We probably should discuss in a wider group what criteria of the pod make sense for `Guaranteed` QoS, and what can be `Burstable` QoS. But in this case it seems that in absolute numbers the difference between "baseline" load and spikes is so low(it's like 10-15 MB) that we can afford to set `requests` to the numbers that actually represent limits.

PR #1548, opened 2021-01-20
Prepare EC2 for Hyperfoil install

trepel:
# Description

- Improved waiting for EC2 to become ready.
- Adding IP address from where there will be allowed full access to EC2 instance
- Adding rule to allow full access from localhost. Useful for Hyperfoil since both agent and controller might be hosted on the same EC2 instance (for now) and they use some non-standard ports and protocols (jgroups)
- Not using Elastic IP for publiIP of EC2 instance. It's not required and since there is a limit of just 5 Elastic IPs on AWS level it better not to use it unless really needed

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

## Verification steps
I used this script to create EC2 instance using 601144055808 AWS account, log in and have a look (sorry, removed already) :
https://console.aws.amazon.com/ec2/v2/home?region=us-east-1#InstanceDetails:instanceId=i-0b8c1403c97faf483

See it's security group and its security group's ingress rules.

To see the output of the script, see the console log of
https://master-jenkins-csb-intly.cloud.paas.psi.redhat.com/job/Public/job/trepel-ec2-deploy/43

**Note** This PR is not for improving of error handling of the script, there is a separate ticket in the backlog for that.

### DIFF
--- a/scripts/ec2_deploy_teardown/deployEC2.sh
+++ b/scripts/ec2_deploy_teardown/deployEC2.sh
@@ -35,6 +35,16 @@ for (( i=0; i<5; i++ )) do
     sleep 5
 done
 
+# For Hyperfoil
+if [[ "${INSTALL_HYPERFOIL}x" == "truex" ]]; then
+    echo "Setting up ingress rules required for Hyperfoil"
+    if [[ "${FULL_ACCESS_IP}x" != "x" ]]; then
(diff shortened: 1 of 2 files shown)


matskiv:
Works as expected. I am glad that we were able to implement this without elastic IPs :) 
/lgtm
/approve

trepel:
@matskiv right, that elasticIP removal was a great suggestion, thank you for the review.

PR #1551, opened 2021-01-21
New console link for grafana customer facing

briangallagher:
# Description
Install RHOAM
Verify that there are now 3 console links are shown in Tiffanys image in: [MGDAPI-544](https://issues.redhat.com/browse/MGDAPI-544)


laurafitzgerald:
👀

pkg/products/grafana/reconciler.go
```diff
@@ -167,6 +176,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
+	if err := r.reconcileConsoleLink(ctx, client, installation); err != nil {
```
laurafitzgerald, line 179: I see we check inside the function if install type is not managed-api and call a delete rather than create it. I can't think of a case where it would be present for an rhmi installation. There might be one I'm not thinking of.

pkg/products/grafana/reconciler.go
```diff
@@ -542,3 +555,53 @@ func buildGrafanaDashboardStrings(softLimits []uint32) (string, string) {
 
 	return graphQueries, dashboardVariables
 }
+
+func (r *Reconciler) reconcileConsoleLink(ctx context.Context, serverClient k8sclient.Client, installation *integreatlyv1alpha1.RHMI) error {
```
laurafitzgerald, line 559: It's being done in multiple locations in this file so feel free to ignore but the installation object is available in the the Reconciler so could be a bit neater to use it rather than passing. Feel free to ignore tho as it's done elsewhere.

laurafitzgerald (commented):
Small few minor comments. Overall looks good. Installing a cluster to verify.

laurafitzgerald:
I also did a verification of the change. The link is showing with the correct image etc. There's only two because of cluster size the instllation didn't complete but happy that this new change for the grafana dashboard link is working as expected. The console link cr is in place also. See image below

Screenshot 2021-01-22 at 17 02 34

pkg/products/grafana/reconciler.go
briangallagher: removed delete

pkg/products/grafana/reconciler.go
briangallagher: removed installation param.

PR #1552, opened 2021-01-21
MDGAPI-1098 - Retry to install product operators if failed install plan

KevFan:
# Description

Install plans can fail for operators that are installed in multiples namespaces such as keycloak. This looks to be due to transient errors / race condition by OLM. As this issue looks to be external to RHOAM and a flaky issue, we should have a way to try re-install the stuck products 

Logged OLM bugzilla ticket that will be investigated in their future sprint
* https://bugzilla.redhat.com/show_bug.cgi?id=1923111

Jira:
* https://issues.redhat.com/browse/MGDAPI-1098

## Verification
* Checkout this branch
* Install RHOAM
```
make cluster/prepare/local
make code/run
```
* Once RHSSO is installed, run script to update keycloak CRD to try reproduce install plan failure for User SSO
```
 while true; do                                                                                  
        oc patch crd keycloakusers.keycloak.org  --type=json -p='[{"op" : "add", "path" : "/metadata/labels/test", "value": "test"}]'
        oc patch crd keycloakusers.keycloak.org  --type=json -p='[{"op" : "remove", "path" : "/metadata/labels/test"}]'
done
```
* Check for failed install plan for user sso
  * If no failed install plan, uninstall the keycloak operator in user sso to re-trigger installation and repeat till failure
image
* Stop the script that updates keycloak CRD
* Verify RHOAM will delete the subscription to retry the install (following log lines should appear in log)
```
WARN[2021-02-01T20:07:58Z] Deleting csv for re-install due to failed install plan  csv=keycloak-operator.v11.0.3 install plan=rhmi-rhsso ns=redhat-rhoam-user-sso-operator product=rhssouser
WARN[2021-02-01T20:07:58Z] Deleting subscription for re-install due to failed install plan  install plan=rhmi-rhsso ns=redhat-rhoam-user-sso-operator product=rhssouser
```
* Verifiy install completes successfully
## Type of change

## Checklist

### DIFF
--- a/pkg/resources/reconciler.go
+++ b/pkg/resources/reconciler.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-
 	productsConfig "github.com/integr8ly/integreatly-operator/pkg/config"
 
 	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/pkg/apis/integreatly/v1alpha1"
@@ -200,7 +199,7 @@ func (r *Reconciler) ReconcileSubscription(ctx context.Context, target marketpla
 	if err != nil && !k8serr.IsAlreadyExists(err) {
 		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("could not create subscription in namespace: %s: %w", target.Namespace, err)
 	}
-	ips, _, err := r.mpm.GetSubscriptionInstallPlans(ctx, client, target.Pkg, target.Namespace)
+	ips, sub, err := r.mpm.GetSubscriptionInstallPlans(ctx, client, target.Pkg, target.Namespace)
 	if err != nil {
 		// this could be the install plan or subscription so need to check if sub nil or not TODO refactor
 		if k8serr.IsNotFound(err) || k8serr.IsNotFound(errors.Unwrap(err)) {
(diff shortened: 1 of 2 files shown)


laurafitzgerald:
@KevFan in terms of fully verifying this one I think it would be good to reproduce the issue on a cluster using 1.1.0 and then run this version against that cluster to see that it resolves. wdyt?

KevFan:
@laurafitzgerald Yeah, that would be ideal, just I'm not how to exactly reproduce this on a cluster :thinking:

pkg/resources/reconciler.go
```diff
@@ -229,6 +235,11 @@ func (r *Reconciler) ReconcileSubscription(ctx context.Context, target marketpla
 		}
 	}
 
+	// If all install plans failed - operator is not installed / upgraded
+	if allIpsFailed {
```
maleck13, line 239: is it possible we introduce a timing issue here? Could we see all IPs markes as failed before OLM creates a new one?

maleck13: cc @philbrookes

philbrookes: this will not abort the install though, if OLM creates another installplan after this returns, it will be picked up on the next reconcile - or have I misread this @KevFan

KevFan: @philbrookes Yes, correct, installation will continue if another install plan is created and does not fail. Though I'm no longer sure will this resolve the issue as I had an install where only one install plan was created even when it failed :thinking: Possibly more of a timing issue

pkg/resources/reconciler.go
```diff
@@ -219,6 +218,24 @@ func (r *Reconciler) ReconcileSubscription(ctx context.Context, target marketpla
...
 		}
 
+		// Workaround to re-install product operator if install plan fails due to https://bugzilla.redhat.com/show_bug.cgi?id=1923111
+		if ip.Status.Phase == operatorsv1alpha1.InstallPlanPhaseFailed {
+			if sub.Status.InstalledCSV != "" {
+				log.Warningf("Deleting csv for re-install due to failed install plan", l.Fields{"ns": target.Namespace, "install plan": target.Pkg, "csv": sub.Status.InstalledCSV})
+				csv := &operatorsv1alpha1.ClusterServiceVersion{
+					ObjectMeta: metav1.ObjectMeta{Namespace: target.Namespace, Name: sub.Status.InstalledCSV},
+				}
+				if err := client.Delete(ctx, csv); err != nil && k8serr.IsNotFound(err) {
```
maleck13, line 228: curious why we have to remove the CSV?

maleck13 (approved):
/lgtm code looks good and reasonable set of tests.

pkg/resources/reconciler.go
KevFan: Before adding this change to remove the csv also, if we remove the subscription without removing the installed csv, a new subscription would be created due but a install plan wont be created since the csv is already/still installed in the namespace preventing re-install from continuing

When the csv is deleted, the operator pod is deleted. Deleting the subscription only cleans up the install plans it seems. 
Uninstalling the operator via the ui looks to perform both these tasks of deleting subscription and csv in the namespace :thinking:

KevFan:
Thanks @maleck13 for the review. I just pushed a new change to fix a condition block and updated the unit test to reflect this if you want to take a look again

maleck13 (approved):
/lgtm

PR #1555, opened 2021-01-22
MGDAPI-557 - Try fix H03 test on nightly pipeline 

KevFan:
# Description

H03 e2e looks to fail consistently on nightly pipelines, but passes consistently when running locally. Going to try using new httpclient in threescale client and logging into 3scale to see does it help

Jira:
* https://issues.redhat.com/browse/MGDAPI-557

## Type of change

## Checklist

### DIFF
--- a/test/common/3scale_crudl.go
+++ b/test/common/3scale_crudl.go
@@ -37,10 +37,15 @@ func Test3ScaleCrudlPermissions(t *testing.T, ctx *TestingContext) {
 	keycloakHost := rhmi.Status.Stages[v1alpha1.AuthenticationStage].Products[v1alpha1.ProductRHSSO].Host
 	redirectUrl := fmt.Sprintf("%v/p/admin/dashboard", host)
 
-	tsClient := resources.NewThreeScaleAPIClient(host, keycloakHost, redirectUrl, ctx.HttpClient, ctx.Client, t)
+	httpClient, err := NewTestingHTTPClient(ctx.KubeConfig)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	tsClient := resources.NewThreeScaleAPIClient(host, keycloakHost, redirectUrl, httpClient, ctx.Client, t)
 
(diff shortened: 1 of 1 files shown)


test/common/3scale_crudl.go
```diff
@@ -37,17 +37,23 @@ func Test3ScaleCrudlPermissions(t *testing.T, ctx *TestingContext) {
...
-	err = loginToThreeScale(t, host, threescaleLoginUser, DefaultPassword, "testing-idp", ctx.HttpClient)
+	err = loginToThreeScale(t, host, threescaleLoginUser, DefaultPassword, "testing-idp", httpClient)
 	if err != nil {
 		dumpAuthResources(ctx.Client, t)
 		// t.Fatalf("[%s] error ocurred: %v", getTimeStampPrefix(), err)
 		t.Skipf("flakey test [%s] error ocurred: %v jira https://issues.redhat.com/browse/MGDAPI-557 ", getTimeStampPrefix(), err)
 	}
 
 	// Make sure 3Scale is available
+	t.Log("testing 3scale is available")
```
Boomatang, line 56: Would it make more sense to have the logs inside the `if err != nil` as there is no point in adding the logs if the stage passes. This would the same for other `t.Log()'s ` you have added.

Boomatang (commented):
I was looking at this and still created a second PR to add more logging LOL. How ever which ever get merged first I am happy. Left one commit on where the logs are added and this applie's to all the logs you have added. 

The pr I have added is #1560

test/common/3scale_crudl.go
KevFan: Yeah, makes sense. Updated :+1:

Boomatang:
@KevFan As we have seen this doesn't full fix the issue. I am wondering if we should close this PR and try some other unknown approach. What do you think?

KevFan:
@Boomatang Yeah, sound good to me, I'll close this PR so :+1:

PR #1561, opened 2021-01-26
[MGDAPI-1016] Add script to configure session expiration and token inactivity timeouts for OSD cluster

iankko:
# Description

    [MGDAPI-1016] Add script to configure OAuth access token
    (session) expiration and token inactivity timeouts for OSD
    cluster(s)
    
    Call it as part of scripts setting up the HTPasswd and SSO IDPs
    
    Signed-off-by: Jan Lieskovsky <jlieskov@redhat.com>

## Type of change

- [V ] Bug fix (non-breaking change which fixes an issue)
- [V] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Note: The change modifies the current behaviour of ```./scripts/setup-htpass-idp.sh``` and ```scripts/setup-sso-idp.sh` scripts to enhance them they to also configure session expiration and token inactivity timeouts for OSD cluster in question.

## Checklist
- [? ] This change requires a documentation update 
- [X ] I have added tests that prove my fix is effective or that my feature works
- [X ] I have added a test case that will be used to verify my changes 
- [X ] Verified independently on a cluster by reviewer

Note: Per previous agreement with @psturc the automated testcase will be created within a follow-up JIRA to this one.

@CathalOConnorRH @psturc @briangallagher PTAL once got a chance
Thanks!
Jan


psturc:
code lgtm, going to verify this against my cluster

scripts/refine-oauth-token-timeouts.sh
```diff
@@ -0,0 +1,209 @@
...
+# inactivity timeouts for all relevant OAuth clients (console, the default OAuth
+# clients, and those provisioned by a RHOAM instance)
+declare -ra OAUTH_CLIENTS=(
+  "console"
+  "openshift-browser-client"
+  "openshift-challenging-client"
+  "redhat-rhoam-3scale"
+  "redhat-rhoam-rhsso"
+  "redhat-rhoam-rhssouser"
+)
```
psturc, line 194: Could this be possibly parametrized?
Let's say we would want to configure the session and token expiration only for SSO IDP users (not kubeadmin and others)...

scripts/refine-oauth-token-timeouts.sh
```diff
@@ -0,0 +1,209 @@
...
+
+  # Enumerate supported resource types
+  declare -ra ALLOWED_RESOURCE_TYPES=( "oauth" "oauthclient" )
+
+  # OpenShift namespace to operate at
+  local NAMESPACE="openshift-config"
+
+  # Convert both resource type and name to lowercase
+  resourceType="${resourceType,,}"
+  resourceName="${resourceName,,}"
```
psturc, line 69: Unfortunately this also wouldn't work on Mac.. It can be replaced with
```suggestion
  resourceType=$(awk '{print tolower($0)}' <<< ${resourceType})
  resourceName=$(awk '{print tolower($0)}' <<< ${resourceName})
```

psturc (commented):
Just a couple of questions/suggestions

iankko:
@psturc Thanks for the review, Pavel! Right, valid points, will update the PR yet.

scripts/refine-oauth-token-timeouts.sh
```diff
@@ -0,0 +1,209 @@
...
+fi
+
+# If OpenShift version is v4 and greater set the token expiration timeout for
+# internal OAuth server.
+#
+# Note: Skip this step for OpenShift server of v3, since in that case the
+# timeouts need to be configured in the 'oauthConfig' section of the master
+# configuration file!
+readonly ocServerVersion=$(
+  oc version -o yaml | grep 'openshiftVersion' | grep -oP '((\d+\.){2}\d+)'
```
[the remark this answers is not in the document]
iankko: Thanks @psturc , replaced with BRE expression to avoid `awk` proces launch, fixed (will squash commits in the final version of the patch).

scripts/setup-htpass-idp.sh
```diff
@@ -10,18 +11,22 @@ if [[ ! -f "htpasswd" ]]; then
...
 oc patch oauth cluster --type=merge -p '{ "spec": { "identityProviders": [{ "name": "htpasswd_provider", "challenge": true, "login": true, "mappingMethod": "claim", "type": "HTPasswd", "htpasswd": { "fileData": { "name": "htpasswd-secret" } } }] } }'
 
 oc adm groups add-users dedicated-admins customer-admin
-oc adm groups add-users osd-sre-admins sre-user-1
\ No newline at end of file
+oc adm groups add-users osd-sre-admins sre-user-1
+
+# Also configure session expiration (OAuth access token duration) and token inactivity timeouts
+# as part of defining new HTPasswd identity provider for the cluster
+source refine-oauth-token-timeouts.sh "${TOKEN_INACTIVITY_TIMEOUT}" "${SESSION_EXPIRATION_TIMEOUT}"
```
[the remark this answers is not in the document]
iankko: Great catch!, thanks, fixed.

scripts/setup-sso-idp.sh
```diff
@@ -161,3 +161,8 @@ until ! oc get deployment oauth-openshift -n openshift-authentication -o yaml |
   echo "\"oauth-openshift\" deployment is still updating, trying again in 10s"
   sleep 10s
 done
+
+# Also configure session expiration (OAuth access token duration) and token inactivity timeouts
+# as part of defining new SSO identity provider for the cluster
+source refine-oauth-token-timeouts.sh "${TOKEN_INACTIVITY_TIMEOUT}" "${SESSION_EXPIRATION_TIMEOUT}"
```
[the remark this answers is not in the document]
iankko: Here too

scripts/refine-oauth-token-timeouts.sh
iankko: @psturc Quick clarification regarding this one yet - do you want all of them to be parametrizable? (no default clients to be updated)

Or should some of them by updated by default, but the final list would be created as a concatenation of the default ones plus those specified on the command line? If the latter, can you possibly suggest which of them should be the default ones?

scripts/refine-oauth-token-timeouts.sh
iankko: Fixed with call of `tr` to avoid time costs of `awk` process launch (not sure if it matters that much).

scripts/refine-oauth-token-timeouts.sh
psturc: These are good questions...

So I suppose this script will be used for both development and testing purposes.

From developer's POV I can imagine that having the session and token expiration the longest would probably make sense, i.e. if I run the SSO IDP script without any extra params, I would expect that the token and session expiration doesn't change at all. (i.e. 24 hours expiration for kubeadmin?)

But from testing POV (where we want to mimic the customer's environment as close as possible), I think that the typical configuration would include configuring the session and token expiration for RHOAM services, i.e. SSO, User-SSO and 3scale.

Else if I specify just the env vars for token and session timeouts, I would expect that all clients would be updated (by default).

So, if it is possible, I would see the final functionality like this:

```bash
# Don't configure token inactivity timeout nor session expiration timeout at all
./scripts/setup-sso-idp.sh

# Set token inactivity and session expiration timeouts for all clients
TOKEN_INACTIVITY_TIMEOUT=300 SESSION_EXPIRATION_TIMEOUT=3600 ./scripts/setup-sso-idp.sh

# Set token inactivity and session expiration timeouts only for 3scale, RHSSO and User SSO
TOKEN_INACTIVITY_TIMEOUT=300 SESSION_EXPIRATION_TIMEOUT=3600 OAUTH_CLIENTS=redhat-rhoam-3scale,redhat-rhoam-rhsso,redhat-rhoam-rhssouser ./scripts/setup-sso-idp.sh
```

Would it make sense?

iankko: > These are good questions...
> 
> So I suppose this script will be used for both development and testing purposes.

Right.

> 
> From developer's POV I can imagine that having the session and token expiration the longest would probably make sense, i.e. if I run the SSO IDP script without any extra params, I would expect that the token and session expiration doesn't change at all. (i.e. 24 hours expiration for kubeadmin?)

Sounds reasonable. Agree that in development environment, where the cluster management is utilized all the time practically, it would make sense to have a way to keep default OpenShift values (e.g. expiration timeout set to one day, and inactivity timeout not set at all), so the re-logins doesn't need to be issued that often.

> 
> But from testing POV (where we want to mimic the customer's environment as close as possible), I think that the typical configuration would include configuring the session and token expiration for RHOAM services, i.e. SSO, User-SSO and 3scale.

Right, ok.

> 
> Else if I specify just the env vars for token and session timeouts, I would expect that all clients would be updated (by default).

Ok.

> 
> So, if it is possible, I would see the final functionality like this:
> 
> ```shell
> # Don't configure token inactivity timeout nor session expiration timeout at all
> ./scripts/setup-sso-idp.sh
> 
> # Set token inactivity and session expiration timeouts for all clients
> TOKEN_INACTIVITY_TIMEOUT=300 SESSION_EXPIRATION_TIMEOUT=3600 ./scripts/setup-sso-idp.sh
> 
> # Set token inactivity and session expiration timeouts only for 3scale, RHSSO and User SSO
> TOKEN_INACTIVITY_TIMEOUT=300 SESSION_EXPIRATION_TIMEOUT=3600 OAUTH_CLIENTS=redhat-rhoam-3scale,redhat-rhoam-rhsso,redhat-rhoam-rhssouser ./scripts/setup-sso-idp.sh
> ```
> 
> Would it make sense?

Yes, sounds reasonable to me. Thanks for the clarification / sketch about different ways / environments the script could be used. Will mirror above expectations into the code yet.

iankko:
@psturc Yet, there's one thing with propagating the specified timeout values into the provisioned Keycloak / RH-SSO instance. The more I think about it, the more sense it makes to me. Currently the script modifies the timeouts just on the underlying OpenShift instance (for the internal OAuth server and various OAuth clients). But IMHO it should also propagate the token / session timeouts [to the following three Keycloak / RH-SSO token properties](https://www.keycloak.org/docs/latest/server_admin/#_timeouts):
* Specified expiration timeout should be used / set as a value for ```SSO Session Max``` setting (and possibly also for the ```Access Token Lifespan``` setting),
* In parallel, token inactivity timeout would be used to set the values of  ```SSO Session Idle`` setting.

This way, IMHO the script could be made a single point of control for both the OpenShift and RH-SSO instances, and configure session limits on both of them (if would configure the aforementioned RH-SSO fields / setting only in the case there's SSO IDP configured already of course).

WDYT?

scripts/refine-oauth-token-timeouts.sh
iankko: @psturc Above behaviour implemented in the last commit. PTAL once got a chance. Thanks

iankko:
The PR got updated to also propagate the specified timeouts to associated SSO IDP if configured. Should be ready for review (again).

@psturc PTAL once got a chance
Thanks!

iankko:
@maleck13 @matskiv @pawelpaszki @briangallagher @KevFan @ciaranRoche @ciaran-byrne Would be great / appreciated to get feedback on this proposal from you too. PTAL once got a chance (not urgent)

Thanks!
Jan

scripts/refine-oauth-token-timeouts.sh
```diff
@@ -0,0 +1,747 @@
...
+function process_curl_response() {
+  local response="${1?Please specify the curl response to process!}"
+  # Trim trailing whitespace from status code
+  # shellcheck disable=SC2155
+  local responseStatusCode=$(sed -n '1s/[[:space:]]\+$//;1p' <<< "${response}")
+  # Collect response content from the first empty line (excluded) till the end
+  # of the response as a reponse body
+  # shellcheck disable=SC2016
+  # shellcheck disable=SC2155
+  local responseBody=$(sed -n '/^.$/,${n;p}' <<< "${response}")
```
psturc, line 214:
```
About to issue the following curl request:
curl --data client_id=admin-cli --data grant_type=password --data username=admin --data-urlencode password=mIQJ4m4DfaUSpg== --include --show-error --silent https://keycloak-edge-redhat-rhoam-rhsso.<REDACTED>/auth/realms/master/protocol/openid-connect/token
sed: 1: "/^.$/,${n;p}": extra characters at the end of p command
```
got an error when running this on mac os

iankko: Right, will have a look / fix. Thanks, Pavel! (sorry, didn't try it on MacOSX).

iankko: @psturc This should be fixed now. Tried the script multiple times on both Fedora 33 & FreeBSD 12.2 (just to ensure `sed` calls syntax is also compliant with *BSD variants of sed), and works for me now on both systems.

Thanks for catching this!

iankko:
@psturc Will fix the conflicts above. 

But for the subcase of `Scenario 1` in the MGDAPI-1016 JIRA, found the following ``` logoutRedirect:``` setting yet:
* https://docs.openshift.com/container-platform/4.6/web_console/configuring-web-console.html
* https://docs.openshift.com/container-platform/4.6/rest_api/config_apis/console-config-openshift-io-v1.html

which could help to address it too. WDYT? Worthy to explore if it would be possible to automate setting of this field too?

psturc:
> @psturc Will fix the conflicts above.
> 
> But for the subcase of `Scenario 1` in the MGDAPI-1016 JIRA, found the following ` logoutRedirect:` setting yet:
> 
> * https://docs.openshift.com/container-platform/4.6/web_console/configuring-web-console.html
> * https://docs.openshift.com/container-platform/4.6/rest_api/config_apis/console-config-openshift-io-v1.html
> 
> which could help to address it too. WDYT? Worthy to explore if it would be possible to automate setting of this field too?

Yes I think it would be great if we could cover the `Scenario 1` with this fix as well 👍

PR #1564, opened 2021-01-28
MGDAPI-1159-add ns label to alerts

MStokluska:
# Description
Addresses https://issues.redhat.com/browse/MGDAPI-1159
Adding a namespace to labels of both alerts. 

# Verification 
Verified by running on stage cluster (pointing to my own image) with alert time reduced to 5 mins, SRE-P confirmed the alert triggered


pmccarthy:
As this is a simple change and has already been verified in stage with SRE-P, I'm happy to approve this.

/lgtm
/approve

PR #1565, opened 2021-01-28
3scale user action failure metric and alert

laurafitzgerald:
# Changes
expose new metric around user creation and deletion in 3scale
allow installation to complete when the user creation fails
add new alert that will fire when the user creation in 3scale fails
test updates

# Description
https://issues.redhat.com/browse/MGDAPI-1154
https://issues.redhat.com/browse/MGDAPI-1155
https://issues.redhat.com/browse/MGDAPI-1156

#### 1.1) Prepare cluster 
`INSTALLATION_TYPE=managed-api IN_PROW=true make cluster/prepare/local`

Create a User with the following yaml and `oc apply -f` you cannot create through the UI without configuring a OAuthClient or 
```
apiVersion: user.openshift.io/v1
kind: User
metadata:
  name: alongusernamethatisabovefourtycharacterslong
```
and
```
apiVersion: user.openshift.io/v1
kind: User
metadata:
  name: auserlessthanfourtycharacters
```

#### 1.2) Install in cluster to verify monitoring

Run the operator in cluster - local run won't provide the metrics
Update operator.yaml quay images with `quay.io/laurafitzgerald/managed-api-service:v1.1.0-1154`
update env INSTALLATION_TYPE='managed-api'

`oc apply -f deploy/operator.yaml`

Add this annotation to the rhmi cr once it's created to allow it to install on ocm stage cluster
```
annotations:
    in_prow: 'true'
```

#### 1.3) Verify that the alert is firing

Example shown below

Screenshot 2021-02-01 at 12 21 47

#### 1.4) Verify that the installation completes

After some time check that the rhmi cr goes to complete status and is not blocked by the failure to create the user in 3scale

You can also check that the user `auserlessthanfourtycharacters` is created in 3scale

Navigate to `https://3scale-admin.<your-cluster>/p/admin/account/users`
You should see `admin` an `auserlessthanfourtycharacters` but not `alongusernamethatisabovefourtycharacterslong` registered in 3scale.

#### 1.5) Verify deletion of users still reflects to 3scale as expected

Delete the user with `oc delete user auserlessthanfourtycharacters`

Check that the user is removed in the 3scale console link above.

### 2) Updating user email gets recreated in 3scale

The installation is complete from the previous step.

#### 2.1) Log in as admin user to 3scale
 
`https://3scale-admin.apps.<your-cluster>/p/admin/account/users`

#### 2.2) Edit any user and update the username

Wait for a while and the user will be set back.

## Type of change

## Checklist
- [x] This change requires a SOP update - we likely need to add a SOP for this

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -66,6 +66,7 @@ func init() {
 	customMetrics.Registry.MustRegister(integreatlymetrics.RHMIStatus)
 	customMetrics.Registry.MustRegister(integreatlymetrics.RHOAMVersion)
 	customMetrics.Registry.MustRegister(integreatlymetrics.RHOAMStatus)
+	customMetrics.Registry.MustRegister(integreatlymetrics.ThreeScaleUserAction)
 	integreatlymetrics.OperatorVersion.Add(1)
 }
 
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -90,6 +90,17 @@ var (
 			"stage",
 		},
 	)
+
+	ThreeScaleUserAction = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "threescale_user_action",
+			Help: "Status of user CRUD action in 3scale",
+		},
+		[]string{
+			"username",
+			"action",
+		},
+	)
 )
 
 // SetRHMIInfo exposes rhmi info metrics with labels from the installation CR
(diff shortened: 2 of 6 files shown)


philbrookes:
This looks good.
/lgtm

Boomatang:
I have verified this is on my cluster. Everything works as describe in the validation steps. There is one question over what should happen if the username letter casing changes but not the username.

The alerts are created for each user with a over length name. Below is a screen shot of the alerts been fired.
image

PR #1566, opened 2021-01-29
README Update

Boomatang:
Updated the command for running the single e2e test


Boomatang:
Updated got added in a different PR

PR #1570, opened 2021-02-02
Update to generate bundles - MGDAPI-1090

MStokluska:
# Description
Addresses https://issues.redhat.com/browse/MGDAPI-1090

NOTE: Please verify this PR with changes in https://github.com/integr8ly/integreatly-operator/pull/1605 as the bundle script used the release/prepare which is fixed in the linked PR.

# To verify
Go through the updates to readme and confirm that the commands generate correct bundles and indexes in your registry.

### DIFF
--- a/scripts/bundle-rhmi-operators.sh
+++ b/scripts/bundle-rhmi-operators.sh
@@ -7,11 +7,9 @@ fi
 case $OLM_TYPE in
   "integreatly-operator")
     LATEST_VERSION=$(grep $OLM_TYPE packagemanifests/$OLM_TYPE/$OLM_TYPE.package.yaml | awk -F v '{print $2}')
-    PACKAGE="integreatly"
     ;;
   "managed-api-service")
     LATEST_VERSION=$(grep $OLM_TYPE packagemanifests/$OLM_TYPE/$OLM_TYPE.package.yaml | awk -F v '{print $3}')
-    PACKAGE="managed-api-service"
     ;;
   *)
     echo "Invalid OLM_TYPE set"
@@ -28,16 +26,19 @@ UPGRADE_RHMI="${UPGRADE:-false}"
 VERSIONS="${BUNDLE_VERSIONS:-$LATEST_VERSION}"
 ROOT=$(pwd)
 INDEX_IMAGE=""
+CATALOG_SOURCE_INSTALL="${OC_INSTALL:-true}"
 
 
 start() {
-  # TODO: validate input
+  clean_up
   create_work_area
   copy_bundles
   check_upgrade_install
   generate_bundles
   generate_index
+  if [ "$CATALOG_SOURCE_INSTALL" = true ] ; then
   create_catalog_source
+  fi
   clean_up
 }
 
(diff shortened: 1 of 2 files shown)


scripts/bundle-rhmi-operators.sh
```diff
@@ -28,17 +26,20 @@ UPGRADE_RHMI="${UPGRADE:-false}"
...
   create_work_area
   copy_bundles
   check_upgrade_install
   generate_bundles
   generate_index
+  if [ "$CATALOGUE_SOURCE_INSTALL" = true ] ; then
   create_catalog_source
-  clean_up
+  fi
+  #clean_up
```
sergioifg94, line 42:
```suggestion
  clean_up
  fi
```

I like the approach of cleaning up in the first place in case the temporal folder was still there (very useful to look into it for debugging purposes), but I'd keep the clean up at the end in order to keep git clean. Other approach would be adding the temp folder to the `.gitignore`

scripts/bundle-rhmi-operators.sh
```diff
@@ -28,17 +26,20 @@ UPGRADE_RHMI="${UPGRADE:-false}"
 VERSIONS="${BUNDLE_VERSIONS:-$LATEST_VERSION}"
 ROOT=$(pwd)
 INDEX_IMAGE=""
+CATALOGUE_SOURCE_INSTALL="${OC_INSTALL:-true}"
```
sergioifg94, line 29:
```suggestion
CATALOG_SOURCE_INSTALL="${OC_INSTALL:-true}"
```

Just to align with the official name

scripts/bundle-rhmi-operators.sh
```diff
@@ -84,48 +84,57 @@ generate_bundles() {
...
 
-  INDEX_IMAGE=$REG/$ORG/${PACKAGE}-index:$NEWEST_VERSION
+  INDEX_IMAGE=$REG/$ORG/${OLM_TYPE}-index:$NEWEST_VERSION
 
   printf 'Pushing index image:'$INDEX_IMAGE'\n'
 
   docker push $INDEX_IMAGE
 }
 
+# creates catalogue source on the cluster
```
sergioifg94, line 129:
```suggestion
# creates catalog source on the cluster
```

Same as above

scripts/bundle-rhmi-operators.sh
```diff
@@ -56,24 +57,23 @@ copy_bundles() {
...
   fi
-
-  echo "versions:::: "$VERSIONS
   # Get the oldest version, example: VERSIONS="2.5,2.4,2.3" oldest="2.3"
   OLDEST_VERSION=${VERSIONS##*,}
 
-  printf "Removing replaces field from CSV \n"
-  file=`ls './'$OLDEST_VERSION | grep .clusterserviceversion.yaml`
-  sed '/replaces/d' './'$OLDEST_VERSION'/'$file > newfile ; mv newfile './'$OLDEST_VERSION'/'$file
+  file=("./$(echo ${OLDEST_VERSION} | sed "s/,/ /g")/$(echo ${OLM_TYPE} | sed "s/,/ /g").v$(echo ${OLDEST_VERSION} | sed "s/,/ /g").clusterserviceversion.yaml")
```
sergioifg94, line 71: This won't work with manifests generated with the new version of the operator-sdk, since the filename of the CSV now is  `managed-api-service.clusterserviceversion.yaml` (missing the release version that is assumed in this line).

To test this scenario, generate new manifests and run the script for that release:
```sh
SEMVER=1.3.0 OLM_TYPE=managed-api-service make release/prepare
OC_INSTALL=false ORG=<org> OLM_TYPE=managed-api-service BUNDLE_VERSIONS=1.3.0 make create/olm/bundle
```

It fails the validation as the replaces field is not removed. That's why I modified that line in the migration commit to use grep, so it works for both pre-migration and post-migration manifests.

sergioifg94 (commented):
Overall looks good. Just a few things to be addressed. I'd also suggest to rebase as #1605 was merged after the last rebase

scripts/bundle-rhmi-operators.sh
MStokluska: ah good spot, I've commented it out just to see how do the bundles look like, forgot to uncomment it! thank you!

scripts/bundle-rhmi-operators.sh
MStokluska: Makes sense, going to rollback the changes here! thanks @sergioifg94

sergioifg94:
Changes look good. Tested it with different combinations of manifest releases and works as expected
/lgtm

PR #1580, opened 2021-02-03
Fix time for instant queries in capture_resource_metrics.sh

matskiv:
# Description
Fixes a bug in the ./scripts/capture_resource_metrics.sh
Now it will run "instant queries" at a time when a run ended,
instead of running them at a time when script runs.
This is useful if you run the script much later after the run finish time.

Edit: I also fix one query (Allocatable RAM on workers)

## Type of change

## Checklist

### DIFF
--- a/scripts/capture_resource_metrics.sh
+++ b/scripts/capture_resource_metrics.sh
@@ -42,7 +42,7 @@ INSTANT_QUERIES=(\
   "sum(kube_node_status_allocatable_cpu_cores * on (node) (kube_node_role{role='worker'} == on (node) group_left () (count by (node) (kube_node_role{}))))"\
   "sum(cluster:capacity_memory_bytes:sum)/1024/1024/1024"\
   "sum(cluster:capacity_memory_bytes:sum{label_node_role_kubernetes_io!~'master|infra'})/1024/1024/1024"\
-  "kube_node_status_allocatable_memory_bytes * on (node) (kube_node_role{role='worker'} == on (node) group_left () (count by (node) (kube_node_role{}))) / 1024 / 1024 / 1024"\
+  "sum(kube_node_status_allocatable_memory_bytes * on (node) (kube_node_role{role='worker'} == on (node) group_left () (count by (node) (kube_node_role{}))) / 1024 / 1024 / 1024)"\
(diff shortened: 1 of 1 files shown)


scripts/capture_resource_metrics.sh
```diff
@@ -84,7 +84,7 @@ LOAD_QUERIES=(\
 #
 for query in "${INSTANT_QUERIES[@]}";
 do
-  curl -s -G -H "Authorization: Bearer $TOKEN" --data-urlencode "query=$query" -H 'Accept: application/json' $PROM_QUERY_ROUTE | jq -r ".data.result[0].value[1]"
+  curl -s -G -H "Authorization: Bearer $TOKEN" --data-urlencode "query=$query" --data-urlencode "time=$endTime" -H 'Accept: application/json' $PROM_QUERY_ROUTE | jq -r ".data.result[0].value[1]"
```
austincunningham, line 87: matches the second curl so I'm good with this change

scripts/capture_resource_metrics.sh
```diff
@@ -42,7 +42,7 @@ INSTANT_QUERIES=(\
   "sum(kube_node_status_allocatable_cpu_cores * on (node) (kube_node_role{role='worker'} == on (node) group_left () (count by (node) (kube_node_role{}))))"\
   "sum(cluster:capacity_memory_bytes:sum)/1024/1024/1024"\
   "sum(cluster:capacity_memory_bytes:sum{label_node_role_kubernetes_io!~'master|infra'})/1024/1024/1024"\
-  "kube_node_status_allocatable_memory_bytes * on (node) (kube_node_role{role='worker'} == on (node) group_left () (count by (node) (kube_node_role{}))) / 1024 / 1024 / 1024"\
+  "sum(kube_node_status_allocatable_memory_bytes * on (node) (kube_node_role{role='worker'} == on (node) group_left () (count by (node) (kube_node_role{}))) / 1024 / 1024 / 1024)"\
```
austincunningham, line 45: query looks good and verified
image

scripts/capture_resource_metrics.sh
austincunningham: /lgtm
/approve

austincunningham (approved):

PR #1596, opened 2021-02-15
add osde2e rhmi support

MStokluska:
# Description
Addresses: https://issues.redhat.com/browse/INTLY-10444

Enabled rhmi osde2e and changed the timeout on the alerts_mechanism tests as the RHMI installation takes longer, and in case if that test is flaky, we can't be giving it 40minutes before it fails as instead of failing single test, we might end up having an entire suite of tests timinig out.

### DIFF
--- a/test/osde2e/managed_api_test.go
+++ b/test/osde2e/managed_api_test.go
@@ -2,7 +2,6 @@ package osde2e
 
 import (
 	"fmt"
-	"os"
 
 	"github.com/integr8ly/integreatly-operator/test/common"
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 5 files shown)


jjkiely (commented):
Verified via osde2e repo on RHMI installation

test/osde2e/managed_api_test.go
```diff
@@ -21,47 +21,26 @@ var _ = Describe("integreatly", func() {
...
-		common.Marin3rProductNamespace = common.NamespacePrefix + "marin3r"
-		common.CustomerGrafanaNamespace = common.NamespacePrefix + "customer-monitoring-operator"
+		// Assign the installation name by WATCH_NAMESPACE - if no installation name has been found fail the test.
+		if watchNs == "redhat-rhmi-operator" {
+			installType = "managed"
+		} else if watchNs == "redhat-rhoam-operator" {
+			installType = "managed-api"
+		} else {
+			t.Fatalf("No installation name has been found")
+		}
```
Boomatang, line 33: I wonder would this be more readable as a switch statement. 
```suggestion
               switch watchNs {
                  case "redhat-rhmi-operator":
                     installType = "managed"
                  case "redhat-rhoam-operator":
                     installType = "managed-api"   
                  default:
                     t.Fatalf("No installation name has been found")
                   }
```
I am not to pushed on this and I know may indents may not be correct in the example.

test/osde2e/suite_test.go
```diff
@@ -86,3 +89,36 @@ var _ = AfterSuite(func() {
 	err := testEnv.Stop()
 	Expect(err).ToNot(HaveOccurred())
 })
+
+func setVars(possibleWN, possibleNS string) {
+	os.Setenv("WATCH_NAMESPACE", possibleWN)
+	os.Setenv("SKIP_FLAKES", "true")
```
Boomatang, line 95: `os.Setenv()` does return an error on failures. Do you think this should be some we can to be checking for. Even if its to log the error so we can debug at a later date

Boomatang (commented):
Few small comments, waiting for the osde2e to run so that we can see if this fixes the current issue seen there.

test/osde2e/suite_test.go
MStokluska: Thanks I've updated it locally, going to push changes when I verify it!

test/osde2e/managed_api_test.go
MStokluska: hmm, now that I think of it...we actually don't need this check. The `installType` is pulled from suite_test.go file here: https://github.com/integr8ly/integreatly-operator/blob/dbd8ceda04ad3123155186078f860f1ca1beeaaf/test/osde2e/suite_test.go#L35, then it gets a value and only then the managed-api-test is triggered. I think we should be ok with removing the entire logic around reading WATCH_NAMESPACE here but need to verify it first!

MStokluska: Just verified it, it works without reading the WATCH_NAMESPACE in the managed-api-test.go and looks much cleaner!

test/osde2e/suite_test.go
```diff
@@ -90,9 +90,15 @@ var _ = AfterSuite(func() {
...
-	os.Setenv("WATCH_NAMESPACE", possibleWN)
-	os.Setenv("SKIP_FLAKES", "true")
+func setVars(possibleWN, possibleNS string, t *testing.T) {
+	err := os.Setenv("WATCH_NAMESPACE", possibleWN)
+	if err != nil {
+		t.Fatalf("Failed to set WATCH_NAMESPACE env var %s", err)
+	}
+	err = os.Setenv("SKIP_FLAKES", "true")
+	if err != nil {
+		t.Fatalf("Failed to set SKIP_FLAKES env var %s", err)
```
Boomatang, line 100: I don't think not setting the `SKIP_FLAKES` should be a fatal error. This would case the test that should be skipped to run but the operator can still be installed. My view would be to log that the `SKIP_FLAKES` did not get set but continue to run the test. You never know the flaky test might pass in the end.

MStokluska: Done, good idea!

PR #1603, opened 2021-02-19
MGDAPI-1289: Modify watches in namespace controller to enqueue correct requests

sergioifg94:
# Description

The namespace controller watches Namespaces and ConfigMaps, which will enqueue different requests since the namespaces have their name in theresource name, while the ConfigMap is namespaced. In order to ensure uniform requests, add an EventHandler for the ConfigMap watch that enqueues only its namespace, and add predicates for both types of resources to watch exclusivelly what we're interested in: the operator namespace and the add-on ConfigMaps.

## Verification steps

1. Install RHOAM/RHMI from this PR
2. Create a ConfigMap called
    * RHOAM: `managed-api-service`
    * RHMI: `rhmi`
3. In the operator logs, verify that the following entry is immediately added (proof that the watch works for the ConfigMap):
    ```
    Reconciling namespace labels
    ```
4. Add the following label to the ConfigMap
    ```yaml
    # RHOAM
    "api.openshift.com/addon-managed-api-service-delete": "true"
    # RHMI
    "api.openshift.com/addon-rhmi-operator-delete": "true"
    ```
5. Verify that the uninstall starts immediately

## Type of change

## Checklist

### DIFF
--- a/controllers/namespacelabel/enqueue.go
+++ b/controllers/namespacelabel/enqueue.go
@@ -1,0 +1,63 @@
+package controllers
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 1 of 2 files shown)


psturc:
LGTM, verified for RHMI and RHOAM.

I'm not much familiar with the integreatly-operator code, so it's probably worth getting someone else for a code review.

PR #1604, opened 2021-02-19
adding sku package

laurafitzgerald:
# Description
Closes  https://issues.redhat.com/browse/MGDAPI-1219, https://issues.redhat.com/browse/MGDAPI-1220

### Verify SKU and ToSKU are set as expected

Run `INSTALLATION_TYPE=managed-api make cluster/prepare/local`

Verify that a dummy secret and config map is in place.
Run `oc get secret addon-managed-api-service-parameters -n redhat-rhoam-operator -o json | jq -r '.data.sku' | base64 --decode` should output DEV_SKU. 
Run  `oc get configmap sku-config -n redhat-rhoam-operator -o json | jq .data` should return an object with key `skuconfigs` and a json object with skus.

Run `INSTALLATION_TYPE=managed-api IN_PROW=true make deploy/integreatly-rhmi-cr.yml`
Run `INSTALLATION_TYPE=managed-api make code/run`
The installation should start successfully.

Once the installation starts check that `toSKU` is set to `DEV_SKU`

`oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq -r '.status.toSKU'` should output `DEV_SKU`
oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq -r '.status.SKU' should output null

Wait for the installation to go to complete status and then 

Check that toSKU is unset and SKU is set to `DEV_SKU`

`oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq -r '.status.toSKU'` -> null
`oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq -r '.status.sku'` -> `DEV_SKU`

### Verify Default SKU and isUpdated

Wait for the installation to go to complete status and then 

`oc delete  secret addon-managed-api-service-parameters -n redhat-rhoam-operator` -> Note: this is just to simulate a SKU change and to check if the defaults are selected and wouldn't be done in production scenario.

Will delete the secret. On the next reconcile (so you may need to wait) this will be recognised as a change to the SKU. As there is no secret it will also use the default. 

`oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq -r '.status.sku'` -> `DEV_SKU`
`oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq -r '.status.toSKU'` -> `default`

Wait for the reconcile to go back to a complete state: after that the values should be 

`oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq -r '.status.sku'` -> `default`
`oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq -r '.status.toSKU'` -> `null`

You can also check that changing the value works as expected again (i.e. it will pick up a new specified value in parameter):

`make cluster/prepare/sku`

This would recreate the secret with a different param. This can be done even before the toSKU has changed to SKU. i.e. before a sku change is fully completed another can be done. At this stage the expected outcomes are:

`oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq -r '.status.toSKU'` -> `DEV_SKU`
`oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq -r '.status.sku'` -> `default`

Again 
Wait for the reconcile to go back to a complete state: after that the values should be 

`oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq -r '.status.sku'` -> `DEV_SKU`
`oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq -r '.status.toSKU'` -> `null`

## Verify things aren't broken when either config map or sku param is not present.

Uninstall by `oc delete rhmi rhoam -n redhat-rhoam-operator`
`oc delete  secret addon-managed-api-service-parameters -n redhat-rhoam-operator`
`oc delete configmap sku-config -n redhat-rhoam-operator`

Don't run cluster/prepare!

Run `INSTALLATION_TYPE=managed-api IN_PROW=true make deploy/integreatly-rhmi-cr.yml`
Run `INSTALLATION_TYPE=managed-api make code/run`

## Type of change

## Checklist

### DIFF
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -260,20 +260,23 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 		if err != nil {
 			return retryRequeue, errors.Wrap(err, "Error getting sku config map")
 		}
-		err = sku.GetSKU(skuSecretName, cm, installationSKU)
-		if err != nil {
-			return retryRequeue, err
-		}
 
 		// if it's the first install and first reconcile there will be no value
-		// set it because we want to be able to compare the sku
-		if installation.Status.ToSKU == "" && installation.Status.SKU == "" {
-			installationSKU.SetUpdated(true)
+		// or if the secretName is not the same as the current SKU
+		// there has been a change to the SKU
+		// so
+		// set ToSKU to the new value
+		// set isUpdate to true
+		isSKUUpdated := false
+		if (installation.Status.ToSKU == "" && installation.Status.SKU == "") ||
+			skuSecretName != installation.Status.SKU {
+			isSKUUpdated = true
 			installation.Status.ToSKU = skuSecretName
 		}
 
-		if skuSecretName != installation.Status.SKU {
-			installationSKU.SetUpdated(true)
+		err = sku.GetSKU(skuSecretName, cm, installationSKU, isSKUUpdated)
+		if err != nil {
+			return retryRequeue, err
 		}
 
 	}
@@ -421,6 +424,8 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 		installation.Status.Version = version.GetVersionByType(installation.Spec.Type)
 		installation.Status.ToVersion = ""
 		metrics.SetRhmiVersions(string(installation.Status.Stage), installation.Status.Version, installation.Status.ToVersion, installation.CreationTimestamp.Unix())
+		installation.Status.SKU = skuSecretName
+		installation.Status.ToSKU = ""
 		log.Info("installation completed successfully")
 	}
 
@@ -432,8 +437,6 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 		if installation.Spec.RebalancePods {
 			r.reconcilePodDistribution(installation)
 		}
-		installation.Status.SKU = skuSecretName
-		installation.Status.ToSKU = ""
 	}
 	metrics.SetRHMIStatus(installation)
 
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -238,25 +238,25 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 		installationCfgMap = installation.Spec.NamespacePrefix + DefaultInstallationConfigMapName
 	}
 
-	skuSecretName := ""
 	installationSKU := &sku.SKU{}
 	if installation.Spec.Type == string(rhmiv1alpha1.InstallationTypeManagedApi) {
+		isSKUUpdated := false
 		skuSecretName, found, err := addon.GetStringParameterByInstallType(context.TODO(), r.Client, rhmiv1alpha1.InstallationTypeManagedApi, request.NamespacedName.Namespace, "sku")
 		if err != nil {
 			return retryRequeue, errors.Wrap(err, "Error checking for SKU secret")
 		}
 
+		//!found means the param wasn't found so we want to default rather than return
+		//but don't do it until the installation object is more than a minute old in case the secret is slow to create
 		if !found && !installation.ObjectMeta.CreationTimestamp.Time.Before(time.Now().Add(-(1 * time.Minute))) {
 			return retryRequeue, nil
 		}
-		//!found means the param wasn't found so we want to default rather than return
-		//but don't do it until the installation object is more than a minute old in case the secret is slow to create
-		if !found && installation.ObjectMeta.CreationTimestamp.Time.Before(time.Now().Add(-(1 * time.Minute))) {
-			skuSecretName = "TWENTY_MILLION_SKU"
+		if !found {
+			skuSecretName = ""
 		}
 		// get a configmap from the cluster
 		cm := &corev1.ConfigMap{}
-		err = r.Get(context.TODO(), types.NamespacedName{Namespace: request.Namespace, Name: "sku-config"}, cm)
+		err = r.Get(context.TODO(), types.NamespacedName{Namespace: request.Namespace, Name: sku.ConfigMapName}, cm)
 		if err != nil {
 			return retryRequeue, errors.Wrap(err, "Error getting sku config map")
 		}
@@ -267,16 +267,26 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 		// so
 		// set ToSKU to the new value
 		// set isUpdate to true
-		isSKUUpdated := false
-		if (installation.Status.ToSKU == "" && installation.Status.SKU == "") ||
-			skuSecretName != installation.Status.SKU {
+
+		// if the secret is deifferent to what's currently set in SKU there has been an updated
+		if skuSecretName != installation.Status.SKU {
 			isSKUUpdated = true
-			installation.Status.ToSKU = skuSecretName
 		}
 
 		err = sku.GetSKU(skuSecretName, cm, installationSKU, isSKUUpdated)
 		if err != nil {
 			return retryRequeue, err
+		}
+
+		// if both are empty it's the first round of installation
+		// or if the secretname is not the same as what the SKU is marked there has been a change
+		// so set toSKU and update the status object
+		if (installation.Status.ToSKU == "" && installation.Status.SKU == "") ||
+			skuSecretName != installation.Status.SKU {
+			installation.Status.ToSKU = installationSKU.GetName()
+			if err := r.Status().Update(context.TODO(), installation); err != nil {
+				return ctrl.Result{}, err
+			}
 		}
 
 	}
@@ -424,7 +434,7 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 		installation.Status.Version = version.GetVersionByType(installation.Spec.Type)
 		installation.Status.ToVersion = ""
 		metrics.SetRhmiVersions(string(installation.Status.Stage), installation.Status.Version, installation.Status.ToVersion, installation.CreationTimestamp.Unix())
-		installation.Status.SKU = skuSecretName
+		installation.Status.SKU = installationSKU.GetName()
 		installation.Status.ToSKU = ""
 		log.Info("installation completed successfully")
 	}
@@ -436,6 +446,10 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 		retryRequeue.RequeueAfter = 5 * time.Minute
 		if installation.Spec.RebalancePods {
 			r.reconcilePodDistribution(installation)
+		}
+		if installationSKU.IsUpdated() {
+			installation.Status.SKU = installationSKU.GetName()
+			installation.Status.ToSKU = ""
 		}
 	}
 	metrics.SetRHMIStatus(installation)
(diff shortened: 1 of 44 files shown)


controllers/rhmi/rhmi_controller.go
```diff
@@ -231,6 +232,22 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
...
+	cm := &corev1.ConfigMap{}
+	err = r.Get(context.TODO(), types.NamespacedName{Namespace: request.Namespace, Name: "sku-config"}, cm)
+	if err != nil {
+		return retryRequeue, errors.Wrap(err, "Error getting sku config map")
+
+	}
+
+	installationSKU := &sku.SKU{}
+	if installation.Spec.Type == string(rhmiv1alpha1.InstallationTypeManagedApi) {
+		installationSKU, err = sku.GetSKU("DEV_SKU", cm)
```
laurafitzgerald, line 245: This needs fixing up as it's hardcoded for now.

controllers/rhmi/rhmi_controller.go
```diff
@@ -768,7 +785,7 @@ func (r *RHMIReconciler) processStage(installation *rhmiv1alpha1.RHMI, stage *St
 		if err != nil {
 			return rhmiv1alpha1.PhaseFailed, fmt.Errorf("could not create server client: %w", err)
 		}
-		product.Status, err = reconciler.Reconcile(context.TODO(), installation, &product, serverClient)
+		product.Status, err = reconciler.Reconcile(context.TODO(), installation, &product, serverClient, sku.GetProduct(productName))
```
philbrookes, line 788: In general, I like this (passing in the specific product SKU config, rather than the whole SKU, but it requires considering how the reconciler will know if the SKU has just changed.

controllers/rhmi/rhmi_controller.go
```diff
@@ -231,6 +232,33 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
...
+			return retryRequeue, nil
+		}
+		//!found means the param wasn't found so we want to default rather than return
+		//but don't do it until the installation object is more than a minute old in case the secret is slow to create
+		if !found && installation.ObjectMeta.CreationTimestamp.Time.Before(time.Now().Add(-(1 * time.Minute))) {
+			skuSecretName = "TWENTY_MILLION_SKU"
+		}
+		// get a configmap from the cluster
+		cm := &corev1.ConfigMap{}
+		err = r.Get(context.TODO(), types.NamespacedName{Namespace: request.Namespace, Name: "sku-config"}, cm)
```
laurafitzgerald, line 252: I'm particularly interested in feedback on this block. If no secret is found on the cluster, aka there is no sku passed via ocm param (will be the case until that functionality is added in ocm) the skuSecretName can be added. I don't like that it's hardcoded but as long as we have a sku with the same name present in the secret in the addon config this will work fine. I think ideally in the longer term, once the functionality is implemented in ocm the sku param should be required. in that case we can just return until we find a secret.

pkg/resources/sku/sku.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+type skuConfigReceiver struct {
+	Name      string                    `json:"name,omitempty"`
+	RateLimit RateLimit                 `json:"rate-limiting,omitempty"`
+	Resources map[string]ResourceConfig `json:"resources,omitempty"`
+}
+
+func GetSKU(SKUId string, SKUConfig *corev1.ConfigMap, retSku *SKU) error {
+	received := &[]skuConfigReceiver{}
```
philbrookes, line 45: Could we rename this `allSKUs` it took me a few moments to work out why we have this as well as `skuReceiver`

pkg/resources/sku/sku.go
```diff
@@ -0,0 +1,157 @@
...
+
+func (s *SKU) GetProduct(productName v1alpha1.ProductName) ProductConfig {
+	return s.productConfigs[productName]
+}
+
+func (s *SKU) IsUpdated() bool {
+	return s.isUpdated
+}
+
+func (s *SKU) SetUpdated(isUpdated bool) {
```
philbrookes, line 102: I don't think this should be public, in fact I think we don't need this

controllers/rhmi/rhmi_controller.go
```diff
@@ -231,6 +232,46 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
...
+		}
+		err = sku.GetSKU(skuSecretName, cm, installationSKU)
+		if err != nil {
+			return retryRequeue, err
+		}
+
+		// if it's the first install and first reconcile there will be no value
+		// set it because we want to be able to compare the sku
+		if installation.Status.ToSKU == "" && installation.Status.SKU == "" {
+			installationSKU.SetUpdated(true)
```
philbrookes, line 265: If we do this in the `sku.GetSKU` it can be kept private so the SKU remains immutable.

pkg/resources/sku/sku.go
```diff
@@ -0,0 +1,157 @@
...
+	AlertLimits     []int64 `json:"alert_limits,omitempty"`
+}
+
+type skuConfigReceiver struct {
+	Name      string                    `json:"name,omitempty"`
+	RateLimit RateLimit                 `json:"rate-limiting,omitempty"`
+	Resources map[string]ResourceConfig `json:"resources,omitempty"`
+}
+
+func GetSKU(SKUId string, SKUConfig *corev1.ConfigMap, retSku *SKU) error {
```
philbrookes, line 44: easiest way might be to add a `isUpdate bool` arg, that can be passed to this constructor, to keep it immutable.

pkg/resources/sku/sku.go
```diff
@@ -0,0 +1,157 @@
...
+
+func (p *ProductConfig) GetResourceConfig(ddcssName string) corev1.ResourceRequirements {
+	return p.resourceConfigs[ddcssName].Resources
+}
+
+func (p ProductConfig) GetReplicas(ddcssName string) int32 {
+	return p.resourceConfigs[ddcssName].Replicas
+}
+
+func (p *ProductConfig) Configure(obj interface{}) error {
```
philbrookes, line 114: This is cool. I'm really happy with how this turned out. Have you tested it is functional?

pkg/resources/sku/sku_test.go
```diff
@@ -0,0 +1,633 @@
...
+			}
+
+			if tt.validate != nil {
+				tt.validate(tt.args.SKU, t)
+			}
+		})
+	}
+}
+
+func TestProductConfig_Configure(t *testing.T) {
```
philbrookes, line 179: This is some awesome looking unit test, at a glance!

philbrookes: It looks like we're only testing deployment configs though? We should check that this also functions for the 3scale CRD, a deployment and a stateful set.

philbrookes: We should also check that a cluster with more resources than necessary is lowered when isUpdated is true.

philbrookes: ~cluster~ I mean deployment, deploymentconfig, statefulset or 3scale CR

philbrookes: ah my mistake, I see we are testing all the types, but using the same name on each one :facepalm: still worth testing the isUpdate lowers values anyway

laurafitzgerald:
/hold
this one should wait to merge until after this sprint release is done

pkg/resources/sku/sku_test.go
laurafitzgerald: ah sorry should have updated the names. I didn't cover 3scale CRD. but I can add that also later. About lowering the values. I have a test that they don't get lowered even when isUpdated is true as I thought we only wanted to update them if they were lower?

pkg/resources/sku/sku.go
laurafitzgerald: I haven't tested it in cluster but the tests cover the whole function. I started working to update the 3scale one but think I should add an apimanager crd object into the switch.

sergioifg94 (commented):
Code is looking good. Just a couple of suggestions

controllers/rhmi/rhmi_controller.go
```diff
@@ -231,6 +232,49 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 		installationCfgMap = installation.Spec.NamespacePrefix + DefaultInstallationConfigMapName
 	}
 
+	skuSecretName := ""
+	installationSKU := &sku.SKU{}
+	if installation.Spec.Type == string(rhmiv1alpha1.InstallationTypeManagedApi) {
+		skuSecretName, found, err := addon.GetStringParameterByInstallType(context.TODO(), r.Client, rhmiv1alpha1.InstallationTypeManagedApi, request.NamespacedName.Namespace, "sku")
```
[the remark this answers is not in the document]
laurafitzgerald: Hi @sergioifg94 Thanks for the review. in the meantime I've moved the skuSecretName inside the if. It's only needed in there because the name gets assigned in the installationSKU object so should be referred to from there. But it's a good point which I didn't spot. 👍

controllers/rhmi/rhmi_controller.go
laurafitzgerald: I did a bit of refactoring in this section so I think I'm happy with it now. I will push it soon. I will leave the comment open for now as I'm still interested in the feedback.

pkg/resources/sku/sku.go
```diff
@@ -0,0 +1,154 @@
...
+		if p.sku.isUpdated && *t.Spec.Replicas < p.resourceConfigs[t.ObjectMeta.Name].Replicas {
+			t.Spec.Replicas = &configReplicas
+		}
+		p.mutate(&t.Spec.Template, t.ObjectMeta.Name)
+		break
+	default:
+		return errors.New(fmt.Sprintf("sku configuration can only be applied to Deployments, StatefulSets or Deployment Configs, found %s", reflect.TypeOf(obj)))
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
laurafitzgerald: Hey @sergioifg94 I'm implemented some of the feedback here and will push the change soon. The one piece I didn't take was that the function would accept a Object rather than an interface. the reason for this is that we would also like this function to handle APIManager crs in the case of 3scale so there will be some refactoring later on this.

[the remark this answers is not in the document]
sergioifg94: Thanks @laurafitzgerald. I think it would still handle APIManager CRs. The `metav1.Object` is implemented by every resource, including Custom Resources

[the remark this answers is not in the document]
laurafitzgerald: ah didn't know that. I will make the update. 👍

laurafitzgerald:
@philbrookes @sergioifg94 this one is ready for review again when you have some time. I have one an question on the Configure section which I've left the comment unresolved. The section where we decide how the default is set has been updated is fixed up. Phil could you take a look at that that it is as you expected. Also that one left me with a question. If there is no skuName found because of no secret or no value in the secret, we use the default but what if there is a value passed in the secret but it's not found in the config map. Is that something that we should error on?

test/common/sku.go
```diff
@@ -0,0 +1,59 @@
+package common
```
laurafitzgerald, line 1: @psturc would you mind taking a look at this test case?

psturc: 👀

laurafitzgerald:
/hold
we'll use this one as a feature branch for merging the sku change.

laurafitzgerald:
Close in favour of https://github.com/integr8ly/integreatly-operator/pull/1688

PR #1605, opened 2021-02-19
Fix make release/prepare

sergioifg94:
# Description

Fix `make release/prepare` to work with Kustomize when it's not installed. The Makefile will automatically install (in a temporal
directory) Kustomize if it's not found on the $PATH. However, when running `release/prepare`, it will invoke the `prepare-release.sh` script, which assumes Kustomize is in the $PATH.

Modify the script to use an optional environment variable with the path to Kustomize, and the Makefile to pass the temporal path when running the script.

## Type of change

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -419,7 +419,7 @@ prepare-patch-release:
 
 .PHONY: release/prepare
 release/prepare: kustomize
-	@./scripts/prepare-release.sh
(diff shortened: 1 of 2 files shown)


MStokluska:
/lgtm 

Verified it on the rhmi upgrade pipeline which uses the prepare/release script. Works as expected.

PR #1606, opened 2021-02-19
[WIP] Print Errors when they may have happened in Test C03

Boomatang:
# Description

Some error could have happened during the test C03 that would not get logged. These errors may help identify why C03 sometimes fail to scale back up the operator.

## Type of change

## Checklist


Boomatang:
Going to include this in a larger improvement PR

PR #1609, opened 2021-02-22
MGDAPI-1170: Improve product operator installation

sergioifg94:
# Description

### Duplicated install plans

Currently, as part of the product operators installation, the retrieval of the InstallPlans associated with the product is done by
listing all InstallPlans in the Subscription namespace. OLM however might create multiple InstallPlans where only one is valid and is referenced in the Subscription status as an `InstallPlanRef` field. The current approach will loop through all the InstallPlans in the namespace, approve them an wait for the status to be `Complete`, but if an invalid InstallPlan comes first in the list, it will remain stuck in progress, never reaching the valid one.

Change the `MarketPlaceInterface` and its `Manager` implementation to retrieve only the InstallPlan that is referenced by the Subscription, and update the unit tests that depend on this method.

### Missing deployment in CSVs

It's been noted that CSVs have been created with missing deployments, signaling the product operator as installed but the operator doesn't start running. Add logic to validate the CSVs and retry the installation if the CSV is found to be invalid

## Verification

1. Install a cluster with Openshift ~4.6.8~
2. Install RHOAM from this PR
    > While RHOAM progresses the products installation, check the InstallPlans in the product operator namespaces, check if there are more than one InstallPlan for namespace, this might happen arbitrarily
3. Verify that the installation completes successfilly

## Type of change

## Checklist

### DIFF
--- a/pkg/resources/reconciler.go
+++ b/pkg/resources/reconciler.go
@@ -239,20 +239,13 @@ func (r *Reconciler) ReconcileSubscription(ctx context.Context, target marketpla
 
 	// Workaround to re-install product operator if install plan fails due to https://bugzilla.redhat.com/show_bug.cgi?id=1923111
 	if ip.Status.Phase == operatorsv1alpha1.InstallPlanPhaseFailed {
+		var csv *operatorsv1alpha1.ClusterServiceVersion
 		if sub.Status.InstalledCSV != "" {
-			log.Warningf("Deleting csv for re-install due to failed install plan", l.Fields{"ns": target.Namespace, "install plan": target.Pkg, "csv": sub.Status.InstalledCSV})
-			csv := &operatorsv1alpha1.ClusterServiceVersion{
+			csv = &operatorsv1alpha1.ClusterServiceVersion{
 				ObjectMeta: metav1.ObjectMeta{Namespace: target.Namespace, Name: sub.Status.InstalledCSV},
 			}
-			if err := client.Delete(ctx, csv); err != nil && !k8serr.IsNotFound(err) {
-				return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to delete csv for re-intstall: %s", err)
-			}
 		}
-		log.Warningf("Deleting subscription for re-install due to failed install plan", l.Fields{"ns": target.Namespace, "install plan": target.Pkg})
-		if err := client.Delete(ctx, sub); err != nil && !k8serr.IsNotFound(err) {
-			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to delete subscription for re-intstall: %s", err)
-		}
-		return integreatlyv1alpha1.PhaseAwaitingOperator, nil
+		return retryInstallation(ctx, client, log, target, csv, sub)
 	}
 
 	//if it's approved but not complete, then it's in progress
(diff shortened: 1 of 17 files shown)


pkg/resources/reconciler.go
```diff
@@ -228,47 +228,82 @@ func (r *Reconciler) ReconcileSubscription(ctx context.Context, target marketpla
...
+	}
+
+	return nil
+}
+
+func retryInstallation(ctx context.Context, client k8sclient.Client, log l.Logger, target marketplace.Target, csv *operatorsv1alpha1.ClusterServiceVersion, sub *operatorsv1alpha1.Subscription) (integreatlyv1alpha1.StatusPhase, error) {
+	if csv != nil {
+		log.Warningf("Deleting csv for re-install due to failed install plan", l.Fields{"ns": target.Namespace, "install plan": target.Pkg, "csv": csv.Name})
+		if err := client.Delete(ctx, csv); err != nil && !k8serr.IsNotFound(err) {
+			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to delete csv for re-intstall: %s", err)
```
jjkiely, line 296: Tiny typo on lines 296 and 302, "intstall" instead of install

pkg/resources/reconciler.go
```diff
@@ -228,47 +228,82 @@ func (r *Reconciler) ReconcileSubscription(ctx context.Context, target marketpla
...
+			return retryInstallation(ctx, client, log, target, ipCSV, sub)
 		}
 	}
 
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
 
+func validateCSV(csv *operatorsv1alpha1.ClusterServiceVersion) error {
+	if csv.Spec.InstallStrategy.StrategyName == operatorsv1alpha1.InstallStrategyNameDeployment && len(csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs) == 0 {
+		return errors.New("No Deployment found in install strategy")
```
jjkiely, line 286: Hey Sergio, not a big issue but I've found out that we should be trying to make sure that error messages begin with lower case letters rather than capitals

pkg/resources/reconciler.go
```diff
@@ -228,47 +228,82 @@ func (r *Reconciler) ReconcileSubscription(ctx context.Context, target marketpla
...
+			log.Warningf("CSV failed validation. Retrying operator installation", l.Fields{"error": err, "install plan": target.Pkg})
+			return retryInstallation(ctx, client, log, target, ipCSV, sub)
 		}
 	}
 
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
 
+func validateCSV(csv *operatorsv1alpha1.ClusterServiceVersion) error {
+	if csv.Spec.InstallStrategy.StrategyName == operatorsv1alpha1.InstallStrategyNameDeployment && len(csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs) == 0 {
```
jjkiely, line 285: Hey @sergioifg94 , so just to confirm, is the issue with the current install plans that they might have a deployment specified but no actual deployment specs inside them?

sergioifg94: Hey Joseph, so there's actually 2 issues addressed in this PR: 1. The duplicated install plans and 2. The possibility of an install plan having no deployment. I believe the second one is indirectly caused the first one, but for safety I've addressed both (that's why they're in 2 separate commits). 
The second issue seems to be caused by the install plan not including a deployment at all in the CSV. A CSV may include many deployments, that's why I'm checking that the length of the array is not zero.

Here's an example of a CSV with 2 deployments: https://github.com/integr8ly/integreatly-operator/blob/master/manifests/integreatly-marin3r/0.7.0/marin3r.v0.7.0.clusterserviceversion.yaml#L481-L559

sergioifg94:
Thanks for the review @jjkiely I just pushed the changes you suggested

jjkiely:
/lgtm
Thanks for the detailed replies Sergio! :+1:

PR #1612, opened 2021-02-24
fix: getting cluster id in sso-idp script

psturc:
# Description
The issue with getting no result for a cluster ID was seen couple of times in nightly pipelines and is causing problems when running e2e test (IDP tests)

The command I'm introducing in this PR is a command that I was using lately in my local scripts and it was working for me every time.

I verified it's also working against the cluster where the issue with sso-idp script can be currently seen:
```
No CLUSTER_ID detected. IDP will be added directly into "cluster" OAuth resource.
```

In the new `ocm` command we're asking ocm directly to search for the specific cluster API URL instead of asking it to get the full list of managed clusters and then filter them. I thought that it doesn't really matter which approach we choose, but in this case it does. (But why - I have no idea.)

### DIFF
--- a/scripts/setup-sso-idp.sh
+++ b/scripts/setup-sso-idp.sh
@@ -78,7 +78,7 @@ echo "Keycloak realm: $REALM"
 
 # If CLUSTER_ID is not passed, find out ID based on currently targeted server
 set +e # ignore errors in environments without ocm command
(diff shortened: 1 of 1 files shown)


psturc:
Hi @matskiv, would you have a sec to take a look at this PR? (Not super important fix but definitely nice to have.)

matskiv:
Oh, that is odd.
I tried the new command with my cluster and it works fine. :) 

/lgtm
/approve

psturc:
cc @philbrookes @Boomatang 
FYI This should have a positive effect on IDP tests, which are failing recently in nightly pipelines, most probably due to fact that SSO is set only on a cluster level (and later removed by hive during test run), not in OCM.

To verify that this fix worked, we should not see the following message in any of the nightly pipelines' log:
```
No CLUSTER_ID detected. IDP will be added directly into "cluster" OAuth resource.
```

PR #1613, opened 2021-02-24
Cro resources dashboard

jackdelahunt:
# Description

Added new Grafana dashboard that displays CRO resource metrics

image

## Type of change

## Checklist

## Verification
- Clone this branch
- Deploy locally on byoc cluster
- Check if CRO resources dashboard exists in middleware monitoring operator grafana
- Verify that the dashboard is displaying data
- Changed is deployed here (mwcollabaz-byoc) to verify visually

### DIFF
--- a/pkg/products/monitoring/dashboards/croResources.go
+++ b/pkg/products/monitoring/dashboards/croResources.go
@@ -22,7 +22,7 @@ const MonitoringGrafanaDBCROResourcesJSON = `{
   "links": [],
   "panels": [
     {
-      "datasource": null,
+      "datasource": "Prometheus",
       "fieldConfig": {
         "defaults": {
           "custom": {},
@@ -96,7 +96,7 @@ const MonitoringGrafanaDBCROResourcesJSON = `{
       "type": "stat"
     },
     {
-      "datasource": null,
+      "datasource": "Prometheus",
       "fieldConfig": {
         "defaults": {
           "custom": {},
@@ -171,7 +171,7 @@ const MonitoringGrafanaDBCROResourcesJSON = `{
     },
     {
       "collapsed": true,
-      "datasource": null,
+      "datasource": "Prometheus",
       "gridPos": {
         "h": 1,
         "w": 24,
@@ -299,7 +299,7 @@ const MonitoringGrafanaDBCROResourcesJSON = `{
     },
     {
       "collapsed": true,
-      "datasource": null,
+      "datasource": "Prometheus",
       "gridPos": {
         "h": 1,
         "w": 24,
@@ -586,7 +586,7 @@ const MonitoringGrafanaDBCROResourcesJSON = `{
     },
     {
       "collapsed": true,
-      "datasource": null,
+      "datasource": "Prometheus",
       "gridPos": {
         "h": 1,
         "w": 24,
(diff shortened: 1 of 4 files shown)


pkg/products/monitoring/dashboards/croResources.go
```diff
@@ -0,0 +1,1349 @@
...
+                    "color": "green",
+                    "value": null
+                  },
+                  {
+                    "color": "red",
+                    "value": 80
+                  }
+                ]
+              },
+              "unit": "decbytes"
```
matskiv, line 774: We should be consistent with the units used to avoid confusion (MB vs MiB in different places).
Please use `bytes` unit for all memory representations.

pkg/products/monitoring/dashboards/croResources.go
```diff
@@ -0,0 +1,1349 @@
...
+          "xaxis": {
+            "buckets": null,
+            "mode": "time",
+            "name": null,
+            "show": true,
+            "values": []
+          },
+          "yaxes": [
+            {
+              "format": "decbytes",
```
matskiv, line 843: Same as above, please use `bytes` unit for all memory representations.

pkg/products/monitoring/dashboards/croResources.go
```diff
@@ -0,0 +1,1349 @@
...
+            "alignLevel": null
+          }
+        },
+        {
+          "aliasColors": {},
+          "bars": false,
+          "dashLength": 10,
+          "dashes": false,
+          "datasource": "Prometheus",
+          "fieldConfig": {
```
matskiv, line 1224: `unit` is missing under this `fieldConfig` section?

pkg/products/monitoring/dashboards/croResources.go
```diff
@@ -0,0 +1,1349 @@
...
+          "xaxis": {
+            "buckets": null,
+            "mode": "time",
+            "name": null,
+            "show": true,
+            "values": []
+          },
+          "yaxes": [
+            {
+              "format": "short",
```
matskiv, line 1289: This displays as `Mil`
I think it needs to be changed to `bytes` to show `MiB` as a unit.

pkg/products/monitoring/dashboards/croResources.go
```diff
@@ -0,0 +1,1349 @@
...
+              }
+            ]
+          }
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 5,
+        "w": 8,
+        "x": 16,
```
matskiv, line 140: Minor one..
This panel displays awkwardly on right side of the screen.
If we change `gridPos.x` to 8, then it should be better.

https://user-images.githubusercontent.com/1605799/109150036-fe371f80-7770-11eb-976b-939d78a62037.mov

KevFan:
@jackdelahunt I think the new dashboard would need to be added to this e2e test :thinking: 
* https://github.com/integr8ly/integreatly-operator/blob/master/test/common/dashboards_exist.go#L133
*  https://github.com/integr8ly/integreatly-operator/blob/master/test/common/dashboards_exist.go#L22

matskiv (changes requested):
Some units need to be fixed, and maybe "Redis connection" panel position.
Otherwise looks pretty good 👍

PR #1614, opened 2021-02-25
MGDAPI-1180: Update CRD and webhook API versions

sergioifg94:
# Description

> Link to JIRA: https://issues.redhat.com/browse/MGDAPI-1180

v1beta1 API version of CRDs and Admission resources (webhooks) will be deprecated soon. Update the project to use v1 instead.

* Update CRD generation options to enable v1 API version of CRD
* Update imports of ValidatingWebhookConfiguration and MutatingWebhookConfiguration to reference v1 API version instead of
v1beta1.

## Type of change

## Checklist

### DIFF
--- a/pkg/webhooks/rules.go
+++ b/pkg/webhooks/rules.go
@@ -1,13 +1,13 @@
 package webhooks
 
-import "k8s.io/api/admissionregistration/v1beta1"
+import v1 "k8s.io/api/admissionregistration/v1"
 
 // The `RuleWithOperations` and `Rule` types redefine the original ones from
-// k8s.io/api/admissionregistration/v1beta1 in order to allow to define methods
+// k8s.io/api/admissionregistration/v1 in order to allow to define methods
 // to build the rule as a fluent interface.
 
 type RuleWithOperations struct {
-	Operations []v1beta1.OperationType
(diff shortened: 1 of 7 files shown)


maleck13 (commented):
Change looks good to me. Have we tried upgrading from current to this version to measure if there is any impact?

pkg/webhooks/rules.go
```diff
@@ -1,21 +1,21 @@
 package webhooks
 
-import "k8s.io/api/admissionregistration/v1beta1"
+import v1 "k8s.io/api/admissionregistration/v1"
```
MStokluska, line 3: Just a question, what is the reason why in the previous file we call the import by 
```admissionv1 "k8s.io/api/admissionregistration/v1"```
and in this file with an import name of `v1` ?

MStokluska:
Just a question left, otherwise lgtm, verified the upgrade scenario from master to this branch and it works fine!
/lgtm

pkg/webhooks/rules.go
sergioifg94: It's just an alias. Probably in the file where it was aliased to `admissionv1` it was done to avoid collision with other packages also named `v1`

PR #1615, opened 2021-02-25
add edge subscriptions checks to subscription controller

briangallagher:
WHAT
Edge should behave the same as Production with regard to auto or manual upgrade of integreatly operator. Hive automatically sets auto upgrade on install plans to true. It's the responsibility of the operator to change this if required. Integreatly does this in the subscription controller. But not currently for edge. 

HOW
Update the subscription controller to check for edge Subscriptions.


CathalOConnorRH:
/lgtm

We will test fully in edge at the next release

PR #1619, opened 2021-02-26
[MGDDOC-47] Docs Testing for 1.3 Release

joaedwar:
# Description

This PR includes docs for the upcoming 1.3 release. 

## Type of change

## Checklist


pawelpaszki:
@joaedwar - I am not involved in any RHOAM testing. did you mean @psturc ?

psturc:
Hi @joaedwar, new tests look good, but I would like to avoid deleting the old RHOAM test cases. Reason is that we're keeping a track of them in Polarion (we're uploading new a test cases every release). And if there's a new test case with the same id as the old one (e.g. `i02`), it will replace it and we loose a track of it.

So if you could keep the old test cases and bump IDs on the new test cases, that would be great 👍  thanks

trepel:
Hi @joaedwar , I am sorry, it is my bad I only get to this this late. Right, as it has been mentioned above it turns out the best thing to do is to just append new test cases and leave the existing ones as they are. Since I had the test cases updated accordingly locally because I generated them out of my local repo, I created an PR myself (https://github.com/integr8ly/integreatly-operator/pull/1658)

You can either lgtm my PR or update this PR accordingly and I can lgtm this then. Whatever works better for you.

trepel:
The https://github.com/integr8ly/integreatly-operator/pull/1658 has been merged so closing this one.

PR #1630, opened 2021-03-03
WIP Hyperfoil stats script refactor

jackdelahunt:
# Description
Currently the hyperfoil-requests-stats.sh script uses /run/:runId/stats/total, after a controller restart the cli does not return data for runs before the restart. The endpoint /run/:runId/stats/all will persist data after controller restart.
## Type of change

## Checklist
- [X] I have added tests that prove my fix is effective or that my feature works


jackdelahunt:
Commits on branches that are not apart of this repo

PR #1632, opened 2021-03-03
Refactored hyperfoil-requests-stats.sh to use /run/stats/all endpoint

jackdelahunt:
# Description
Currently the hyperfoil-requests-stats.sh script uses /run/:runId/stats/total, after a controller restart the cli does not return data for runs before the restart. The endpoint /run/:runId/stats/all will persist data after controller restart.

## Type of change

## Checklist

## Verification
- Clone this branch
- Deploy a cluster 
- Install hyperfoil on cluster
- Run a load test on the cluster
- Run this script pointing to the hyperfoil url 
- Verify that the output matches the script before the changes

### DIFF
--- a/scripts/hyperfoil-request-stats.sh
+++ b/scripts/hyperfoil-request-stats.sh
@@ -11,16 +11,19 @@
 
 RUN=$1
 hundred=100
+H1='Accept: application/json'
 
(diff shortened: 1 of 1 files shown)


matskiv:
I compared the output of new implementation with the old one, and it matches exactly. 👍 
/lgtm
I'll add approval label after merge blocker is lifted (once release is signed off).

PR #1641, opened 2021-03-05
MGDAPI-1261: Automate H22 test (rate limit redis counter)

sergioifg94:
# Description

> Link to JIRA: https://issues.redhat.com/browse/MGDAPI-1261

Automate test H22 using Golang, create script to launch testing and modify test steps to indicate script usage and expected output.

Extend `ThreeScaleInterface` with more API calls that are used by the test to create an API

> _Now with 500% more emojis in test output 🔥_

## Verification steps

1. Install RHOAM on a cluster
2. Perform the updated steps for H22 and ensure the test finishes successfully:
    * https://github.com/sergioifg94/integreatly-operator/blob/MGDAPI-1261/test-cases/tests/products/h22-validate-that-rate-limit-service-is-working-as-expected.md

## Type of change

## Checklist

### DIFF
--- a/test/scripts/products/h22-validate-that-rate-limit-service-is-working-as-expected/h22-validate-that-rate-limit-service-is-working-as-expected.go
+++ b/test/scripts/products/h22-validate-that-rate-limit-service-is-working-as-expected/h22-validate-that-rate-limit-service-is-working-as-expected.go
@@ -1,0 +1,578 @@
+package main
+
+import (
+	"context"
+	"crypto/tls"
+	"errors"
+	"flag"
+	"fmt"
+	"net/http"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	crov1alpha1 "github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1"
+	"github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1/types"
+	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/apis/v1alpha1"
+	"github.com/integr8ly/integreatly-operator/pkg/products/threescale"
+	"github.com/integr8ly/integreatly-operator/pkg/resources"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/logger"
+	testcommon "github.com/integr8ly/integreatly-operator/test/common"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/kubernetes"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func main() {
+	var namespacePrefix string
+	var iterations int
+	var cleanupAPI bool
+
+	flag.StringVar(&namespacePrefix, "namespace-prefix", "redhat-rhoam-", "Namespace prefix of RHOAM. Defaults to redhat-rhoam-")
+	flag.IntVar(&iterations, "iterations", 10, "Number of times to perform requests and test the counter")
+	flag.BoolVar(&cleanupAPI, "cleanup-api", true, "Whether to clean up the test API or not")
+
+	flag.Parse()
+
+	namespace := fmt.Sprintf("%soperator", namespacePrefix)
+
+	scheme := runtime.NewScheme()
+
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(crov1alpha1.SchemeBuilder.AddToScheme(scheme))
+	utilruntime.Must(integreatlyv1alpha1.AddToScheme(scheme))
+
+	config := ctrl.GetConfigOrDie()
+	client, err := k8sclient.New(config, k8sclient.Options{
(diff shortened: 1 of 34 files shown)


psturc:
👀

test/scripts/products/h22-validate-that-rate-limit-service-is-working-as-expected/h22-validate-that-rate-limit-service-is-working-as-expected.go
```diff
@@ -0,0 +1,504 @@
...
+		fmt.Println("Run: oc delete redis throw-away-redis-pod -n redhat-rhoam-operator")
+	}
+}
+
+// testCountIncreases performs numberOfRequests simultaneously and asserts that
+// the counter in Redis is increased at least that amount of times. Returns
+// true with no error if the test succeeds. false with no error if the test failed
+// and an error if an unexpected error occurred when performing the test
+func testCountIncreases(r *redisCounter, endpoint string, numberOfRequests int) (bool, error) {
+	if err := r.Flush(); err != nil {
```
MStokluska, line 230: hey @sergioifg94 the test looks brilliant!

I've just a question around flushing redis - if we are flushing before each series of pings, do the "previous count" matter - in my understanding from looking at the code - it would be equivalent to 0 every time. 
Or, we could get the current count before flushing? as it seems that the GetCount will always get 0 value?

test/scripts/products/h22-validate-that-rate-limit-service-is-working-as-expected/h22-validate-that-rate-limit-service-is-working-as-expected.go
```diff
@@ -0,0 +1,504 @@
...
+		r.PodName,
+		r.Namespace,
+		"redis",
+	)
+	if err != nil {
+		return 0, err
+	}
+	count = strings.TrimSpace(count)
+
+	return strconv.Atoi(count)
```
psturc, line 472: I got an error after running this test against RHOAM installed from master branch:

```
...
  ☁️  Waiting for throwaway Redis container to complete. Current phase: in progress...
  ☁️  Waiting for throwaway Redis container to complete. Current phase: in progress...
 ️ℹ️  Throw away Redis Pod ready
  Redis pod name: throw-away-redis-pod-57f8ddf94f-hsk76
Previous count: 0 | Number of requests: 10 | Current count: 10 | [PASS] 🎉
Previous count: 0 | Number of requests: 14 | Current count: 14 | [PASS] 🎉
strconv.Atoi: parsing "ERR wrong number of arguments for 'get' command": invalid syntax
```

psturc: After I removed the redis CR manually, I tried to rerun the test, but it failed with
```bash
  ☁️  Waiting for throwaway Redis container to complete. Current phase: in progress...
INFO[0004] Looking for RHMI CR                           ns=redhat-rhoam-operator
  🔑  Found access token: <REDACTED>
unexpected response code: 422
exit status 1
```

sergioifg94: Hmmm I was getting this kind of errors randomly but after adding the flush call it was working consistently, but seems like it's still occurring. I'll investigate what's causing this.

I think it's failing to re-run because the API was already created. A workaround is changing the base name in the code, so it will just create a new one:
https://github.com/sergioifg94/integreatly-operator/blob/MGDAPI-1261/test/scripts/products/h22-validate-that-rate-limit-service-is-working-as-expected/h22-validate-that-rate-limit-service-is-working-as-expected.go#L82

test/scripts/products/h22-validate-that-rate-limit-service-is-working-as-expected/h22-validate-that-rate-limit-service-is-working-as-expected.go
sergioifg94: Thanks Michal, I initially hadn't added the flush call but it was failing at random when calling `GetCount`. The flush appeared to make it reliable but yes, after the flush it would always be 0

MStokluska: I think the reason why it was unreliable is we are keeping track not only of the current minute but few past minutes too, where each minute has a new key-value pair in redis. I see you are filtering the liveness probe from the redis call but there's also the "old" minutes and since we are getting the values back by the "*", we are getting the results of current + couple last minutes - flush clears it so there's always only 1 minute. 
The issue I could see we could possibly have, is around if the tests are run between 2 minutes. For example, if you start hitting an enpoint at 00:59 and finish at 01:00 - there will be 2 ratelimit counts and a liveness probe.
The workaround this in manual testing was to flush every 1min mark or, when there's more than 2 entries in redis.

test/scripts/products/h22-validate-that-rate-limit-service-is-working-as-expected/h22-validate-that-rate-limit-service-is-working-as-expected.go
sergioifg94: @psturc  I just pushed some changes that fix all these reliability issues. I tested it with 50 iterations a few times and it's working consistently now

sergioifg94: @psturc I just pushed the changes you suggested about cleaning up the testing API. It will clean up the API after the test, but seems like 3scale might take a few minutes to pick up those changes so I left a note to wait around 5 minutes before retesting.
To know when it's not yet ready to re-test, in the dashboard it would still show the test backend in the dropdown below, but when clicking on it we get 404. After a few minutes that is gone and re-launching the test will work
image

psturc:
/lgtm
/approve
works perfectly ⭐

psturc:
Hi @laurafitzgerald, could you approve this one once you get a chance? Thanks!

PR #1647, opened 2021-03-08
Small improvement to LDAP test case

trepel:
# Description

Small improvements to LDAP test case.

**Verification Steps**
Eye review only.
You can also go to OCM UI and try to create a cluster (no need to actually create it, just see the form)


trepel:
@psturc I am sorry, I forgot to run prettier for this one. Should be fixed now

PR #1648, opened 2021-03-08
MGDAPI-1296: Ability to dynamically select GRPc or ConfigMap based installation

sergioifg94:
# Description

> Link to JIRA: https://issues.redhat.com/browse/MGDAPI-1296

Create `products/installation.yaml` file that declares the product operator installation strategy. Allows to point to either a manifests directory or an index image.

Modify marketplace package to use this file to instantiate the `CatalogSourceReconciler` implementations

## Type of change

## Checklist

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -4,12 +4,15 @@ ENV OPERATOR=/usr/local/bin/rhmi-operator \
     USER_UID=1001 \
     USER_NAME=integreatly-operator \
     TEMPLATE_PATH=/usr/local/bin/templates/monitoring \
-    MANIFEST_DIR=/usr/local/bin/manifests
+    MANIFEST_DIR=/usr/local/bin/manifests \
+    PRODUCT_DECLARATION=/usr/local/bin/products/installation.yaml
 
 # install operator binary
 COPY build/_output/bin/integreatly-operator ${OPERATOR}
 # copy templates
 COPY templates /usr/local/bin/templates
+
+COPY products /usr/local/bin/products
 
 COPY manifests ${MANIFEST_DIR}
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,6 +33,8 @@ COPY templates /templates
 
 COPY manifests /manifests
 
+COPY products /products
+
 COPY build/bin /usr/local/bin
 RUN /usr/local/bin/user_setup
 
--- a/Dockerfile.functional
+++ b/Dockerfile.functional
@@ -13,6 +13,7 @@ COPY vendor ./vendor
 COPY test ./test
 COPY Makefile ./
 COPY manifests/ ./manifests
+COPY products ./products
 
 RUN make test/compile/functional
 
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -90,6 +90,8 @@ type RHMIReconciler struct {
 	controller      controller.Controller
 	restConfig      *rest.Config
 	customInformers map[string]map[string]*cache.Informer
+
+	productsInstallationLoader marketplace.ProductsInstallationLoader
 }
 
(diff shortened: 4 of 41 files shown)


KevFan:
Looks good :+1: 

Unit tests looks to be failing due to be being unable to find the `installation.yaml` file as the test executable file path is not the project root directory but at the directory level of the tests, so you can proabably do something like how it's done in `templateHelper.go` to fix this. Not pretty but it worked on their unit tests :man_shrugging: 
* https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/monitoring/templateHelper.go#L36-L45 

One other thing is that as it's a file included in the operator, think we'd need to include it in the operator dockerfile to correctly install :thinking: 
(Not sure about the test image dockerfiles)
* https://github.com/integr8ly/integreatly-operator/blob/master/Dockerfile

sergioifg94:
Thanks for the review @KevFan 

I'm aware of the failing unit tests, I didn't go through the effort in case my approach wasn't the right one. The solution I have in my mind is to decouple that configuration from the file and mock it in the unit tests. I'll give it a go later.

Good spot on the Dockerfile. I was initially playing with the idea of mounting this file from a ConfigMap, but I think it's better to include it in the build as that'll bind a RHOAM release with specific product index tags, similar to the manifests. I'll include it in the Dockerfile

sergioifg94:
@KevFan I pushed the changes and removed the WIP tag. Made some other changes beside of the fix of the unit tests and the Dockerfile:
* The ProductDeclaration for each product is passed to the Reconciler constructor, so now the product reconciler doesn't have to read the file. This meant modifying a bunch of function signatures
* The product key in the file uses the product name as specified in the `rhmi_types` constants, instead of that pretty redundant constant that I had created in each product reconciler
* Created a separate interface to load the data from the products file. The goal of this was to mock it in the unit tests so it doesn't depend on the actual file but this wasn't necessary at the end. Still left it decoupled as there's no harm
* Created some unit tests for the `(ProductDeclaration).PrepareTarget` function

sergioifg94:
Tests still timing out. I'll log in the testing cluster and see what's going on
/retest

sergioifg94:
The Dockerfile that's used to build the testing image is the one in the `build` directory. I had missed the `COPY` command there for `products`

/test e2e
/test rhoam-e2e

build/Dockerfile
```diff
@@ -11,6 +11,8 @@ COPY build/_output/bin/integreatly-operator ${OPERATOR}
 # copy templates
 COPY templates /usr/local/bin/templates
 
+COPY products /usr/local/bin/products
```
KevFan, line 14:
```suggestion
COPY products /products
```
I think this is what's preventing prow tests to pass / run as prow seems to be using this dockerfile to build image :thinking:

sergioifg94: Thanks @KevFan it was indeed related to the Docker container of the e2e tests. I used the same approach that is being used for the manifests folder, getting it from an env var and defaulting to the relative path if not set. Seems to be working now, hope the tests pass now :crossed_fingers:

KevFan: Oh cool, that should work :+1: Though it may not be actually needed since we don't seem to be setting that env var for the release dockerfile anymore and have been installing fine :thinking: 
* https://github.com/integr8ly/integreatly-operator/blob/master/Dockerfile#L34

Maybe we should eventually revisit the differences between these 2 dockerfiles and decide whether the prow one is still needed / should be updated. Be good to keep them consistent enough or build from the release one :thinking: No harm in having the ability to set via env var though :+1:

sergioifg94: Totally agree. If I remember correctly we ended up with these 2 Dockerfiles in order to unblock us when testing the operator-sdk bump in prow, but we should look into having only one Dockerfile

PR #1649, opened 2021-03-09
fix hardcoded addon value in controllers/rhmi/promethuesRules.go

Priyankasaggu11929:
# Description

**Related JIRA ticket:** https://issues.redhat.com/browse/INTLY-10528
**Files updated:** `controllers/rhmi/prometheusRules.go`

- The PR make changes to fix the value of `addon` field in the `rhoam/rhmi-installation-alerts` & `rhoam/rhmi-upgrade-alerts` in the `openshift-monitoring` project/namespace. 
- It changes the hardcoded value `"addon": "managed-api-service"` to `"addon": installation.Spec.Type`, so as to fetch the value of addon based on the installation type on-the-go.

## Type of change

## Checklist

- [X] Verified independently on a cluster by reviewer


controllers/rhmi/prometheusRules.go
```diff
@@ -57,7 +57,7 @@ func (r *RHMIReconciler) newAlertsReconciler(installation *integreatlyv1alpha1.R
 						},
 						Expr:   intstr.FromString(fmt.Sprintf(`absent(%s_status{stage='complete'} == 1) and absent(%s_version{version=~".+"})`, installationName, installationName)),
 						For:    "120m",
-						Labels: map[string]string{"severity": "critical", "product": installationName, "addon": "managed-api-service", "namespace": "openshift-monitoring"},
+						Labels: map[string]string{"severity": "critical", "product": installationName, "addon": installation.Spec.Type, "namespace": "openshift-monitoring"},
```
laurafitzgerald, line 60: installation.Spec.Type would be either `managed` or `managed-api` which doesn't exactly replace the value that is currently present. The two types for addons are RHOAM -> `managed-api-service` and RHMI -> `integreatly-operator`
I'm not sure if these values are consumed by anything or if they are just informational for SRE. The `product` label is using the `installationName` which would contain either rhmi/rhoam.

laurafitzgerald (commented):
Hi @Priyankasaggu11929 Thank you for the pr. 

One question/comment inline about the value but the change itself is working fine for both RHOAM and RHMI.

Screenshot 2021-03-09 at 09 43 58

Screenshot 2021-03-09 at 09 52 57

controllers/rhmi/prometheusRules.go
Priyankasaggu11929: @laurafitzgerald, thank you for pointing the above things.

As a followup fix to that:
- I have added a new internal function `getAddonName()`. This function will return `managed-api-service` for RHOAM and `integreatly-operator` for RHMI.
-  And so, in place of the hardcoded value "addon": "managed-api-service", now the field value will be evaluated using the above function like `"addon": getAddonName(installation)`.

controllers/rhmi/prometheusRules.go
```diff
@@ -19,6 +19,14 @@ var (
 	}
 )
 
+func getAddonName(installation *integreatlyv1alpha1.RHMI) string {
+	if installation.Spec.Type == "managed-api" {
+		return "managed-api-service"
+	} else {
+		return "integreatly-operator"
+	}
```
laurafitzgerald, line 27: A small suggestion just in case there is a different entry to the Type, we can handle that with an empty string (or something else if you think there is a better value) rather than always defaulting to `integreatly-operator` 

```suggestion
      switch installation.Spec.Type {
	case "managed-api":
		return "managed-api-service"
	case "integreatly-operator":
		return "integreatly-operator"
	default:
		return ""
	}
```

Priyankasaggu11929: Thank you @laurafitzgerald. Adding a default case makes more sense.

I'll keep that as a note for future. Also I've committed the suggestion patch.

laurafitzgerald:
@Priyankasaggu11929 the pr looks fine to me. I see the JIRA is with the Customer Success Team though so happy for someone there to also review. @jessesarn if you could route this to someone appropriate that would be great. 👍

jessesarn:
Linked this PR to the JIRA and set the ticket to In Review.  Somebody from Customer Success will review it soon.

controllers/rhmi/prometheusRules.go
pepedocs: @laurafitzgerald good point on `consumed by anything or if they are just informational for SRE`. Do you think it's safe to remove this label? I'm thinking of even removing this label if it's not consumed by anyone since `product` is already available.

Priyankasaggu11929: Hi @laurafitzgerald,

As Jose pointed above, I was also wondering if we could remove `addon` label provided it is not consumed by anything else. Since the `product` label is giving information on what type of addon is installed (`rhoam` or `rhmi`), that might be enough.

Please give your views. Thanks!

CathalOConnorRH:
I'm happy with the comments above, and have verified both installs on cluster
/LGTM

laurafitzgerald:
/approve 
as per approval above.

PR #1653, opened 2021-03-11
add options step to use IN_PROW variable for local dev

laurafitzgerald:
# Description
Adding step to use IN_PROW for local dev steps for clusters that are smaller than the required

## Type of change

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,6 @@ Ensure that the cluster satisfies minimal requirements:
 - RHOAM (managed-api): 18 vCPU. More details can be found in the [service definition](https://access.redhat.com/articles/5534341) 
   under the "Resource Requirements" section
 
-Consider using IN_PROW [optional variable](#variables-table) in case of limited resources.  
 ### 1. Clone the integreatly-operator
 Only if you haven't already cloned. Otherwise, navigate to an existing copy. 
 ```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -74,7 +74,17 @@ Include the `INSTALLATION_TYPE`. See [here](#variables-table) about this and oth
...
 
-### 3. Run integreatly-operator
+### 3. Create a cr (optional)
+
+If you are running RHOAM against a cluster which is smaller than the requirements listed above, you 
+should use the IN_PROW variable, otherwise the installation will not complete. 
+If you have a cluster which meets the requirements, please skip this step.
+
+```shell script
+INSTALLATION_TYPE=managed-api IN_PROW=true make deploy/integreatly-rhmi-cr.yml
```
makslion, line 84: https://github.com/integr8ly/integreatly-operator/blob/master/README.md#variables-table at the beginning (line 111 - have no idea how to link the readme lines of code, sorry) of the section I've mentioned creating CR, but it's obvious that nobody can find it there. Wouldn't it be better to more that "section" here instead? So not only IN_PROW variable but all others will be honoured. And, maybe, link the table. Also, in case we are moving section, we shall update/remove the link from `Consider using IN_PROW [optional variable](#variables-table) in case of limited resources.` in here  https://github.com/integr8ly/integreatly-operator/blob/master/README.md#local-development

laurafitzgerald: Yes, make sense.

makslion: https://github.com/integr8ly/integreatly-operator#2-prepare-your-cluster and here as well in line `See [here](#variables-table) about this and other optional configuration variables.` <- we won't need that line then as the "optional configuration" will be just after the section

makslion:
There are two links that won't work: 
1. `Consider using IN_PROW [optional variable](#variables-table) in case of limited resources.` [here](https://github.com/integr8ly/integreatly-operator/blob/master/README.md#local-development)
2. `See [here](#variables-table) about this and other optional configuration variables.` [here](https://github.com/integr8ly/integreatly-operator/blob/master/README.md#2-prepare-your-cluster)

The first could be updated, while the second, probably, shall be deleted along with that sentence.

laurafitzgerald:
@makslion thanks for review. I've removed one command as the user would now see it in step 3 and updated the other one. 
Let me know if this makes sense now or if there are any other updates required. 👍

README.md
```diff
@@ -69,12 +68,31 @@ cd integreatly-operator
...
+
+### 3. Configuration (optional)
+
+If you are running RHOAM against a cluster which is smaller than the requirements listed above, you 
+should use the IN_PROW variable, otherwise the installation will not complete. 
+If you have a cluster which meets the requirements, this step can be skipped.
+Please see the table below for other configuration options.
+
+```shell script
+INSTALLATION_TYPE=managed-api IN_PROW=true USE_CLUSTER_STORAGE=<true/false> make deploy/integreatly-rhmi-cr.yml
```
makslion, line 85: It looks.... rather feels strange now. Hear (read?) me out. Before with just one variable, it was more like an example. Now with all 3, it looks like a suggestion. INSTALLATION_TYPE and IN_PROW with only one "option" can be interpreted as "we recommend to do so" while USE_CLUSTER_STORAGE with `<.../...>` could be "take a look at the table and chose". But I'm not even sure this is a thing - could be too paranoid.

README.md
```diff
@@ -69,12 +68,31 @@ cd integreatly-operator
...
 INSTALLATION_TYPE=<managed/managed-api> make cluster/prepare/local
 ```
 
-### 3. Run integreatly-operator
+
+### 3. Configuration (optional)
+
+If you are running RHOAM against a cluster which is smaller than the requirements listed above, you 
+should use the IN_PROW variable, otherwise the installation will not complete. 
+If you have a cluster which meets the requirements, this step can be skipped.
```
makslion, line 81: Now, when it is not only about IN_PROW it looks like configuration with ll the variables only for clusters that do not meet requirements. The human might skip the whole step if a cluster is big. And we write about IN_PROW in two places now: here and in the table. We are saying here that the variable for a `cluster smaller than the requirements` and in table `for small clusters` - those are not equal (should have spotted it earlier, sorry). Not sure what is a proposition here. Probably, remove line 81 or write "if you are not happy with defaults". As for the IN_PROW, probably, move all details to the table, leaving here a warning (in bold?) with "installation might fail because of the cluster size - check IN_PROW in table"

README.md
laurafitzgerald: Yes, For this case IN_PROW is mostly used for RHOAM(managed-api) situation. I'm not even sure if it can be used in the case of RHMI. 🤔  so it was intentional that i didn't give the two options here but was really specific about what should be run. In the first instance I left out USE_CLUSTER_STORAGE but when I moved the table down I decided to include it but in the small cluster case it doesn't matter really what the user choses for that value.

README.md
laurafitzgerald: Yes, I take your point. I'm trying to be specific about the IN_PROW as it's the common one that catches folks out but we should also be generic as well. Also see my comment above. If you have any suggestion feel free to include it. I'm not sure what's best here tbh.

PR #1655, opened 2021-03-12
[MGDAPI-1348] reduced severity of misleading errors

makslion:
# Description
https://issues.redhat.com/browse/MGDAPI-1348

## Type of change

## Checklist

### DIFF
--- a/pkg/products/grafana/reconciler.go
+++ b/pkg/products/grafana/reconciler.go
@@ -373,7 +373,7 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, client k8sclient.C
 		if !k8serr.IsNotFound(err) {
 			return integreatlyv1alpha1.PhaseFailed, err
 		}
-		return integreatlyv1alpha1.PhaseAwaitingComponents, fmt.Errorf("waiting for prometheus service for grafana datasource to become available, %w", err)
+		return integreatlyv1alpha1.PhaseAwaitingComponents, nil
 	}
 
 	var upstreamPort int32
--- a/pkg/products/monitoring/reconciler.go
+++ b/pkg/products/monitoring/reconciler.go
@@ -211,7 +211,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 
 	isMultiAZCluster, err := resources.IsMultiAZCluster(ctx, serverClient)
 	if err != nil {
-		r.Log.Error("Error when deciding if the cluster is multi-az or not. Defaulted to false:", err)
+		r.Log.Warning("Failure when deciding if the cluster is multi-az or not. Defaulted to false: " + err.Error())
 	}
 
 	phase, err = r.ReconcileNamespace(ctx, operatorNamespace, installation, serverClient, r.Log)
@@ -250,7 +250,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 	r.Log.Infof("ReconcileAlertManagerConfigSecret", l.Fields{"phase": phase})
 	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
 		events.HandleError(r.recorder, installation, phase, "Failed to reconcile alert manager config secret", err)
-		r.Log.Error("failed to reconcile alert manager config secret", err)
+		r.Log.Warning("failed to reconcile alert manager config secret " + err.Error())
 		return phase, err
 	}
 
@@ -264,7 +264,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 	phase, err = r.reconcileDashboards(ctx, serverClient)
 	r.Log.Infof("reconcileDashboards", l.Fields{"phase": phase})
 	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
-		r.Log.Error("Error reconciling dashboards", err)
+		r.Log.Warning("Failure reconciling dashboards " + err.Error())
 		events.HandleError(r.recorder, installation, phase, "Failed to reconcile dashboards", err)
 		return phase, err
 	}
@@ -575,7 +575,7 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, serverClient k8scl
 
 	antiAffinityRequired, err := resources.IsAntiAffinityRequired(ctx, serverClient)
 	if err != nil {
-		r.Log.Error("Error when deciding if monitoring pod anti affinity is required. Defaulted to false:", err)
+		r.Log.Warning("Failure when deciding if monitoring pod anti affinity is required. Defaulted to false: " + err.Error())
 	}
 
 	owner.AddIntegreatlyOwnerAnnotations(m, r.installation)
--- a/pkg/products/rhsso/reconciler.go
+++ b/pkg/products/rhsso/reconciler.go
@@ -261,7 +261,7 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
 	host := r.Config.GetHost()
 	if host == "" {
 		r.Log.Info("URL for Keycloak not yet available")
-		return integreatlyv1alpha1.PhaseAwaitingComponents, fmt.Errorf("Host for Keycloak not yet available")
+		return integreatlyv1alpha1.PhaseAwaitingComponents, nil
 	}
 
 	r.Log.Infof("Operation result", l.Fields{"keycloak": kc.Name, "result": or})
--- a/pkg/products/threescale/reconciler.go
+++ b/pkg/products/threescale/reconciler.go
@@ -280,7 +280,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 	r.log.Infof("reconcileOutgoingEmailAddress", l.Fields{"phase": phase})
 	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
 		if err != nil {
-			r.log.Error("Failed to reconcileOutgoingEmailAddress", err)
+			r.log.Warning("Failed to reconcileOutgoingEmailAddress: " + err.Error())
 			events.HandleError(r.recorder, installation, phase, "Failed to reconcileOutgoingEmailAddress", err)
 		}
 		return phase, err
@@ -952,7 +952,7 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, serverClient k8scl
 
 	antiAffinityRequired, err := resources.IsAntiAffinityRequired(ctx, serverClient)
 	if err != nil {
-		r.log.Error("error when deciding if pod anti affinity is required. Defaulted to false", err)
+		r.log.Warning("Failure when deciding if pod anti affinity is required. Defaulted to false: " + err.Error())
 		antiAffinityRequired = false
 	}
 
@@ -1391,7 +1391,7 @@ func (r *Reconciler) reconcileOutgoingEmailAddress(ctx context.Context, serverCl
 	} else {
 		accessToken, err := r.GetAdminToken(ctx, serverClient)
 		if err != nil {
-			r.log.Error("Failed to get admin token in reconcileOutgoingEmailAddresss", err)
+			r.log.Info("Failed to get admin token in reconcileOutgoingEmailAddresss: " + err.Error())
 			return integreatlyv1alpha1.PhaseInProgress, err
 		}
 		_, err = r.tsClient.SetFromEmailAddress(existingSMTPFromAddress, *accessToken)
@@ -1448,12 +1448,12 @@ func (r *Reconciler) reconcileRHSSOIntegration(ctx context.Context, serverClient
 
 	accessToken, err := r.GetAdminToken(ctx, serverClient)
 	if err != nil {
-		r.log.Error("Failed to get admin token", err)
+		r.log.Info("Failed to get admin token: " + err.Error())
 		return integreatlyv1alpha1.PhaseInProgress, err
 	}
 	_, err = r.tsClient.GetAuthenticationProviderByName(rhssoIntegrationName, *accessToken)
 	if err != nil && !tsIsNotFoundError(err) {
-		r.log.Error("Failed to get authentication provider:", err)
+		r.log.Info("Failed to get authentication provider:" + err.Error())
 		return integreatlyv1alpha1.PhaseInProgress, err
 	}
 	if tsIsNotFoundError(err) {
@@ -1468,7 +1468,7 @@ func (r *Reconciler) reconcileRHSSOIntegration(ctx context.Context, serverClient
 			"published":                         "true",
 		}, *accessToken)
 		if err != nil || res.StatusCode != http.StatusCreated {
-			r.log.Error("Failed to add authentication provider", err)
+			r.log.Info("Failed to add authentication provider:" + err.Error())
 			return integreatlyv1alpha1.PhaseInProgress, err
 		}
 	}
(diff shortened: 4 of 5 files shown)


makslion:
@laurafitzgerald @briangallagher @philbrookes @KevFan Please, take a look at changes here. As persons whose code I'm modifying you are most likely to tell me if I'm doing a horrible mistake. Please, comment on any line you would find inappropriate.

pkg/products/grafana/reconciler.go
```diff
@@ -373,7 +373,7 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, client k8sclient.C
 		if !k8serr.IsNotFound(err) {
 			return integreatlyv1alpha1.PhaseFailed, err
 		}
-		return integreatlyv1alpha1.PhaseAwaitingComponents, fmt.Errorf("waiting for prometheus service for grafana datasource to become available, %w", err)
+		return integreatlyv1alpha1.PhaseAwaitingComponents, nil
```
Boomatang, line 376: I think we should still pass out the error in this case. We need to handle this error where ever it's been returned to.

pkg/products/monitoring/reconciler.go
```diff
@@ -211,7 +211,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 
 	isMultiAZCluster, err := resources.IsMultiAZCluster(ctx, serverClient)
 	if err != nil {
-		r.Log.Error("Error when deciding if the cluster is multi-az or not. Defaulted to false:", err)
+		r.Log.Warning("Error when deciding if the cluster is multi-az or not. Defaulted to false: " + err.Error())
```
Boomatang, line 214: Remove the word `Error` from the warning message. This was one of the issues with the word "error" showing up in greps of the logs.

pkg/products/monitoring/reconciler.go
```diff
@@ -575,7 +575,7 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, serverClient k8scl
 
 	antiAffinityRequired, err := resources.IsAntiAffinityRequired(ctx, serverClient)
 	if err != nil {
-		r.Log.Error("Error when deciding if monitoring pod anti affinity is required. Defaulted to false:", err)
+		r.Log.Warning("Error when deciding if monitoring pod anti affinity is required. Defaulted to false: " + err.Error())
```
Boomatang, line 578: Same here remove the word "error" from the warning message

pkg/products/rhsso/reconciler.go
```diff
@@ -261,7 +261,7 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
 	host := r.Config.GetHost()
 	if host == "" {
 		r.Log.Info("URL for Keycloak not yet available")
-		return integreatlyv1alpha1.PhaseAwaitingComponents, fmt.Errorf("Host for Keycloak not yet available")
+		return integreatlyv1alpha1.PhaseAwaitingComponents, nil
```
Boomatang, line 264: Same as before, we should still pass out the error message and let what ever is receiving the error log it or not.

pkg/products/threescale/reconciler.go
```diff
@@ -952,7 +952,7 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, serverClient k8scl
 
 	antiAffinityRequired, err := resources.IsAntiAffinityRequired(ctx, serverClient)
 	if err != nil {
-		r.log.Error("error when deciding if pod anti affinity is required. Defaulted to false", err)
+		r.log.Warning("error when deciding if pod anti affinity is required. Defaulted to false: " + err.Error())
```
Boomatang, line 955: Remove the word "error" from the warning message

pkg/products/threescale/reconciler.go
```diff
@@ -1391,7 +1391,7 @@ func (r *Reconciler) reconcileOutgoingEmailAddress(ctx context.Context, serverCl
 	} else {
 		accessToken, err := r.GetAdminToken(ctx, serverClient)
 		if err != nil {
-			r.log.Error("Failed to get admin token in reconcileOutgoingEmailAddresss", err)
+			r.log.Info("Failed to get admin token in reconcileOutgoingEmailAddresss: " + err.Error())
```
Boomatang, line 1394: Why was this changed to `r.log.Info` and not `r.log.Warning`? I have no preferences just asking the question. And this goes for the other instances in this file also.

pkg/products/threescale/reconciler.go
```diff
@@ -1698,47 +1698,47 @@ func (r *Reconciler) reconcileServiceDiscovery(ctx context.Context, serverClient
 func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, installation *integreatlyv1alpha1.RHMI, client k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
 	cfg, err := r.ConfigManager.ReadMonitoring()
 	if err != nil {
-		return integreatlyv1alpha1.PhaseInProgress, fmt.Errorf("error reading monitoring config: %w", err)
+		return integreatlyv1alpha1.PhaseInProgress, nil
```
Boomatang, line 1701: We should still return the error message and let what ever consumes the massage decide if it should be logged.

pkg/products/threescale/reconciler.go
```diff
@@ -1698,47 +1698,47 @@ func (r *Reconciler) reconcileServiceDiscovery(ctx context.Context, serverClient
...
 
 	err = monitoring.CreateBlackboxTarget(ctx, "integreatly-3scale-admin-ui", monitoringv1alpha1.BlackboxtargetData{
 		Url:     r.Config.GetHost() + "/" + r.Config.GetBlackboxTargetPathForAdminUI(),
 		Service: "3scale-admin-ui",
 	}, cfg, installation, client)
 	if err != nil {
-		r.log.Error("Error creating threescale blackbox target", err)
-		return integreatlyv1alpha1.PhaseInProgress, fmt.Errorf("error creating threescale blackbox target: %w", err)
+		r.log.Info("Error creating threescale blackbox target: " + err.Error())
+		return integreatlyv1alpha1.PhaseInProgress, nil
```
Boomatang, line 1710: We should still return the error message and last few changes should still return the error messages

Boomatang (commented):
In the place where we were returning the error message I think we should still do that but it would be not harm to check what is receiving the error messages.

Do have a question why `Warning` is used in some places and `Info` is used in others.

pkg/products/monitoring/reconciler.go
makslion: for `Log.Error` it shows and `ERRO` message. 
For example: `ERRO[2021-03-15T13:55:16Z] Rollout system-app deployment                 error="deploymentconfigs.apps.openshift.io \"system-app\" not found" product=3scale
` 
So for the grep it should be fine

pkg/products/grafana/reconciler.go
makslion: it returns it here https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/grafana/reconciler.go#L144-L148
and `err` get passed here https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/events/events.go#L28-L31 
and never seen after 
Would it be better to log it as INFO/WARNING instead?

pkg/products/threescale/reconciler.go
makslion: I've been following this idea ->  https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/threescale/reconciler.go#L1120-L1123
`INFO` for the cases when it is expected and nothing happened. 
`WARNING` for the cases when something went wrong (missing the value)  but I will be handled (defaulting to values) or when reconciling and got an error but the stage is not `StageFailed` yet

pkg/products/monitoring/reconciler.go
Boomatang: That would be correct but if someone is using grep incorrectly for example `grep -i 'erro' log.txt` This would also show up this warning message as it starts with "Error".

pkg/products/grafana/reconciler.go
makslion: Found an example of this phase -> https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/cloudresources/reconciler.go#L350-L352 
I assumed this is how we want to handle it

Boomatang: The function for this message can only pass out an error or nil. It's up to the `HandleError` to decide what to do with the error message once it gets it. I still feel we should pass this message back out to the receiver. Maybe we need a second opinion on this!

pkg/products/threescale/reconciler.go
makslion: Was following this pattern https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/cloudresources/reconciler.go#L327-L330 
But it seems that There should be a `Log.INFO` to be able to track it

pkg/products/grafana/reconciler.go
makslion: It makes sense, but should then all other `PhaseAwaitingComponents` be changed to return an error? For the name of consistency

pkg/products/monitoring/reconciler.go
makslion: Oh, yes, then it definitely should be changed. Will `failure` fit in here? Technically: Error -> Fault -> Failure and only the "Failure" we can observe.

makslion:
There was a proposition to modify `events.HandleEvent(...)` so it can consume and log not only errors on the `PhaseFailed` but also warnings or info messages for other phases: 
```
if err != nil {
	switch phase {
	case integreatlyv1alpha1.PhaseFailed:
		reconsiler.Log.Error(...)
	case integreatlyv1alpha1.PhaseAwaitingComponents:
		reconsiler.Log.Info(...)
	default:
		reconsiler.Log.Warning(...)
	}
}
```
So we would have one place to log it all.
It is related to the current PR as I had modified `Log.` in recosilers, but might fall out of the scope of this ticket.

makslion:
removed `error` in strings for `Log.Info` and `Log.Warning` so they will not come from greps

KevFan (commented):
Overall looks good :+1: 

However, want to point out some stuff that may provide some conversation.

I think part of the problem here is that:
1. There is some reliance of returning an error also when the phase returned is not `integreatlyv1alpha1.PhaseFailed`. I think this is due to this error is eventually propagated upwards to the the CR `lastError` field that we may also see / use as some sort of last status field also on certain scenarios that is not an error state but could become one if the reconciler never gets passed this part of the code. 

So I think we may want to be careful on if some `err` is no longer returned, things may also become harder to debug as this error is no longer displayed on the CR and we would now need to dig through the logs (also not sure would CSSRE be okay with this) :thinking: 
* https://github.com/integr8ly/integreatly-operator/blob/master/controllers/rhmi/rhmi_controller.go#L358-L362

2. In certain cases we often return `integreatlyv1alpha1.PhaseInProgress, err` if we hit an unexpected error. This I think is due to that some errors are transient during an install and we believe it will eventually reconcile back. However, a case can also be made that technically it's an unexpected failed state from a code perspective, so maybe `integreatlyv1alpha1.PhaseFailed` should be used instead and return with an error. 

If we're assuming that `integreatlyv1alpha1.PhaseInProgress` and some other states are just  waiting for resources / install, then error should be nil, otherwise maybe it would be better to evaluate is the correct phase being returned if we want to return an error :thinking: 

I think these 2 points apply especially with the 3scale log changes here, but maybe this is out of scope for this ticket. Some food for thought :shallow_pan_of_food:  :thinking:

pkg/products/threescale/reconciler.go
```diff
@@ -1698,47 +1698,47 @@ func (r *Reconciler) reconcileServiceDiscovery(ctx context.Context, serverClient
...
-		return integreatlyv1alpha1.PhaseInProgress, fmt.Errorf("error creating threescale blackbox target (system-developer): %w", err)
+		r.log.Info("Failed to create blackbox target (system-developer): " + err.Error())
+		return integreatlyv1alpha1.PhaseInProgress, nil
 	}
 
 	// Create a blackbox target for the master console ui
 	threescaleRoute, err = r.getThreescaleRoute(ctx, client, "system-master", nil)
 	if err != nil {
-		return integreatlyv1alpha1.PhaseInProgress, fmt.Errorf("error getting threescale system-master threescaleRoute: %w", err)
+		return integreatlyv1alpha1.PhaseInProgress, nil
```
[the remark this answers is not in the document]
makslion: As mentioned before, `log.Info` or `log,Warning` logically should be here, but when I came to check on how it is done in the rest of the code: just return `nil` and that's it. Personally, I would prefer to have as many logs as possible, but does it means that we (I) need to refactor all returns with Phases to log something if `err != nil` ? It seems like a following up ticket to be created, but I'm not sure as don't have much experience.

pkg/products/threescale/reconciler.go
makslion: In this particular case, if I'm not wrong, `err` means "not yet, just waiting for the thing" (or could mean), so, technically, it is expected (?). On a conceptual level, this ticket was not about the reduction of the severity of errors but about making severity consistent withing the codebase. If this to be changed to `warning` it means that in every other place where we saying "not ready yet, just waiting` it also should be a warning (at this moment it is `Info` everywhere). Would like to highlight, that there was a brilliant idea of making `event.HandleError` do the thing for us: always return an error and make sure we always call `event.HandleError` and there we can unify the way we treat each of the states (for example, always `log.Warning` for the `PhaseInProgress`

pkg/products/threescale/reconciler.go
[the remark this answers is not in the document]
KevFan: I understand, but I think this is an error state. Referencing my main review comment, perhaps it's better to more so evaluate whether `integreatlyv1alpha1.PhaseInProgress` is the correct state to use here. In other parts of the code where a BlackBoxTarget is created, `integreatlyv1alpha1.PhaseFailed` is returned with an error. 
* https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/rhssocommon/reconciler.go#L529-L543
* https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/ups/reconciler.go#L298-L313

If we decide to keep using `integreatlyv1alpha1.PhaseInProgress` and an error is not returned here. The error is not logged anywhere and if the reconciler becomes stuck at this point for some reason, it becomes hard to debug / find the problem so there is loss also here for CS SRE :man_shrugging:

pkg/products/threescale/reconciler.go
KevFan: Well it depends on what caused the `err`. To me the core function used would only return an error if there was an issue in creating the `BlackBoxTarget` so it's not waiting anything if an error is returned :thinking: 
* https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/monitoring/reconciler.go#L816-L864

This to me is at least a warning since it's not expected or should happen frequently. Again referencing my main review comment, maybe `integreatlyv1alpha1.PhaseInProgress` is not the right phase to return here if there is an error, we should have a conversation with CS SRE on whether this loss of propagation of the error to the CR `lastError` field is okay on the off chance the reconciler is stuck at this point for some reason :thinking:

pkg/products/threescale/reconciler.go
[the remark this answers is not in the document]
makslion: Oh, yes. Then it definitely should be a `phaseFailed`

makslion:
Changed all blackbox target creations to return `PhaseFailed` along with an error.

PR #1664, opened 2021-03-19
Remove patch to CatalogSource in RBAC resource

sergioifg94:
# Description

Remove unnecessary patch that renamed the resource `rhmi-registry-cs` to
`rhoam-registry-cs` in RHOAM installations. This is not needed because
the operator always uses `rhmi-registry-cs` regardless of installation
type.

## Verification steps

1. Run `cluster/prepare/local`
    ```sh
    ```
2. Check that the role applied to the operator SA references `rhmi-registry-cs`

## Type of change

## Checklist

### DIFF
--- a/config/rbac-rhoam/kustomization.yaml
+++ b/config/rbac-rhoam/kustomization.yaml
@@ -8,11 +8,3 @@ patchesStrategicMerge:
   - patches/leader_election_role_binding.yaml
   # - patches/auth_proxy_role_binding.yaml
   - patches/role_binding.yaml
-
-patchesJSON6902:
-- target:
-    name: manager-role
-    group: rbac.authorization.k8s.io
-    version: v1
-    kind: ClusterRole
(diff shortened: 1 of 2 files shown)


MStokluska:
Verified with CRO and RHOAM on byoc cluster. Cataloge source is updated as expected!
Only question I have, is there any advantage in keeping it commented out instead of removing it along with the cataloge source patch file?

/lgtm

sergioifg94:
@MStokluska @matskiv There's a ticket to narrow the scope of the permissions given to either RHOAM or RHMI. The solution might be similar to what this PR is reverting (patching the manifest) so if that's the case it can be used for reference:
https://issues.redhat.com/browse/MGDAPI-688

sergioifg94:
@matskiv Sounds good. Just pushed the changes to the PR

matskiv:
Great. Thank you @sergioifg94 !

/lgtm
/approve

PR #1669, opened 2021-03-22
allow uninstall during installation

makslion:
# Description
JIRA: https://issues.redhat.com/browse/MGDAPI-1253
There is a bug that blocks uninstallation if it was triggered during the installation process. It must be a precise time and only on a fresh cluster. 
The reason for this: https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/rhssocommon/reconciler.go#L104-L108
Those lines of code are producing an error `no matches for kind \"KeycloakUser\" in version \"keycloak.org/v1alpha1\`  if there is no Keycloak CRD in a cluster and uninstallation loops. The way to fix it right now is to remove `finalizer.rhsso.integreatly.org` finalized from RHOAM CR 

## Verification steps
For complete verification, there will be two deletions: "traditional" with CRDs created and the "edge case". To make verification easier I propose to modify the code in `pkg/resources/marketplace/manager.go` at the line 86 to be as following: 
```
log.Infof("Creating subscription in ns if it doesn't already exist", l.Fields{"ns": t.Namespace})
if !(t.Namespace == "redhat-rhoam-rhsso-operator") {
	err = serverClient.Create(ctx, sub)
}
```
This change will not proceed the installation with "Edge Case" further than needed. Otherwise, there will be a tight window to trigger an uninstall. 
Additionally, If the "Edge Case" being tested not on the fresh cluster, please, consider running the following command: 
```
 oc delete -f manifests/integreatly-rhsso/12.0.3/ 
```
Otherwise, the uninstallation will never reach the "Edge Case" covered in this PR. 

The "Edge Case" verification: 
1. Install RHOAM: 
```
make cluster/prepare/local
SELF_SIGNED_CERTS=false IN_PROW=true make deploy/integreatly-rhmi-cr.yml
make code/run
```
2. Trigger an uninstall when you see the following output: 
```
INFO[2021-03-22T14:45:37Z] Creating subscription in ns if it doesn't already exist  component=marketplace ns=redhat-rhoam-rhsso-operator
```
If you have not modified `manager.go` and the message appeared more then once you would need to run  ` oc delete -f manifests/integreatly-rhsso/12.0.3/ ` and make another attempt to test this change
3. Verify uninstallation succeeds

Traditional Uninstallation 
1. same as before
2. Trigger uninstallation when the `Keycloak` resource appears in the UI search tab (After the second time `Creating subscription in ns if it doesn't already exist` appears during the installation) 
3. Verify installation succeeds 

## Type of change

## Checklist

### DIFF
--- a/apis/v1alpha1/addtoscheme_integreatly_v1alpha1.go
+++ b/apis/v1alpha1/addtoscheme_integreatly_v1alpha1.go
@@ -3,6 +3,7 @@ package v1alpha1
 import (
 	upsv1alpha1 "github.com/aerogear/unifiedpush-operator/pkg/apis/push/v1alpha1"
 	monitoring "github.com/integr8ly/application-monitoring-operator/pkg/apis/applicationmonitoring/v1alpha1"
+	apiextensionv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 
 	envoyconfigv1 "github.com/3scale/marin3r/apis/marin3r/v1alpha1"
 	discoveryservicev1 "github.com/3scale/marin3r/apis/operator/v1alpha1"
@@ -90,5 +91,6 @@ func init() {
 		envoyconfigv1.SchemeBuilder.AddToScheme,
 		discoveryservicev1.SchemeBuilder.AddToScheme,
 		apiextensionv1beta1.SchemeBuilder.AddToScheme,
+		apiextensionv1.SchemeBuilder.AddToScheme,
 	)
 }
--- a/pkg/products/rhssocommon/reconciler.go
+++ b/pkg/products/rhssocommon/reconciler.go
@@ -3,12 +3,12 @@ package rhssocommon
 import (
 	"context"
 	"fmt"
+	apiextensionv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"strings"
 
 	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
-	grafanav1alpha1 "github.com/integr8ly/grafana-operator/v3/pkg/apis/integreatly/v1alpha1"
-
 	monitoringv1alpha1 "github.com/integr8ly/application-monitoring-operator/pkg/apis/applicationmonitoring/v1alpha1"
+	grafanav1alpha1 "github.com/integr8ly/grafana-operator/v3/pkg/apis/integreatly/v1alpha1"
 	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/apis/v1alpha1"
 	"github.com/integr8ly/integreatly-operator/pkg/config"
 	"github.com/integr8ly/integreatly-operator/pkg/products/monitoring"
@@ -101,9 +101,22 @@ func (r *Reconciler) CleanupKeycloakResources(ctx context.Context, inst *integre
 		Namespace: ns,
 	}
 
+	keycloakCRD := &apiextensionv1.CustomResourceDefinition{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "keycloaks.keycloak.org",
+		},
+	}
+	crdExists, err := resources.Exists(ctx, serverClient, keycloakCRD)
+	if err != nil {
+		r.Log.Error("Error checking Keycloak CRD existence: ", err)
+		return integreatlyv1alpha1.PhaseFailed, err
+	} else if !crdExists {
+		return integreatlyv1alpha1.PhaseCompleted, nil
+	}
+
 	// Delete all users
 	users := &keycloak.KeycloakUserList{}
-	err := serverClient.List(ctx, users, opts)
+	err = serverClient.List(ctx, users, opts)
 	if err != nil {
 		return integreatlyv1alpha1.PhaseFailed, err
 	}
--- a/pkg/products/rhssocommon/reconciler_test.go
+++ b/pkg/products/rhssocommon/reconciler_test.go
@@ -5,6 +5,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/integr8ly/integreatly-operator/pkg/resources"
+	apiextensionv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"testing"
 
 	"github.com/integr8ly/integreatly-operator/pkg/resources/constants"
@@ -15,8 +17,6 @@ import (
 	controllerruntime "sigs.k8s.io/controller-runtime"
 
 	keycloakCommon "github.com/integr8ly/keycloak-client/pkg/common"
-	//k8serr "k8s.io/apimachinery/pkg/api/errors"
-	//"k8s.io/apimachinery/pkg/runtime/schema"
 
 	threescalev1 "github.com/3scale/3scale-operator/pkg/apis/apps/v1alpha1"
 	kafkav1alpha1 "github.com/integr8ly/integreatly-operator/apis-products/kafka.strimzi.io/v1alpha1"
@@ -24,7 +24,6 @@ import (
 	moqclient "github.com/integr8ly/integreatly-operator/pkg/client"
 	"github.com/integr8ly/integreatly-operator/pkg/config"
 
-	//"github.com/integr8ly/integreatly-operator/pkg/resources"
 	"github.com/integr8ly/integreatly-operator/pkg/resources/marketplace"
 	keycloak "github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1"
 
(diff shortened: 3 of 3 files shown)


makslion:
there was a typo. Fixed

pkg/products/rhssocommon/reconciler.go
```diff
@@ -100,9 +100,21 @@ func (r *Reconciler) CleanupKeycloakResources(ctx context.Context, inst *integre
...
+	keycloakCR := &keycloak.Keycloak{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      string(integreatlyv1alpha1.ProductRHSSO),
+			Namespace: inst.Namespace,
+		},
+	}
+	_, err := resources.Exists(ctx, serverClient, keycloakCR)
+	if err != nil && !k8serr.IsForbidden(err) {
+		return integreatlyv1alpha1.PhaseCompleted, nil
+	}
```
KevFan, line 113: Would be nice to have a unit test for this :thinking:

KevFan (commented):
Looks good ! :+1: 
What do you think on adding a unit test for testing the correct phase is returned for the added code ? :thinking:

pkg/products/rhssocommon/reconciler.go
makslion: Sorry, that it took me so long :sweat: . I've added unit tests and also refactored code to directly check the presence of CRD instead of relying on created CRs. Tested on the cluster - it works. Verification steps remain the same

pkg/products/rhssocommon/reconciler.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	grafanav1alpha1 "github.com/integr8ly/grafana-operator/v3/pkg/apis/integreatly/v1alpha1"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
```
KevFan, line 10: Looks good ! :+1: Though I think it would be good to use `v1` of `apiextensions` to align with this PR https://github.com/integr8ly/integreatly-operator/pull/1685 :thinking:

makslion: I've tried `v1` and `apiextensions` and none of them was recognized by the live cluster. It seemed odd once I found this line in the CRD `apiVersion: apiextensions.k8s.io/v1`  
The error I"m getting while trying with anything but v1beta1: `no kind is registered for the type v1.CustomResourceDefinition in scheme \"pkg/runtime/scheme.go:101\"` :thinking:

KevFan: Sounds like maybe https://github.com/integr8ly/integreatly-operator/blob/master/apis/v1alpha1/addtoscheme_integreatly_v1alpha1.go#L57-L92 needs to be added to / updated to work with `v1` :thinking:

makslion: isn't it already? https://github.com/integr8ly/integreatly-operator/blob/master/apis/v1alpha1/addtoscheme_integreatly_v1alpha1.go#L92 :thinking:

KevFan: `v1beta1` is but `v1` is not from what I can see, which might be why it didn't work when changed to `v1` :thinking: 
(side note: If this does work using `v1`, would be good to use `apiextensionv1` as the import name to follow #1685)

makslion: sorry for being so silly. Updated and it was tested on a cluster. Also updated unit tests to work with v1

apis/v1alpha1/addtoscheme_integreatly_v1alpha1.go
```diff
@@ -3,6 +3,7 @@ package v1alpha1
 import (
 	upsv1alpha1 "github.com/aerogear/unifiedpush-operator/pkg/apis/push/v1alpha1"
 	monitoring "github.com/integr8ly/application-monitoring-operator/pkg/apis/applicationmonitoring/v1alpha1"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
KevFan, line 6:
```suggestion
	apiextensionv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```

apis/v1alpha1/addtoscheme_integreatly_v1alpha1.go
```diff
@@ -90,5 +91,6 @@ func init() {
 		envoyconfigv1.SchemeBuilder.AddToScheme,
 		discoveryservicev1.SchemeBuilder.AddToScheme,
 		apiextensionv1beta1.SchemeBuilder.AddToScheme,
+		v1.SchemeBuilder.AddToScheme,
```
KevFan, line 94:
```suggestion
		apiextensionv1.SchemeBuilder.AddToScheme,
```

pkg/products/rhssocommon/reconciler.go
```diff
@@ -3,12 +3,12 @@ package rhssocommon
 import (
 	"context"
 	"fmt"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
KevFan, line 6:
```suggestion
	apiextensionv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```

pkg/products/rhssocommon/reconciler.go
```diff
@@ -101,9 +101,22 @@ func (r *Reconciler) CleanupKeycloakResources(ctx context.Context, inst *integre
 		Namespace: ns,
 	}
 
+	keycloakCRD := &v1.CustomResourceDefinition{
```
KevFan, line 104:
```suggestion
	keycloakCRD := &apiextensionv1.CustomResourceDefinition{
```

pkg/products/rhssocommon/reconciler_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/integr8ly/integreatly-operator/pkg/resources"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
KevFan, line 9:
```suggestion
	apiextensionv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```

pkg/products/rhssocommon/reconciler_test.go
```diff
@@ -879,3 +880,101 @@ type mockClientContext struct {
 func getLogger() l.Logger {
 	return l.NewLoggerWithContext(l.Fields{l.ProductLogContext: integreatlyv1alpha1.ProductRHSSO})
 }
+
+func TestReconciler_CleanupKeycloakResources(t *testing.T) {
+	scheme, err := getBuildScheme()
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = v1.AddToScheme(scheme)
```
KevFan, line 889:
```suggestion
	err = apiextensionv1.AddToScheme(scheme)
```

pkg/products/rhssocommon/reconciler_test.go
```diff
@@ -879,3 +880,101 @@ type mockClientContext struct {
...
+						DeletionTimestamp: &metav1.Time{},
+					},
+				},
+				serverClient: moqclient.NewSigsClientMoqWithScheme(scheme),
+			},
+			want:    integreatlyv1alpha1.PhaseCompleted,
+			wantErr: false,
+		},
+		{
+			name: "Test uninstallation: missing v1beta1 API returns phaseFailed",
```
KevFan, line 938:
```suggestion
			name: "Test uninstallation: missing v1 API returns phaseFailed",
```

pkg/products/rhssocommon/reconciler_test.go
```diff
@@ -46,7 +48,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
-	//fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
```
KevFan, line 49: Lets remove this instead of commenting out

KevFan (commented):
Looks good overall :+1: 

Just have some small suggestions mainly to use `apiextensionv1` as import name to align with https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/marketplace/manifestHelper.go#L19 and make the import easier to read if it ends up used elsewhere

pkg/products/rhssocommon/reconciler_test.go
makslion: Done. There couple of commented imports as well. https://github.com/makslion/integreatly-operator/blob/MGDAPI-1253/pkg/products/rhssocommon/reconciler_test.go#L20-L29 They also to be removed manually? :thinking:

KevFan: Yes, I think they can be removed also. They can later be re-imported if they end up being used again :thinking:

makslion: Removed all unused imports from changed files

KevFan:
Nice work ! :+1: 

Verified edge case uninstall and normal uninstall 

/lgtm

PR #1671, opened 2021-03-23
Rhoam 1 4 0 test case updates

psturc:
# Description
https://issues.redhat.com/browse/MGDAPI-1407

### DIFF
--- a/test-cases/tests/backup-restore/j05b-verify-3scale-backup-and-restore.md
+++ b/test-cases/tests/backup-restore/j05b-verify-3scale-backup-and-restore.md
@@ -10,7 +10,7 @@ products:
       - 0.1.0
       - 0.2.0
       - 1.0.0
-      - 1.4.0
+      - 1.5.0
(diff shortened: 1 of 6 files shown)


psturc:
ping @trepel

trepel:
@psturc It looks ok to me, I just don't understand the combination of both `target` and `manual-selection`. If test case is tagged by `manual-selection` it won't be picked every three releases by `tools.sh plan...`. Is it intentional?

I think each test case tagged by `manual-selection` should contain some information about when it should be selected.

psturc:
> @psturc It looks ok to me, I just don't understand the combination of both `target` and `manual-selection`. If test case is tagged by `manual-selection` it won't be picked every third release by `tools.sh plan...`. Is it intentional?
> 
> I think each test case tagged by `manual-selection` should contain some information about when it should be selected.

That is a good point, I don't know what I was thinking when creating these test cases :) Probably a copy-pasta issue

PR #1673, opened 2021-03-23
Added if statement to backup/restore script to handle script being run without aws cli

jjkiely:
# Description
While testing the functionality of this script, I ran into an issue where after the script had run for some time, it eventually failed due to an AWS error, this is because I did not have the aws cli installed at the time. I have updated the script so that it will abort if it does not detect aws cli, to save time as well as avoid any data being left on the cluster.


MStokluska:
@jjkiely going to close it for now, feel free to re-open please
/close

PR #1679, opened 2021-03-26
fix: test: add product names to all test cases

psturc:
# Description
I noticed that some test cases are not correctly added to Polarion, because they are missing important metadata (product name), so our tooling is no longer updating/uploading these test cases.

Adding the product names fixed the issue.


psturc:
ping @trepel

trepel:
@psturc looks good, I think this testcase is missing?
https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/uninstallation/o03-verify-sendgrid-credentials-are-removed.md

psturc:
oops, missed that one, thanks 👍

PR #1680, opened 2021-03-26
mgdapi-1346 - update to C03 test

MStokluska:
# Description
Jira https://issues.redhat.com/browse/MGDAPI-1346

## Type of change

Updated the alert mechanism works test to reduce the time it takes for the test to complete.
Also, updated the tests output (added some successful messages):
image

## Verification
Verify that the e2e tests on this PR passed, or run locally

### DIFF
--- a/test/common/alerts_mechanism.go
+++ b/test/common/alerts_mechanism.go
@@ -7,18 +7,13 @@ import (
 	"strings"
 	"time"
 
-	appsv1 "github.com/openshift/api/apps/v1"
 	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	"gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/util/retry"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
-
-type repeatFunc func()
 
 type alertManagerConfig struct {
 	Global struct {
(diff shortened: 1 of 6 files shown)


test/common/alerts_mechanism.go
```diff
@@ -31,73 +26,79 @@ type alertManagerConfig struct {
...
+	originalOperatorReplicas, err := getNumOfReplicasDeployment(keycloakOperatorDeploymentName, ctx.KubeClient)
 	if err != nil {
 		t.Errorf("failed to get number of replicas: %s", err)
 	}
 
 	// verify that alert to be tested is not firing before starting the test
-	err = getThreescaleAlertState(ctx)
+	err = getKeycloakAlertState(ctx)
 	if err != nil {
 		t.Fatal("failed to get threescale alert state", err)
```
Boomatang, line 51:
```suggestion
		t.Fatal("failed to get keycloak alert state", err)
```

test/common/alerts_mechanism.go
```diff
@@ -31,73 +26,79 @@ type alertManagerConfig struct {
...
 	}
 
-	if threescaleAlertsFiring {
-		t.Log("3Scale alerts firing unexpectedly")
+	if keycloakAlertsFiring {
+		t.Log("Keycloak alerts firing unexpectedly")
 		t.FailNow()
 	}
 
-	// scale down Threescale operator and product pods and verify that threescale alert is firing
```
Boomatang, line 75: Update comment
`scale down keycloak operator and product pods and verify that keycloak alert is firing`

test/common/alerts_mechanism.go
```diff
@@ -31,73 +26,79 @@ type alertManagerConfig struct {
...
+		if keycloakAlertState != "none" {
+			keycloakAlertsFiring = true
+			t.Errorf("%s alert should not be firing", keycloakAlertName)
 		}
 	}
 
-	if threescaleAlertsFiring {
-		t.Log("3Scale alerts firing unexpectedly")
+	if keycloakAlertsFiring {
+		t.Log("Keycloak alerts firing unexpectedly")
```
Boomatang, line 64: If I am understanding correctly at this point we don't expect to see alerts firing. Something that might be more clear in the logs is `Keycloak alerts firing unexpectedly before deployments are scaled down`

test/common/alerts_mechanism.go
```diff
@@ -31,73 +26,79 @@ type alertManagerConfig struct {
...
 		t.FailNow()
 	}
 
-	// scale down Threescale operator and product pods and verify that threescale alert is firing
+	t.Log("Scaling down keycloak operator deployment")
+	scaleDeployment(t, keycloakOperatorDeploymentName, 0, ctx.KubeClient)
+
+	t.Log("Keycloak alerts are not firing - performing tests")
```
Boomatang, line 71: To be more accurate in the logs this statement would need to go before the `t.Log("Scaling down keycloak operator deployment")` on line 68

test/common/alerts_mechanism.go
```diff
@@ -31,73 +26,79 @@ type alertManagerConfig struct {
...
-	err = checkThreescaleOperatorReplicasAreReady(ctx, t, originalOperatorReplicas)
+	err = checkKeycloakOperatorReplicasAreReady(ctx, t, originalOperatorReplicas)
 	if err != nil {
-		t.Log("Error During verify the operator has been scaled backup")
-		t.Fatal(err)
+		t.Fatalf("Error During verify the operator has been scaled backup with error: %s", err)
 	}
+	t.Log("Keycloak operator deployment scaled back up")
 
 	// verify that threescale alert is not firing
```
Boomatang, line 88: update comment

test/common/alerts_mechanism.go
```diff
@@ -31,73 +26,79 @@ type alertManagerConfig struct {
...
+	t.Log("Keycloak alerts are not firing - performing tests")
 	err = performTest(t, ctx, originalOperatorReplicas)
 	if err != nil {
-		t.Log("Error During scale down 3Scale operator and product pods and verify that 3scale alert is firing")
+		t.Log("Error During scale down keycloak operator and product pods and verify that keycloak alert is firing")
 		t.Fatal(err)
 	}
 
+	t.Log("Scaling up keycloak operator deployment")
+	scaleDeployment(t, keycloakOperatorDeploymentName, originalOperatorReplicas, ctx.KubeClient)
```
Boomatang, line 79: This function needs to happen even if the test fails at an earlier stage such as the error check on line 73. Please add this as a deferred function.

test/common/tests.go
```diff
@@ -95,9 +96,7 @@ var (
 		{"F08 - Verify Replicas Scale correctly in User SSO", TestReplicasInUserSSO},
 	}
 
-	FAILURE_TESTS = []TestCase{
-		/*FLAKY*/ {"C03 - Verify that alerting mechanism works", TestIntegreatlyAlertsMechanism},
-	}
+	FAILURE_TESTS = []TestCase{}
```
Boomatang, line 99: Why was this moved out of the failure tests. This is not a happy path test as it test functionally of a install that is not in a healthy state.

Boomatang (commented):
Most of this looks good but there is a large concern. 

The changes can leave the deployments scaled down if there is a failure in test which will leave the cluster in a bad state. This may have unexpected side affects on other tests. No matter how or when the test finishes the cluster should be returned to the same state as before the test started. 

This is only the code review and won't be running the code till the concern above is addressed.

test/common/alerts_mechanism.go
MStokluska: This is addressed by line @68

test/common/tests.go
MStokluska: Reason for this is that I had a different understanding of what "FAILURE_TESTS" are. Thanks for explaining what it is!

test/common/alerts_mechanism.go
MStokluska: I moved the scaleDeployment into perform tests instead. Thanks!

test/common/tests.go
MStokluska: I've also updated the prow e2e to run the FAILURE_TESTS as it seems that we were not executing this test on a PR submission.

MStokluska:
> Most of this looks good but there is a large concern.
> 
> The changes can leave the deployments scaled down if there is a failure in test which will leave the cluster in a bad state. This may have unexpected side affects on other tests. No matter how or when the test finishes the cluster should be returned to the same state as before the test started.
> 
> This is only the code review and won't be running the code till the concern above is addressed.

Thanks for review. I've addressed the comments and updated the PR, going to re-run rhoam-e2e and e2e to verify.

PR #1684, opened 2021-03-30
Readme - how to deploy the operator from a custom image

jjaferson:
# Description

Adds how to deploy the operator from a custom image to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,10 @@ The operator can now be run locally:
 ```shell
 INSTALLATION_TYPE=<managed/managed-api> make code/run
 ```
+If you want to run the operator from a specific image, you can specify the image and run `make cluster/deploy`
+```shell
+IMAGE_FORMAT=<image-registry-address> INSTALLATION_TYPE=managed-api  make cluster/deploy
+```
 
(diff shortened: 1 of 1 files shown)


Boomatang:
I don't know, I only found out about this command. Can you use the `IN_PROW` flag with this?

README.md
```diff
@@ -114,6 +114,13 @@ Once the installation completed the command wil result in following output:
 "complete"
 ```
 
+### 6. Deploy a custom image of the operator
```
laurafitzgerald, line 117: I would say this isn't step 6. of the install but rather an alternative to step 4 of running the operator.

jjaferson: Yep, it make more sense actually.

jjaferson:
> I don't know, I only found out about this command. Can you use the `IN_PROW` flag with this?

there is a section for that already https://github.com/integr8ly/integreatly-operator#3-configuration-optional

Boomatang:
> > I don't know, I only found out about this command. Can you use the `IN_PROW` flag with this?
> 
> there is a section for that already https://github.com/integr8ly/integreatly-operator#3-configuration-optional

I can expect that answer

PR #1689, opened 2021-04-01
test: test-case for sku package

psturc:
# Description
test case for SKU package


psturc:
ping @laurafitzgerald

PR #1691, opened 2021-04-01
allow unistall while postgress crd in not present

makslion:
# Description
JIRA: https://issues.redhat.com/browse/MGDAPI-1406
Uninstallation fails in CRD is not present (not complete installation process). 
Added checks for that case and if CRD is missing, then return `PhaseCompleeted`. 
Replica of this PR https://github.com/integr8ly/integreatly-operator/pull/1669 

## Verification steps
To verify you will need two installations: with and without CRD. If you have the `Postgres` and `PostgresSnapshots` CRSs, consider running the following command: 
```
oc delete -f manifests/integreatly-cloud-resources/0.23.0   
```
This error is hard to catch, so I suggest you modifying [this](https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/marketplace/manager.go#L86) line to be as follows: 
```
if !(t.Namespace == "redhat-rhoam-cloud-resources-operator") {
	err = serverClient.Create(ctx, og)
}
if (t.Namespace == "redhat-rhoam-cloud-resources-operator"){
	log.Warning("DELETION TIME!")
}
```

### Uninstallation without CRD
1. run the following command to start the installation process: 
```
ocm_cluster_login <cluster> 
INSTALLATION_TYPE=managed-api IN_PROW=true make deploy/integreatly-rhmi-cr.yml
INSTALLATION_TYPE=managed-api IN_PROW=true make code/run
```
2. Once you see a warning (you have modified the manager.go) OR info log with `Creating subscription in ns if it doesn't already exist ... ns=redhat-rhoam-cloud-resources-operator` (if you have not modified manager.go) trigger an uninstallation. 
3. Verify that CRD is not present on a cluster by `oc get -f manifests/integreatly-cloud-resources/0.23.0 | grep postgres`
4. Verify uninstallation succeded 

### Uninstallation with CRD 
0. Undo the `manager.go` modification if it was made
1. install as before 
2. wait for the second info log with `Creating subscription in ns if it doesn't already exist ... ns=redhat-rhoam-cloud-resources-operator`
3. trigger uninstallation 
4. Verify CRD is present 
5. Verify uninstallation succeded 
 
## Type of change

## Checklist

### DIFF
--- a/pkg/products/cloudresources/reconciler_test.go
+++ b/pkg/products/cloudresources/reconciler_test.go
@@ -1,0 +1,244 @@
+package cloudresources
+
+import (
+	"context"
+	threescalev1 "github.com/3scale/3scale-operator/pkg/apis/apps/v1alpha1"
+	monitoringv1 "github.com/integr8ly/application-monitoring-operator/pkg/apis/applicationmonitoring/v1alpha1"
+	crov1 "github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1"
+	kafkav1alpha1 "github.com/integr8ly/integreatly-operator/apis-products/kafka.strimzi.io/v1alpha1"
+	"github.com/integr8ly/integreatly-operator/apis/v1alpha1"
+	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/apis/v1alpha1"
+	moqclient "github.com/integr8ly/integreatly-operator/pkg/client"
+	"github.com/integr8ly/integreatly-operator/pkg/config"
+	"github.com/integr8ly/integreatly-operator/pkg/resources"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/logger"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/marketplace"
+	keycloak "github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1"
+	oauthv1 "github.com/openshift/api/oauth/v1"
+	projectv1 "github.com/openshift/api/project/v1"
+	routev1 "github.com/openshift/api/route/v1"
+	usersv1 "github.com/openshift/api/user/v1"
+	coreosv1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1"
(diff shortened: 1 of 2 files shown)


KevFan:
Verified uninstallation without CRD and uninstallation with CRD :+1: 

/lgtm

matskiv:
Could there be a scenario when PostgresSnapshot CRD doesn't exist, but RedisSnapshot CRs do exist?

makslion:
only if crd.yaml is missing from the folder. 
https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/marketplace/manifestHelper.go#L27-L56
In theory, there is no need for the second CRD detection (inside CleanupResources) since it never will be reached for the edge case 
https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/cloudresources/reconciler.go#L109-L132
But I've decided to include it anyway to be double sure :thinking:

matskiv:
@makslion Sorry, I didn't describe the scenario in my question very well. But after reading your response and checking the code I realized that it can not happen...

Thanks for adding the unit test 👍 
/approve

P.S: Is that typo in unit test intentionally to test if the reviewers read the test code? 😄

makslion:
Oh lord, what typo? You mean typo here https://github.com/makslion/integreatly-operator/blob/MGDAPI-1406/pkg/products/cloudresources/reconciler_test.go#L134
Where it should be "Postgres Snapshots"? :thinking:

pkg/products/cloudresources/reconciler_test.go
```diff
@@ -0,0 +1,244 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
+)
+
+func TestReconciler_cleanupResources(t *testing.T) {
+	scheme, err := getBuildScheme()
+	if err != nil {
+		t.Fatalf("Error obtaining sheme")
```
matskiv, line 34: I only noticed this one^ :) 
`sheme` => `scheme`

makslion: Shame on me :disappointed:

PR #1694, opened 2021-04-06
Mgdapi 1355 Updating the Golang Version to 1.16.2

Boomatang:
# Description

Update the go version being used in the operator to 1.16.2

**Jira** https://issues.redhat.com/browse/MGDAPI-1355

## Verify 
* code builds locally
* all docker images are created

## Type of change

## Checklist


Makefile
```diff
@@ -112,8 +112,8 @@ GOBIN=$(shell go env GOBIN)
 endif
 
 .PHONY: setup/moq
-setup/moq:
-	go install github.com/matryer/moq
+setup/moq: code/gen
+	GO111MODULE=off go get github.com/matryer/moq
```
matskiv, line 116: I think we shouldn't be changing `go install` to `go get` 
AFAIK:
`go install` installs moq binary to make it available via CLI
`go get` installs moq as library dependency

matskiv: Also, why is `code/gen` added above?

make/osde2e.mk
```diff
@@ -15,4 +15,4 @@ image/osde2e/build/push: image/osde2e/build image/osde2e/push
 
 .PHONY: test/compile/osde2e
 test/compile/osde2e:
-	CGO_ENABLED=0 go test -v -c -o managed-api-test-harness.test ./test/osde2e -ginkgo.noColor
```
matskiv, line 18: This change seems unrelated to Go version bump,
highlighting in case it was unintentionally added to the commit.

Makefile
Boomatang: I am sure Oleg, but on Monday this week or last week you came up with that fix we where having around moq. :confused:

make/osde2e.mk
Boomatang: On go 1.16, go test will not compile if extra flags are give to the `-c` flag. This was an intentionally commit.

matskiv: Okay, good to know.
Thank you for explanation 👍

Makefile
matskiv: Yes, that's true, but later I realized my mistake and I thought we discussed it.
Sorry for the confusion.

PR #1696, opened 2021-04-06
fix potential nil pointers when logging

KevFan:
# Description

There can be potential nil pointer when calling `err.Error()` for logging in certain conditional blocks that can cause a panic 

## Type of change

## Validation
* `e2e` and `rhoam-e2e` prow tests should pass 

## Checklist


KevFan:
@makslion @Boomatang

makslion:
looks great! Not sure I'm a reliable person for an `lgtm` though :thinking:

Boomatang:
Looks good to me. I feel bad that we didn't see that the first time. I think we should wait for @makslion to add the `lgtm` label. This is as good a place to start adding the labels.

makslion:
/lgtm
but with trembling hands because of the weight of responsibility

PR #1698, opened 2021-04-07
make dynamic sku work rhmi safe

laurafitzgerald:
# Description

## Type of change

## Checklist


controllers/rhmi/rhmi_controller.go
```diff
@@ -408,9 +410,11 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 		if installation.Spec.RebalancePods {
 			r.reconcilePodDistribution(installation)
 		}
-		if installationSKU.IsUpdated() {
-			installation.Status.SKU = installationSKU.GetName()
-			installation.Status.ToSKU = ""
+		if installation.Spec.Type == string(rhmiv1alpha1.InstallationTypeManagedApi) {
+			if installationSKU.IsUpdated() {
```
jjaferson, line 414: Small change here - I think we can avoid having two if statement here

```suggestion
		if installation.Spec.Type == string(rhmiv1alpha1.InstallationTypeManagedApi)  && installationSKU.IsUpdated() {
```

jjaferson:
Verified the changes on an OSD cluster installing the managed install type and everything worked as expected.

PR #1699, opened 2021-04-07
[WIP] rate limit alert dashboard config

laurafitzgerald:
# Description
https://issues.redhat.com/browse/MGDAPI-1237

## Changes in this pr

- update grafana dashboard to only contain the hard limit
- remove soft limit from dashboard
- remove soft limit alerts from prometheus
- update envoy configmap based on skuconfig

### Verification

Install the operator using the following commands

```
INSTALLATION_TYPE=managed-api IN_PROW=true make deploy/integreatly-rhmi-cr.yml
```
Wait for the installation to fully complete. You can verify this by
oc get rhmi rhoam -n redhat-rhoam-operator -o yaml | yq - r 'status.toSKU' -> nil

##### Grafana Dashboard

The grafana dashboard should no longer show any soft limits and have a hard limit which matches the hard limit set in the config. In the case of the first default installation the value should be 347 - see image below

Screenshot 2021-04-07 at 15 12 44

##### Alerts

Navigate to prometheus 
- verify that there are no softlimit related alerts
- verify that the percentage of hard limit alerts are still present - see image below

Screenshot 2021-04-07 at 15 14 19

##### Rate Limit Configuration 

Verify that the rate limit configuration gets update to match the rate limit specified in the sku

`oc get configmap ratelimit-config -n redhat-rhoam-marin3r -o yaml | yq r - 'data'`

Requests per limit should be 347

##### Validate that changing the SKU updates all of the above

Change the sku using the following command 
`INSTALLATION_TYPE=managed-api QUOTA=20 make cluster/prepare/sku`

Step through all the previous validation confirming that the values are now `1389`

## Type of change

## Checklist


pkg/resources/sku/sku.go
```diff
@@ -172,6 +176,9 @@ func (p AProductConfig) Configure(obj metav1.Object) error {
 		return errors.New(fmt.Sprintf("sku configuration can only be applied to Deployments, StatefulSets or Deployment Configs, found %s", reflect.TypeOf(obj)))
 	}
 
+	if replicas == nil {
+		*replicas = defaultReplicas
```
makslion, line 180: Shouldn't it be `replicas = &defaultReplicas`? Now we are "if the pointer is nill, store in that nill location my value".  IT should be "you have a nill? here is an address you need"

psturc:
👀

makslion:
Failing e2e tests - it is known and to be fixed https://issues.redhat.com/browse/MGDAPI-1488
Here we have A13 not being happy and this is expected

psturc:
/lgtm

verified locally. just noticed `Marin3rRateLimitServiceEndpointDown` alert firing, which will be addressed in https://issues.redhat.com/browse/MGDAPI-1390

image

laurafitzgerald:
@psturc thanks for the verification. The issue you saw will be addressed by -> https://github.com/integr8ly/integreatly-operator/pull/1712

PR #1690, opened 2021-04-01
MDGAPI-1450 - Fix 3scale user updated back to original none lower cased username

KevFan:
# Description

Jira:
* https://issues.redhat.com/browse/MGDAPI-1450
## Verification
**Note: Requires setting up GIthub IDP and signing in with Github User with uppercase letter in username. Reach out if you don't have one**
* Checkout this branch 
* Install RHOAM
```
export INSTALLATION_TYPE=managed-api  
make cluster/prepare/local
SELF_SIGNED_CERTS=false IN_PROW=true make deploy/integreatly-rhmi-cr.yml
make code/run
```
* Once installed, set up Github IDP on cluster via ocm
* Sign into 3scale as admin using `system-seed` secret credentials
* In incognito tab, sign in to 3scale as github user 
* Verify github username is lowercased
* Promote github user to admin
* As github user, update username to any value
* Verify username is updated back and is **lowercased** 
* Verify admin permissions are **not** reverted

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [ ] Verified independently on a cluster by reviewer

PR #1700, opened 2021-04-08
fix: update test case h27

psturc:
# Description
Update h27 test case based on https://github.com/integr8ly/integreatly-operator/pull/1690


psturc:
ping @trepel

PR #1702, opened 2021-04-09
SKU metric

jackdelahunt:
# Description
Added a new metric for the currrent SKU configuration for RHOAM. The metric stores the current config, and if the config is changing what it is changing to. This metric also has an annotation in Grafana so when the SKU config is changed it can be seen in Grafana.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Checklist
- [ ] This change requires a documentation update 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added a test case that will be used to verify my changes 
- [ ] Verified independently on a cluster by reviewer

## Verify
- Log into a cluster with the changes (ask me for details)
- Change the SKU configuration of RHOAM -> [here](https://github.com/jackdelahunt/integreatly-operator/blob/170b7aaed0c4573b02c386365766038ae602af71/Makefile#L363)
- Verify that the annotation appears on Grafana by checking a dashboard

image

PR #1706, opened 2021-04-12
SKU metric

jackdelahunt:
# Description
Added a new metric for the currrent SKU configuration for RHOAM. The metric stores the current config, and if the config is changing what it is changing to. This metric also has an annotation in Grafana so when the SKU config is changed it can be seen in Grafana.

## Type of change

## Checklist

## Verify
- Install RHOAM based on an image with new changes (ask me for image details)
   - Run make cluster/deploy with IMAGE_FORMAT set to the image details I will supply
   - Run make cluster/prepare/sku to create the sku secret in the operator namespace
- Change the SKU configuration
   - Select a value to change the sku value to. Any value that 'name' is set to in the sku-config config map in the operator namespace is a valid entry.
   - Run make cluster/prepare/sku again with a QUOTA set to the new sku value (5 should work fine).
- Verify that the annotation appears on Grafana by checking a dashboard

image

### DIFF
--- a/pkg/products/monitoring/dashboards/clusterResources.go
+++ b/pkg/products/monitoring/dashboards/clusterResources.go
@@ -37,12 +37,12 @@ func GetMonitoringGrafanaDBClusterResourcesJSON(installationName string) string 
 				"hide": false,
 				"iconColor": "#FADE2A",
 				"limit": 100,
-				"name": "Active SKU",
+				"name": "SKU",
 				"showIn": 0,
 				"step": "",
 				"tagKeys": "stage,sku,toSKU",
 				"tags": "",
-				"titleFormat": "Active SKU",
+				"titleFormat": "SKU Change (million per day)",
 				"type": "tags",
 				"useValueForTime": false
 			}
--- a/pkg/products/monitoring/dashboards/clusterResources.go
+++ b/pkg/products/monitoring/dashboards/clusterResources.go
@@ -33,7 +33,7 @@ func GetMonitoringGrafanaDBClusterResourcesJSON(installationName string) string 
 			{
 				"datasource": "Prometheus",
 				"enable": true,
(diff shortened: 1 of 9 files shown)


jackdelahunt:
This is a cleaned up version of [this](https://github.com/integr8ly/integreatly-operator/pull/1702) PR, check it out for some history and context.

pkg/products/monitoring/dashboards/clusterResources.go
```diff
@@ -29,6 +29,22 @@ func GetMonitoringGrafanaDBClusterResourcesJSON(installationName string) string
...
+				"expr": "count by (stage,sku,toSKU)(active_sku{toSKU!=\"\"})",
+				"hide": false,
+				"iconColor": "#FADE2A",
+				"limit": 100,
+				"name": "SKU",
+				"showIn": 0,
+				"step": "",
+				"tagKeys": "stage,sku,toSKU",
+				"tags": "",
+				"titleFormat": "SKU Change (million per day)",
```
Boomatang, line 45: Why is this saying "(million per day)" that seems like a lot or a misleading title?  I see this is also set in a few other places.

Boomatang (commented):
Code looks okay, will validate next. There is just one question on the text of the `titleFromat` and I wonder is this something that docs may need to know about.

pkg/products/monitoring/dashboards/clusterResources.go
jackdelahunt: The annotation in the grafana dashboard only displays single digits (the same numbers in the sku config). Having "million" per day shows that when the metric is "5" it means "5 million per day".

Boomatang: Ok that makes sense, now the "SKU Change" does not seem to describe what it's titling but I am happy to let someone from docs comment on it. This is something that can be changed later if needed.

PR #1712, opened 2021-04-13
updated sku-config.yaml

makslion:
# Description
https://issues.redhat.com/browse/MGDAPI-1390

### Verification steps 
- Compare values in the .yaml with corresponding from a ticket linked above 

## Type of change

## Checklist

### DIFF
--- a/config/configmap/sku-config.yaml
+++ b/config/configmap/sku-config.yaml
@@ -4,101 +4,244 @@ metadata:
   name: sku-config
 data:
   sku-configs: |-
-    [
-    {
-              "name": "20",
-              "rate-limiting": {
-                  "unit": "minute",
-                  "requests_per_unit": 1389
-              },
-              "resources": {
-                  "backend_listener": {
-                      "replicas": 3,
-                      "resources": {
-                          "requests": {
-                              "cpu": 0.25,
-                              "memory": 450
-                          },
-                          "limits": {
-                              "cpu": 0.3,
-                              "memory": 500
-                          }
-                      }
-                  },
-                  "backend_worker": {
-                      "replicas": 3,
-                      "resources": {
-                          "requests": {
-                              "cpu": 0.15,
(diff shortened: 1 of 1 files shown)


config/configmap/sku-config.yaml
```diff
@@ -4,89 +4,243 @@ metadata:
...
+                            "memory": "500Mi"
+                        }
+                    }
+                },
+                "backend_worker": {
+                    "replicas": 3,
+                    "resources": {
+                        "requests": {
+                            "cpu": 0.15,
+                            "memory": "100Mi
```
laurafitzgerald, line 111:
```suggestion
                            "memory": "100Mi"
```

laurafitzgerald:
@makslion this looks fine. we can verify that the values are propagated as expected for each configuration as a verification step.

makslion:
Discussed with @laurafitzgerald -> propagation of values will be verified as part of each product PR. This one only needs to have correct values in the .yaml

KevFan:
Overall looks good :+1: . Just have 2 questions:

1. Grafana block seems to still in the 5 million and 1 million config, but removed from the rest, wondering are they still needed or kept for a reason ? (ticket does not have values for grafana) :thinking: 
2. Is there a list of expected `requests_per_unit` value for each config to compare or is okay to just multiple that value by `60 * 24` ? They were aproximately equal to the total sku value but just wondering is there a formal list to compare against :thinking:

makslion:
1. Updated the config. Well spotted. It was an artefact and there is no need to keep it there now. 
2. `requests_per_unit` = SKU_Quota / 24 / 60 -> this is how they were calculated, yes. They are approximately equal to the SKU quota, but showing requests per minute (while SKU quota is per day and this is the only "formal list of requirements")

KevFan:
Thanks @makslion 
/lgtm

PR #1713, opened 2021-04-13
Remove flaky RHMIConfig check from B04

Boomatang:
# Description

This feature is being deprecated in rhmi and not used in rhoam.

## Type of change

## Checklist

### DIFF
--- a/test/common/user_dedicated_admin_permissions.go
+++ b/test/common/user_dedicated_admin_permissions.go
@@ -103,9 +103,6 @@ func TestDedicatedAdminUserPermissions(t TestingTB, ctx *TestingContext) {
 	// Verify Dedicated admins permissions around secrets
 	verifyDedicatedAdminSecretPermissions(t, openshiftClient, rhmi.Spec.Type)
 
-	// Verify Dedicated admin permissions around RHMI Config
-	verifyDedicatedAdminRHMIConfigPermissions(t, openshiftClient)
-
 	verifyDedicatedAdmin3ScaleRoutePermissions(t, openshiftClient)
 
 	if rhmi.Spec.Type != string(integreatlyv1alpha1.InstallationTypeManagedApi) {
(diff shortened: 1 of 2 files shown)


Boomatang:
@KevFan Sorry I am only getting back to this now. There is going to be an update here in slight change in how its going to be addressed. Thanks for having a look and I hope you will have a look once the change is pushed.

test/common/user_dedicated_admin_permissions.go
```diff
@@ -128,6 +128,9 @@ func TestDedicatedAdminUserPermissions(t TestingTB, ctx *TestingContext) {
 		// Verify dedicated admin Role / Role binding for AMQ Online resources
 		verifyDedicatedAdminAMQOnlineRolePermissions(t, ctx)
 	}
+
+	// Verify Dedicated admin permissions around RHMI Config
+	verifyDedicatedAdminRHMIConfigPermissions(t, openshiftClient)
```
KevFan, line 133: Not quite for I understand the change, this looks to me to just move the function call down but the test will still be run for RHOAM install ? :confused:

Boomatang: Yes maybe I should have explained this a bit more. We are ok with this check failing the test. When we see this fail is on the addon installs but never a master install. With the test failing earlier because of this, we don't know if there is any other checks failing.

KevFan: Ah I see, so we want the this test to be last even for RHMI installs :+1: 
In this case, I think it would be better to just removed the commented out function call at:
* https://github.com/integr8ly/integreatly-operator/blob/92878a28424723f88b7bc6e224c617e10f5a3a29/test/common/user_dedicated_admin_permissions.go#L106-L107

I think the reason why it fails for addon installs is described here:
* https://issues.redhat.com/browse/INTLY-10156?focusedCommentId=15323980&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15323980

PR #1715, opened 2021-04-13
fix in the logic which was causing cr to not be udpated

laurafitzgerald:
# Description

## Verification

Install the operator using the following steps:

`INSTALLATION_TYPE=managed-api make cluster/prepare `
`INSTALLATION_TYPE=managed-api IN_PROW=true make deploy/integreatly-rhmi-cr.yml`
`INSTALLATION_TYPE=managed-api make code/run `

Verify that the toSKU value is set on initial install.

## Type of change

## Checklist


makslion:
:eyes:

PR #1681, opened 2021-03-26
A32 Automation

makslion:
# Description
Automating test A32 - Validate SSO config

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist
- [ ] This change requires a documentation update 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added a test case that will be used to verify my changes 
- [ ] Verified independently on a cluster by reviewer

## Verification steps 
1. Make sure you have access to the cluster with RHOAM installed
2. Login to the cluster 
3. Run command: 
`INSTALLATION_TYPE=managed-api TEST=A32 make test/e2e/single`
4. Verify the test succeeds 

OR
just comment with 
`/test rhoam-e2e`

PR #1688, opened 2021-04-01
Mgdapi 1192 3

laurafitzgerald:
See https://github.com/integr8ly/integreatly-operator/pull/1604 for details but use the branch on integreatly as the feature branch

PR #1718, opened 2021-04-14
A32 Automation

makslion:
# Description
Automating test A32 - Validate SSO config
Reopened this PR https://github.com/integr8ly/integreatly-operator/pull/1681 against a feature branch implementing extraction of values from currently active SKU. 

Verification steps remain the same: 
1. Make sure you have access to the cluster with RHOAM installed
2. log in to the cluster
3. Run command:
`INSTALLATION_TYPE=managed-api TEST=A32 make test/e2e/single`
4. Verify the test succeeds
OR
just comment with
`/test rhoam-e2e`

## Type of change

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
 	github.com/go-openapi/spec v0.19.12
 	github.com/golang/protobuf v1.4.3
+	github.com/google/go-cmp v0.4.0
 	github.com/google/go-querystring v1.0.0
 	github.com/googleapis/gnostic v0.5.3 // indirect
 	github.com/headzoo/surf v1.0.0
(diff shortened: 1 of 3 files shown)


laurafitzgerald:
Hey @psturc myself and @makslion are chatting through this one. We are also in the process of automating the sku configuraiton testing which would cover the resource requests and limits for usersso in the case of RHOAM. This test case could apply in both cases (RHMI and RHOAM) wdyt about having this one enabled for both cases but the resource checks would check the hard code values that get applied to RHMI only. This is the change that would be covered by this and the SKU e2e tests -> https://github.com/integr8ly/integreatly-operator/pull/1688/files#diff-7a53ba6e3e9d30b7c003c462bb3887e0941309ec30341f32d62b494052aa7d4aR318-R334

makslion:
It seems that sku tests shall check that values were passed from config to the CR. 
Checking that values then were propagated from CR to the endpoint is the responsibility of each of the components. So in this case we need to retrieve values from Keycloak CR and from pods and compare them. What your thoughts on this?

makslion:
Updated test. Changed the logic of resources checks to compare Keycloak CR against each keycloak pod and validate requests and limits for CPU and memory.

jackdelahunt:
Installed on new cluster, test passed

PR #1730, opened 2021-04-22
Valery mgdapi 1516

valerymo:
# Description
Update availability alert for SSO to use 5xx error rate
Jira: https://issues.redhat.com/browse/MGDAPI-1516

## Type of change

## Checklist


valerymo:
Hello Craig @maleck13 , Hello Peter @pb82 , Could you please review this my PR
https://github.com/integr8ly/integreatly-operator/pull/1730
Jira: https://issues.redhat.com/browse/MGDAPI-1516
Thank you
Valery

templates/monitoring/jobs/openshift_monitoring_federation.yaml
```diff
@@ -13,6 +13,7 @@
     - 'kubelet_volume_stats_used_bytes{endpoint="https-metrics",namespace=~"{{ index .Params "namespace-prefix" }}.*"}'
     - 'kubelet_volume_stats_available_bytes{endpoint="https-metrics",namespace=~"{{ index .Params "namespace-prefix" }}.*"}'
     - 'kubelet_volume_stats_capacity_bytes{endpoint="https-metrics",namespace=~"{{ index .Params "namespace-prefix" }}.*"}'
+    - 'haproxy_backend_http_responses_total{route=~"^keycloak.*", exported_namespace=~"redhat-rhoam.*sso"}'
```
pb82, line 16: very slight improvement: `redhat-rhoam.*sso` -> `redhat-rhoam.*sso$` (ensure it ends in 'sso').

pkg/products/rhssouser/prometheusRules.go
```diff
@@ -72,6 +72,77 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
...
+				GroupName: "rhoam-user-sso-slo-availability.rules",
+				Rules: []monitoringv1.Rule{
+					{
+						Alert: "RHOAMUserSsoAvailability5mto1hErrorBudgetBurn",
+						Annotations: map[string]string{
+							"sop_url": resources.SopUrlRHOAMUserSsoAvailabilityAlert,
+							"message": "High 5m and 1h error budget burn for RHOAM SSO User",
+						},
+						Expr: intstr.FromString(`
+							sum( sum(rate(haproxy_backend_http_responses_total{route=~"^keycloak.*", exported_namespace="redhat-rhoam-user-sso", code="5xx"}[5m]))
```
KevFan, line 91: Out of curiousity, are these alerts for RHMI also ? :thinking: 

If they are, then currently I dont think they would work for an RHMI install with the hardcoded `redhat-rhoam` for namespace references so might want to use `r.Config.GetNamespace()` call instead like :thinking: 
* https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/rhsso/prometheusRules.go#L30

templates/monitoring/jobs/openshift_monitoring_federation.yaml
valerymo: Hi Peter, thank you very much for review and comments.
Fix done,  + new Alerts changed to support RHMI, so the change was:
redhat-rhoam.*sso -> redhat-.*sso$
Testing done, so hope it's ok.

pkg/products/rhssouser/prometheusRules.go
valerymo: Hi Kevin, thank you very much for your Review and comments.
New Alerts updated now to support RHMI.

pkg/products/rhssouser/prometheusRules.go
```diff
@@ -72,6 +73,77 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
...
+			//SLO-availability-alerts:
+			//https://sre.google/workbook/alerting-on-slos/
+			//https://promtools.dev/alerts/errors
+			{
+				AlertName: "user-sso-slo-availability-alerts",
+				Namespace: r.Config.GetOperatorNamespace(),
+				GroupName: "user-sso-slo-availability.rules",
+				Rules: []monitoringv1.Rule{
+					{
+						Alert: fmt.Sprintf("%sUserSsoAvailability5mto1hErrorBudgetBurn", strings.ToUpper(installationName)),
```
pb82, line 86: why prefix this with the installation type? Are there different installation types expected on the same cluster?

pb82 (approved):
Looks good, just one comment

templates/monitoring/jobs/openshift_monitoring_federation.yaml
```diff
@@ -13,6 +13,7 @@
     - 'kubelet_volume_stats_used_bytes{endpoint="https-metrics",namespace=~"{{ index .Params "namespace-prefix" }}.*"}'
     - 'kubelet_volume_stats_available_bytes{endpoint="https-metrics",namespace=~"{{ index .Params "namespace-prefix" }}.*"}'
     - 'kubelet_volume_stats_capacity_bytes{endpoint="https-metrics",namespace=~"{{ index .Params "namespace-prefix" }}.*"}'
+    - 'haproxy_backend_http_responses_total{route=~"^keycloak.*", exported_namespace=~"redhat-.*sso$"}'
```
KevFan, line 16:
```suggestion
    - 'haproxy_backend_http_responses_total{route=~"^keycloak.*", exported_namespace=~"{{ index .Params "namespace-prefix" }}.*sso$"}'
```
Should we use the `namepsace-prefix`  template parameter here like the others ? :thinking:

KevFan (commented):
Looks good overall :+1:  Just have one question :thinking:

templates/monitoring/jobs/openshift_monitoring_federation.yaml
valerymo: Thank you very much for comment. I will test it and apply. Sorry for late response, as I found mail in my private mail ... (not work, although I had both mails : ). Sorry & Thanks!

pkg/products/rhssouser/prometheusRules.go
valerymo: Thank you very much for Review. Yes, it could be 2 installation types, as according to https://github.com/integr8ly/integreatly-operator  Readme.md:  "...An Integreatly Operator can be installed using two different flavours: managed or managed-api", where "managed" is for RHMI, and "managed-api" - for RHOAM.
According to another comment, I need try check another syntax, - 'haproxy_backend_http_responses_total{route=~"^keycloak.*", exported_namespace=~"{{ index .Params "namespace-prefix" }}.*sso$"}'. I'm sorry for problems, I will send update, if any change. Thanks a lot

valerymo: Federation of metrics -- updated and verified, working correctly,  same way as previous. Current/latest syntax: exported_namespace=~"{{ index .Params "namespace-prefix" }}.*sso$"
Thank you very much for Review.

templates/monitoring/jobs/openshift_monitoring_federation.yaml
valerymo: Done - Updated and Checked for Rhoam.
Will need check all for RHMI

pkg/products/rhsso/prometheusRules.go
```diff
@@ -71,6 +72,77 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
...
+				GroupName: "rhsso-slo-availability.rules",
+				Rules: []monitoringv1.Rule{
+					{
+						Alert: fmt.Sprintf("%sRhssoAvailability5mto1hErrorBudgetBurn", strings.ToUpper(installationName)),
+						Annotations: map[string]string{
+							"sop_url": resources.SopUrlSloRhssoAvailabilityAlert,
+							"message": "High 5m and 1h error budget burn for RHSSO",
+						},
+						Expr: intstr.FromString(fmt.Sprintf(`
+							sum( sum(rate(haproxy_backend_http_responses_total{route=~"^keycloak.*", exported_namespace='%s', code="5xx"}[5m]))
```
maleck13, line 91: we may not need the regex here. This was to cover that in MAS SSO there are to keycloak routes in staging

maleck13: same for the others below

pkg/products/rhsso/prometheusRules.go
```diff
@@ -71,6 +72,77 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
...
+								/sum(rate(haproxy_backend_http_responses_total{route=~"^keycloak.*", exported_namespace='%s'}[5m])) > (14.40 * (1-0.99000)))
```
maleck13, line 92: 99% availability is an ok starting point for this SLO but our SLA is 99.95 and we would want to up this SLO after we see that it is not failing and causing a lot of noise

pkg/products/rhssouser/prometheusRules.go
```diff
@@ -72,6 +73,77 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
...
+						Annotations: map[string]string{
+							"sop_url": resources.SopUrlSloUserSsoAvailabilityAlert,
+							"message": "High 5m and 1h error budget burn for SSO User",
+						},
+						Expr: intstr.FromString(fmt.Sprintf(`
+							sum( sum(rate(haproxy_backend_http_responses_total{route=~"^keycloak.*", exported_namespace='%s', code="5xx"}[5m]))
```
maleck13, line 92: same applies here I don't think we need the regex and 99% is ok as a starting point but will soon want to up to 99.95 once we see not making too much noise. We will also want to move them to critical that point

pkg/resources/sop_url.go
```diff
@@ -35,4 +35,6 @@ const (
 	SopUrlThreeScaleDeveloperUIBBT                            = "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/rhoam/alerts/ThreeScaleDeveloperUIBBT.asciidoc"
 	SopUrlThreeScaleSystemAdminUIBBT                          = "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/rhoam/alerts/ThreeScaleSystemAdminUIBBT.asciidoc"
 	SopUrlPodDistributionIncorrect                            = "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/multi-az/pod_distribution.md"
+	SopUrlSloRhssoAvailabilityAlert                           = "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/rhoam/alerts/SloRhssoAvailabilityAlert.md"
```
maleck13, line 38: I think it would be worth looking at the MAS SSO SOP also and seeing do we want to take any of that and apply it here

maleck13:
couple of comments left. I think updating the SOP can be a follow on from this work but we should check do we actually need the regex for the route. Also could we create a follow up that calls out updating the availability target for these alerts. 
Finally we should look at adding a set of tests to prove these alerts do what we expect.

pkg/products/rhsso/prometheusRules.go
valerymo: I'm checking Prometheus in RHOAM cluster,
using query: haproxy_backend_http_responses_total {route=~"^keycloak.*", code="5xx"}
There are two keycloak routes: route="keycloak-edge" and route="keycloak"
So seems  current code is correct (?) : route=~"^keycloak.*"

pkg/products/rhssouser/prometheusRules.go
valerymo: I wanted to ask - maybe we could progress to approval/merge with current Alerts task and this PR?
Current development was verified both for RHOAM and RHMI.
We could open new tasks for improvements including testing automation.

pkg/products/rhsso/prometheusRules.go
valerymo: ok, will continue to follow

pkg/products/rhsso/prometheusRules.go
valerymo: There are only two routes for haproxy_backend_http_responses_total 5xx metrics at RHOAM: 
route="keycloak" and route="keycloak-edge"
Could I remove "route=~"^keycloak.*" expession from alert? So remains following:
'haproxy_backend_http_responses_total{exported_namespace=~"redhat-.*sso$"}'
Is it ok?

KevFan:
We would need to add the new alert rules to the C04 alert exists tests case, otherwise the e2e tests would fail 
* https://github.com/integr8ly/integreatly-operator/blob/master/test/common/alerts_exist.go#L345

test/common/alerts_exist.go
```diff
@@ -530,6 +530,24 @@ func commonExpectedRules() []alertsTestRule {
...
+			},
+		},
+		{
+			File: NamespacePrefix + "user-sso-operator-user-sso-slo-availability-alerts.yaml",
+			Rules: []string{
+				"RHOAMUserSsoAvailability5mto1hErrorBudgetBurn",
+				"RHOAMUserSsoAvailability30mto6hErrorBudgetBurn",
+				"RHOAMUserSsoAvailability2hto1dErrorBudgetBurn",
+				"RHOAMUserSsoAvailability6hto3dErrorBudgetBurn",
+			},
```
KevFan, line 549: I think this would fail for RHMI e2e but pass with RHOAM e2e test due to the name is different depending on install.
https://github.com/integr8ly/integreatly-operator/blob/36b9deaa1e97db98a8d5268d00c13a797e822d6b/pkg/products/rhssouser/prometheusRules.go#L86

We should pass the installation name into this function and format the rule names similar to this but using the upper case name :thinking: 
https://github.com/integr8ly/integreatly-operator/blob/36b9deaa1e97db98a8d5268d00c13a797e822d6b/test/common/alerts_exist.go#L555-L564

valerymo: fix done, Thank you Kevin. Hope it's better now

test/common/alerts_exist.go
```diff
@@ -350,7 +350,8 @@ func managedApiSpecificRules() []alertsTestRule {
 }
 
 // Common to all install types
-func commonExpectedRules() []alertsTestRule {
+func commonExpectedRules(installationName string) []alertsTestRule {
+	titledName := strings.Title(installationName)
```
KevFan, line 354:
```suggestion
	upperName := strings.ToUpper(installationName)
```
Can just use initalise uppercase name here instead since there is no usage of the title name in this fucntion. But this is pretty small, so am happy for it to be addressed later

KevFan:
`e2e` and `rhoam-e2e` tests passed :+1: 

/lgtm

PR #1731, opened 2021-04-22
Remove the IN_PROW annotation from RHMI sample

carlkyrillos:
# Description
Removed the IN_PROW annotation from the RHMI CR sample and removed reference to IN_PROW from Makefile because it wont be needed after the SKU work is completed.

**JIRA Ticket:** https://issues.redhat.com/browse/MGDAPI-1472

## Verify
- Checkout this branch
- Run the relevant make command: `make deploy/integreatly-rhmi-cr.yml`
- Confirm that the generated sample, config/samples/integreatly-rhmi-cr.yml, doesn't have the in_prow annotation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -410,9 +410,11 @@ deploy/integreatly-rhmi-cr.yml:
 	sed "s/INSTALLATION_PREFIX/$(INSTALLATION_PREFIX)/g" | \
 	sed "s/INSTALLATION_SHORTHAND/$(INSTALLATION_SHORTHAND)/g" | \
 	sed "s/SELF_SIGNED_CERTS/$(SELF_SIGNED_CERTS)/g" | \
-	sed "s/IN_PROW/'$(IN_PROW)'/g" | \
 	sed "s/OPERATORS_IN_PRODUCT_NAMESPACE/$(OPERATORS_IN_PRODUCT_NAMESPACE)/g" | \
 	sed "s/USE_CLUSTER_STORAGE/$(USE_CLUSTER_STORAGE)/g" > config/samples/integreatly-rhmi-cr.yml
+	# Workaround until in_prow annotation can be removed from prow
+	yq w -i config/samples/integreatly-rhmi-cr.yml metadata.annotations.in_prow "IN_PROW"
+	$(SED_INLINE) "s/IN_PROW/'$(IN_PROW)'/g" config/samples/integreatly-rhmi-cr.yml
 	@-oc create -f config/samples/integreatly-rhmi-cr.yml
 
 .PHONY: prepare-patch-release
--- a/Makefile
+++ b/Makefile
@@ -410,7 +410,6 @@ deploy/integreatly-rhmi-cr.yml:
 	sed "s/INSTALLATION_PREFIX/$(INSTALLATION_PREFIX)/g" | \
 	sed "s/INSTALLATION_SHORTHAND/$(INSTALLATION_SHORTHAND)/g" | \
 	sed "s/SELF_SIGNED_CERTS/$(SELF_SIGNED_CERTS)/g" | \
-	sed "s/IN_PROW/'$(IN_PROW)'/g" | \
 	sed "s/OPERATORS_IN_PRODUCT_NAMESPACE/$(OPERATORS_IN_PRODUCT_NAMESPACE)/g" | \
 	sed "s/USE_CLUSTER_STORAGE/$(USE_CLUSTER_STORAGE)/g" > config/samples/integreatly-rhmi-cr.yml
 	@-oc create -f config/samples/integreatly-rhmi-cr.yml
--- a/Makefile
+++ b/Makefile
@@ -410,6 +410,7 @@ deploy/integreatly-rhmi-cr.yml:
 	sed "s/INSTALLATION_PREFIX/$(INSTALLATION_PREFIX)/g" | \
 	sed "s/INSTALLATION_SHORTHAND/$(INSTALLATION_SHORTHAND)/g" | \
 	sed "s/SELF_SIGNED_CERTS/$(SELF_SIGNED_CERTS)/g" | \
+	sed "s/IN_PROW/'$(IN_PROW)'/g" | \
 	sed "s/OPERATORS_IN_PRODUCT_NAMESPACE/$(OPERATORS_IN_PRODUCT_NAMESPACE)/g" | \
 	sed "s/USE_CLUSTER_STORAGE/$(USE_CLUSTER_STORAGE)/g" > config/samples/integreatly-rhmi-cr.yml
 	@-oc create -f config/samples/integreatly-rhmi-cr.yml
--- a/config/samples/integreatly-rhmi-cr.yaml
+++ b/config/samples/integreatly-rhmi-cr.yaml
@@ -2,8 +2,6 @@ apiVersion: integreatly.org/v1alpha1
 kind: RHMI
 metadata:
   name: INSTALLATION_NAME
-  annotations:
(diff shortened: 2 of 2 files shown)


laurafitzgerald:
hi @carlkyrillos I believe there are other references to IN_PROW in the makefile which relate to e2e tests. We plan to replace the need for this by having the default QUOTA be the 1 million SKU which should fit on a prow cluster and in stage as the replica counts are 1 for some of the components now.
A good way to test this is that when we have the e2e tests passing on the feature branch that you can rebase it against that and then check if the prow tests also pass here. 
To verify that the 1 million sku works in place of the IN_PROW use in stage it would just require an install without specifying IN_PROW once that's removed from the Makefile.

carlkyrillos:
> hi @carlkyrillos I believe there are other references to IN_PROW in the makefile which relate to e2e tests running IN_PROW. We plan to replace the need for this by having the default QUOTA be the 1 million SKU which should fit on a prow cluster and in stage as the replica counts are 1 for some of the components now.
> A good way to test this is that when we have the e2e tests passing on the feature branch that you can rebase it against that and then check if the prow tests also pass here.
> To verify that the 1 million sku works in place of the IN_PROW use in stage it would just require an install without specifying IN_PROW once that's removed from the Makefile.

Hi @laurafitzgerald a separate [ticket](https://issues.redhat.com/browse/MGDAPI-1625) was created to address this work per our conversation so I believe this is resolved for this ticket.

Makefile
```diff
@@ -410,7 +410,6 @@ deploy/integreatly-rhmi-cr.yml:
 	sed "s/INSTALLATION_PREFIX/$(INSTALLATION_PREFIX)/g" | \
 	sed "s/INSTALLATION_SHORTHAND/$(INSTALLATION_SHORTHAND)/g" | \
 	sed "s/SELF_SIGNED_CERTS/$(SELF_SIGNED_CERTS)/g" | \
-	sed "s/IN_PROW/'$(IN_PROW)'/g" | \
```
laurafitzgerald, line 413: Hi @carlkyrillos as we are doing a separate ticket we should keep this line in for now as it's required by the existing logic.

KevFan:
Looks good :+1: Just the `make deploy/integreatly-rhmi-cr.yml` would still require the `in_prow` annotation until it is addressed in the other ticket as @laurafitzgerald mentioned

Makefile
carlkyrillos: Hi @laurafitzgerald I added the line back to the Makefile and re-verified that the fix is still working as expected, thank you for catching that!

carlkyrillos:
> Looks good  Just the `make deploy/integreatly-rhmi-cr.yml` would still require the `in_prow` annotation until it is addressed in the other ticket as @laurafitzgerald mentioned

Thanks @KevFan I made the change that @laurafitzgerald pointed out so it should be good to go now.

KevFan:
@carlkyrillos @laurafitzgerald  Thinking on this now, just keeping the `IN_PROW` sed in the makefile won't work with the removal of the annotion from the sample since there is nothing the replace :thinking: 
* https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L412

Might want to do something like this for `make deploy/integreatly-rhmi-cr.yml:`:
```
	@echo "selfSignedCerts = $(SELF_SIGNED_CERTS)"
	sed "s/INSTALLATION_NAME/$(INSTALLATION_NAME)/g" config/samples/integreatly-rhmi-cr.yaml | \
	sed "s/INSTALLATION_TYPE/$(INSTALLATION_TYPE)/g" | \
	sed "s/INSTALLATION_PREFIX/$(INSTALLATION_PREFIX)/g" | \
	sed "s/INSTALLATION_SHORTHAND/$(INSTALLATION_SHORTHAND)/g" | \
	sed "s/SELF_SIGNED_CERTS/$(SELF_SIGNED_CERTS)/g" | \
	sed "s/OPERATORS_IN_PRODUCT_NAMESPACE/$(OPERATORS_IN_PRODUCT_NAMESPACE)/g" | \
	sed "s/USE_CLUSTER_STORAGE/$(USE_CLUSTER_STORAGE)/g" > config/samples/integreatly-rhmi-cr.yml
	# Workaround until in_prow annotation can be removed from prow
	yq w -i config/samples/integreatly-rhmi-cr.yml metadata.annotations.in_prow "IN_PROW"
	$(SED_INLINE) "s/IN_PROW/'$(IN_PROW)'/g" config/samples/integreatly-rhmi-cr.yml
	@-oc create -f config/samples/integreatly-rhmi-cr.yml
```

This will add the `in_prow` annotation when running in prow and locally if people want it until the other ticket is resolved but  remove the annotation from the sample (i.e. also annotation won't be in the samples in the generated csv when running `make prepare/release`). What do you think ? :thinking:

carlkyrillos:
> @carlkyrillos @laurafitzgerald Thinking on this now, just keeping the `IN_PROW` sed in the makefile won't work with the removal of the annotion from the sample since there is nothing the replace 
> 
> * https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L412
> 
> Might want to do something like this for `make deploy/integreatly-rhmi-cr.yml:`:
> 
> ```
> 	@echo "selfSignedCerts = $(SELF_SIGNED_CERTS)"
> 	sed "s/INSTALLATION_NAME/$(INSTALLATION_NAME)/g" config/samples/integreatly-rhmi-cr.yaml | \
> 	sed "s/INSTALLATION_TYPE/$(INSTALLATION_TYPE)/g" | \
> 	sed "s/INSTALLATION_PREFIX/$(INSTALLATION_PREFIX)/g" | \
> 	sed "s/INSTALLATION_SHORTHAND/$(INSTALLATION_SHORTHAND)/g" | \
> 	sed "s/SELF_SIGNED_CERTS/$(SELF_SIGNED_CERTS)/g" | \
> 	sed "s/OPERATORS_IN_PRODUCT_NAMESPACE/$(OPERATORS_IN_PRODUCT_NAMESPACE)/g" | \
> 	sed "s/USE_CLUSTER_STORAGE/$(USE_CLUSTER_STORAGE)/g" > config/samples/integreatly-rhmi-cr.yml
> 	# Workaround until in_prow annotation can be removed from prow
> 	yq w -i config/samples/integreatly-rhmi-cr.yml metadata.annotations.in_prow "IN_PROW"
> 	$(SED_INLINE) "s/IN_PROW/'$(IN_PROW)'/g" config/samples/integreatly-rhmi-cr.yml
> 	@-oc create -f config/samples/integreatly-rhmi-cr.yml
> ```
> 
> This will add the `in_prow` annotation when running in prow and locally if people want it until the other ticket is resolved but remove the annotation from the sample (i.e. also annotation won't be in the samples in the generated csv when running `make prepare/release`). What do you think ?

@KevFan @laurafitzgerald Good catch thank you for spotting this. I added the changes and verified that everything is still working as expected.

KevFan:
* Verified running the `make deploy/integreatly-rhmi-cr.yml` command still deploys with annotation
image
* Verified annotation is **removed in csv** samples when generating a new release using `SEMVER=1.6.0-rc1 OLM_TYPE=managed-api-service make release/prepare` and changing https://github.com/integr8ly/integreatly-operator/blob/master/scripts/prepare-release.sh#L36 to `KUSTOMIZE=$KUSTOMIZE_PATH` to run locally

/lgtm 

Woulld be good to have `e2e` and `rhoam-e2e` prow test passing to ensure there's no side effects from using `yq` in the make command now :+1:

jjaferson:
/approve 

e2e tests are passing and based on @KevFan verification PR is good to be merged

PR #1739, opened 2021-04-26
remove check for alerts already removed

laurafitzgerald:
# Description
Test changes related to https://issues.redhat.com/browse/MGDAPI-1488

H11 - Fixes
F08 - Removes for managed api
F05 - Fixes 

## Verification 
The e2e tests should pass and the rhoam-e2e tests will fail on A15 and A13 which are being address by @makslion in a sperate pr.

Install using the following commands

`INSTALLATION_TYPE=managed-api make cluster/prepare/local`
`INSTALLATION_TYPE=managed-api IN_PROW=true make deploy/integreatly-rhmi-cr.yml`
`INSTALLATION_TYPE=managed-api make code/run`

Wait for the installation to go complete

Run H11 with `INSTALLATION_TYPE=managed-api TEST=H11 make test/e2e/single` It should pass
Run  F05 with ` INSTALLATION_TYPE=managed-api TEST=F05 make test/e2e/single`

## Type of change

## Checklist

### DIFF
--- a/pkg/config/threescale.go
+++ b/pkg/config/threescale.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	"errors"
-
 	"github.com/integr8ly/integreatly-operator/test/resources"
 
 	threescaleapps "github.com/3scale/3scale-operator/pkg/apis/apps"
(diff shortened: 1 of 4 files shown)


pkg/config/threescale.go
```diff
@@ -123,6 +122,31 @@ func (t *ThreeScale) GetReplicasConfig(inst *integreatlyv1alpha1.RHMI) map[strin
 		setDefaultNumberOfReplicas(2, threeScaleComponents)
 	}
 
+	if inst.Spec.Type == string(integreatlyv1alpha1.InstallationTypeManagedApi) {
+		switch inst.Status.Quota {
+		case "1":
+			threeScaleComponents["apicastProd"] = 1
+			threeScaleComponents["backendListener"] = 1
+			threeScaleComponents["backendWorker"] = 1
```
trepel, line 130: I am not sure if these values are ok, I guess I need to get definitive answer if 1M SKU has to be HA. But these can be changed later easily. I would go with all "3"s here though.

test/common/tests.go
```diff
@@ -106,7 +106,7 @@ var (
 
 		{
 			[]TestCase{
-				{"A34 - Verify SKU values", TestQuotaValues},
+				{"A34 - Verify QUOTA values", TestQuotaValues},
```
trepel, line 109: This looks ok but it is is dependent on https://github.com/integr8ly/integreatly-operator/pull/1688

trepel (approved):

trepel:
I haven't run through verification steps (don't have suitable cluster, I need to create one first), but the code looks ok to me.
/lgtm

pkg/config/threescale.go
laurafitzgerald: Hi @trepel I believe there's an ask to confirm the 1 million sku. Once that's confirmed it needs updating in the config map and here. As I understand it that ask is with you. Can you let me know if it's not on your backlog so that we can make sure someone has it.

test/common/tests.go
laurafitzgerald: @trepel that on should be passing now as the branch i'm based this pr on is the branch for the pr you've linked above.

pkg/config/threescale.go
trepel: Yeah, I am working on that, but I don't play with replica counts, just with requests/limits for containers. We use 3 as minimum due to HA (in multiAZ setup there are three zones so we can have one pod per zone). But yes, for now we can use whatever here, I just think '3's are more probable the right values than '1's

laurafitzgerald: Probably yes but we need to have a quota which is low enough to run in prow. @philbrookes proposed a 0 sku for that. so we can update the 1 to 3 if required.

trepel:
I followed the verification steps, the H11 and F05 passed for me when running locally. I assume this can be merged now @laurafitzgerald

PR #1741, opened 2021-04-27
fixed A13 and A15 tests

makslion:
# Description
Part of the 1488 JIRA
Fixed A13 and A15 tests to get replicas from CRs and compare them against a number of pods running on a cluster for products, configured by Quota 

## Verification steps
1. ensure you have access to a cluster with RHOAM installed from a feature branch: log in to the cluster.
2. run the following commands in your terminal:
```
INSTALLATION_TYPE=managed-api TEST=A13 make test/e2e/single
INSTALLATION_TYPE=managed-api TEST=A14 make test/e2e/single
INSTALLATION_TYPE=managed-api TEST=A15 make test/e2e/single
```

## Type of change

## Checklist

### DIFF
--- a/test/common/deployment_types.go
+++ b/test/common/deployment_types.go
@@ -2,6 +2,13 @@ package common
 
 import (
 	goctx "context"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/quota"
+	"github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1"
+	"golang.org/x/net/context"
+	k8sappsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	k8sError "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/integr8ly/integreatly-operator/pkg/config"
 	"github.com/integr8ly/integreatly-operator/test/resources"
@@ -11,7 +18,6 @@ import (
 	"github.com/integr8ly/integreatly-operator/pkg/resources/constants"
 	appsv1 "github.com/openshift/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -38,17 +44,17 @@ var (
 	}
 )
 
-func getDeploymentConfiguration(deploymentName string, inst *integreatlyv1alpha1.RHMI) Namespace {
+func getDeploymentConfiguration(deploymentName string, inst *integreatlyv1alpha1.RHMI, t TestingTB, ctx *TestingContext) Namespace {
 	threescaleConfig := config.NewThreeScale(map[string]string{})
 	replicas := threescaleConfig.GetReplicasConfig(inst)
 	deployment := map[string]Namespace{
-		"threeScaleDeployment": Namespace{
+		"threeScaleDeployment": {
 			Name: ThreeScaleOperatorNamespace,
 			Products: []Product{
 				{Name: "3scale-operator", ExpectedReplicas: 1},
 			},
 		},
-		"aMQOnlineOperatorDeployment": Namespace{
+		"aMQOnlineOperatorDeployment": {
 			Name: AMQOnlineOperatorNamespace,
 			Products: []Product{
 				{Name: "address-space-controller", ExpectedReplicas: 1},
@@ -58,19 +64,19 @@ func getDeploymentConfiguration(deploymentName string, inst *integreatlyv1alpha1
 				{Name: "standard-authservice", ExpectedReplicas: 1},
 			},
 		},
-		"cloudResourceOperatorDeployment": Namespace{
+		"cloudResourceOperatorDeployment": {
 			Name: CloudResourceOperatorNamespace,
 			Products: []Product{
 				{Name: "cloud-resource-operator", ExpectedReplicas: 1},
 			},
 		},
-		"codeReadyOperatorDeployment": Namespace{
+		"codeReadyOperatorDeployment": {
 			Name: CodeReadyOperatorNamespace,
 			Products: []Product{
 				{Name: "codeready-operator", ExpectedReplicas: 1},
 			},
 		},
-		"codereadyWorkspacesDeployment": Namespace{
+		"codereadyWorkspacesDeployment": {
 			Name: NamespacePrefix + "codeready-workspaces",
 			Products: []Product{
 				{Name: "codeready", ExpectedReplicas: 1},
@@ -78,13 +84,13 @@ func getDeploymentConfiguration(deploymentName string, inst *integreatlyv1alpha1
 				{Name: "plugin-registry", ExpectedReplicas: 1},
 			},
 		},
-		"fuseOperatorDeployment": Namespace{
+		"fuseOperatorDeployment": {
 			Name: FuseOperatorNamespace,
 			Products: []Product{
 				{Name: "syndesis-operator", ExpectedReplicas: 1},
 			},
 		},
-		"monitoringOperatorDeployment": Namespace{
+		"monitoringOperatorDeployment": {
 			Name: MonitoringOperatorNamespace,
 			Products: []Product{
 				{Name: "application-monitoring-operator", ExpectedReplicas: 1},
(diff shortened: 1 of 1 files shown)


test/common/deployment_types.go
```diff
@@ -2,6 +2,13 @@ package common
 
 import (
 	goctx "context"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/quota"
+	"github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1"
+	"golang.org/x/net/context"
+	v12 "k8s.io/api/apps/v1"
+	v13 "k8s.io/api/core/v1"
```
matskiv, line 9: Can you rename these imports please?
`v13` -> `corev1`
and `v12` to something like `k8sappsv1` ?

test/common/deployment_types.go
```diff
@@ -183,6 +189,28 @@ func getDeploymentConfiguration(deploymentName string, inst *integreatlyv1alpha1
...
+			t.Fatalf("Error getting key from ratelimit Deployment: %v", err)
+		}
+
+		err = ctx.Client.Get(context.TODO(), key, ratelimitCR)
+		if err != nil {
+			if !k8sError.IsNotFound(err) {
+				t.Fatalf("Error obtaining ratelimit CR: %v", err)
+			}
+		}
+		deployment["marin3rDeployment"].Products[1].ExpectedReplicas = *ratelimitCR.Spec.Replicas
```
KevFan, line 211:
```suggestion
		deployment["marin3rDeployment"] = Namespace{
			Name: Marin3rProductNamespace,
			Products: []Product{
				{Name: "prom-statsd-exporter", ExpectedReplicas: 1},
				{Name: "ratelimit", ExpectedReplicas: *ratelimitCR.Spec.Replicas},
			},
		}
```
What do you think on just adding the `deployment["marin3rDeployment"]` block from https://github.com/integr8ly/integreatly-operator/blob/3ae07fc47fd73ee86a7e18c86054b8d2f40c709b/test/common/deployment_types.go#L149-L155 to  here instead of just changing the replica value ?  I feel it would be slighly safer on the off chance someone in the future adds to this and messes up the array order causing the wrong struct replica value to be reassigned. Also this will make it a single place where it is added and not changed somewhere else in the code :thinking:

test/common/deployment_types.go
```diff
@@ -183,6 +189,28 @@ func getDeploymentConfiguration(deploymentName string, inst *integreatlyv1alpha1
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      quota.RateLimitName,
+				Namespace: Marin3rProductNamespace,
+			},
+		}
+
+		key, err := k8sclient.ObjectKeyFromObject(ratelimitCR)
+		if err != nil {
+			t.Fatalf("Error getting key from ratelimit Deployment: %v", err)
+		}
```
matskiv, line 203: If I am looking at the correct source code for `ObjectKeyFromObject()`,
then we don't need any error checking as it never returns errors:
https://github.com/kubernetes-sigs/controller-runtime/blob/b2c90ab82af89fb84120108af14f4ade9df5d787/pkg/client/interfaces.go#L33-L36

test/common/deployment_types.go
```diff
@@ -376,6 +478,25 @@ func TestStatefulSetsExpectedReplicas(t TestingTB, ctx *TestingContext) {
 		rhssoExpectedReplicas = 1
 		rhssoUserExpectedReplicas = 1
 	}
+	if rhmi.Spec.Type == string(integreatlyv1alpha1.InstallationTypeManagedApi) {
+		keycloakCR := &v1alpha1.Keycloak{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      quota.KeycloakName,
+				Namespace: NamespacePrefix + "user-sso",
```
KevFan, line 485:
```suggestion
				Namespace: RHSSOUserProductNamespace,
```
Assuming rebase with the latest changes from feature branch

test/common/deployment_types.go
makslion: here is an implementation https://github.com/makslion/integreatly-operator/blob/MGDAPI-1488-A13/vendor/sigs.k8s.io/controller-runtime/pkg/client/interfaces.go#L33-L39

test/common/deployment_types.go
```diff
@@ -416,6 +537,30 @@ func TestStatefulSetsExpectedReplicas(t TestingTB, ctx *TestingContext) {
 				continue
 			}
 
+			if namespace.Name == RHSSOUserProductOperatorNamespace && product.Name == "keycloak" {
+				pods := &v13.PodList{}
+				selector, _ := labels.Parse("component=keycloak")
+				keycloakListOpts := []k8sclient.ListOption{
+					k8sclient.InNamespace(NamespacePrefix + "user-sso"),
```
KevFan, line 544:
```suggestion
					k8sclient.InNamespace(RHSSOUserProductNamespace),
```
Assuming rebase with the latest changes from feature branch

test/common/deployment_types.go
matskiv: Oh, I was looking at newer version of controller-runtime...
Sorry for spam.. 😅

KevFan (commented):
Overall looks good :+1: Left some comments / questions

test/common/deployment_types.go
makslion: Yes, that's definitely the change to be made. Thanks

test/common/deployment_types.go
makslion: Missed those :see_no_evil:

test/common/deployment_types.go
```diff
@@ -266,39 +294,113 @@ func TestDeploymentExpectedReplicas(t TestingTB, ctx *TestingContext) {
...
+						k8sclient.InNamespace(ThreeScaleProductNamespace),
+						k8sclient.MatchingLabelsSelector{
+							Selector: selector,
+						},
+					}
+					err = ctx.Client.List(context.TODO(), pods, listOpts...)
+					if err != nil {
+						t.Fatalf("failed to get backend listener pods for 3scale: %v", err)
+					}
+					checkPods(t, pods, product, namespace, deployment)
```
KevFan, line 356: Bit confused on why there's these blocks for checking 3scale deployment configs here ? :thinking: 
This specific test `TestDeploymentExpectedReplicas` is for `deployments` and I don't think it would ever get 3scale deployment configs in this test 
* https://github.com/integr8ly/integreatly-operator/blob/3ae07fc47fd73ee86a7e18c86054b8d2f40c709b/test/common/deployment_types.go#L261
* https://github.com/integr8ly/integreatly-operator/blob/3ae07fc47fd73ee86a7e18c86054b8d2f40c709b/test/common/deployment_types.go#L385-L405
* https://github.com/integr8ly/integreatly-operator/blob/3ae07fc47fd73ee86a7e18c86054b8d2f40c709b/test/common/deployment_types.go#L24-L45

Should also fail here since deployment resource would not exist for 3scale deployment configs :thinking: 
* https://github.com/integr8ly/integreatly-operator/blob/3ae07fc47fd73ee86a7e18c86054b8d2f40c709b/test/common/deployment_types.go#L280-L285

Were you meaning to put these code blocks into the `TestDeploymentConfigExpectedReplicas` test ? :thinking: 
* https://github.com/integr8ly/integreatly-operator/blob/3ae07fc47fd73ee86a7e18c86054b8d2f40c709b/test/common/deployment_types.go#L407-L451

makslion: Now I'm wondering how I had those passing. Well spotted. Moved 3scale checks to the `TestDeploymentConfigExpectedReplicas`.

makslion:
rebased against feature branch to include latest changes, implemented suggestion from reviewers and removed few redundant declarations.

test/common/deployment_types.go
```diff
@@ -266,39 +300,80 @@ func TestDeploymentExpectedReplicas(t TestingTB, ctx *TestingContext) {
...
+				if deployment.Status.AvailableReplicas < product.ExpectedReplicas {
+					t.Errorf("Deployment %s in namespace %s doesn't match the number of expected available replicas. Available Replicas: %v / Expected Replicas: %v",
+						product.Name,
+						namespace.Name,
+						deployment.Status.AvailableReplicas,
+						product.ExpectedReplicas,
+					)
+					continue
+				}
 			}
```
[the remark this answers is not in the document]
makslion: and in this case those two `if` statements should be combined into one, yes :thinking:

test/common/deployment_types.go
```diff
@@ -93,61 +99,61 @@ func getDeploymentConfiguration(deploymentName string, inst *integreatlyv1alpha1
...
 				{Name: "marin3r-controller-webhook", ExpectedReplicas: 2},
 			},
 		},
-		"marin3rDeployment": Namespace{
+		"marin3rDeployment": {
 			Name: Marin3rProductNamespace,
 			Products: []Product{
 				{Name: "prom-statsd-exporter", ExpectedReplicas: 1},
 				{Name: "ratelimit", ExpectedReplicas: 3},
 			},
```
KevFan, line 154: We could remove this since it's later assigned and only applies to managed api

test/common/deployment_types.go
```diff
@@ -334,8 +407,53 @@ func TestDeploymentConfigExpectedReplicas(t TestingTB, ctx *TestingContext) {
 				)
 				continue
 			}
-
-			if deploymentConfig.Status.AvailableReplicas < product.ExpectedReplicas {
+			if rhmi.Spec.Type == string(integreatlyv1alpha1.InstallationTypeManagedApi) {
```
KevFan, line 410: Do we need this `if` block in general ? Seems we are just checking number of pods running vs expected number ? I don't there'll be much harm in having this run in RHMI e2e also unless I'm missing something ? :thinking:

test/common/deployment_types.go
```diff
@@ -334,8 +407,53 @@ func TestDeploymentConfigExpectedReplicas(t TestingTB, ctx *TestingContext) {
...
+						k8sclient.MatchingLabelsSelector{
+							Selector: selector,
+						},
+					}
+					err = ctx.Client.List(context.TODO(), pods, listOpts...)
+					if err != nil {
+						t.Fatalf("failed to get backend listener pods for 3scale: %v", err)
+					}
+					checkDeploymentConfigPods(t, pods, product, namespace, deploymentConfig)
+
```
KevFan, line 425: What do you think on having an helper function for getting returning `listOpts` ? Something like
```
func getListOptsForSelectorInNamespace(label, namespace string) []k8sclient.ListOption {
	selector, _ := labels.Parse(label)
	listOpts := []k8sclient.ListOption{
		k8sclient.InNamespace(namespace),
		k8sclient.MatchingLabelsSelector{
			Selector: selector,
		},
	}
	
	return listOpts
}
```
Then the `if` blocks would be smaller in size :thinking: 
```suggestion
					pods := &corev1.PodList{}
					listOpts := getListOptsForSelectorInNamespace("deploymentconfig=apicast-production", ThreeScaleProductNamespace)
					err = ctx.Client.List(context.TODO(), pods, listOpts...)
					if err != nil {
						t.Fatalf("failed to get backend listener pods for 3scale: %v", err)
					}
					checkDeploymentConfigPods(t, pods, product, namespace, deploymentConfig)

```

KevFan (commented):
Nice :+1: Just a few final questions / comments :technologist:

test/common/deployment_types.go
makslion: Not sure about that one. The reason why we decided to bother ourselves with checking values in pods in the first place was an introduction of Quota. Idea was to check not only that we set values in CRs but also that we are setting values that could be accepted by various products and could be propagated down to pods. 
For RHMI nothing gets changed so the default CR check should be sufficient. In theory, nothing holds us from checking CPU and memory in pods for RHMI (the code to do so is already here), but I think that if we decide to refactor RHMI-related things they, probably, should be addressed in a separate ticket :thinking: 
What do you think?

test/common/deployment_types.go
makslion: yes, it would make it prettier. Let me try to make it up.

test/common/deployment_types.go
KevFan: Ah okay, just wondering overall. I'm good keeping it as it if we want to isolate this to just RHOAM :+1:

makslion:
Implemented more suggestions. Also modified description to check A14 as it is also got affected by this PR. 
Introduced `GetListOptions(...)` function to obtain list options for getting pods. Made it public as it will be used during the testing week to refactor existing code in the `test` package that as well utilizes list options for pods retrieval

test/common/deployment_types.go
makslion: Removed. You are absolutely right.

makslion:
/test
hey, bot, would you do the thing? :thinking:

trepel:
@makslion I have a cluster installed from https://github.com/integr8ly/integreatly-operator/pull/1739 since I was reviewing that one. I ran the A13-15 tests against that cluster from this PR locally and all passed.

I am adding lgtm but I'd be good if @matskiv and @KevFan have another look if their comments were addressed properly.

/lgtm

KevFan:
Code looks good to me :+1: 
/lgtm

PR #1742, opened 2021-04-27
MGDAPI-1620: Add creation of quota config map at operator startup

sergioifg94:
# Description

> Link to JIRA: https://issues.redhat.com/browse/MGDAPI-1620

This is a temporal solution to include the quota configuration in the installation. Add the logic to create the quota ConfigMap at
operator startup, after the RHMI CR is created/discovered.

## Verification steps

Run the operator locally, ensuring that the quota ConfigMap doesn't exist in the cluster, and ensure that the installation progresses and finishes successfully

1. Checkout this PR
2. Prepare the cluster
    ```sh
    ```
3. Ensure the ConfigMap doesn't exist
    ```sh
    oc delete cm/quota-config-managed-api-service -n redhat-rhoam-operator
    ```
4. Run the operator
    ```sh
    ```
5. Verify that the ConfigMap is created and the installation progresses
    ```sh
    oc get cm/quota-config-managed-api-service -n redhat-rhoam-operator
    ```
    ```
    NAME                               DATA   AGE
    quota-config-managed-api-service   1      8s
    ```

## Type of change

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -358,7 +358,6 @@ cluster/prepare/dms:
 
 .PHONY: cluster/prepare/quota
 cluster/prepare/quota:
-	@-oc apply -n $(NAMESPACE) -f config/configmap/quota-config-managed-api-service.yaml
 	@-oc delete  -n $(NAMESPACE) secret addon-managed-api-service-parameters
 	@-oc process -n $(NAMESPACE) QUOTA=$(QUOTA) -f config/secrets/quota-secret.yaml | oc apply -f -
 
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -47,6 +47,7 @@ import (
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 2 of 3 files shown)


matskiv:
@sergioifg94 @laurafitzgerald Should we also remove the `config/configmap/quota-config-managed-api-service.yaml` and update Makefile?

sergioifg94:
@matskiv I'd say it might be good to keep it for development purposes (e.g. to change the values in the ConfigMap using the make command), but it could still be possible to do it by changing the hardcoded value. Also this solution is only temporal so it might be good to keep the make command for when this is reverted

matskiv:
@sergioifg94 okay, but I think we need to remove implicit Configmap creation from the usual "cluster/prepare" flows that are used by all the tests. If we keep it, then we don't really know if it's created by make commands or new code added in this PR.

laurafitzgerald:
We should try to keep local install as close to managed tenants as possible. If we don't get config map from managed tenants we should remove it from make cluster prepare.

sergioifg94:
@makslion @laurafitzgerald I just updated the PR, removed the ConfigMap creation from the make command, and also modified the ConfigMap creation code to use the variables `ConfigMapName` and `ConfigMapData` that were already defined

matskiv:
I verified that ConfigMap gets created by the operator and that the operator can still successfully switch to a different SKU.
/lgtm

@laurafitzgerald please merge this whenever suitable..

laurafitzgerald:
/hold

sergioifg94:
Just rebased from the feature branch @laurafitzgerald @matskiv

PR #1743, opened 2021-04-27
Fix error in operator logs when updating SKU from 1 to 20.

carlkyrillos:
# Description
Removed oc command from cluster/prepare/quota target in Makefile that deleted the quota secret before updating and re-creating the secret with a new quota value. Now the target just updates the existing secret which prevents the error that would occur because the operator was looking for a secret that no longer existed.

**JIRA Ticket:** https://issues.redhat.com/browse/MGDAPI-1592

## Verify

- Checkout this branch
- Install the RHOAM operator
- Run Make command to update quota value: `INSTALLATION_TYPE=managed-api QUOTA=20 make cluster/prepare/quota`
- Confirm that the addon-managed-api-service value in the addon-managed-api-service-parameters secret updates to 20.
- Confirm that there are no errors in the operator logs.

## Type of change

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -358,7 +358,6 @@ cluster/prepare/dms:
 
 .PHONY: cluster/prepare/quota
 cluster/prepare/quota:
-	@-oc delete  -n $(NAMESPACE) secret addon-managed-api-service-parameters
 	@-oc process -n $(NAMESPACE) QUOTA=$(QUOTA) -f config/secrets/quota-secret.yaml | oc apply -f -
 
 .PHONY: cluster/prepare/delorean


laurafitzgerald:
/hold

makslion:
Verified changing 1 -> 20 -> 1 -> 5 quotas succeeds and not produces error messages in the reconciler. Looks good and ready to drop lgtm once the merge conflict resolved. 
However, it seems that the problem was caused by the fact that we were deleting `addon-managed-api-service-parameters`. The reconciler was starting to panic as it failed to find the secret, but once `oc apply...` recreated it, the quota change event progresses as intended. We are still using 'oc apply...` that recreates secret and, we might still catch an error due to extremely unlikely timings. I've been wondering if it is possible to use `oc patch...` instead to only update the `addon-managed-api-service` in the field :thinking:

carlkyrillos:
> Verified changing 1 -> 20 -> 1 -> 5 quotas succeeds and not produces error messages in the reconciler. Looks good and ready to drop lgtm once the merge conflict resolved.
> However, it seems that the problem was caused by the fact that we were deleting `addon-managed-api-service-parameters`. The reconciler was starting to panic as it failed to find the secret, but once `oc apply...` recreated it, the quota change event progresses as intended. We are still using `oc apply...` that recreates secret and we might still catch an error due to extremely unlikely timings. I've been wondering if it is possible to use `oc patch...` instead to only update the `addon-managed-api-service` field in the secret

After discussing with Sergio, I think the current implementation can be kept because `oc patch` will fail if the secret doesn't yet exist.

makslion:
/lgtm
then. Thanks, Sergio, for a help

PR #1744, opened 2021-04-27
Refactor: moved initial setup of quota to bootstrap stage

jackdelahunt:
# Description
JIRA: https://issues.redhat.com/browse/MGDAPI-1599
Before we were processing quota in the reconciler but it has now been moved to the bootstrap reconciler.

## Type of change

## Verify
- [Create a cluster](https://github.com/integr8ly/delorean/blob/master/docs/ocm/README.md#using-ocm-for-installation-of-rhmi-or-managed-api-service-rhoam)
- Install RHOAM based on an image with new changes (ask me for image details)
    - Run make cluster/deploy with IMAGE_FORMAT set to the image details I will supply
    - Run make cluster/prepare/quota to create the quota secret in the operator namespace (may already be there)
- Change the Quota configuration
    - Select a value to change the sku value to. Any value that `name` is set to in the `quota-config-managed-api-service` config map in the operator namespace is a valid entry.
Run make cluster/prepare/quota again with a `DEV_QUOTA` set to the new value (5 should work fine).
- Verify Quota values
   - You can access prometheus from the route that is exposed in the *middleware-monitoring-operator* namespace
   - Use the query 'rhoam_quota' and you should see the values `Quota` and `toQuota` change as expected 

## Checklist

### DIFF
--- a/controllers/rhmi/bootstrapReconciler.go
+++ b/controllers/rhmi/bootstrapReconciler.go
@@ -5,8 +5,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/rand"
+	"os"
 	"strings"
 	"time"
+
+	"github.com/integr8ly/integreatly-operator/pkg/addon"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/quota"
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/pkg/errors"
 
@@ -22,6 +27,7 @@ import (
 	"github.com/integr8ly/integreatly-operator/pkg/resources/marketplace"
 	"github.com/integr8ly/integreatly-operator/pkg/resources/owner"
 
+	rhmiv1alpha1 "github.com/integr8ly/integreatly-operator/apis/v1alpha1"
 	oauthv1 "github.com/openshift/api/oauth/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -33,6 +39,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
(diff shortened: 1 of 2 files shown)


laurafitzgerald:
/hold

carlkyrillos:
/lgtm
Verified that the quota and toQuota values correctly update when the rhoam_quota query is executed in prometheus after running QUOTA=<5|10|20> make cluster/prepare/quota.

matskiv:
@jackdelahunt Looks okay. Just one question: was this moved for "code style" reasons or for functional reasons(to fix some unwanted behavior?)

laurafitzgerald:
It was suggested by @jjaferson in this pr -> https://github.com/integr8ly/integreatly-operator/pull/1683#discussion_r608695994 
but i'm actually starting to question the change now. I have some recollection of someone saying that the bootstrap reconciler doesn't always run if it's coompleted. In this case we need to check on each reconcile that the sku param is updated so it may not be suitable to make this change.
/hold

matskiv:
@laurafitzgerald It seems like it runs on every reconcile:
https://github.com/integr8ly/integreatly-operator/blob/a95f79cef64962f6980c2de1262fafed9e9e60d5/controllers/rhmi/rhmi_controller.go#L368-L370 

Thanks for linking that comment. I think we can merge this.

matskiv:
@jackdelahunt It seems there was a misunderstanding because [MGDAPI-1599 JIRA](https://issues.redhat.com/browse/MGDAPI-1599) links to a piece of code in a PR, but it's not a permanent link, so line references got out of date...
The lines that were meant to be moved were mentioned in the comment that Laura linked above - https://github.com/integr8ly/integreatly-operator/pull/1683#discussion_r608695994

laurafitzgerald:
let's revisit this one after test week.

laurafitzgerald:
@jackdelahunt there are some tests failing on this one. You can fix them by running `make code/fix` and `make test-cases/fix`

carlkyrillos:
/lgtm
Reviewed the changes to each file since I last reviewed this PR. Verified that the quota and toQuota values correctly update after running INSTALLATION_TYPE=managed-api DEV_QUOTA=<5|10|20> make cluster/prepare/quota

philbrookes:
This needs a rebase @jackdelahunt

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -496,3 +511,67 @@ func generateSecret(length int) string {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getSecretQuotaParam(installation *rhmiv1alpha1.RHMI, quotaParam string, found bool) (string, error) {
+
+	if found || quotaParam != "" {
```
carlkyrillos, line 558: Should this if statement be checking that the `quotaParam` is found _and_ that it doesn't contain an empty value? Right now if `found` is True then the function will return without checking if `quotaParam` actually contains a value.
```suggestion
	if found && quotaParam != "" {
```

jackdelahunt: Okay just had to look into this more and I think you are right, if this is left in it might even revert the quota values to "" when they are valid so I think this does need to be changed.

carlkyrillos:
/lgtm
Reviewed the changes to each file since PR was last reviewed. Verified that the quota and toQuota values correctly update after running INSTALLATION_TYPE=managed-api DEV_QUOTA=<5|10> make cluster/prepare/quota

PR #1738, opened 2021-04-26
add QUOTA env to operator

laurafitzgerald:
# Description
Adding an env to handle cases where the user is upgrading from lower versions.
Setting the default to 20

# Verification

## Upgrade Scenario

Install from 1.5.0 tag -> using all the fancy bundling stuff

Verify there is no QUOTA env in the rhmi-operator pod
Verify there is no quota or toquota entry in the rhmi cr.
Verify there is no rhoam_quota metric in prometheus

Verify that there is no QUOTA env and no reference to toQuota or Quota in the cr.

upgrade to this branch -> using all the fancy bundling stuff

verify that the QUOTA value is present in the deployment/pod and has the value 20. 
Verify that cr now has toQuota: 20 or Quota: 20 depending on how long you wait.
Verify there is a rhoam_quota metric in prometheus

## Installation Scenario

Install from this branch using 

`INSTALLATION_TYPE=managed-api make cluster/prepare/local`
`INSTALLATION_TYPE=managed-api IN_PROW=true make deploy/integreatly-rhmi-cr.yml`
`INSTALLATION_TYPE=managed-api make code/run`

Verify that the installation defaults to `1` as this is the value provided by the secret in make cluster/prepate/local
If the value is 20 then it has taken from the env and there is a problem. 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist
- [ ] This change requires a documentation update 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added a test case that will be used to verify my changes 
- [ ] Verified independently on a cluster by reviewer

PR #1748, opened 2021-04-28
MGDAPI-1600 Add preflight check for quota parameter

sergioifg94:
# Description

> Link to JIRA: https://issues.redhat.com/browse/MGDAPI-1600

* Add a preflight check to ensure the quota parameter exists
* Move the preflight checks to be evaluated *before* the quota is processed

#### 1.1 Without falling back to env var

1. Checkout this PR
2. Prepare your cluster
    ```sh
    
    # Ensure the parameter doesn't exist
    oc delete  -n redhat-rhoam-operator secret addon-managed-api-service-parameters
    ```

3. Remove the `QUOTA=$(DEV_QUOTA)` env var from the `make code/run` target in the `Makefile` so that the operator doesn't find the env var
3. Run the operator locally
    ```sh
    ```
4. Check the RHMI CR is created but the installation doesn't progress as the preflight checks fail with the message
    ```
    quota parameter not found, waiting 1 minute before defaulting to env var
    ```
5. Wait a minute
6. Verify that the preflight error message changes to:
    ```
    quota parameter not found from add-on or env var
    ```
7. Delete the RHMI CR
8. Stop the operator

#### 1.2. Falling back to the env var

1. Restore the `QUOTA=$(DEV_QUOTA)` env var from the `make code/run` target in the `Makefile` that was removed in a previous step
2. Run the operator locally
4. Check the RHMI CR is created but the installation doesn't progress as the preflight checks fail with the message
    ```
    quota parameter not found, waiting 1 minute before defaulting to env var
    ```
5. Wait a minute
6. Verify that the preflight check passes and the `toQuota` field is set to `1` (the default in the Makefile)

### 2. Verify the preflight check passes

1. Uninstall RHOAM
2. Stop the operator
3. Restore the parameter secret
    ```sh
    ```
3. Run the operator locally
    ```sh
    ```
4. Verify that the preflight check passes immediately as the parameter secret was found

## Type of change

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ OPERATOR_SDK_VERSION=1.2.0
 AUTH_TOKEN=$(shell curl -sH "Content-Type: application/json" -XPOST https://quay.io/cnr/api/v1/users/login -d '{"user": {"username": "$(QUAY_USERNAME)", "password": "$(QUAY_PASSWORD)"}}' | jq -r '.token')
 TEMPLATE_PATH="$(shell pwd)/templates/monitoring"
 IN_PROW ?= "false"
-QUOTA ?= "1"
+DEV_QUOTA ?= "1"
 TYPE_OF_MANIFEST ?= master
 
 CONTAINER_ENGINE ?= docker
@@ -358,7 +358,7 @@ cluster/prepare/dms:
 
 .PHONY: cluster/prepare/quota
 cluster/prepare/quota:
-	@-oc process -n $(NAMESPACE) QUOTA=$(QUOTA) -f config/secrets/quota-secret.yaml | oc apply -f -
+	@-oc process -n $(NAMESPACE) QUOTA=$(DEV_QUOTA) -f config/secrets/quota-secret.yaml | oc apply -f -
 
 .PHONY: cluster/prepare/delorean
 cluster/prepare/delorean: cluster/prepare/delorean/pullsecret
--- a/Makefile
+++ b/Makefile
@@ -130,7 +130,7 @@ setup/git/hooks:
 
 .PHONY: code/run
 code/run: code/gen cluster/prepare/smtp cluster/prepare/dms cluster/prepare/pagerduty setup/service_account
-	@KUBECONFIG=TMP_SA_KUBECONFIG WATCH_NAMESPACE=$(NAMESPACE) go run ./main.go
+	@KUBECONFIG=TMP_SA_KUBECONFIG WATCH_NAMESPACE=$(NAMESPACE) QUOTA=$(DEV_QUOTA) go run ./main.go
 
 .PHONY: code/rerun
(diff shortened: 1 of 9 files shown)


controllers/rhmi/rhmi_controller.go
```diff
@@ -688,6 +689,20 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
...
+		ok, err := addon.ExistsParameterByInstallation(context.TODO(), r.Client, installation, addon.QuotaParamName)
+		if err != nil || !ok {
+			var preflightMessage string
+			if err != nil {
+				preflightMessage = fmt.Sprintf("failed to retrieve addon parameter %s: %v", addon.QuotaParamName, err)
+			} else {
+				preflightMessage = fmt.Sprintf("parameter %s could not be found", addon.QuotaParamName)
+			}
+
+			log.Info(preflightMessage)
```
KevFan, line 702: Should there be a return here ? Preflght checks looks to progress and succeed even if the parameter is not found :thinking:

sergioifg94: @KevFan  It should. Missed that after reverting the initial refactor :stuck_out_tongue:

sergioifg94:
@KevFan I pushed some changes in the PR to apply the preflight check commit on top of @laurafitzgerald commit from #1738 and updated the verification steps to include the logic to fallback to the environment variable. Should be working now

sergioifg94:
Just updated the PR after the recent changes to #1738. Only relevant change is the change to `DEV_QUOTA` in the `code/run` make target to set the env var

KevFan:
* Verified the preflight check fails - Without falling back to env var
Screenshot from 2021-04-29 18-30-18

* Verified falling back to env var if addon parameter is present
image

* Verified the preflight check succeeds immediately if addons parameter is present
image

/lgtm

Makefile
```diff
@@ -130,7 +130,7 @@ setup/git/hooks:
 
 .PHONY: code/run
 code/run: code/gen cluster/prepare/smtp cluster/prepare/dms cluster/prepare/pagerduty setup/service_account
-	@KUBECONFIG=TMP_SA_KUBECONFIG WATCH_NAMESPACE=$(NAMESPACE) go run ./main.go
+	@KUBECONFIG=TMP_SA_KUBECONFIG WATCH_NAMESPACE=$(NAMESPACE) QUOTA=$(DEV_QUOTA) go run ./main.go
```
laurafitzgerald, line 133: In this case I would expect the installation to fallback on the secret. If the user has run make cluster/prepare they will have the secret in place. If they didn't (I think this would be unusual) but it would fall back to the env. No need to change it but just wanted to mention it.

laurafitzgerald:
Closing in favour of above linked pr.

PR #1756, opened 2021-04-30
update test replicas for 3scale under 1 milion quota

laurafitzgerald:
# Description

## Type of change

## Checklist

### DIFF
--- a/pkg/config/threescale.go
+++ b/pkg/config/threescale.go
@@ -125,9 +125,9 @@ func (t *ThreeScale) GetReplicasConfig(inst *integreatlyv1alpha1.RHMI) map[strin
 	if inst.Spec.Type == string(integreatlyv1alpha1.InstallationTypeManagedApi) {
 		switch inst.Status.Quota {
 		case "1":
-			threeScaleComponents["apicastProd"] = 1
-			threeScaleComponents["backendListener"] = 1
-			threeScaleComponents["backendWorker"] = 1
+			threeScaleComponents["apicastProd"] = 2
+			threeScaleComponents["backendListener"] = 2
+			threeScaleComponents["backendWorker"] = 2
 		case "5":
 			threeScaleComponents["apicastProd"] = 3
 			threeScaleComponents["backendListener"] = 3


pkg/config/threescale.go
```diff
@@ -125,9 +125,9 @@ func (t *ThreeScale) GetReplicasConfig(inst *integreatlyv1alpha1.RHMI) map[strin
 	if inst.Spec.Type == string(integreatlyv1alpha1.InstallationTypeManagedApi) {
 		switch inst.Status.Quota {
 		case "1":
-			threeScaleComponents["apicastProd"] = 1
-			threeScaleComponents["backendListener"] = 1
-			threeScaleComponents["backendWorker"] = 1
+			threeScaleComponents["apicastProd"] = 2
+			threeScaleComponents["backendListener"] = 2
+			threeScaleComponents["backendWorker"] = 2
```
Boomatang, line 130: Should these values not be matching what is in the config file or am I misunderstanding this.

https://github.com/laurafitzgerald/integreatly-operator/blob/3scale-replicas-test-one-update/config/configmap/quota-config-managed-api-service.yaml#L315:L358

Boomatang (commented):
I have not ran any of the test yet but it looks right. There is one question below.

pkg/config/threescale.go
KevFan: @Boomatang The config map is now created from https://github.com/integr8ly/integreatly-operator/blob/master/pkg/addon/quota.go#L306-L352 since the PR https://github.com/integr8ly/integreatly-operator/pull/1742 merged. I don't think that config map is applied anymore :thinking:

laurafitzgerald: The config map is moved from that location. you can check it on master here -> https://github.com/integr8ly/integreatly-operator/blob/master/pkg/addon/quota.go#L314-L352

Boomatang: If that config map is not used any more should be not be removed? But I am happy with the changes in this PR.

KevFan:
Verified `F05` e2e test past locally with 1 million quota locally

image
/lgtm

pkg/config/threescale.go
laurafitzgerald: Ah, the command to create it in cluster prepare is removed. I thought we would removed the config map itself as part of that change but looks like it's not. @sergioifg94 could the config map itself be remove in order to avoid any confusion.

sergioifg94: @laurafitzgerald Yes, I left it in case it could be useful to keep it for development, but if it generates confusion I'm happy to have it removed. I can create a PR for that

PR #1757, opened 2021-04-30
feat: Add preflight check for quota param

laurafitzgerald:
# Description
Copy of https://github.com/integr8ly/integreatly-operator/pull/1748 rebased on master
## Type of change

## Checklist


KevFan:
Verified rebase with master following verification steps from original PR :+1: 

* Verified the preflight check fails - Without falling back to env var
Screenshot from 2021-04-30 11-01-32

* Verified falling back to env var if addon parameter is not present
image

* Verified the preflight check succeeds immediately if addons parameter is present
Screenshot from 2021-04-30 11-09-07

/lgtm

PR #1765, opened 2021-05-04
MGDAPI-1744 Remove quota ConfigMap from config/ folder

sergioifg94:
# Description

Remove this ConfigMap as it's not created by the make target anymore but reconciled by the operator.

## Type of change

## Checklist


makslion:
/lgtm

This one caused me lots of confusions recently. This config map is not used for our Quota configuration and should be removed.

PR #1769, opened 2021-05-05
set image pull policy to IfNotPresent

briangallagher:
# Description
Current pull policy on the integreatly operator is "Always". Set to "IfNotPresent" to mitigate against the risk of quay being down
See JIRA: https://issues.redhat.com/browse/MGDAPI-1473

# Verify
Install RHMI from this branch
Check the YAML on the Cronjobs in ns redhat-rhmi-amq-online for imagePullPolicy: IfNotPresent
Modify the cron time in the YAML 
Verify a new job runs as the specified time
Verify that the IfNotPresent is set in the pod yaml
Verify the the image pulls successfully and runs.


briangallagher:
@CathalOConnorRH Can you review please?

PR #1777, opened 2021-05-10
fix go sum

MStokluska:
# Description
Fixes issue around go sum and CRO 0.24.0 failing on checksum


MStokluska:
/test e2e

Failure on cluster creation

MStokluska:
/test e2e

`ailed to create pod sandbox: rpc error: code = Unknown desc = Kubelet may be retrying requests that are timing out in CRI-O due to system load`

PR #1779, opened 2021-05-11
MGDAPI-1606: Change the expression in RhoamOperatorInstallDelayed to no longer use absent

austincunningham:
# Description

Use of absent in the alert RhoamOperatorInstallDelayed can trigger if the metric is not present rather that the metric being in a certain state 
https://issues.redhat.com/browse/MGDAPI-1606

## Type of change

## Verification
**Successful Installation**
- Deploy rhoam via  addon flow 
- patch the images in the csv
```bash
oc patch ClusterServiceVersion managed-api-service.v1.6.0 -n redhat-rhoam-operator --patch '{"metadata": {"annotations": {"containerImage": "quay.io/austincunningham/managed-api-service:v1.6.0" }}}' --type=merge
oc patch ClusterServiceVersion managed-api-service.v1.6.0 -n redhat-rhoam-operator --type='json' -p='[{"op": "replace", "path": "/spec/install/spec/deployments/0/spec/template/spec/containers/0/image", "value":"quay.io/austincunningham/managed-api-service:v1.6.0"}]'
```
- patch the 'UserClusterStorage
```bash
oc patch rhmi rhoam -n redhat-rhoam-operator --patch '{"spec": {"useClusterStorage": "true"}}' --type=merge
```
- In openshift-monitoring Prometheus the alert should go into a pending state during install
image
if the install takes more that 2 hours it will trigger
- Check the alert exists and is not firing after the install has completed in openshift-monitoring Prometheus
image

## Checklist

### DIFF
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -288,6 +288,7 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 		})
 	}
 	metrics.SetRHMIStatus(installation)
+	metrics.SetRhmiVersions(string(installation.Status.Stage), installation.Status.Version, installation.Status.ToVersion, installation.CreationTimestamp.Unix())
 
 	configManager, err := config.NewManager(context.TODO(), r.Client, request.NamespacedName.Namespace, installationCfgMap, installation)
 	if err != nil {
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -288,7 +288,6 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 		})
 	}
 	metrics.SetRHMIStatus(installation)
-	metrics.SetRhmiVersions(string(installation.Status.Stage), installation.Status.Version, installation.Status.ToVersion, installation.CreationTimestamp.Unix())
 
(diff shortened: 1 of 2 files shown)


controllers/rhmi/rhmi_controller.go
```diff
@@ -288,6 +288,7 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 		})
 	}
 	metrics.SetRHMIStatus(installation)
+	metrics.SetRhmiVersions(string(installation.Status.Stage), installation.Status.Version, installation.Status.ToVersion, installation.CreationTimestamp.Unix())
```
MStokluska, line 291: hey @austincunningham just a question, why is this required? 
We are currently setting the metric on 
a) first reconcile
b) when stage reports as "completed"
The description on the b) check is `Check for stage complete to avoid setting the metric when installation is happening` - is there a reason why we want to avoid setting the metric? If not, I would suggest removing the logic around setting the metric when the "completed" stage is reported, if we are setting the metric every time anyway. 

IMO I don't think we have to set metrics on every reconcile but instead, on first reconcile + after completed is reported (unless I'm missing something!)

austincunningham: @MStokluska the alert was currently checking for the absent stage="completed", but the absent of the metric would trigger this alert as well.

austincunningham: Looks like I was miss interpreting the data as clock on cluster was an hour behind

image

I can remove this :+1:

MStokluska:
/lgtm

Verified on the cluster and it works as expected.

PR #1784, opened 2021-05-12
removed "ksm-monitoring-arelts" and "rate-limit-soft-limits" alerts

makslion:
# Description
MiddlewareMonitoringPodCount alert (serverity = warning) are firing during installation and gets resolved in about 15 minutes. It was decided to remove the alert as it has no value for an SRE team rather then fixing it. 
rate-limit-soft-limits Prometheus rule was removed from creation as part of 1.6.0, but we newer ensured that this alert will not be present on a cluster while upgrading from 1.5.0 to 1.6.0, so it stayed on a cluster.

## Verification steps
1. Ensure you have a cluster with 'ksm-monitoring-alerts" PrometheusRule [present](https://i.imgur.com/ZLhHLE0.png) and with "rate-limit-soft-limits" (you can run master of 1.5.0 against a cluster to obtain those) 
2. Run the reconciler from this PR 
3. Verify that the PrometheusRule CRs were [removed](https://i.imgur.com/QU70t5p.png) 

## Type of change

## Checklist

Optional step: grub a cup of coffee or a tea and enjoy the day for 1 minute 43 seconds

### DIFF
--- a/pkg/resources/prometheusRules.go
+++ b/pkg/resources/prometheusRules.go
@@ -63,6 +63,28 @@ func (r *AlertReconcilerImpl) ReconcileAlerts(ctx context.Context, client k8scli
 		}
 	}
 
+	rule := &monitoringv1.PrometheusRule{}
+	err := client.Get(ctx, k8sclient.ObjectKey{
+		Name:      "ksm-monitoring-alerts",
+		Namespace: "redhat-rhoam-middleware-monitoring-operator",
+	}, rule)
+	if !errors.IsNotFound(err) {
+		if err := client.Delete(ctx, rule); err != nil {
+			r.Log.Info("Alert \"ksm-monitoring-alerts\" was deleted")
+		}
+	}
+
+	rule = &monitoringv1.PrometheusRule{}
+	err = client.Get(ctx, k8sclient.ObjectKey{
+		Name:      "rate-limit-soft-limits",
+		Namespace: "redhat-rhoam-marin3r",
+	}, rule)
+	if !errors.IsNotFound(err) {
+		if err := client.Delete(ctx, rule); err != nil {
+			r.Log.Info("Alert \"rate-limit-soft-limits\" was deleted")
+		}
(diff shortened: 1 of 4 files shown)


pkg/resources/prometheusRules.go
```diff
@@ -63,6 +63,17 @@ func (r *AlertReconcilerImpl) ReconcileAlerts(ctx context.Context, client k8scli
...
+	err := client.Get(ctx, k8sclient.ObjectKey{
+		Name:      "ksm-monitoring-alerts",
+		Namespace: "redhat-rhoam-middleware-monitoring-operator",
+	}, rule)
+	if !errors.IsNotFound(err) {
+		if err := client.Delete(ctx, rule); err != nil {
+			r.Log.Info("Alert \"ksm-monitoring-alerts\" was deleted")
+		}
+	}
+
```
Boomatang, line 76: I would like to add the removal of the rate limiting soft alerts in to this PR, the changes required are below. This is will address the jira: https://issues.redhat.com/browse/MGDAPI-1705

```suggestion
	err := client.Get(ctx, k8sclient.ObjectKey{
		Name:      "rate-limit-soft-limits,
		Namespace: "redhat-rhoam-marin3r",
	}, rule)
	if !errors.IsNotFound(err) {
		if err := client.Delete(ctx, rule); err != nil {
			r.Log.Info("Alert \"rate-limit-soft-limits\" alert was deleted")
		}
	}

```

makslion: included removal of the soft limits alert

carlkyrillos:
Verified that both PrometheusRules (`ksm-monitoring-alerts` in the `redhat-rhoam-middleware-monitoring-operator` namespace and `rate-limit-soft-limits` in the `redhat-rhoam-marin3r` namespace) were removed after running the reconciler from this PR against a cluster that already had RHOAM 1.5.0 installed.

/lgtm

PR #1786, opened 2021-05-13
Add script to patch the csv with dev image and rhmi cr useClusterStorage during addon flow

austincunningham:
# Description

helper script that  during addon flow
- patches csv with a developer image 
- patches rhmi cr with useClusterStorage

## Type of change

## Verification
- build/push a development image e.g.
```bash
ORG=austincunningham OLM_TYPE=managed-api-service INSTALLATION_TYPE=managed-api make image/build
ORG=austincunningham OLM_TYPE=managed-api-service INSTALLATION_TYPE=managed-api make image/push
```
- export the vars of the script e.g.
```bash
export VERSION=v1.6.0
export ADDON_VERSION=v1.6.0
export ORG=austincunningham
```
- run the script (need to be logged in to oc for the cluster)
- Install RHOAM via the addon flow
```bash
./patch-image-csv.sh
clusterserviceversion.operators.coreos.com/managed-api-service.v1.6.0 patched
clusterserviceversion.operators.coreos.com/managed-api-service.v1.6.0 patched
Verificaton the CSV has been patched :
    containerImage: quay.io/austincunningham/managed-api-service:v1.6.0
                image: quay.io/austincunningham/managed-api-service:v1.6.0
 
Patching the RHMI CR when its created with useClusterStorage
rhmi.integreatly.org/rhoam patched
Verification USE_CLUSTER_STORAGE is set :
        f:useClusterStorage: {}
  useClusterStorage: "true"

```

## Checklist

### DIFF
--- a/scripts/patch-image-csv.sh
+++ b/scripts/patch-image-csv.sh
@@ -6,15 +6,36 @@
 # VERSION=<your image tag>
 # ORG=<you quay org>
 # ADDON_VERSION=<version set https://gitlab.cee.redhat.com/service/managed-tenants/-/blob/main/addons/managed-api-service/metadata/stage/addon.yaml#L25>
-# Set the varse.g.
+# USE_CLUSTER_STORAGE=<true/false optional will default true if not set>
+# Set the vars.g.
 # export VERSION=v1.6.0
 # export ADDON_VERSION=v1.6.0
 # export ORG=austincunningham
+# export USE_CLUSTER_STORAGE=false
 # ./patch-image-csv.sh
+#
+# Check csv exists
+oc get csv managed-api-service.$VERSION -n redhat-rhoam-operator
+while [ $? -ne 0 ]; do
+  oc get csv managed-api-service.$VERSION -n redhat-rhoam-operator
+done
+# Patch the csv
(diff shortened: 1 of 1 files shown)


MStokluska:
@austincunningham this is a great idea! I just wonder, would it be possible to make the script check for the presence of the csv and if present patch it? 
So the flow would be:
- trigger the script
- click on install addon
- the script reports completed

The reason for this is that I've hit some timing issues, this has to be run after the install is pressed, but before the operator pod is created. Unless? The operator pod is re-deployed after patching the csv? But I don't think this is the case.

austincunningham:
> would it be possible to make the script check for the presence of the csv and if present patch it?
> So the flow would be:
> 
> * trigger the script
> * click on install addon
> * the script patches csv and reports completed

@MStokluska Yes it is possible , could also do the same for patching the rhmi CR, although may need an extra var for UseClusterStorage

austincunningham:
@MStokluska updated the pr with your suggestions, have verified and works as expected.

jackdelahunt:
- Ran the script with variables set
- Installed RHOAM via addon
- The addon used the image specified 
- Working as expected

/lgtm

jackdelahunt:
> * Ran the script with variables set
> * Installed RHOAM via addon
> * The addon used the image specified
> * Working as expected
> 
> /lgtm

/lgtm

PR #1790, opened 2021-05-14
add trial quota option

philbrookes:
# Description
Add a zero option to the quota configs configmap.

JSON validation: https://jsonformatter.org/json-parser/23933a


laurafitzgerald:
@philbrookes i've sanity checked the change and it looks good. It's a copy of the 1 million quota, is that correct? 
otherwise
/lgtm

PR #1800, opened 2021-05-18
Silence Alerts for Keycloak on uninstall

CathalOConnorRH:
# Description
Silence critical keycloak alert on uninstall. 
https://issues.redhat.com/browse/MGDAPI-1612

## Type of change

## Verification

1. Install RHOAM or RHMI from this branch
2. Scale down keycloak operator and keycloak pods
3. verify alert is firing in openshift-monitoring alertmanager and AMO alertmanager
4. Start uninstall 
5. verify silence gets created in both alertmanagers


controllers/rhmi/rhmi_controller.go
```diff
@@ -434,6 +436,85 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
...
+	stdout, stderr, err := resources.ExecuteRemoteCommand(namespace, podName, command+host+endpoint, log, containerName)
+	if err != nil {
+		log.Error("Failed to curl alerts", err)
+		return silenceExists, err
+	} else if stderr != "" {
+		err := errors.New(stderr)
+		log.Error("Error attempting curl alerts", err)
+		return silenceExists, err
+	} else {
+		//log.Infof("Alerts result", l.Fields{"stdout": stdout})
```
briangallagher, line 510: Do we need this log?

controllers/rhmi/rhmi_controller.go
```diff
@@ -434,6 +436,85 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
...
+	openshiftMonitoringNamespace := "openshift-monitoring"
+	openshiftPodName := "alertmanager-main-0"
+	containerName := "alertmanager"
+
+	integreatlyMonitoringNamespace := installation.Spec.NamespacePrefix + "middleware-monitoring-operator"
+	integreatlyPodName := "alertmanager-application-monitoring-0"
+
+	exists, err := silenceExists(openshiftMonitoringNamespace, openshiftPodName, host, endpoint, log, containerName, alertName)
+	if err != nil {
+		log.Error("err", err)
```
briangallagher, line 455: Can we add a more descriptive error. Should we return at this point?

controllers/rhmi/rhmi_controller.go
```diff
@@ -434,6 +436,85 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
...
+	}
+	if !exists {
+		silenceAlert(openshiftMonitoringNamespace, openshiftPodName, host, endpoint, log, containerName, alertName)
+	}
+
+	exists, err = silenceExists(integreatlyMonitoringNamespace, integreatlyPodName, host, endpoint, log, containerName, alertName)
+	if err != nil {
+		log.Error("err", err)
```
briangallagher, line 463: Can we add a more descriptive error.

controllers/rhmi/rhmi_controller.go
```diff
@@ -509,6 +590,13 @@ func (r *RHMIReconciler) handleUninstall(installation *rhmiv1alpha1.RHMI, instal
 		}
 	}
 
+	log.Info("Creating Silence for Keycloak alerts")
```
briangallagher, line 593: Is it for all alerts?

controllers/rhmi/rhmi_controller.go
```diff
@@ -434,6 +436,85 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
...
+		log.Error("err", err)
+	}
+	if !exists {
+		silenceAlert(integreatlyMonitoringNamespace, integreatlyPodName, host, endpoint, log, containerName, alertName)
+	}
+
+	return nil
+}
+
+func silenceAlert(namespace string, podName string, host string, endpoint string, log l.Logger, containerName string, alertName string) error {
```
briangallagher, line 472: No actioning the error, so maybe no need to return?

controllers/rhmi/rhmi_controller.go
CathalOConnorRH: nope, will remove.

controllers/rhmi/rhmi_controller.go
CathalOConnorRH: Updated.

controllers/rhmi/rhmi_controller.go
CathalOConnorRH: Updated description. Its only one alert that needs to be silenced

controllers/rhmi/rhmi_controller.go
CathalOConnorRH: Updated.

controllers/rhmi/rhmi_controller.go
CathalOConnorRH: updated

config/rbac/role.yaml
```diff
@@ -12,6 +12,21 @@ rules:
...
+  - pods
+  - pods/exec
+  - secrets
+  verbs:
+  - create
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
matskiv, line 29: Why are we adding these new cluster wide permissions?

CathalOConnorRH: We need access to the alert manager pods in two other namespaces to exec commands within them. Is there a nicer way to do it?

matskiv: Ideally we should restrict permissions via `resourceNames`, e.g.:
https://github.com/integr8ly/integreatly-operator/blob/15f3aa133793cd859cb8630508b99f520256fb09/config/rbac/role.yaml#L30-L37 
Do we really need access to secrets, configmaps, events for pod/exec?

Why can't we do the http call from integreatly-operator using go? Why does it have to be curl executed in the alert manager pod?

CathalOConnorRH: I can adjust the resources, we only need the pods/exec. 

We need to curl within the pod because alertmanager is behind a oauth proxy and it listens on 127.0.0.1/9093

matskiv: I am not 100% sure, but I think we would be able to make a call with integreatly-operator ServiceAccount token passed in `Authorization: Bearer` header. We just need to check that SA has the permissions that are checked by oauth proxy(probably view on some ns).

controllers/rhmi/rhmi_controller.go
```diff
@@ -434,6 +436,81 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 	return retryRequeue, err
 }
 
+func (r *RHMIReconciler) createSilence(installation *rhmiv1alpha1.RHMI) error {
+
+	host := "localhost:9093"
+	endpoint := "/api/v1/silences"
+
+	alertName := "KeycloakInstanceNotAvailable"
```
matskiv, line 444: I noticed we are silencing only this one alert^
And the JIRA is talking about more general problem - `Please suppress alerts that are caused by a cluster uninstall`. No other alerts can fire during uninstallation?

CathalOConnorRH: The other alerts within Keycloak are warning alerts and as far as Im aware shouldnt page SRE.
 All other integreatly alerts are removed as part of uninstall https://github.com/integr8ly/integreatly-operator/blob/master/controllers/rhmi/rhmi_controller.go#L499:L510

config/rbac/role.yaml
CathalOConnorRH: I'll have a look into that and see if it's possible.

controllers/rhmi/rhmi_controller.go
matskiv: Cool, thanks for explanation.
So this means that any critical alerts created by other operators need to be silenced during uninstall?
I know that we plan to allow 3scale-operator create the alerts - [MGDAPI-428](https://issues.redhat.com/browse/MGDAPI-428), should we expand this JIRA to make sure those alerts become silenced on uninstall?

config/rbac/role.yaml
CathalOConnorRH: Updated to work with Bearer auth and removed unnecessary permissions.

controllers/rhmi/rhmi_controller.go
CathalOConnorRH: updated to allow additional alerts to be added. it should only be a matter of adding the names of critical alerts to the `alertsToSilence` array.

CathalOConnorRH:
@briangallagher Can you re-review when you get a chance?

controllers/rhmi/rhmi_controller.go
```diff
@@ -434,6 +465,191 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
...
+
+	url, err := r.getURLFromRoute(route, namespace, rc)
+	if err != nil {
+		return fmt.Errorf("error getting route : %w", err)
+
+	}
+	url = url + endpoint
+	serviceAccount := r.getServiceAccount(installation)
+
+	token, err := r.getBearerToken(installation, serviceAccount)
```
matskiv, line 510: The token should be in the rest config instance that is attached to this reconciler: `r.restConfig.BearerToken`

Docs: https://pkg.go.dev/k8s.io/client-go/rest#Config

controllers/rhmi/rhmi_controller.go
```diff
@@ -78,13 +84,35 @@ const (
...
 
 var (
 	productVersionMismatchFound bool
 	log                         = l.NewLoggerWithContext(l.Fields{l.ControllerLogContext: "installation_controller"})
+	alertsToSilence             = []string{
+		"KeycloakInstanceNotAvailable",
+	}
 )
 
+type Silence struct {
```
matskiv, line 99: Minor: this type can be imported from here I believe:
https://github.com/prometheus/alertmanager/blob/22ac6dff21901bfce14545da59b37a1aaca0db3a/api/v2/models/silence.go#L32-L55

controllers/rhmi/rhmi_controller.go
```diff
@@ -434,6 +465,191 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
...
+	endpoint := "/api/v1/silences"
+	startsAt := time.Now().Format(time.RFC3339)
+	endsAt := time.Now().Add(time.Hour * 1).Format(time.RFC3339)
+
+	var silence = []byte(`{` +
+		`"endsAt": "` + endsAt + `",` +
+		`"startsAt": "` + startsAt + `",` +
+		`"matchers":[{"isRegex": false, "name": "alertname","value": "` + alertName + `"}],` +
+		`"Comment":"Silence alert due to uninstall",` +
+		`"createdBy":"Integreatly Operator"}`)
```
matskiv, line 500: Would be nicer to use the `Silence` type for this as well.

CathalOConnorRH: updated to use `silence` type

controllers/rhmi/rhmi_controller.go
CathalOConnorRH: imported, thanks :D

controllers/rhmi/rhmi_controller.go
```diff
@@ -174,6 +186,9 @@ func New(mgr ctrl.Manager) *RHMIReconciler {
 // Permission to get the ConfigMap that embeds the CSV for an InstallPlan
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get
 
+// Permission to get the serviceaccount
+// +kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch
+
```
matskiv, line 191: Do we still need to add this permission? Now that you use bearer token from restConfig...

controllers/rhmi/rhmi_controller.go
matskiv: This suggestion is implemented now. Marking as resolved.

matskiv:
Code looks good. Just one last comment.
Thanks for implementing all suggestions 👍

controllers/rhmi/rhmi_controller.go
CathalOConnorRH: Good spot, No we don't. I've removed it now.

PR #1818, opened 2021-05-19
update product yaml and add removal of common lines

MStokluska:
# Description
- Updated manifests manually due to an issue with product.yaml around kc version introduced in recent kc bump
- Added removal of common lines


Boomatang:
To verify this I was on call with @MStokluska looking at the playground runs in jenkins and descussing the changes made. For the issue of the extra "master" entries in the production manifest, we were not able to track down the cause or reproduce the issue.

\lgtm

PR #1832, opened 2021-05-24
fix link for ROSA pipeline in test case A37

austincunningham:
# Description

Link to ROSA pipeline doesn't resolve.
Fix the link to the ROSA pipeline in test case A37

## Type of change

## Checklist


trepel:
Not sure why that format check did not pass, but the change looks ok

/lgtm

PR #1836, opened 2021-05-25
skip functional aws tests when useClusterStorage is true

makslion:
# Description
If useClusterStorage in RHMI CR is set to true, skip functional AWS tests. 

## Verification steps 
1. Obtain cluster with RHOAM installed and `useClusterStorage` set to `true` (true by default) 
2. Run following command:
```
INSTALLATION_TYPE=managed-api make test/functional
``` 
3. Observe tests getting skipped with a following message 
```
Aborting functional tests: "UseClusterStorage" is set to true. 
Please, run another testing suite or reinstall operator with "UseClusterStorage" set to false
```

Optional: 
- Reinstall cluster with `useClusterStorage` set to false and see functional tests being executed 

## Type of change

## Checklist

### DIFF
--- a/test/functional/aws_rds_resources_exist.go
+++ b/test/functional/aws_rds_resources_exist.go
@@ -16,7 +16,6 @@ func AWSRDSResourcesExistTest(t common.TestingTB, ctx *common.TestingContext) {
 	if err != nil {
 		t.Fatalf("error getting RHMI CR: %v", err)
 	}
-
 	// build an array of postgres resources to check and an array of test errors
 	rdsResourceIDs, testErrors := GetRDSResourceIDs(goContext, ctx.Client, rhmi)
 
--- a/test/functional/suite_test.go
+++ b/test/functional/suite_test.go
@@ -53,6 +53,19 @@ func TestAPIs(t *testing.T) {
 		Groups:   []string{"system:authenticated"},
 	}
 
+	// get RMI CR and if cluster storage set to true, skip the suite
+	context, err := common.NewTestingContext(cfg)
+	if err != nil {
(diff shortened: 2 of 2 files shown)


MStokluska:
Well done @makslion, your logic will work however I think it could be done a little bit more efficiently. 
I also left a comment on the ticket itself about the purpose of this. 
If we need to provide this logic, I think it would be better to move the logic here: 
https://github.com/integr8ly/integreatly-operator/blob/master/test/functional/integreatly_test.go
And follow a similar approach to what MULTI_AZ does - https://github.com/integr8ly/integreatly-operator/blob/master/test/functional/integreatly_test.go#L58 instead of checking same spec field in each of the tests.

makslion:
I've been thinking about this approach and there is one thing that bothers me a bit: In functional.FUCNTIONAL_TESTS there is N06 that shall not be skipped (but I could just create another group, I guess :thinking: , however, it seems like creating more mess then organizing things)

makslion:
Letting the note (as it confused me when I've revisited this PR). This work is paused and will be resumed during the testing week. Changes to be made: 
- We will not skip tests
- At the beginning of functional tests, we would check RHMI CR to have useClusterStorage set to false. If it is true we abort tests with a suggestion to run another suite or reinstall the operator with the variable set to false

makslion:
Updated PR according to changes in the ticket

jackdelahunt:
/lgtm
- Installed rhoam locally on a cluster
- Set `USE_CLUSTER_STORAGE` to true and waited for install to complete
- Ran the following command `INSTALLATION_TYPE=managed-api make test/functional`
- Recieved this output when ran `Aborting functional tests: "UseClusterStorage" is set to true etc.`

PR #1858, opened 2021-06-06
Remove ami public filter

tonyxrmdavidson:
Removed ami public filter from command. It is not needed.


trepel:
Verified, tried locally and indeed just one image is found if using default value for IMAGE var:
```
$ aws ec2 describe-images --filters "Name=name,Values=RHEL-8.2.0_HVM-*-x86_64*Hourly2-GP2" --region us-east-1 --output text --query 'Images[0].ImageId'
ami-098f16afa9edf40be
$ aws ec2 describe-images --filters "Name=name,Values=RHEL-8.2.0_HVM-*-x86_64*Hourly2-GP2" --region us-east-1 --output text --query 'Images[1].ImageId'
None
```

/lgtm

PR #1861, opened 2021-06-07
Updates of test cases post RHOAM v1.7.0 GA

trepel:
# Description
Test case updated due to RHOAM v1.7.0.

Most notable modifications:
* obsolete tasks tagged as 'manual-selection', not sure of a better way
* using 'external' env where appropriate
  * we might start creating 'external' epic since we have quite a few such tasks now
* a31 - tagged per-release because we should be testing with current and -1 and -2 OSD versions if possible

Note
We have four rate limit related test cases now, I feel there is a room for optimization (merging) but this is out of scope for this PR.


test-cases/tests/installation/a39-installation-verify-rhoam-on-private-cluster.md
```diff
@@ -0,0 +1,21 @@
...
+  - per-release
+---
+
+# A39 - installation - verify RHOAM on private cluster
+
+## Steps
+
+1. Create a clean OSD cluster as usual
+2. Follow [this guide](https://docs.google.com/document/d/1BwjzezNFtE7gd2y6FY6v2W6KRXCn0jMZk58ilJ8zSa8/edit) to make it private
+3. Use [addon-flow](https://master-jenkins-csb-intly.apps.ocp4.prod.psi.redhat.com/job/ManagedAPI/job/managed-api-install-addon-flow/) pipeline to run the automated tests against the cluster
```
psturc, line 17: Is it possible to run automated tests against private cluster via Jenkins? I was under the impression that a private cluster can be accessed only via VPN atm (afaik we don't have VPN config for those cases in Jenkins)

psturc:
/lgtm
just one question

test-cases/tests/installation/a39-installation-verify-rhoam-on-private-cluster.md
trepel: Thanks for explanation. It is not (due to private cluster being behind AWS VPN and VPN client need to be configured on a machine from which we want to access the cluster. I will rewrite to say how to run the test locally.

PR #1863, opened 2021-06-08
remove deprecated alerts deletion

makslion:
# Description
Closes MGDAPI-1797
There were rules to be deleted. This PR cleans up the code after rule deletion. 

## Verification steps 
- Ensure that rules deletion lines of code were removed  

## Type of change

## Checklist

### DIFF
--- a/pkg/resources/prometheusRules.go
+++ b/pkg/resources/prometheusRules.go
@@ -62,29 +62,6 @@ func (r *AlertReconcilerImpl) ReconcileAlerts(ctx context.Context, client k8scli
 			r.Log.Infof("Operation result", l.Fields{"productName": r.ProductName, "alertName": alert.AlertName, "result": string(or)})
 		}
 	}
-
-	rule := &monitoringv1.PrometheusRule{}
-	err := client.Get(ctx, k8sclient.ObjectKey{
-		Name:      "ksm-monitoring-alerts",
-		Namespace: "redhat-rhoam-middleware-monitoring-operator",
-	}, rule)
-	if !errors.IsNotFound(err) {
-		if err := client.Delete(ctx, rule); err != nil {
-			r.Log.Info("Alert \"ksm-monitoring-alerts\" was deleted")
-		}
-	}
-
-	rule = &monitoringv1.PrometheusRule{}
-	err = client.Get(ctx, k8sclient.ObjectKey{
-		Name:      "rate-limit-soft-limits",
-		Namespace: "redhat-rhoam-marin3r",
-	}, rule)
-	if !errors.IsNotFound(err) {
-		if err := client.Delete(ctx, rule); err != nil {
-			r.Log.Info("Alert \"rate-limit-soft-limits\" was deleted")
-		}
-	}
-
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
 


makslion:
@KevFan you asked me to ping you on this PR :thinking:

KevFan:
> @KevFan you asked me to ping you on this PR

@makslion Can't say I remember this :sweat_smile:  but I'll have a look anyway :+1:

pkg/resources/prometheusRules.go
```diff
@@ -62,29 +62,6 @@ func (r *AlertReconcilerImpl) ReconcileAlerts(ctx context.Context, client k8scli
 			r.Log.Infof("Operation result", l.Fields{"productName": r.ProductName, "alertName": alert.AlertName, "result": string(or)})
 		}
 	}
-
-	rule := &monitoringv1.PrometheusRule{}
-	err := client.Get(ctx, k8sclient.ObjectKey{
-		Name:      "ksm-monitoring-alerts",
-		Namespace: "redhat-rhoam-middleware-monitoring-operator",
```
KevFan, line 69: Since this was hardcoded to use `redhat-rhoam-middleware-monitoring-operator`, would this mean that the associated RHMI alert one is kept ? :thinking:

pkg/resources/prometheusRules.go
```diff
@@ -62,29 +62,6 @@ func (r *AlertReconcilerImpl) ReconcileAlerts(ctx context.Context, client k8scli
...
-	err = client.Get(ctx, k8sclient.ObjectKey{
-		Name:      "rate-limit-soft-limits",
-		Namespace: "redhat-rhoam-marin3r",
-	}, rule)
-	if !errors.IsNotFound(err) {
-		if err := client.Delete(ctx, rule); err != nil {
-			r.Log.Info("Alert \"rate-limit-soft-limits\" was deleted")
-		}
-	}
-
```
KevFan, line 87: General question, should this be only code merged once 1.7.0 has been released to all production clusters or we good to just merged into 1.8.0 release ? I'm just thinking that if not all production environments are on 1.7.0 and say a patch version was made that skipped that version and went straight to 1.8.0, these alerts may still be kept on those clusters with this change :thinking:

KevFan (commented):
Looks good overall, just have 2 questions

pkg/resources/prometheusRules.go
makslion: If I'm not wrong, the only way to skip 1.7.0 is ti have 1.8.0 to mark it as `skipped` or something :thinking:

pkg/resources/prometheusRules.go
makslion: At the moment of doing it, the focus was only on RHOAM. It is hard to recall all details of discussions, but the JIRA description suggests removing those alerts from RHOAM with this line: `Upgrading to an RHOAM version with the alert removed`, but the more I'm thinking about it the less confident I am :thinking: 
If the RHMI alert also were supposed to be removed then there would be a new JIRA ticket created with its own cleanup afterwards. What do you think about this? :thinking:

pkg/resources/prometheusRules.go
Boomatang: So as @makslion said the only that a customer could install 1.8.0 with out installing 1.7.0 is if 1.8.0 is marked to skip 1.7.0. All production clusters will have to install 1.7.0 first. But it is not a bad question to ask release coordinator just to confirm that confirm the 1.8.0 release will not be skipping the 1.7.0 release. 

Max you should ask that question on the jira. If you need help finding the release coordinator I can help with that.

pkg/resources/prometheusRules.go
Boomatang: The reason this was only done for RHOAM is the alert was falsely firing during the installs of the operator. As there will be no fresh installs of RHMI there was no need to remove it from the current installs.

pkg/resources/prometheusRules.go
makslion: It was confirmed that 1.7.x will not be skipped

pkg/resources/prometheusRules.go
KevFan: Cool, if there is no need for the alert to be removed in RHMI, then this comment can be ignored, was just wondering is all :+1:

pkg/resources/prometheusRules.go
KevFan: Great, thanks, was just wondering in case it's something to consider :+1:

KevFan:
Based on comments discussion, this looks good to me :+1: 

/lgtm

pkg/resources/prometheusRules.go
laurafitzgerald: I think we can still verify if this is a consideration for RHMI There is a 2.9 release about to happen so if we need it in until after that release it's just a matter of waiting. And there's no problem to have it sit in 1.8 and remove in 1.9. cc @briangallagher as he's co-ordinating the 2.9 release

briangallagher: There does not seem to be an equivalent rhmi rule so I think this is fine to delete. @laurafitzgerald

PR #1867, opened 2021-06-10
MGDAPI-1728 - Consume addon parameter for the STS role in RHOAM and p…

valerymo:
Jira: https://issues.redhat.com/browse/MGDAPI-1728

# Description
integreatly-operator should check if it is running in a cluster that uses STS authentication, and if that's the case:
a) in a preflight check validate that the "STS role ARN" parameter value contains a valid AWS role ARN
b) pass the ARN value into a Secret in CRO namespace

## Type of change

## Checklist

### DIFF
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -21,17 +21,19 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/go-openapi/strfmt"
-	routev1 "github.com/openshift/api/route/v1"
-	appsv1Client "github.com/openshift/client-go/apps/clientset/versioned/typed/apps/v1"
-	"github.com/prometheus/alertmanager/api/v2/models"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/go-openapi/strfmt"
+	routev1 "github.com/openshift/api/route/v1"
+	appsv1Client "github.com/openshift/client-go/apps/clientset/versioned/typed/apps/v1"
+	"github.com/prometheus/alertmanager/api/v2/models"
 
 	"github.com/integr8ly/integreatly-operator/pkg/resources/events"
 	"github.com/integr8ly/integreatly-operator/pkg/resources/quota"
@@ -73,6 +75,8 @@ import (
 	l "github.com/integr8ly/integreatly-operator/pkg/resources/logger"
 	"github.com/integr8ly/integreatly-operator/pkg/resources/marketplace"
 	"github.com/integr8ly/integreatly-operator/version"
+
+	cloudcredentialv1 "github.com/openshift/api/operator/v1"
 )
 
 const (
@@ -88,6 +92,7 @@ const (
 	priorityClassNameEnvName         = "PRIORITY_CLASS_NAME"
 	managedServicePriorityClassName  = "rhoam-pod-priority"
 	routeRequestUrl                  = "/apis/route.openshift.io/v1"
+	addonStsArnParameterName         = "sts-role-arn"
 )
 
 var (
@@ -223,6 +228,8 @@ func New(mgr ctrl.Manager) *RHMIReconciler {
 // +kubebuilder:rbac:groups=marin3r.3scale.net,resources=envoyconfigs,verbs=get;list;watch;create;update;delete,namespace=integreatly-operator
 
 // +kubebuilder:rbac:groups=operator.marin3r.3scale.net,resources=discoveryservices,verbs=get;list;watch;create;update;delete,namespace=integreatly-operator
+
+// +kubebuilder:rbac:groups=operator.openshift.io,resources=cloudcredentials,verbs=get;list;watch
 
 func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 	_ = context.Background()
@@ -957,6 +964,21 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
 		}
 	}
 
+	log.Info("preflightChecks: checking if STS mode")
+	sts, err := checkIfStsClusterByCredentialsMode(context.TODO(), r.Client, installation.Namespace)
+	if err != nil {
+		log.Error("Error checking STS mode", err)
+		return result, err
+	}
+	if sts {
+		log.Info("validation of STS role ARN parameter ")
+		validArn, err := validateAddOnStsRoleArnParamaterPattern(r.Client, installation.Namespace)
+		if err != nil || !validArn {
+			log.Error("STS role ARN parameter pattern validation failed", err)
+			return result, err
+		}
+	}
+
 	installation.Status.PreflightStatus = rhmiv1alpha1.PreflightSuccess
 	installation.Status.PreflightMessage = "preflight checks passed"
 	err = r.Status().Update(context.TODO(), installation)
@@ -1418,4 +1440,64 @@ func (r *RHMIReconciler) processQuota(installation *rhmiv1alpha1.RHMI, namespace
 func isInstallationOlderThan1Minute(installation *rhmiv1alpha1.RHMI) bool {
 	aMinuteAgo := time.Now().Add(-time.Minute)
 	return aMinuteAgo.After(installation.CreationTimestamp.Time)
+}
+
+func validateAddOnStsRoleArnParamaterPattern(client k8sclient.Client, namespace string) (bool, error) {
+	// function is checking if STS addon parameter Pattern is valid
+	// Parameter is Valid only in case:
+	// 1.	Parameter exists and value matching AWS Role ARN pattern
+	// Parameter is Not valid  in other cases:
+	// 2.	parameter exists and value is NOT matching AWS Role ARN pattern
+	// 3.	parameter exists and value is empty
+	// 4.	parameter does not exists
+
+	awsArnPattern := "arn:aws(?:-us-gov)?:iam:\\S*:\\d+:role\\/\\S+"
+	r, err := regexp.Compile(awsArnPattern)
+	if err != nil {
+		log.Error("regexp Compile error: ", err)
+		return false, err
+	}
+
+	stsRoleArn, stsFound, err := addon.GetStringParameterByInstallType(
+		context.TODO(),
+		client,
+		rhmiv1alpha1.InstallationTypeManagedApi,
+		namespace,
+		addonStsArnParameterName,
+	)
+	if err != nil {
+		log.Error("failed while retrieving addon parameter", err)
+		return false, err
+	} else if stsFound && r.MatchString(stsRoleArn) { //case 1 - VALID
+		log.Info("ARN pattern is valid")
+		return true, nil
+	} else if stsFound && len(stsRoleArn) == 0 { //case 2 - NOT valid
+		log.Info("AWS STS role ARN parameter validation failed - parameter value is empty")
+		return false, nil
+	} else if !stsFound { //case 3 - NOT valid
+		log.Info("AWS STS role ARN parameter validation failed - parameter not found")
+		return false, nil
+	} else if stsFound && !r.MatchString(stsRoleArn) { //case 4 - NOT valid
+		log.Info("AWS STS role ARN parameter validation failed - parameter pattern is not matching to AWS ARN standard")
+		return false, nil
+
+	} else {
+		log.Info("AWS STS role ARN parameter validation failed")
+		return false, nil
+	}
+}
+
+func checkIfStsClusterByCredentialsMode(ctx context.Context, client k8sclient.Client, operatorNamespace string) (bool, error) {
+	cloudCredential := &cloudcredentialv1.CloudCredential{}
+	err := client.Get(ctx, k8sclient.ObjectKey{Name: "cluster"}, cloudCredential)
+	if err != nil {
+		log.Error("failed to get cloudCredential whle checking if STS mode", err)
+		return false, err
+	}
+	if cloudCredential.Spec.CredentialsMode == cloudcredentialv1.CloudCredentialsModeManual {
+		log.Info("STS mode")
+		return true, nil
+	}
+	log.Info("non STS mode")
+	return false, nil
 }
--- a/controllers/rhmi/rhmi_controller_test.go
+++ b/controllers/rhmi/rhmi_controller_test.go
@@ -1,0 +1,77 @@
+package controllers
+
+import (
+	"context"
+	"testing"
+
+	cloudcredentialv1 "github.com/openshift/api/operator/v1"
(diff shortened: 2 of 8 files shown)


controllers/rhmi/rhmi_controller.go
```diff
@@ -224,6 +229,8 @@ func New(mgr ctrl.Manager) *RHMIReconciler {
 
 // +kubebuilder:rbac:groups=operator.marin3r.3scale.net,resources=discoveryservices,verbs=get;list;watch;create;update;delete,namespace=integreatly-operator
 
+// +kubebuilder:rbac:groups=operator.openshift.io,resources=cloudcredentials,verbs=get;list;watch
```
matskiv, line 232: Considering that there will be just one CloudCredentials CR on the cluster, we can leave this code as is.
So just for future reference: we can limit the permissions to just `get` on a CR with specific name("cluster"):
```
// +kubebuilder:rbac:groups=operator.openshift.io,resources=cloudcredentials,verbs=get,resourceNames=cluster
```
That will work if you create a new client that doesn't use cache, e.g.:
```
serverClient, err := k8sclient.New(r.restConfig, k8sclient.Options{
  Scheme: r.mgr.GetScheme(),
})
// handle err
```
and pass that^ client to `checkIfStsClusterByCredentialsMode`

controllers/rhmi/rhmi_controller.go
```diff
@@ -957,6 +964,21 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
...
+	sts, err := checkIfStsClusterByCredentialsMode(context.TODO(), r.Client, installation.Namespace)
+	if err != nil {
+		log.Error("Error checking STS mode", err)
+		return result, err
+	}
+	if sts {
+		log.Info("validation of STS role ARN parameter ")
+		validArn, err := validateAddOnStsRoleArnParamaterPattern(r.Client, installation.Namespace)
+		if err != nil || !validArn {
+			log.Info("STS role ARN parameter pattern validation failed")
```
matskiv, line 977: Minor
```suggestion
			log.Error("STS role ARN parameter pattern validation failed")
```

controllers/rhmi/rhmi_controller.go
```diff
@@ -1419,3 +1441,63 @@ func isInstallationOlderThan1Minute(installation *rhmiv1alpha1.RHMI) bool {
...
+func validateAddOnStsRoleArnParamaterPattern(client k8sclient.Client, namespace string) (bool, error) {
+	// function is checking if STS addon parameter Pattern is valid
+	// Parameter is Valid only in case:
+	// 1.	Parameter exists and value matching AWS Role ARN pattern
+	// Parameter is Not valid  in other cases:
+	// 2.	parameter exists and value is NOT matching AWS Role ARN pattern
+	// 3.	parameter exists and value is empty
+	// 4.	parameter does not exists
+
+	awsArnPattern := "arn:aws:iam:\\S*:\\d+:role\\/\\S+"
```
matskiv, line 1454: We can future-proof this regexp pattern if we take other AWS partitions into account.
Particularly US GovCloud - https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-arns.html
```suggestion
	awsArnPattern := "arn:aws(?:-us-gov)?:iam:\\S*:\\d+:role\\/\\S+"
```

controllers/rhmi/rhmi_controller.go
```diff
@@ -1419,3 +1441,63 @@ func isInstallationOlderThan1Minute(installation *rhmiv1alpha1.RHMI) bool {
...
+	} else if stsFound && r.MatchString(stsRoleArn) { //case 1 - VALID
+		log.Info("ARN pattern is valid")
+		return true, nil
+	} else if stsFound && len(stsRoleArn) == 0 { //case 2 - NOT valid
+		log.Info("AWS STS role ARN parameter validation failed - parameter value is empty")
+		return false, nil
+	} else if !stsFound { //case 3 - NOT valid
+		log.Info("AWS STS role ARN parameter validation failed - parameter not found")
+		return false, nil
+	} else if stsFound && !r.MatchString(stsRoleArn) { //case 4 - NOT valid
```
matskiv, line 1480: Minor: this can be just `} else {` I believe.

matskiv: Oh, I now I see there is `else` block below.
What kind of other, "unknown", issue could there be? 🤔

pkg/products/cloudresources/reconciler_test.go
```diff
@@ -15,17 +17,24 @@ import (
...
 	oauthv1 "github.com/openshift/api/oauth/v1"
+	cloudcredentialv1 "github.com/openshift/api/operator/v1"
 	projectv1 "github.com/openshift/api/project/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	usersv1 "github.com/openshift/api/user/v1"
 	coreosv1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1"
 	operatorsv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	apiextensionv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
matskiv, line 28: Minor: naming the import `metav1` would better match our import naming convention.

controllers/rhmi/rhmi_controller_test.go
```diff
@@ -0,0 +1,147 @@
...
+	oauthv1 "github.com/openshift/api/oauth/v1"
+	cloudcredentialv1 "github.com/openshift/api/operator/v1"
+	projectv1 "github.com/openshift/api/project/v1"
+	routev1 "github.com/openshift/api/route/v1"
+	usersv1 "github.com/openshift/api/user/v1"
+	coreosv1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1"
+	operatorsv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	apiextensionv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
matskiv, line 22: Same as in other test file - Minor: naming the import metav1 would better match our import naming convention.

matskiv:
btw: `ci/prow/kubebuilder` check failed because of changes in the `config/rbac/role.yaml` file.
You need to run `make code/gen` and then commit changes to `config/rbac/role.yaml`.

controllers/rhmi/rhmi_controller.go
valerymo: Thank you very much for comment. Will use for future.

controllers/rhmi/rhmi_controller.go
valerymo: Done. Thank you

controllers/rhmi/rhmi_controller.go
valerymo: Done. Thank you
Just syntax is not so clear for me (probably I need look more at regular expressions),
for me something like following is more clear/readable : )        (aws)|(aws-us-gov)
But your syntax is working fine , I checked, so using it : )

controllers/rhmi/rhmi_controller.go
valerymo: I just removed word "unknown" : ), all remains same, as logic seems to me is ok. The order of cases in code could be any actually. But if you think that it's not so correct, maybe we could talk and I will update. Thank you.

pkg/products/cloudresources/reconciler_test.go
valerymo: Yes, sure, Thanks a lot. Done.

controllers/rhmi/rhmi_controller_test.go
valerymo: Done

valerymo:
Thank you , I didn't know about config/rbac/role.yaml.   Done.

controllers/rhmi/rhmi_controller.go
matskiv: You are right, this way - `(aws)|(aws-us-gov)`
it is probably more readable. Feel free to change it.

controllers/rhmi/rhmi_controller_test.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return nil, err
+	}
+	err = cloudcredentialv1.AddToScheme(scheme)
+	if err != nil {
+		return nil, err
+	}
+
+	return scheme, err
+}
```
matskiv, line 96: You don't need to add all this packages to schema.
The test will only use `cloudcredentialv1`.

pkg/products/cloudresources/reconciler.go
```diff
@@ -504,3 +534,75 @@ func (r *Reconciler) reconcileCIDRValue(ctx context.Context, client k8sclient.Cl
...
+	credSec := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      croSecretName,
+			Namespace: operatorNamespace,
+		},
+		Data: map[string][]byte{},
+	}
+
+	_, err := controllerutil.CreateOrUpdate(context.TODO(), client, credSec, func() error {
+		credSec.Data["AWS_ROLE_ARN"] = []byte(stsRoleArn)
```
matskiv, line 550:
```suggestion
		credSec.Data["role_arn"] = []byte(stsRoleArn)
		credSec.Data["web_identity_token_file"] = []byte("/var/run/secrets/openshift/serviceaccount/token")
```
Can you implement this^ change, please?
That^ is how CRO expects it at this point.
I should have communicated secret format upfront, sorry about that.

controllers/rhmi/rhmi_controller.go
```diff
@@ -1419,3 +1441,63 @@ func isInstallationOlderThan1Minute(installation *rhmiv1alpha1.RHMI) bool {
...
+		log.Info("AWS STS role ARN parameter validation failed - parameter pattern is not matching to AWS ARN standard")
+		return false, nil
+
+	} else {
+		log.Info("AWS STS role ARN parameter validation failed")
+		return false, nil
+	}
+}
+
+func checkIfStsClusterByCredentialsMode(ctx context.Context, client k8sclient.Client, operatorNamespace string) (bool, error) {
```
matskiv, line 1490: This is duplicating same function that is in pkg/products/cloudresources/reconciler.go

controllers/rhmi/rhmi_controller.go
```diff
@@ -1419,3 +1441,63 @@ func isInstallationOlderThan1Minute(installation *rhmiv1alpha1.RHMI) bool {
 	aMinuteAgo := time.Now().Add(-time.Minute)
 	return aMinuteAgo.After(installation.CreationTimestamp.Time)
 }
+
+func validateAddOnStsRoleArnParamaterPattern(client k8sclient.Client, namespace string) (bool, error) {
```
matskiv, line 1445: And in this function there is significant overlap with `getStsRoleArn` from the pkg/products/cloudresources/reconciler.go
Here we could just call that `getStsRoleArn` function and check if it return error or empty value.

pkg/products/cloudresources/reconciler_test.go
```diff
@@ -240,5 +249,151 @@ func getBuildScheme() (*runtime.Scheme, error) {
...
+			},
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		err := passArnIntoSecretInCroNamespace(tt.args.ctx, tt.args.client, operatorNamespace, tt.ARN)
+		if (err != nil) != tt.wantErr {
+			t.Errorf("passArnIntoSecretInCroNamespace() error = %v, wantErr %v", err, tt.wantErr)
+			return
+		}
```
matskiv, line 345: Here it would be good to check that secret is actually created.
You can do a .Get call with the mock client used in the test - `tt.args.client`
Example - https://github.com/integr8ly/integreatly-operator/blob/0bd1f9ea75a032a51eeb2d84ac070a0497aa3945/pkg/resources/pullSecret_test.go#L71-L76

matskiv:
We should add one more thing - a way to set ARN when running `make code/prepare`
So there would be a new variable for that in the Makefile, and it should ultimately end up in the `sts-role-arn` field of the`addon-managed-api-service-parameters` Secret.
Currently that Secret is created with `make cluster/prepare/quota`, so we might need to rename a few things, because it's not going to be just about quota.

controllers/rhmi/rhmi_controller.go
valerymo: Could we discuss please. We talked before that it will be useful for SRE to know exact issue with ARN parameter. So in ```preflightChecks(..)```  - it's carefully checked.  In  products/cloudresources/reconciler.go  -  this detailed check seems already  not needed (as it was already checked in preflightChecks, in rhmi_controller.go. This is the reason that I thought that in ```preflightChecks``` - it should be detailed check, and in ```reconciler.go``` in ```getStsRoleArn``` - just brief check (sorry if my explanation is a bit confusing)

controllers/rhmi/rhmi_controller.go
valerymo: Could we please discuss it.  We do validation of ARN parameters in two places - in ```rhmi_controller.go preflightChecks()``` and in ```reconciler.go```.  In both cases, before ARN parameter validation - it's required to know if it's STS mode.  (The difference just - that in ```preflightChecks``` -  ARN param validation is more detailed, as it could be helpful/informative for SRE. I added notes about this in next comment).

pkg/products/cloudresources/reconciler_test.go
valerymo: Done. Thank you for comment.

pkg/products/cloudresources/reconciler.go
valerymo: Done. Thank you for comment.

controllers/rhmi/rhmi_controller_test.go
valerymo: Done. Thank you for comment.

valerymo:
Hello @matskiv  , Thank you for review and comments. Regarding Makefile changes - I put some changes, but can't check it as Cluster creation stuck (case opened). I will check it as soon as cluster available. Thank you.

valerymo:
Makefile checked locally (CodeReady Containers).  Secret contains ARN parameter. Just not sure if parameter name is ok (sts-role-arn), also for quota (?)
$INSTALLATION_TYPE=managed-api make cluster/prepare/addon-params
$ oc get secret/addon-managed-api-service-parameters -oyaml
apiVersion: v1
data:
  addon-managed-api-service: MQ==
  sts-role-arn: YXJuOmF3czppYW06OjQ4NTAyNjI3ODI1ODpyb2xlLzEyMzQ1
kind: Secret
.....

PR #1905, opened 2021-06-23
increase upgrade alert time to 60 minutes

laurafitzgerald:
# Description

Update the upgradExpectedDurationExceeded alert to 60 minutes from 10 minutes
This is to accomodate for a longer upgrade in 1.8.0 where a forced upgrade of RDS engine version happens.

# Verification
checkout master

install everything - check the prometheus rule has 10 mintues set in the cr and in the Openshift Prometheus

Stop the operator

check the prometheus rule has 60 minutes set in the cr and in the Openshift Prometheus.

## Checklist


MStokluska:
/lgtm

Verified on Laura's cluster. Alert time is increased to 60mins and did not fired during upgrade.

PR #1927, opened 2021-06-29
Small fixes to test cases

trepel:
# Description

Small fixes to test cases. Mostly just said how to check whether workload webapp has been deployed already. Other changes are:
- fixed link to self managed apicast documentation
- how to check the health status of the cluster before installing RHOAM addon
- how to create request against 3scale and check the response (200 OK code)
- reverting changes made to RHOAM config done as part of the test case after finishing

Verification Steps
Just eye review mostly.

### DIFF
--- a/test-cases/tests/alerts/c18-validate-api-usage-alerts.md
+++ b/test-cases/tests/alerts/c18-validate-api-usage-alerts.md
@@ -2,7 +2,7 @@
 products:
   - name: rhoam
     environments:
-      - osd-fresh-install
+      - external
     targets:
       - 1.0.0
       - 1.3.0
@@ -26,11 +26,12 @@ estimate: 90m
    - ocmClusterLifespan: 36
    - openshiftVersion: (search for the version in the test plan document)
    - useByoc: false
+   - quota: 1
    - multiAZ: false
    - patchCloudResAwsStrCM: false
    - clusterID: api-usage-alerts
    - emailRecipients: <your@email.address>
-   - stepsToDo: provision + install
+   - pipelineSteps: provisionCluster, installProduct, setupIdp
 
 3. Valid SMTP credentials have been added to the `redhat-rhoam-smtp` secret
 
@@ -200,7 +201,3 @@ estimate: 90m
 
     In an earlier step the presence of the RHOAMApiUsageOverLimit was verified and a time noted. If 30 minutes has passed
     since then please verify that the alert is firing and that an email has been received to the BU and Customer email.
-
-18. Revert the value changed as part of this test case
-
-Use the commands from Step #4 and Step #5, just use the original values. Only required if the cluster is to be used for other test cases.
--- a/test-cases/tests/alerts/c18-validate-api-usage-alerts.md
+++ b/test-cases/tests/alerts/c18-validate-api-usage-alerts.md
@@ -70,7 +70,7 @@ estimate: 90m
 
     This updates the per minute rate limit to 694
 
-4.  Modify the `rate-limit-alerts` to allow alerts to fire on a per minute basis. Note the original values to be able to revert back once finished:
+4.  Modify the `rate-limit-alerts` to allow alerts to fire on a per minute basis:
 
     ```shell script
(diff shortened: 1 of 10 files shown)


test-cases/tests/alerts/c18-validate-api-usage-alerts.md
```diff
@@ -200,3 +200,7 @@ estimate: 90m
 
     In an earlier step the presence of the RHOAMApiUsageOverLimit was verified and a time noted. If 30 minutes has passed
     since then please verify that the alert is firing and that an email has been received to the BU and Customer email.
+
+18. Revert the value changed as part of this test case
+
+Use the commands from Step #4 and Step #5, just use the original values. Only required if the cluster is to be used for other test cases.
```
psturc, line 206: Not sure if reverting the changes is needed, because this test case is performed on a separate cluster (see the step n. 2)

Would it make sense to change the environment to `external` for this test case instead?

test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
```diff
@@ -51,6 +51,14 @@ Host prefix: /26
 
 9. Click on `Create cluster` (cluster creation takes ~40 minutes)
 
+Run following command:
+
+```
+ocm get subs --parameter search="cluster_id = '<your-cluster-ID>'" | jq -r .items[0].metrics[0].health_state
```
psturc, line 57:
```suggestion
# Copy your cluster's name from OCM UI ("test-ldap-idp" by default) and assign it to the env var CLUSTER_NAME
CLUSTER_NAME=<your-cluster-name>
# Get cluster's CID
CID=$(ocm get clusters --parameter search="display_name like '$CLUSTER_NAME'" | jq -r '.items[0].id')
ocm get subs --parameter search="cluster_id = '${CID}'" | jq -r .items[0].metrics[0].health_state
```

I copied those commands from lines 77-80, so if you agree with the suggestion, those lines below can be deleted.

test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
```diff
@@ -51,6 +51,14 @@ Host prefix: /26
...
+```
+
+Provide your cluster ID (e.g. 1lblvj4k7qgsqimqhbkc64n9q350748n). The command should return `healthy`.
```
psturc, line 60:
```suggestion
The command should return `healthy`.
```
(Please apply this if you agree with the suggestion above).
Also would it be worth to add an advice on what to do in case the cluster is not healthy?

psturc (commented):
Just couple of questions and suggestions, otherwise LGTM

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
trepel: Agree, better to have this external. Will update.

test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
trepel: Much nicer, I will do as suggested.

test-cases/tests/installation/a30-validate-installation-of-rhoam-addon-and-integration-with-ld.md
trepel: It would, but I have no piece of advice. The cluster should get `healthy` eventually. If not, the only thing I can think of is to try again

trepel:
@psturc suggestions addressed. Could you please have another look? Thanks

test-cases/tests/alerts/c18-validate-api-usage-alerts.md
```diff
@@ -69,15 +70,15 @@ estimate: 90m
...
     ```
 
 5.  Patch the `rhoam` CR to specify BU, SRE and Customer email addresses:
 
-    _NOTE:_ Replace `<rh_username>` references in the below commands with a valid Red Hat username. For example: `pamccart+BU@redhat.com`
+    _NOTE:_ Replace `<rh_username>` references in the below commands with a valid Red Hat username. For example: `pamccart+BU@redhat.com`. Note the original values to be able to revert back once finished.
```
psturc, line 81: should [the changes on these lines](https://github.com/integr8ly/integreatly-operator/pull/1927/files#diff-9925ccbb89aaa85d0090bc55d19296fda08506de84e32faba7bae2eb95012278R73-R81) be reverted as well?

trepel: Ah, right, good catch, completely forgot on those two lines.

trepel: Addressed. @psturc Could you please have yet another look?

PR #1931, opened 2021-07-01
Add go.sum vendor check & update go.sum file

Patryk-Stefanski:
# Description
Addition of git diff to check that go.sum gets updated or has no changes in pr.
Link to Jira: https://issues.redhat.com/browse/MGDAPI-2161

## Type of change

- Addition  of git diff for go.sum in the vendor check prow check 
-  Updated go.sum file

## Verification

- Check out this branch and replace the local go.sum file with the master go.sum file .
   Running the vendor check should then return an error


Boomatang:
The `make vendor/check` fails when the go.sum file has being modified.
/lgtm

PR #1938, opened 2021-07-02
Fix prep/release script

MStokluska:
# Description
Fixes issue with building bundles when using csv generated by the prepare/release script

# Verification
- create a new rhoam release and confirm that a multiline string in json pattern is created for containerImages in metadata.annotation of rhoam CSV
- try creating a bundle and index using following guide lines (it should not fail) 
https://github.com/RHCloudServices/integreatly-help/blob/master/guides/olm/installing-rhmi-bundle-format.md


carlkyrillos:
Verified that a json string was created for containerImages in metadata.annotations of the rhoam csv. Verified that I could successfully create a bundle and index using the new release.
/lgtm

PR #1972, opened 2021-07-14
MGDAPI-1846 - format code as part of release script

KevFan:
# Description

The release script updates the version in `version.go` by replacing the whole version string and thereby removing any whitespaces provided by `gofmt`
* https://github.com/integr8ly/integreatly-operator/blob/master/scripts/prepare-release.sh#L68-L75

This works for RHOAM but not for RHMI as RHOAM version string is longer with no extra whitespaces and so doesn't not fail the `prow` code format check when the release is generated.
* https://github.com/integr8ly/integreatly-operator/blob/master/version/version.go#L16-L17

To fix this, the code is formated at the end of the release script 

## Verification
* Checkout master
* Run `OLM_TYPE=integreatly-operator SEMVER=2.10.0 make release/prepare`
* Verify version string whitespace is removed and unformated in `version/version.go`
image
* Checkout this banch
* Run `OLM_TYPE=integreatly-operator SEMVER=2.10.0 make release/prepare`
* Verifiy whitespace is kept and formated in in `version/version.go`
image

## Type of change

## Checklist

### DIFF
--- a/scripts/prepare-release.sh
+++ b/scripts/prepare-release.sh
@@ -36,6 +36,13 @@ else
   KUSTOMIZE="/usr/local/bin/kustomize"
 fi
 
+# Path to gofmt
+if [[ -z $GOROOT ]]; then
+  GOFMT="/usr/local/go/bin/gofmt"
+else
+  GOFMT="$GOROOT/bin/gofmt"
+fi
+
 create_new_csv() {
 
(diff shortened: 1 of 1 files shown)


matskiv:
Verified 👍 
/lgtm
/approve

PR #1975, opened 2021-07-15
setup redis service maintenance critical security update alert

austincunningham:
# Description

MGDAPI-2172 add a new alert for redis critical security updates

## Verfication

- deploy this branch locally using a CCS/BYOC cluster
```bash
INSTALLATION_TYPE=managed-api USE_CLUSTER_STORAGE=false make deploy/integreatly-rhmi-cr.yml
```
- Go to the Prometheus route in the redhat-rhoam-middleware-monitoring-operator ns
- Check the alert is present

image

Currently deployed on `mw-collab-byoc`

## Checklist

### DIFF
--- a/pkg/resources/metrics.go
+++ b/pkg/resources/metrics.go
@@ -151,7 +151,7 @@ func ReconcileRedisAlerts(ctx context.Context, client k8sclient.Client, inst *v1
 		return v1alpha1.PhaseFailed, fmt.Errorf("failed to create redis prometheus cpu usage high alerts for %s: %w", cr.Name, err)
 	}
 
-	// create Redis Cpu Usage High Alert
+	// create Redis Service Maintenance Alert
 	if err = CreateRedisServiceMaintenanceAlerts(ctx, client, inst, cr, log); err != nil {
 		return v1alpha1.PhaseFailed, fmt.Errorf("failed to create redis prometheus service maintenance critical alerts for %s: %w", cr.Name, err)
 	}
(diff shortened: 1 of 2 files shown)


matskiv:
Code changes are fine.
I tested the query on our eng cluster with historical data and it looks as expected:
Screenshot 2021-07-19 at 14 01 30

matskiv:
@pat-cremin @Nanyte25 Hey guys,
we are adding a new alert for the Elasticache service updates, would you mind giving it a look?
There are two things that I want to run by you:
a) currently it's written to alert whenever Elasticache instance has service updates in "not complete" state. But there is a number of other states, so I want to double check if we should exclude some of them too, and not fire in those states.
List of all states: "not-applied", "waiting-to-start", "in-progress", "stopping", "stopped", "complete", "scheduling", "scheduled", "not-applicable".
b) There is no SOP link for this alert, and instead we are just saying that you should contact engineering, is that okay with you?

Nanyte25:
@matskiv 

I would certainly included some other Elastic-cache status at a minimum ``` "waiting-to-start", "in-progress", "stopping", "stopped", "complete", "scheduling", "scheduled" ```

There is an existing SOP to access the AWS account and check the Elastic-cache Status but All Critical alerts would require a SOP?

matskiv:
@Nanyte25 
Okay, we can exclude "`waiting-to-start`", "`in-progress`", "`scheduled`" and "`stopping`".
I don't think we should exclude "`scheduling`" status, as in that state it is still uncertain when the instance would be updated. We've seen it be in the status for a long time until it was updated.
Then the "`stopped`" status, I think it would go into this state if the update fails, so we shouldn't exclude it. I checked eng-rhoam cluster, and it never reported this state, so it doesn't seem like it would go into it during successful upgrade.

This is not critical alert, it's just a "warning". And I am not sure what the content of the SOP would be. We just need to release a new version of RHOAM to let CRO apply the update, I don't think SRE would do anything manually. But if needed, we can write a SOP that will say that engineering needs to be notified, instead of putting that note in the place where we usually have SOP url.

pat-cremin:
I would agree with Mark, there should still be a SOP attached if is a critical, it can lay out the process down as to who and where and how to contact Engineering create Jiras if required, etc.

austincunningham:
The Alert is warning level . I have no objections to writing a SOP if still required.

PR #1987, opened 2021-07-20
MGDAPI-2236: update 3scale-operator image

austincunningham:
# Description

MGDAPI-2236 update 3scale-operator image
## Type of change
3scale operator image update

## Verification

- checkout this branch
- install rhoam locally
- confirm 3scale is up and running and check the following containers to see if they have the correct image
  - in the redhat-rhoam-3scale-operator namespace the operator pod and the 3scale-operator container has the following sha sha256:5063f838a8a7649626231edabf47debc3d4e36f69de37675d126b19fdbeb69c3 in the ui
image
```bash
# you can also use the following command  
kubectl get pods --all-namespaces -o=jsonpath='{range .items[*]}{"\n"}{.metadata.name}{":\t"}{range .spec.containers[*]}{.image}{", "}{end}{end}' | sort | grep 3scale-rhel7-operator
# should return something similar to 
3scale-operator-6b4849b6f8-gtv65:       registry.redhat.io/3scale-amp2/3scale-rhel7-operator@sha256:5063f838a8a7649626231edabf47debc3d4e36f69de37675d126b19fdbeb69c3,
```

> *NOTE*: The rhoam upgrade scenario will not upgrade this image

## Checklist

### DIFF
--- a/manifests/integreatly-3scale/0.7.0/3scale-operator.v0.7.0.clusterserviceversion.yaml
+++ b/manifests/integreatly-3scale/0.7.0/3scale-operator.v0.7.0.clusterserviceversion.yaml
@@ -318,7 +318,7 @@ metadata:
     capabilities: Full Lifecycle
     categories: Integration & Delivery
     certified: "false"
-    containerImage: registry.redhat.io/3scale-amp2/3scale-rhel7-operator@sha256:655062177bc53b155b87876dc1096530c365f18f9be3ceb0d32aa2d343968f9a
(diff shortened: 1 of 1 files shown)


jackdelahunt:
/lgtm 
Ran the command above and it returned the image expected with the same sha so everything looks good. Just interested if the rhoam upgrade should update the image.
`

austincunningham:
@jackdelahunt thanks for the review. No the rhoam upgrade wont update image, there is a note in the verification regarding this.

PR #1994, opened 2021-07-22
[WIP] Adds a function to fetch MT users

jjaferson:
# Description

This PR adds a helper function to the user package to retrieve users that are not in the admins groups, this is required as part of the work to enable multi tenancy in RHOAM

### DIFF
--- a/pkg/resources/user/userHelper.go
+++ b/pkg/resources/user/userHelper.go
@@ -203,3 +203,20 @@ func getUsersFromAdminGroups(ctx context.Context, serverClient k8sclient.Client,
 
 	return adminUsers, nil
 }
+
+func GetIdentitiesByProviderName(ctx context.Context, serverClient k8sclient.Client, providerName string) (*usersv1.IdentityList, error) {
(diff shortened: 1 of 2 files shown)


pkg/resources/user/userHelper.go
```diff
@@ -153,3 +155,51 @@ func GetIdentities(ctx context.Context, serverClient k8sclient.Client, user user
 	}
 	return identities, nil
 }
+
+func GetMultiTenantUsers(ctx context.Context, serverClient k8sclient.Client) (*usersv1.UserList, error) {
+
+	exclusionGroups = append(exclusionGroups, "dedicated-admin")
```
briangallagher, line 161: Do we want to exclude some other groups: osd-sre-admins, cluster-admins, osd-sre-cluster-admins

jjaferson: those are the current excluded groups https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/user/userHelper.go#L25 I am not sure we need to add more groups to this list, @philbrookes what do you think?

PR #2016, opened 2021-08-05
Mgdapi 2446

jjaferson:
# Description

Adds the integration of the new 3scale tenant with RHSSO

## Verification steps

1. Install RHOAM with the multi tenant install type
`INSTALLATION_TYPE=multitenant-managed-api make cluster/prepare/local && INSTALLATION_TYPE=multitenant-managed-api make code/run`
2. Install the dev sandbox on the cluster and hook up RHSSO 
3. Register to the cluster via the dev sandbox
4. Verify if the new 3scale account is create for the new user with the auth provider configured


pkg/products/threescale/reconciler.go
```diff
@@ -1340,35 +1343,147 @@ func (r *Reconciler) updateKeycloakUsersAttributeWith3ScaleUserId(ctx context.Co
...
+
+	r.log.Infof("retrieving list of MT accounts available ",
+		l.Fields{"accounts": accounts},
+	)
+
+	for _, account := range accounts {
+		if account.State == "approved" {
+			for _, user := range account.Users.User {
+				if user.State == "pending" {
+					r.log.Infof("activating user", l.Fields{"user Activaded": user.Username})
```
briangallagher, line 1376: typo

pkg/products/threescale/reconciler.go
```diff
@@ -1340,35 +1343,147 @@ func (r *Reconciler) updateKeycloakUsersAttributeWith3ScaleUserId(ctx context.Co
...
+
+					err = r.tsClient.ActivateUser(*accessToken, account.Id, user.Id)
+					if err != nil {
+						r.log.Error("Error activating users in new tenant account:", err)
+						return integreatlyv1alpha1.PhaseFailed, err
+					}
```
briangallagher, line 1382: Add a break here?

pkg/products/threescale/reconciler.go
```diff
@@ -1340,35 +1343,147 @@ func (r *Reconciler) updateKeycloakUsersAttributeWith3ScaleUserId(ctx context.Co
...
+					Name:      "mt-signupaccount-3scale-access-token",
+					Namespace: r.Config.GetNamespace(),
+				},
+			}
+			err = serverClient.Get(ctx, k8sclient.ObjectKey{Name: signUpAccountsSecret.Name, Namespace: signUpAccountsSecret.Namespace}, signUpAccountsSecret)
+			if err != nil {
+				return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("Error getting access token secret: %w", err)
+			}
+
+			if val, ok := signUpAccountsSecret.Data[string(account.OrgName)]; ok {
```
briangallagher, line 1397: Could Data be nil here?

pkg/products/threescale/reconciler.go
```diff
@@ -1340,35 +1343,147 @@ func (r *Reconciler) updateKeycloakUsersAttributeWith3ScaleUserId(ctx context.Co
...
+
+		err = serverClient.Get(ctx, k8sclient.ObjectKey{Name: signUpAccountsSecret.Name, Namespace: signUpAccountsSecret.Namespace}, signUpAccountsSecret)
+		if !k8serr.IsNotFound(err) && err != nil {
+			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("Error getting access token secret: %w", err)
+		} else if k8serr.IsNotFound(err) {
+			signUpAccountsSecret.Data = map[string][]byte{}
+		}
+
+		signUpAccountsSecret.ObjectMeta.ResourceVersion = ""
+		signUpAccountsSecret.Data[string(account.OrgName)] = []byte(newSignupAccount.AccountAccessToken.Value)
```
briangallagher, line 1455: Could data be nil here?

pkg/products/threescale/reconciler.go
```diff
@@ -1379,30 +1494,149 @@ func (r *Reconciler) reconcile3scaleMultiTenancy(ctx context.Context, serverClie
...
+		Name:      oauthClientSecrets.Name,
+		Namespace: oauthClientSecrets.Namespace},
+		oauthClientSecrets,
+	)
+	if err != nil {
+		r.log.Errorf("could not find secret", l.Fields{"secret": oauthClientSecrets.Name, "operatorNamespace": oauthClientSecrets.Namespace}, err)
+		return fmt.Errorf("could not find %s Secret: %w", oauthClientSecrets.Name, err)
+	}
+
+	secret, ok := oauthClientSecrets.Data[tenantID]
```
briangallagher, line 1558: Could data be nil here?

jjaferson: it could but as the creation of the secret is handled in the bootstrap stage https://github.com/integr8ly/integreatly-operator/blob/master/controllers/rhmi/bootstrapReconciler.go#L312 and it makes sure data is created I think we don't need to verify it here

pkg/products/threescale/reconciler.go
jjaferson: I don't think so, there is an if statement above that checks whether the secret exists or not and if not it assigns a map to Data

https://github.com/integr8ly/integreatly-operator/pull/2016/files#diff-c99ff6dd3de866743c817f7b1dc230b4875e42eba21e5768f34fd92b9f873e32R1451

pkg/products/threescale/reconciler.go
jjaferson: it's considered when the secret is created

https://github.com/integr8ly/integreatly-operator/pull/2016/files#diff-c99ff6dd3de866743c817f7b1dc230b4875e42eba21e5768f34fd92b9f873e32R1451

PR #2011, opened 2021-08-03
Add Support for Multi Tenant User SSO

briangallagher:
# Description
Add Support for Multi Tenant User SSO

# Verify
Install RHOAM with this branch
Immediately after rhmi CR creation, stop the operator and change type from "managed-api" to "multitenant-managed-api". 
Restart Operator
Add the testing-idp after RHSSO installed: 
``` NUM_REGULAR_USER=10 NUM_ADMIN=5 DEDICATED_ADMIN_PASSWORD=Password1 PASSWORD=Password1 ./scripts/setup-sso-idp.sh ```
Create an "sre-admin" user in the testing-idp and follow this [SOP](https://github.com/RHCloudServices/integreatly-help/blob/master/sops/2.x/cssre_info/info_cssre_user_sso_access_master_admin_console_2x.asciidoc) to get admin access to user-sso. NOTE: Do not log in prematurely to user-sso (as mentioned in the SOP)!!
Create dev sandbox user namespaces
``` oc new-project test-user01-dev && oc new-project test-user02-dev ``` 
Log into OpenShift with test-user01 and test-user02
Wait for new KCRealm and KCUser CRs to be created 
Give permissions for users for their namespaces:
```oc adm policy add-role-to-user admin test-user01 -n test-user01-dev```
```oc adm policy add-role-to-user admin test-user02 -n test-user02-dev```

## Verify Creation
Log into Openshift as test-user01. Navigate to the Administration view (default is Developer). 
In the Top Right Corner of main pane there should be a link to SSO.
This will be the Realm login
Login and verify that the user is brought to their realm i.e. test-user01 Realm
User should have admin access to this realm but no other realm. 
Verify User and IDP configured in Realm

## Verify as admin
In a new private window login to usersso as sre-admin
Verify 2 new Realms, test-user01 and test-user02

## Verify Delete
`oc delete user test-user01 (deleted related identity also)`
`oc delete user test-user02 (deleted related identity also)`

Wait a period of time for reconcile.
Verify KCRealm and KCUser CRs gone
Verify user `consolelink` and user `oauthclient` CRs are also deleted.
`oc get consolelinks`
`oc get oauthclients`

## Verify delete as admin
In a new private window login to usersso as sre-admin
Verify 2 new Realms, test-user01 and test-user02, no longer exist

PR #2017, opened 2021-08-05
Mgdapi 2458 - Sanitise openshift username before being processed as a tenant

briangallagher:
# Description
sanitise username before creating tenant resources

# Verify
Follow the same steps as in [PR 2011](https://github.com/integr8ly/integreatly-operator/pull/2011) with the following change:
Instead of using test-user01 and test-user02 create 2 new users in the testing IDP with underscore and capital characters, for example, TEST_USER_1, TEST_USER_2


briangallagher:
invalid

PR #2019, opened 2021-08-05
sanitise tentant user names

briangallagher:
# Description
sanitise tentant user names before resources are created

# Verify
Verify in the same way as [PR 2011](https://github.com/integr8ly/integreatly-operator/pull/2011) except use "invalid usernames".
Create users TEST_USER_1 and TEST_USER_2 in the testing IDP and use them in place of test-user01 and test-user02

### DIFF
--- a/pkg/resources/user/userHelper.go
+++ b/pkg/resources/user/userHelper.go
@@ -29,6 +29,13 @@ var (
 	}
 )
 
+type MultiTenantUser struct {
+	Username   string
+	TenantName string
+	Email      string
+	UID        string
+}
+
 func GetUserEmailFromIdentity(ctx context.Context, serverClient k8sclient.Client, user usersv1.User) (string, error) {
 	email := ""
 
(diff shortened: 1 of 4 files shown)


pkg/resources/user/userHelper.go
```diff
@@ -232,3 +208,38 @@ func GetIdentitiesByProviderName(ctx context.Context, serverClient k8sclient.Cli
...
+	}
+	return users, nil
+}
+
+func SanitiseTenantUserName(username string) string {
+	// Regex for only alphanumeric values
+	reg, _ := regexp.Compile("[^a-zA-Z0-9]+")
+
+	// Replace all non-alphanumeric values with the replacement character
+	processedString := reg.ReplaceAllString(strings.ToLower(username), invalidCharacterReplacement)
```
austincunningham, line 241: If I am following this correctly this set every character to lower case and will replace any non-alphanumeric with a hyphen ? which should fix the issue with realm creation 
I will attempt to verifiy

briangallagher: yep, thats correct

PR #2028, opened 2021-08-09
New PR template draft

matskiv:
# JIRA link
N/A

# What
Rough proposal for the new PR template.
This is done as a follow-up from our retrospective discussion.

# Verification steps
N/A

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,16 +1,10 @@
-# Description
-<!-- Please include a summary of the change and a link to the JIRA ticket. Please add any additional motivation and context as needed -->
+# JIRA link
+<!-- insert a link to the JIRA ticket -->
 
-## Type of change
+# What
+<!-- describe a summary of the change, add any additional motivation and context as needed -->
 
-<!-- Please delete options that are not relevant. -->
-
-- [ ] Bug fix (non-breaking change which fixes an issue)
-- [ ] New feature (non-breaking change which adds functionality)
-- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-
-## Checklist
-- [ ] This change requires a documentation update <!-- Update JIRA with Affects -> Documentation -->
-- [ ] I have added tests that prove my fix is effective or that my feature works
-- [ ] I have added a test case that will be used to verify my changes 
-- [ ] Verified independently on a cluster by reviewer
+# Verification steps
+<!-- describe verification steps with sufficient level of detail -->
+<!-- take into consideration upgrade scenario, RHMI 2.x, etc. -->
+<!-- OR state "Not applicable" or "N/A" if your type of change doesn't require verification -->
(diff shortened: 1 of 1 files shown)


makslion:
I remember a discussion a bit ago about avoiding JIRA links in PRs :thinking: It was something to do with an open-source community. As far as I know, all PRs are from our team, so not sure it is relevant. wdyt? :thinking:

matskiv:
@makslion Generally, it's not good to link to private JIRAs in open source projects. But in our case there are no outside contributors, so I think it's acceptable, as it makes work much easier (especially when you open old PR and need more information).

There was also discussion about not mentioning CVEs in the PRs, so maybe you are mixing up the two? But it might be just me who doesn't remember the discussion you are referring to.

laurafitzgerald:
/test e2e
@matskiv i'm running tests here as a test. I'm seeing a failure before the tests even get run on another pr. 
Shouldn't impact what you are doing with this pr. Can be ignored essentially.

makslion:
Not sure how to share the link to GMail conversation, but it could be found by `No Red Hat or "internal" JIRA references in upstream bf2 repo`. The long story short: for public repos it would be nice to put all information in the PR description rather than a single link to the JIRA. However, having something to find an original JIRA also wouldn't hurt :thinking:  What do you think about this?

matskiv:
@makslion Oh yes, I remember that one. To be honest, for that org/repo they have much bigger community growing ambitions...
But yes, putting all information in the PR description would be good. Except for the security sensitive cases (especially non-public CVEs).

makslion:
I really want to see in a close future a large and growing community for RHOAM. And for those, theoretical future people, it would be nice to have more details in out current PRs, so when they would have a question "why and who did this?" they would be able to find descriptive PRs. For sure, CVEs and all confident-like information should be left as a secret. 
I guess just seeing the `#` header on the first like saying "JIRA link" makes me think that this is one of the most important sections :thinking: 
What about having something like: 
```
# Change
## What
## Why
<optional, could include JIRA link>
(or "JIRA link" could be optional another "##" header below) 
# Verfication 

# Further actions
<optional, e.g update docs, reinstall yq with v4, etc....) 
```
:thinking:

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,16 +1,10 @@
-# Description
-<!-- Please include a summary of the change and a link to the JIRA ticket. Please add any additional motivation and context as needed -->
+# JIRA link
+<!-- insert a link to the JIRA ticket -->
```
Boomatang, line 2: I understand @makslion  point but in the case where it is a community PR it would also be nice to see a link to an issue which was raised. Maybe a bit of rewording might be better.

```suggestion
# Issue link
<!-- insert a link to the JIRA ticket, GitHub issue or discussion thread -->
```

makslion: I'm not against this template in any way, I'm just trying to ensure that our templates are ready for a big and happy community. An `Issue link` would be lovely (as well as a new and fancy PR template :tada: ) It already looks good to me and with that change, I would be the happiest person in my room

matskiv: I like this suggestion. 
I'll not update the template just yet, as I want to give people more time for feedback. 
But if there is no opposition to this suggestion, consider it included. :)

PR #2049, opened 2021-08-19
MGDAPI-2489 - 3scale api error

jjaferson:
# Description

There is a issue when creating new tenant accounts in 3scale and smtp is not configured correctly the 3scale API create the tenant account but returns an error, to avoid this from happening the smtp secret needs to be created with empty values so the 3scale API doesn't attempt to send an email after the creation of the new tenant account.

https://issues.redhat.com/browse/MGDAPI-2489

## Type of change

## Verification steps
1) Install the operator from this branch INSTALLATION_TYPE=multitenant-managed-api make cluster/prepare/local && INSTALLATION_TYPE=multitenant-managed-api make code/run
2) Go to the 3scale API under the master account and create a new tenant account or run this curl command below
```
curl -v  -X POST "https://master.<CLUSTER-URL>/master/api/providers.xml" -d 'access_token=<MASTER-TOKEN>&org_name=test2&username=test2&email=test2%40test2.com&password=MT'
```
3) Make sure an error is not returned from the 3scale API.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -338,11 +338,11 @@ cluster/check/operator/deployment:
 .PHONY: cluster/prepare/smtp
 cluster/prepare/smtp:
 	@-oc create secret generic $(NAMESPACE_PREFIX)smtp -n $(NAMESPACE) \
-		--from-literal=host=smtp.example.com \
-		--from-literal=username=dummy \
-		--from-literal=password=dummy \
-		--from-literal=port=587 \
-		--from-literal=tls=true
+		--from-literal=host= \
+		--from-literal=username= \
+		--from-literal=password= \
+		--from-literal=port= \
+		--from-literal=tls=
 
 .PHONY: cluster/prepare/pagerduty
 cluster/prepare/pagerduty:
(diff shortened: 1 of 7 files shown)


Patryk-Stefanski:
1.Installed operator locally  from `3scale-api-error` branch
2.Logged into 3 scale API  as a master.
3.Created a new account successfully without any errors.

/lgtm

Boomatang:
Adding lgtm on behalf of @Patryk-Stefanski 

/lgtm

pmccarthy:
@jjaferson @MStokluska  would it be possible to limit the creation of the smtp secret to the `multitenant-managed-api` installation type? 

We currently have an alert checking that the smtp secret exists which is used by SRE in production to determine if the integration with the SendGrid service in OCM is working as expected. If we're always creating the smtp secret it means that this alert will always be green and we can no longer identify issues with the SendGrid service not creating legitimate credentials on cluster.

jjaferson:
> @jjaferson @MStokluska would it be possible to limit the creation of the smtp secret to the `multitenant-managed-api` installation type?
> 
> We currently have an alert checking that the smtp secret exists which is used by SRE in production to determine if the integration with the SendGrid service in OCM is working as expected. If we're always creating the smtp secret it means that this alert will always be green and we can no longer identify issues with the SendGrid service not creating legitimate credentials on cluster.

@pmccarthy this PR only removes the dummy data set in the smtp secret to avoid the 3scale api to return errors when the smtp account is not valid, if it's the intention to remove the creation of the smtp I would suggest we create another ticket to address that as it will bring up other issues.

MStokluska:
@pmccarthy I've verified it again, turns out it was a false alert and this PR is ready to be approved. Sorry for holding it up @jjaferson

PR #2051, opened 2021-08-19
MGDAPI-2526: Introduce polling for SA token

sergioifg94:
# Description

The test has failed ocassionally when retrieving the token that's created for a testing ServiceAccount. The reason of this is that the token is attempted to be retrieved immediately after the SA is created, which might result if the token not being created by the time that code is executed.
Fix by introducing polling which waits for the token secret to be set in the service account `secrets` list.

# Verification

* The E05 test passes

## Type of change

## Checklist

### DIFF
--- a/test/common/grafana_routes.go
+++ b/test/common/grafana_routes.go
@@ -9,11 +9,14 @@ import (
 	"net/http/httputil"
 	"os"
 	"strings"
+	"time"
 
 	v12 "github.com/openshift/api/authorization/v1"
 	v1 "github.com/openshift/api/route/v1"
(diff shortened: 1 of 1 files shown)


Boomatang:
To verify I check the logs in the prow e2e test runs for both rhoam and rhmi. The code looks good and nothing is jumping out at me.
As this was a bit of a flaky test proving this fixed the issue will be hard but over the next few week close attention will be give to this test in the pipeline watches.

/lgtm

sergioifg94:
Sounds like a good approach, thanks @Boomatang

PR #2079, opened 2021-08-30
fix: run C19 before H11

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-1260

# What

Running H11 before C19 can cause C19 to timeout waiting for alert to fire (seen in nightly runs)

The reason for this is that at the end of H11, the smtp secret is [restored](https://github.com/integr8ly/integreatly-operator/blob/master/test/common/3scale_smtp.go#L102-L106) and this causes 3scale deployments to be [rescaled](https://github.com/integr8ly/integreatly-operator/blob/62da48db2fbb22bcbb61e63338a9fbc7277ebef1/pkg/products/threescale/reconciler.go#L522-L532). However, C19 will imediately run, but at 3scale reconcile, it will wait until these deployments are fully scaled before it can reach the code for reconciling openshift users. This can cause a timeout on C19 for waiting the alert to fire due to this.

Quickest fix for this is running C19 before H11 

# Verification steps

Sample Run of Test : [test-run.log](https://github.com/integr8ly/integreatly-operator/files/7076486/test-run.log)

* Install RHOAM using image
```
IN_PROW=true make cluster/deploy
```
* Once complete, install IDP using script
```
PASSWORD=Password1 DEDICATED_ADMIN_PASSWORD=Password1 ./scripts/setup-sso-idp.sh
```
* Run IDP tests against cluster with RHOAM installed
```
INSTALLATION_TYPE=managed-api BYPASS_STORAGE_TYPE_CHECK=true TEST="IDP" make test/e2e/single 
```

### DIFF
--- a/test-cases/tests/alerts/c19-validate-creation-of-invalid-username-triggers-alert.md
+++ b/test-cases/tests/alerts/c19-validate-creation-of-invalid-username-triggers-alert.md
@@ -18,4 +18,71 @@ estimate: 15m
 
 # C19 - Validate creation of invalid username triggers alert
 
-https://github.com/integr8ly/integreatly-operator/blob/master/test/common/alerts_invalid_username.go
+## Prerequisites
+
+1. Login to OpenShift console and via `oc` as a user with **cluster-admin** role (kubeadmin):
+
+   ```shell script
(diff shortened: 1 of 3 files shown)


Boomatang:
I have ran the verification steps, all looks good to me.

/lgtm

PR #2083, opened 2021-08-30
Fixed Quota name in 3scale.go

trepel:
# What

The A14 and F05 tests fails if using 1M quota. The reason is that due to mismatch in quota names the default value (3 replicas) is expected for apicast-production. But the actual value is two so the tests fails (since the actual is less than expected).

# Verification steps
Assuming you have the RHOAM cluster ready just run the A14 and F05 tests. It is enough to do this for 100k, 1M and 50M quotas (other quotas has the same replicas as the default values so no real change here).

* change the quota via OCM
* oc login into the cluster
* clone this PR
* INSTALLATION_TYPE=managed-api TEST=A14 make test/e2e/single
* INSTALLATION_TYPE=managed-api TEST=F05 make test/e2e/single

Or maybe just check that the values matches the quota from rhoam RHMI CR would be sufficient.

### DIFF
--- a/pkg/config/threescale.go
+++ b/pkg/config/threescale.go
@@ -128,23 +128,23 @@ func (t *ThreeScale) GetReplicasConfig(inst *integreatlyv1alpha1.RHMI) map[strin
 			threeScaleComponents["apicastProd"] = 2
 			threeScaleComponents["backendListener"] = 2
 			threeScaleComponents["backendWorker"] = 2
-		case "10":
+		case "1 Million":
 			threeScaleComponents["apicastProd"] = 2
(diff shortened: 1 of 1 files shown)


trepel:
Since this is not technically a code in tests, I am also running the fresh installation from this PR.

trepel:
The fresh install from this PR finished successfully.

psturc:
changes lgtm, and based on the verification that was done I'm ok with merging this

/lgtm

laurafitzgerald:
changes look good to me. Apologies for missing this in the quota change. 
/approve

PR #2087, opened 2021-08-31
use tenant 3scale email from identitiy and set pod priority to 0 for …

briangallagher:
…multitenancy

# Issue link
https://issues.redhat.com/browse/MGDAPI-2565
and
https://issues.redhat.com/browse/MGDAPI-2557

# What

# Verification steps - email 
Install multi-tenant RHOAM from this branch
` INSTALLATION_TYPE=multitenant-managed-api make code/run`

Install the testing-idp with "rhd" name
`NUM_REGULAR_USER=5 NUM_ADMIN=0 DEDICATED_ADMIN_PASSWORD=Password1 PASSWORD=Password1 REALM_DISPLAY_NAME="rhd" REALM="rhd" ./scripts/setup-sso-idp.sh`

Log in as test-user01 and then test-user02 to OpenShift
Verify that the identity CR has an email test-user01@example.com and test-user02@example.com

Wait for the tenants route to be created under redhat-rhoam-3Scale
Log in using SSO as both users
Go to Account => Users and verify that the email is `test-user01@example.com` and `test-user02@example.com` respectively 

# Verification steps - pod priority
Check APIcast staging and Production pods and verify pod priority is set to 0

### DIFF
--- a/pkg/products/threescale/reconciler.go
+++ b/pkg/products/threescale/reconciler.go
@@ -1429,12 +1429,12 @@ func (r *Reconciler) reconcile3scaleMultiTenancy(ctx context.Context, serverClie
 	}
 
 	// creating new MT accounts in 3scale
-	accountsToBeCreated := getMTAccountsToBeCreated(mtUserIdentities.Items, accounts)
+	accountsToBeCreated, emailAddrs := getMTAccountsToBeCreated(mtUserIdentities.Items, accounts)
 	r.log.Infof("retrieving new MT accounts to be created",
 		l.Fields{"accountsToBeCreated": accountsToBeCreated},
 	)
 
-	for _, account := range accountsToBeCreated {
+	for idx, account := range accountsToBeCreated {
 
 		pw, err := r.getTenantAccountPassword(ctx, serverClient, account)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/products/threescale/reconciler.go
```diff
@@ -1625,9 +1625,15 @@ func getMTAccountsToBeCreated(usersIdentity []usersv1.Identity, accounts []Accou
 				Name:    identity.User.Name,
 				OrgName: identity.User.Name,
 			})
+			if identity.Extra["email"] != "" {
+				email = identity.Extra["email"]
+			} else {
+				email = fmt.Sprintf("%s@rhmi.io", identity.User.Name)
+			}
+			emailAddrs = append(emailAddrs, email)
```
MStokluska, line 1633: Hi Brian, I don't see an issue with it, but more of a question, do you think it would be better to have email set as part of AccountDetail object? In here: https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/threescale/types.go#L67
Maybe we could add it as part of future tasks?

briangallagher: I'm not sure if its a good place as AccountDetail has multiple users associated. We could experiment with it in future though.

MStokluska:
Verified that multitenancy logic works correctly, users got created with correct email.
Verified that the priority is set to 0.
/lgtm

pkg/products/threescale/reconciler.go
MStokluska: Thanks Brian, as for our call yesterday, we have agreed that this is possibly something we can't do atm but might investigate further at some later stage.

PR #2095, opened 2021-09-02
test: mark c19 flaky for nightly runs

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-1260
# What

C19 is failing on nightly runs which looks to be only on fresh installs as subsequents runs succeed consistently. For now, skip the test if it fails at the flaky stage and include command to run manually in test case

This will be investigated in a follow up ticket:
* https://issues.redhat.com/browse/MGDAPI-2581
# Verification steps

N/A


Boomatang (commented):
Not use what is required to get that linter to pass but code looks good.

/lgtm

PR #2098, opened 2021-09-03
Include suggested-namespace annotation in base CSV

sergioifg94:
# Issue link

# What

The `operatorframework.io/suggested-namespace` is supported by OLM and operator-sdk to specify the default/recommended namespace where the operator might be installed. This provides a better user experience, as well as advantages when testing using the scorecard format.

Include this annotation in the base CSVs of both RHOAM and RHMI. The CSV generated will also include the annotation.

# Verification steps

Verify that generating a new set of manifests will result in the generated CSV containing the new annotation

> Perform this test both for RHMI and RHOAM 

1. Run the prepare release script
2. Check that the generated CSV has the corresponding value for the `operatorframework.io/suggested-namespace` annotation

### DIFF
--- a/config/manifests-rhmi/bases/integreatly-operator.clusterserviceversion.yaml
+++ b/config/manifests-rhmi/bases/integreatly-operator.clusterserviceversion.yaml
@@ -2,6 +2,7 @@ apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 metadata:
   annotations:
+    operatorframework.io/suggested-namespace: redhat-rhmi-operator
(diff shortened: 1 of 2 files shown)


Patryk-Stefanski:
Ran release script for RHOAM and RHMI , both contain new  annotation.
/lgtm

KevFan:
* Verified `operatorframework.io/suggested-namespace: redhat-rhmi-operator` annotation added to RHMI csv
* Verified `operatorframework.io/suggested-namespace: redhat-rhoam-operator` annotation added to RHOAM csv

/lgtm

PR #2121, opened 2021-09-13
Update 3scale bundle in installation-cpaas.yaml

sergioifg94:
# What

The wrong image had been set for this field (operator image instead of the bundle image). Revert the change as we know now that the registry will be replaced for a valid one when building in OSBS

# Verification steps


carlkyrillos:
Reviewed changed files.
/lgtm

PR #2122, opened 2021-09-13
[MGDAPI-2585] fix: replaces being removed when csv is re-generated

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-2585
# What

With the bump to `yq v4`, `null` seems to be returned when trying to get a field that does not exist in a yaml file.

This had caused the `-z ` check to always be true in  the `check_csv_replaces_field()` function, which caused the `replaces` field to be missing whenever existing CSVs need to be re-generated such as from RC releases to final releases.

 To fix this:
* The above checks for null being returned instead

Additionally:
* Removed some unneeded extra whitespaces

# Verification steps

##  Verify issue in master
* Checkout master
* Generate new release
```
SEMVER=1.13.0 OLM_TYPE=managed-api-service ORG=kevfan scripts/prepare-release.sh
```
* Verify `replaces` field is **present**
* Re-generated release
```
SEMVER=1.13.0 OLM_TYPE=managed-api-service ORG=kevfan scripts/prepare-release.sh
```
* Verify `replaces` field is **missing**
* Revert / Delete generated release
## Verify branch fixes issue
* Checkout this branch
* Generated a new release
```
SEMVER=1.13.0 OLM_TYPE=managed-api-service ORG=kevfan scripts/prepare-release.sh
```
* Verify `replaces` field is **present**
* Re-generate release
```
SEMVER=1.13.0 OLM_TYPE=managed-api-service ORG=kevfan scripts/prepare-release.sh
```
* Verify `replaces` field is still **present** (will be the last field at the bottom of the csv yaml file)

### DIFF
--- a/scripts/prepare-release.sh
+++ b/scripts/prepare-release.sh
@@ -97,7 +97,7 @@ set_installation_type() {
           echo "using default INSTALLATION_TYPE found in deploy/operator.yaml"
           ;;
         "managed-api-service")
-          yq e -i '(.spec.install.spec.deployments[0].spec.template.spec.containers[0].env.[] | select(.name=="INSTALLATION_TYPE") | .value) = "managed-api" '  packagemanifests/$OLM_TYPE/${VERSION}/$OLM_TYPE.clusterserviceversion.yaml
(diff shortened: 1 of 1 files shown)


carlkyrillos:
Reviewed changes to the `prepare-release.sh` script and verified that the issue does exist in master and that this PR fixes that issue.
/lgtm

PR #2128, opened 2021-09-15
[MGDAPI-2654] fix: do not append default domain to users with username that are email compliant

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-2654

# What

When a user signs into openshift via IDP, the `Identity` resource created on openshift may not have an `email` field assoicated with it. In this case, a default email using their username and appending `@rhmi.io` is used when creating the keycloak user.

This normally is fine, but when the username is an email (e.g. `test@test.com`), `@rhmi.io` is still appended to it (e.g. `test@test.com@rhmi.io`) and when the operator attempts to create the user in 3scale, it will fail due to the double domain appended email address as it is is not valid. This will give a `422` error during the 3scale user creation and cause the `ThreeScaleUserCreationFailed` alert to fire

To further prevent this scenario:
* Use username directly as email if following standard email format
* Otherwise sanitise username before appending default domain for use as generated email address

# Verification steps

## Authomated Verification
* Install RHOAM from this branch
```
make cluster/prepare/local
make code/run
```
* Once installed, run IDP script
```
PASSWORD=Password1 DEDICATED_ADMIN_PASSWORD=Password1 ./scripts/setup-sso-idp.sh    
```
* Run the e2e test:
```
INSTALLATION_TYPE=managed-api BYPASS_STORAGE_TYPE_CHECK=true TEST="B06" make test/e2e/single
```
* Verify test completes successfully

## Manual Verification
If you want to do a full manual verification involving checking 3scale and alerts:
* In RHSSO, create a new user with an email as their username (`test@test.com`) in testing-idp
* Set a password for this user, ensuring the temporary checkbox is switched off
* Sign into openshift through testing idp as this user
* Verify 3scale reconcile completes successfully
* Verify the `email` of the generated `KeycloakUser` for this user does not contain an extra `@rhmi.io` and is just using the username as email
* Verify this user can sign into 3scale
* Verify `ThreeScaleUserCreationFailed` alert is **not** firing

### DIFF
--- a/test/common/default_email.go
+++ b/test/common/default_email.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	goctx "context"
@@ -14,9 +15,12 @@ import (
 )
 
 const (
-	usernameNoEmail   = "autotest-user01"
-	usernameWithEmail = "autotest-user02"
-	existingEmail     = "autotest-nondefault-user02@hotmail.com"
+	usernameNoEmail              = "autotest-user01"
+	usernameWithEmail            = "autotest-user02"
+	usernameUsingEmail           = "autotest-user03@test.com"
+	usernameRequireSanitiseEmail = "autotest$user04"
+	existingEmail                = "autotest-nondefault-user02@hotmail.com"
+	defaultDomain                = "@rhmi.io"
 )
 
 // TestDefaultUserEmail verifies that a user syncronized from the IDP have a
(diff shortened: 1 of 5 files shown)


carlkyrillos:
Reviewed changed files and verified the fix works using both the automatic and manual verification methods.
/lgtm

test/common/default_email.go
```diff
@@ -66,8 +88,20 @@ func TestDefaultUserEmail(t TestingTB, ctx *TestingContext) {
...
 
+	keycloakUser3, err := waitForKeycloakUser(ctx, 5*time.Minute, rhssoNamespace, usernameUsingEmail)
+	if err != nil {
+		t.Fatalf("Unexpected error querying KeycloakUser %s: %v", usernameUsingEmail, err)
+		return
+	}
+
+	keycloakUser4, err := waitForKeycloakUser(ctx, 5*time.Minute, rhssoNamespace, usernameRequireSanitiseEmail)
+	if err != nil {
+		t.Fatalf("Unexpected error querying KeycloakUser %s: %v", usernameUsingEmail, err)
```
KevFan, line 99:
```suggestion
		t.Fatalf("Unexpected error querying KeycloakUser %s: %v", usernameRequireSanitiseEmail, err)
```
Copy / Paste typo. Going to push a fix for this :man_technologist:

KevFan:
@briangallagher Could you take a quick look at this also ? I've included this change to rhoam multitenant logic since I think it could potentially hit this issue there also :thinking:

carlkyrillos:
Re-adding /lgtm after typo fix.

PR #2131, opened 2021-09-16
[MGDAPI-2474] feat: marin3r 0.8.0

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-2474

# What

Upgrade marin3r operator to v0.8.0

# Verification steps

## Verifiy Fresh Install
* Checkout this branch
* Install RHOAM
```
make cluster/prepare/local
make code/run
```
* Verify install complete successfully
* Sign into 3scale as `admin` to complete 3scale tutorial to create echo example endpoint
* Using endpoint created (under Products -> API -> Integration -> Configuration), generate traffic to verify rate limiting still works correctly
  * https://github.com/rakyll/hey
  * Shoudl have 30 or 31 out of 100 requests that are rate limited - 429 status code
```
hey -c 1 -n 100 "<3scale_endpoint>"
```

## Verify Upgrade
* Unintall RHAOM
* Checkout master
* Install from master
```
make cluster/prepare/local
make code/run
```
* Once installation complete, stop the opertor
* Checkout this branch
* Run operator
```
make code/run
```
* Verify marin3r upgrade successfully completes and apicast / backend-listener pods are redeployed
```
 watch "echo Marin3r; oc get csvs -n redhat-rhoam-marin3r-operator; oc get pods -n redhat-rhoam-marin3r; oc get pods -n redhat-rhoam-3scale | grep -E 'apicast|backend-listener'; echo "RHMI CR Status:"; oc get rhmi rhoam -n redhat-rhoam-operator -o json | jq '.status.stage'"
```
* Complete 3scale tutorial to create echo example endpoint
* Sign into 3scale as `admin` to complete 3scale tutorial to create echo example endpoint
* Using endpoint created (under Products -> API -> Integration -> Configuration), generate traffic to verify rate limiting still works correctly
  * https://github.com/rakyll/hey
  * Shoudl have 30 or 31 out of 100 requests that are rate limited - 429 status code
```
hey -c 1 -n 100 "<3scale_endpoint>"
```


pkg/config/marin3r.go
```diff
@@ -1,7 +1,7 @@
 package config
 
 import (
-	marin3rv1alpha1 "github.com/3scale/marin3r/apis/marin3r/v1alpha1"
```
Boomatang, line 4: the adding of the `-ops` into these paths, what other affects does this mean? I am thinking does this change paths to quay, does this affect the EWS? These are open ended questions and I am not sure what answers I am looking for. Any other information you can give would be useful.

Boomatang (commented):
Only the one question related to the `-ops` suffix which is not affecting the changes. Starting to do the verification steps.

pkg/config/marin3r.go
KevFan: I dont think it has much affect unless these paths are hardcoded elsewhere as this is a dependency change :thinking: 

I believe this change is due to the repo is now https://github.com/3scale-ops/marin3r but previous it was https://github.com/3scale/marin3r
* https://github.com/3scale-ops/marin3r/blob/v0.7.0/bundle/manifests/marin3r.clusterserviceversion.yaml#L139
* https://github.com/3scale-ops/marin3r/blob/v0.8.0/bundle/manifests/marin3r.clusterserviceversion.yaml#L139

In terms of quay, they are still using the same quay repo https://quay.io/repository/3scale/marin3r?tag=latest&tab=tags so there doesn't look to be changes needed there. 

In terms of EWS, in not sure is marin3r being watched it and am not familiar with the system itselft to comment whether this affects it :thinking: :shrug:

Boomatang:
After doing the verification steps. Every thing looks goog
/lgtm

pkg/config/marin3r.go
Boomatang: If this is the same quay.io repo I want to check one of the pipelines that use the products.yaml file. I still your changes are correct but the pipeline may need adjusting.

Boomatang: I had a look into what I felt might have needed some follow on work and it doesn't. We are all good.

PR #2137, opened 2021-09-21
add prow vendor check with config to pull private repos

laurafitzgerald:
# What
Adding an additional prow check in order to provide the steps required by prow to configure git/go to be able to pull a private repository. The reason we need this change is because the observability-operator repo is private. This will be needed on master soon.

# Verification steps
Run `make vendor/check/prow`
If you do not have the variables exported you the script should tell you.
export the values.
The don't really matter for now.
Run the script
There should be a new block of configuration in you `~/.gitconfig` file which looks like

```
[url "https://<the username>:<the token>@github.com@github.com"]
        insteadOf = https://github.com
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -461,6 +461,11 @@ vendor/check: vendor/fix
 	git diff --exit-code vendor/
 	git diff --exit-code go.sum
 
+.PHONY: vendor/check/prow
+vendor/check/prow:
+	sh scripts/setup-private-git-access.sh
+	make vendor/check
+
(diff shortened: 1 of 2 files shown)


laurafitzgerald:
Going to merge this as it's isolated to the new make command. need to merge in order to get the related rehearsals running on https://github.com/openshift/release/pull/22098

Boomatang (commented):
A part from the miss match in help everything worked as expected

PR #2139, opened 2021-09-21
test: test-case: h23: update machine IP

psturc:
The previous rate-limit-testing machine was accidentally deleted. New instance was created and the IP has changed.


trepel:
The ssh-ing to that IP worked

/lgtm
/approve

PR #2146, opened 2021-09-23
[WIP] MGDAPI-2663: Improve admin reconciliation in user SSO

sergioifg94:
# Issue link
https://issues.redhat.com/browse/MGDAPI-2663

# What
Improve the logic of the User SSO reconciliation

## Group membership

The User SSO reconciler keeps in sync the dedicated admins group in Openshift with the dedicated admins group in Keycloak. It does so by reconciling the corresponding KeycloakUser resources and delegating on the Keycloak operator to propagate the changes to Keycloak. However, in the case of group membership, the Keycloak operator only reconciles group membership when creating a new user, so if the user already exists on Keycloak, updating the groups field in the resource won't result in
the membership being added in Keycloak.

This PR adds the logic to ensure the membership is in sync
in every reconcile by adding the user to the groups in the Keycloak API.

## Conflict errors

Avoid possible conflict issues when the Keycloak operator attempts to create a user that already exists, by setting the ID in the
KeycloakUser resource if the user is found on the API.

## Error messaging

When the Keycloak operator fails to reconcile a KeycloakUser resource, it sets the status to "failing". This has been currently ignored by the RHOAM operator, which results in errors silently ignored. Add the logic to the resource reconciliation to check the status and fail the reconciliation if the resource status is not positive

# Verification steps

> Coming soon


psturc:
Verified

Screenshot 2021-10-05 at 15 58 19

@sergioifg94 
just to confirm - is it expected behaviour that when user gets deleted from the dedicated-admins group in openshift, this change doesn't get synced with user sso?

psturc:
@sergioifg94 thanks, so in that case after PR checks are passing, this PR lgtm

PR #2150, opened 2021-09-24
Update 3scale bundle to 2.11 on installation-cpaas

sergioifg94:
# What

The current bundle being referenced is from 2.10, which includes images in the CSV that cause IIB to fail when running the Managed Service release pipeline. 

Update the 3scale bundle on the CPaaS installation spec to the 2.11 build from `3scale-operator-bundle-container-2.11.0-3`


carlkyrillos:
Reviewed change to installation-cpaas.yaml config file.
/lgtm

PR #2173, opened 2021-10-07
fix: remove manual creation of DMS alert

KevFan:
# Issue link

# What

DMS creation will be handled by Observability Operator itself 
- https://github.com/redhat-developer/observability-operator/pull/87

If we do not remove the manually created one, there will be 2 DMS alerts created

# Verification steps

Just a code review should be good enough as this is just a deletion of the alert


KevFan:
/cc @laurafitzgerald PR for removing the `DeadMansSnitch` Alert

If we choose to merge this is a later release, then I think additional once off cleanup code will need to be added to remove the prometheus rule from the cluster :thinking:

PR #2188, opened 2021-10-10
MGDAPI-1882 - Update/Create automated e2e tests/test cases (master based)

valerymo:
# Issue link
https://issues.redhat.com/browse/MGDAPI-1882

Current PR replaces old one: https://github.com/valerymo/integreatly-operator/pull/25 (oo-install based)

# What
Automation tests update for Observability Operator.
- Updated tests list: E01,E02,E03,E05,E09,"Verify Alerts are not firing...", A03,A05-A09,A13,A15, C01,C03,C04.

# Verification steps
- Check from command line that all updated Tests are passed.
- Check in pipeline that tests are passed.


test/common/types.go
```diff
@@ -21,6 +21,9 @@ var (
 	RHMIOperatorNamespace             = NamespacePrefix + "operator"
 	MonitoringOperatorNamespace       = NamespacePrefix + "middleware-monitoring-operator"
 	MonitoringFederateNamespace       = NamespacePrefix + "middleware-monitoring-federate"
+	ObservabilityOperatorNamespace    = NamespacePrefix + "observability-operator"
+	ObservabilityProductNamespace     = NamespacePrefix + "observability"
+	ObservabilityNamespacePrefix      = NamespacePrefix + "observability-"
```
MStokluska, line 26: Hey @valerymo wdyt if we would remove the observabilityNamespacePrefix and instead use the observabilityProductNamespace? We would add the `-` at each string where it's required. 
Not a big deal if you would like this to stay as it is, however, namespacePrefix seems more like "redhat-rhoam" or "redhat-sandbox" etc rather than adding a `-` to ProductNs.

test/common/deployment_types.go
```diff
@@ -90,10 +90,9 @@ func getDeploymentConfiguration(deploymentName string, inst *integreatlyv1alpha1
 				{Name: "syndesis-operator", ExpectedReplicas: 1},
 			},
 		},
-		"monitoringOperatorDeployment": {
```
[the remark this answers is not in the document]
valerymo: Hey @MStokluska , seems it looks ok, we are checking OO deployment. This is code at lines 93-99: https://github.com/integr8ly/integreatly-operator/blob/c8cfb7ad2307b28efd69fe2ef6053cc70668c9f2/test/common/deployment_types.go#L93-L99

		"observabilityDeployment": {
			Name: ObservabilityProductNamespace,
			Products: []Product{
				{Name: "grafana-deployment", ExpectedReplicas: 1},
				{Name: "grafana-operator", ExpectedReplicas: 1},
				{Name: "prometheus-operator", ExpectedReplicas: 1},
			},
		},
Thank you

test/common/types.go
valerymo: Hey @MStokluska , thank you very much for your comment. I also thought about this - add "-" to code at each place or just define this as prefix.  But main thing, as seems to me, code maintenance;   "-" at 34  places - it's a source of mistakes. Also, there are another definitions in alerts_exist.go : File: NamespacePrefix + "...." , so I wanted to do similar: File: ObservabilityNamespacePrefix + "..." .  The issue that NamespacePrefix is not "redhat-rhoam", it's "redhat-rhoam-" (defined at GetNamespacePrefix()) 
We can see also "statistics" : ) here:
[common (vm_MGDAPI-1882 *)]$ grep "File: ObservabilityNamespacePrefix +" alerts_exist.go |wc -l
34
[common (vm_MGDAPI-1882 *)]$ grep "File: NamespacePrefix +" alerts_exist.go |wc -l
37
[common (vm_MGDAPI-1882 *)]$ grep "File: NamespacePrefix + \"-" alerts_exist.go |wc -l
0
But agree, that in types.go file it looks a bit not usual. Seems to me - two ways 1) return ObservabilityNamespacePrefix back to file. 2nd option - leave it as is (...maybe move it to the end of the var list, to "separate" from namespaces ... ) 
Thank you very much for your comment!

PR #2248, opened 2021-10-29
error handling on WriteConfig()

briangallagher:
# Issue link
https://issues.redhat.com/browse/MGDAPI-2905

# What
Tidy up code with no error handling on WriteConfig() calls.

# Verification steps
Scan the code for any issues
e2e tests passing are sufficient for verification


briangallagher:
> hey, @briangallagher after a look into unit tests - they should be updated to something like [this](https://github.com/makslion/integreatly-operator/commit/ef52e8b74f2203a662db79498853ae4f9e2afdf8) since now we are failing creation of the reconciler if we can't write config. And a curiosity-based question: is there a reason for capitalized error messages (in contrast to all existing `.WriteConfig` errors)? 🤔 Apart from this, all looks grand. Once e2e tests will start to behave I don't see other reasons to hold this work

Thanks @makslion Updated now

makslion:
looks good! And unless I missed something 

/lgtm

laurafitzgerald:
Verified install
/approve

PR #2253, opened 2021-11-01
[WIP] MGDAPI-2910 - Update UIBBT alert for threescale route to allow for ro…

valerymo:
[WIP]
Update UIBBT alert for threescale route to allow for routes without "3scale."

# Issue link
https://issues.redhat.com/browse/MGDAPI-2910

# What
1) 3 scale Routes , that have labels "system-developer" or "system-provider", could have Hostname without "3scale" prefix. The route will be recognized by Label.
2) If getThreescaleRoute(ctx, client, "system-developer"...) returns nil Route - it should be handled as PhaseInProgress

# Verification steps
1) Unit test (UT) (in progress) - create UT for reconcileBlackboxTargets() method; 2 cases:
  - phase completed - when all routes called in reconcileBlackboxTargets() are available
  - phase inProgress - when route that has label "system-developer" is missing (reconcileBlackboxTargets returns nil for "system-developer" route ).

2) E2E . In Task - existing end to end tests should cover us as alerts should be firing (? discuss)

### DIFF
--- a/pkg/products/threescale/reconciler.go
+++ b/pkg/products/threescale/reconciler.go
@@ -107,6 +107,13 @@ var (
 		"zync-database",
 		"zync-que",
 	}
+	uiBlackboxLabelName            = "uibbt"
+	systemProviderServiceLabel     = "system-provider"
+	systemProviderMonitoringLabel  = "system-provider-admin"
+	systemProviderRoutePrefix      = "3scale-admin"
+	systemDeveloperServiceLabel    = "system-developer"
+	systemDeveloperMonitoringLabel = "system-developer-admin"
+	systemDeveloperRoutePrefix     = "3scale"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/products/threescale/reconciler.go
```diff
@@ -2023,13 +2021,15 @@ func (r *Reconciler) reconcileBlackboxTargets(ctx context.Context, client k8scli
 		}
 
 		// Create a blackbox target for the developer console ui
-		threescaleRoute, err := r.getThreescaleRoute(ctx, client, "system-developer", func(r routev1.Route) bool {
-			return strings.HasPrefix(r.Spec.Host, "3scale.")
-		})
+		threescaleRoute, err := r.getThreescaleRoute(ctx, client, "system-developer", nil)
```
philbrookes, line 2024: Has this been tested on a running cluster - that no UIBBT alerts are firing? I was under the impression that this would return 2 routes?

valerymo:
solution was changed, and implemented in another PR - https://github.com/integr8ly/integreatly-operator/pull/2274

PR #2208, opened 2021-10-14
MGDAPI-2892 - fix tests - suite_test.go

valerymo:
# Issue link
https://issues.redhat.com/browse/MGDAPI-2892

# What
- fixed name of the deployment of the 3scale operator for RHOAM, to be "threescale-operator-controller-manager"
- added waitForProductDeployment(...) and waitForInstallationStageCompletion (...) for Observability Operator for RHMO; added condition to Support both AMO and OO

# Verification steps
run e2e tests for RHMO (and RHMI ? )

PR #2255, opened 2021-11-02
Fix Service monitors

Boomatang:
# Issue link

https://issues.redhat.com/browse/MGDAPI-3026

# What

Add messing service monitors and set the name space prefix

# Verification steps

* install the operator 
* Run `BYPASS_STORAGE_TYPE_CHECK=true INSTALLATION_TYPE=managed-api TEST="Verify servicemonitors are cloned in monitoring namespace and rolebindings are created" make test/e2e/single`
* Expected: Test should pass


psturc:
/lgtm

verified against my cluster

btw I think rhoam-e2e job won't pass until https://github.com/integr8ly/integreatly-operator/pull/2208 is merged

PR #2258, opened 2021-11-03
test: a25: remove updating cidr range from test

psturc:
# Issue
https://issues.redhat.com/browse/MGDAPI-3011
https://issues.redhat.com/browse/MGDAPI-2309

# What
This test fails frequently in our pipelines on following error:
```
    Update Cidr Block in AWS Strategy Map errors:
    	failed to update aws strategy map: Operation cannot be fulfilled on configmaps "cloud-resources-aws-strategies": the object has been modified; please apply your changes to the latest version and try again
```

This is caused by the the unsuccessful attempt to update `cloud-resources-aws-strategies` configmap by the test. The object has changed because rhmi-operator updated the configmap first. It can be fixed by fetching the configmap content just before the update, but there's still a slight chance that rhmi-operator might still be able update it between get & update actions and cause the error again.

I rather decided to remove the flaky "verifyCidrBlockUpdate" part from the test, because it's not a scenario any customer could hit, because they are not able to modify "cloud-resources-aws-strategies" configmap.

# To verify
```
go clean -testcache && MULTIAZ=true BYPASS_STORAGE_TYPE_CHECK=true WATCH_NAMESPACE=redhat-rhoam-operator go test -v ./test/functional -timeout=80m -ginkgo.focus="A25" -ginkgo.v -ginkgo.progress -ginkgo.trace
```

### DIFF
--- a/test-cases/tests/installation/a25-verify-standalone-rhmi-vpc-exists-and-is-configured-properly.md
+++ b/test-cases/tests/installation/a25-verify-standalone-rhmi-vpc-exists-and-is-configured-properly.md
@@ -69,5 +69,3 @@ aws ec2 describe-vpcs --filter "Name=tag-value,Values=$CID"
 > - Standalone VPC route table exists, has the correct tag and has a route targeting the peering connection
 >
 > - Cluster route tables with the correct tag exist and contain a route to the peering connection and the standalone vpc
->
-> Verify no new VPC is created when the CIDR block is updated in the \_network section of the `cloud-resources-aws-strategies` config map
--- a/test/functional/aws_standalone_vpc.go
+++ b/test/functional/aws_standalone_vpc.go
@@ -6,9 +6,6 @@ import (
 	"fmt"
 	"net"
 	"strings"
-	"time"
-
-	"k8s.io/apimachinery/pkg/util/wait"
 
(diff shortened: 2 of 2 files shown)


psturc:
Hi @Boomatang I'd like to get your opinion on removing this part of the test from "A25"

Boomatang:
I go agree that check was adding flakiness to the test. It is my personal opinion that some tests are trying to test to much and do give a clear reason for failure or test components with no direct link to the others in the test case. So I am all for create clean test.

In this case should there be steps some steps added to the written test case so this may be checked manual from time to time.  Or maybe a new test case written which would only be for this, then if it is tried to be automated in the future again it would be adding back in the flakiness to this test but a new one. 

https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/installation/a25-verify-standalone-rhmi-vpc-exists-and-is-configured-properly.md

Go to get my hands on a mutli-az cluster and run the verification step now

laurafitzgerald:
this change looks fine for me. happy to have this test removed. It may be the case that it's covered in cro also. The config map in question is specific to cro
/lgtm

psturc:
I checked cloud-resource-operator repo, but couldn't find such e2e test.
I just removed the step from the manual test case.

PR #2272, opened 2021-11-10
modified test-containers-managed-api for running test suite

martinhesko:
# Issue link
[MGDAPI-2076](https://issues.redhat.com/browse/MGDAPI-2076)
[Main Gitlab MR](https://gitlab.cee.redhat.com/integreatly-qe/ci-cd/-/merge_requests/593)

# What
modified the .yaml config to be used for running the 3scale test suite as a part of our install pipeline

### DIFF
--- a/test-containers-managed-api.yaml
+++ b/test-containers-managed-api.yaml
@@ -1,9 +1,10 @@
 ---
 tests:
   - name: 3scale-test
-    image: quay.io/rh_integration/3scale-testsuite:2.9.0
+    image: quay.io/rh_integration/3scale-testsuite
     timeout: 3600
     imagePullSecretEnvVar: "RH_INTEGRATION"
+    argument: "smoke"
     envVars:
       - name: NAMESPACE
-        value: redhat-rhmi-3scale
+        value: redhat-rhoam-3scale


test-containers-managed-api.yaml
```diff
@@ -1,9 +1,10 @@
...
   - name: 3scale-test
-    image: quay.io/rh_integration/3scale-testsuite:2.9.0
+    image: quay.io/rh_integration/3scale-testsuite
     timeout: 3600
     imagePullSecretEnvVar: "RH_INTEGRATION"
+    argument: "smoke"
     envVars:
       - name: NAMESPACE
-        value: redhat-rhmi-3scale
+        value: redhat-rhoam-3scale
```
trepel, line 10: @martinhesko @psturc is there a way to make this parameterized? I am asking because on EDGE and for Dev Sandbox the different namespace prefix is used. There is a low chance that we will need to run this against EDGE or Dev Sandbox but still.

psturc: Parametrizing the config is probably not a bad idea. I think it's possible that might want to use different configs in a future (for instance to run a different test suite than just "smoke", as mentioned in the related MR)

What do you think about moving this yaml config to the pipeline library, like we have it in case of [functional tests](https://gitlab.cee.redhat.com/integreatly-qe/ci-cd/-/blob/86700845c18c1967d276a7c06430063a1c6fa65d/vars/runFunctionalTest.groovy#L22-55)?
That way we could parametrize the config way easier. Plus we get rid of maintaining a code in another repo (this one).

trepel: @martinhesko @psturc It sounds like a great idea to me. Assuming this file is not used for anything else so it is safe to remove from this repo I would vote for using the same approach as for functional tests - as you suggested.

trepel: You are correct, we will want to run more tests in the future, not just smoke tests.

martinhesko: Alright, will look into creating the config dynamically a similar way we did in the functional tests. For now I'm going to add extra parameters to Jenkins for the 3scale namespace and arguments to be passed, leaving defaults as they are now.

After I'll get that to work this is no longer needed and we can close it without merging

trepel: I think it makes sense to remove the file if it is no longer used?

PR #2279, opened 2021-11-15
Mgdapi 3087 - User with no identities - Edge case

Boomatang:
# Issue link

https://issues.redhat.com/browse/MGDAPI-3087

# What

During the upgrade from 1.12 to 1.13 there was two unexpected scenario found which the cluster users could be in.
1. There may be more identity CR's than there are user CR's
2. There may be user CR's with no existing identity CR

The second scenario will cause the operator to go into a loop it can not recover from and this stops any upgrade from completing. 

This PR adds a unit test to ensure changes that affect the first scenario will be caught in the future. Changes in the code base was needed to fix the second scenario. This is also now covered with a unit test.  The number api calls have being reduced to help proformance in dev sandbox

# Verification steps

* Install rhoam 1.14 using this index `quay.io/jfitzpat/managed-api-service-index:v1.14.0`
* Wait for install to complete
* Set up the testing IDP
* Log into 3scale with one of the testing users. (wait for account to be created)
* Remove the identity for that that user (`oc get identities` returns the list of current identities) 
* A error message like the following should be seen in the rhoam operator logs. Note this error will stop the operator from upgrading. The rhoam CR will also change its status to `authentication`
```
2021-11-15T11:58:40.503Z	DEBUG	controller-runtime.manager.events	Warning	{"object": {"kind":"RHMI","namespace":"redhat-rhoam-operator","name":"rhoam","uid":"bc065302-9d86-4d27-a423-f9189d89f303","apiVersion":"integreatly.org/v1alpha1","resourceVersion":"153963"}, "reason": "ProcessingError", "message": "Failed to reconcile components:\nfailed to synchronize the users: could not get users in active IDPs: could not get identities for user test-user01: could not get identity testing-idp:5107cc8b-da5d-4aca-87d7-fc8336687f62 for user test-user01: identities.user.openshift.io \"testing-idp:5107cc8b-da5d-4aca-87d7-fc8336687f62\" not found"}

```
* Fix the current identity issue by logging in as the user once again.
* Upgrade the operator to use the image for PR `oc patch catalogsources rhoam-operators -n openshift-marketplace --type=json -p='[{"op": "replace", "path": "/spec/image", "value": "quay.io/jfitzpat/managed-api-service-index:1.15.0-3087"}]'`
* Wait of upgrade to complete
* Remove the identity CR for the test user once again
* Expected: Operator should print a warning message about the missing identity but the reconcile loop will continue.
```
WARN[2021-11-15T12:07:26Z] could not get identities for user test-user01  product=rhsso
```

### DIFF
--- a/pkg/products/rhsso/reconciler.go
+++ b/pkg/products/rhsso/reconciler.go
@@ -386,7 +386,7 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
 	}
 
 	// Sync keycloak with openshift users
-	users, err := syncronizeWithOpenshiftUsers(ctx, keycloakUsers, serverClient, r.Config.GetNamespace(), r.Installation)
+	users, err := syncronizeWithOpenshiftUsers(ctx, keycloakUsers, serverClient, r.Config.GetNamespace(), r.Installation, r.Log)
 	if err != nil {
 		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to synchronize the users: %w", err)
 	}
@@ -486,11 +486,11 @@ func getUserDiff(keycloakUsers []keycloak.KeycloakAPIUser, openshiftUsers []user
 	return added, deleted
 }
 
-func syncronizeWithOpenshiftUsers(ctx context.Context, keycloakUsers []keycloak.KeycloakAPIUser, serverClient k8sclient.Client, ns string, installation *integreatlyv1alpha1.RHMI) ([]keycloak.KeycloakAPIUser, error) {
+func syncronizeWithOpenshiftUsers(ctx context.Context, keycloakUsers []keycloak.KeycloakAPIUser, serverClient k8sclient.Client, ns string, installation *integreatlyv1alpha1.RHMI, logger l.Logger) ([]keycloak.KeycloakAPIUser, error) {
 	var openshiftUsers *usersv1.UserList
 	var err error
 
-	openshiftUsers, err = userHelper.GetUsersInActiveIDPs(ctx, serverClient)
+	openshiftUsers, err = userHelper.GetUsersInActiveIDPs(ctx, serverClient, logger)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not get users in active IDPs")
 	}
--- a/pkg/resources/user/userHelper.go
+++ b/pkg/resources/user/userHelper.go
@@ -125,12 +125,18 @@ func GetUsersInActiveIDPs(ctx context.Context, serverClient k8sclient.Client, lo
 		}
 	}
 
+	clusterIdentities := &usersv1.IdentityList{}
+	err = serverClient.List(ctx, clusterIdentities)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not list cluster identities")
+	}
+
 	activeUsers := &usersv1.UserList{}
 
 	//go over each user
 	for _, user := range openshiftUsers.Items {
 		// get  their identities - can be multiple?
-		identities, err := GetIdentities(ctx, serverClient, user)
+		identities, err := GetIdentities(user, clusterIdentities)
 		if err != nil {
 			// it is possible for user CRs to exist without any identity CR's
 			logger.Warning(fmt.Sprintf("could not get identities for user %v", user.Name))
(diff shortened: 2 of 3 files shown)


pkg/products/rhsso/reconciler.go
```diff
@@ -486,11 +486,11 @@ func getUserDiff(keycloakUsers []keycloak.KeycloakAPIUser, openshiftUsers []user
 	return added, deleted
 }
 
-func syncronizeWithOpenshiftUsers(ctx context.Context, keycloakUsers []keycloak.KeycloakAPIUser, serverClient k8sclient.Client, ns string, installation *integreatlyv1alpha1.RHMI) ([]keycloak.KeycloakAPIUser, error) {
+func syncronizeWithOpenshiftUsers(ctx context.Context, keycloakUsers []keycloak.KeycloakAPIUser, serverClient k8sclient.Client, ns string, installation *integreatlyv1alpha1.RHMI, logger l.Logger) ([]keycloak.KeycloakAPIUser, error) {
 	var openshiftUsers *usersv1.UserList
```
MStokluska, line 490: WDYT if we clean this up a bit? It looks like it's the remains from not fully reverted multitenant work (same for line 491)

pkg/resources/user/userHelper.go
```diff
@@ -125,14 +125,23 @@ func GetUsersInActiveIDPs(ctx context.Context, serverClient k8sclient.Client) (*
...
 	for _, user := range openshiftUsers.Items {
 		// get  their identities - can be multiple?
-		identities, err := GetIdentities(ctx, serverClient, user)
+		identities, err := GetIdentities(user, clusterIdentities)
 		if err != nil {
-			return nil, errors.Wrapf(err, "could not get identities for user %v", user.Name)
+			// it is possible for user CRs to exist without any identity CR's
+			logger.Warning(fmt.Sprintf("could not get identities for user %v", user.Name))
+			activeUsers.Items = append(activeUsers.Items, user)
+			continue
```
MStokluska, line 144: Maybe I am missing something, but essentially, is there any difference now between the `openshiftUsers` and the `activeUsers` ?

pkg/products/rhsso/reconciler.go
Boomatang: I am not sure what you are asking.

pkg/resources/user/userHelper.go
Boomatang: The `openshiftUsers` are the list of user that have accounts in sso and are cluster users. While the  `activeUsers` have accounts in sso, are cluster users and have an active IDP. They are a bit more refined.

pkg/products/rhsso/reconciler.go
MStokluska: There is no point in declaring openshiftUsers and error as vars since the value will be assigned in line https://github.com/integr8ly/integreatly-operator/pull/2279/files#diff-ea0e91bd9f2b961ab416fa3821815794ee1f0a283090fff9c72af19d81a42b72R493 and nowhere else. 
I would remove code from 490 and 491

pkg/resources/user/userHelper.go
MStokluska: After the refactor yes, initially, we were returning all users, regardless of whether their identity is found or not. But it looks like the logic got updated. Thanks

briangallagher:
@philbrookes After discussion with @MStokluska we defined the following scenarios:

3. Differs from what was discussed previously. We're suggesting to assume the identity is not valid, rather than assuming it is valid. Reason being, it is better to prevent access to a valid user than the give access to an invalid user. 

wdyt? 

Scenarios:
1. User has no Identity ID on user CR => not an active user.
2. User has identity ID on user CR, Identity CR exist and are part of an active IDP => Is an active user.
3. User has identity ID on user CR, Identity CR do not exist => Not an active user. We should assume the identity CR is associated with a non active IDP
4. User has identity ID on user CR, Identity CR exist and are NOT part of an active IDP => Is not an active user.

philbrookes:
Hey Brian, 

Sorry for the slow reply. This agrees with my initial response to this scenario. What convinced me to change my mind was that apparently valid users can somehow have their identity deleted - apparently reported to @Boomatang from CSSRE resulting in valid users being deleted.

I do see the point of preferring to disallow valid users over allowing invalid users though, and am happy to agree with your proposal.

Boomatang:
@briangallagher 

I am going to have to disagree with your point 3 "User has identity ID on user CR, Identity CR do not exist => Not an active user. We should assume the identity CR is associated with a non active IDP". There is no why to tell if they are or are not an active user. If they are an active user as soon as they log into the cluster the identity is recreated and back to its normal state. 

Treating the user as inactive will remove their account from 3scale. If the user was given elevated permissions these will also be removed as when the user logs back into 3scale a new account is created for the user in 3scale. 

I would agree with marking the cluster user as inactive in point 3,  if we were able to mark the 3scale user as inactive and not remove their account for 90 days or something like.  

As the PR #2287 has being merged I am happy to close this PR. But I do want to highlight how we are creating a possible bad user experience

PR #2281, opened 2021-11-15
Add exception for CPaaS subscription name

briangallagher:
# Issue link
[JIRA](https://issues.redhat.com/browse/MGDAPI-3089)

# What
Validating the RHOAM install failed because the subscription name generated by CPaaS is random. 

# Verification steps
Unit tests sufficient.
Will be validated through CPaaS

### DIFF
--- a/pkg/addon/operator_run.go
+++ b/pkg/addon/operator_run.go
@@ -3,10 +3,10 @@ package addon
 import (
 	"context"
 	"fmt"
-
 	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/apis/v1alpha1"
 	k8sresources "github.com/integr8ly/integreatly-operator/pkg/resources/k8s"
 	operatorsv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	"github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/addon/operator_run.go
```diff
@@ -51,6 +51,16 @@ func InferOperatorRunType(ctx context.Context, client k8sclient.Client, installa
...
 		return runTypesBySubscription[subscription.Name], nil
+	} else {
+		// If subscription is nil, it could be because the CPaaS build has put a non deterministic name on it.
+		// This code may be temporary if CPaaS changes to deterministic Subscription names
+		exists, err := RhoamCPaaSSubscriptionExists(ctx, client, installation.Namespace)
+		if err != nil {
+			return "", err
+		}
+		if exists {
+			return AddonRunType, nil
```
psturc, line 62: Is it addon in case of install type that CVP uses? I thought CVP uses OLM

briangallagher: @sergioifg94 because this is CVP and not added through ManagedTenants maybe this should be OLM?

sergioifg94 (commented):
I'm not sure if these changes would fix the issue. I think that if that code is added to the [`GetSubscription`](https://github.com/integr8ly/integreatly-operator/blob/master/pkg/addon/operator_run.go#L122) function (instead of the `InferOperatorRunType` one) it would work though. I'd probably add it after the for loop has ended, if it still hasn't found the subscription using the known names then default to the logic you implemented

pkg/addon/operator_run.go
briangallagher: @psturc We won't need to set the type with the new approach. Can you have another quick review of this?

pkg/addon/operator_run.go
```diff
@@ -65,6 +65,37 @@ func InferOperatorRunType(ctx context.Context, client k8sclient.Client, installa
...
+			logrus.Infof("Found subscription %s", sub.Name)
+			labels := sub.GetLabels()
+			// Looking for specific label that should be on the CPaaS generated subscription
+			_, ok := labels["operators.coreos.com/managed-api-service.redhat-rhoam-operator"]
+			if ok {
+				logrus.Info("Found Rhoam operator subscription")
+				return &sub, nil
+			}
+		}
+	}
```
sergioifg94, line 93:
```suggestion

	for _, sub := range subs.Items {
		logrus.Infof("Found subscription %s", sub.Name)
		labels := sub.GetLabels()
		// Looking for specific label that should be on the CPaaS generated subscription
		_, ok := labels["operators.coreos.com/managed-api-service.redhat-rhoam-operator"]
		if ok {
			logrus.Info("Found Rhoam operator subscription")
			return &sub, nil
		}
	}

```

sergioifg94 (approved):
Looks good, just that tiny suggestion but not going to block it because of it
/lgtm
/approve

PR #2283, opened 2021-11-16
removed test-containers-managed-api because it is no longer needed

martinhesko:
in [this PR](https://github.com/integr8ly/integreatly-operator/pull/2272) we agreed that this file will be generated dynamically and outside of our use it's no longer needed. This is the PR to remove that file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,13 +270,6 @@ test/products:
 	mkdir -p $(TEST_RESULTS_DIR)
 	delorean pipeline product-tests --test-config ./test-containers.yaml --output $(TEST_RESULTS_DIR) --namespace test-products
 
-.PHONY: test/rhoam/products
-test/rhoam/products: export WATCH_NAMESPACE := $(NAMESPACE)
-test/rhoam/products:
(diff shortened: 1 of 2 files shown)


psturc:
I'm not sure if this change is required since our pipelines are no longer dependant on the content of integreatly-operator regarding running product tests.

Otherwise I think it would make sense to also include the change for deletion of [the make target](https://github.com/integr8ly/integreatly-operator/blob/7533de7ddccfb08fa5388b1fc47a817820f31da9/Makefile#L273-L277)

martinhesko:
It's definitely not required but as Tomáš said in the previous PR: if it's not going to be used anymore anywhere and we have a different workflow for it now I think it makes sense to remove it.

psturc:
ah, ok, I forgot about that. that's fine then

KevFan:
@martinhesko Would you still like this PR to be merged ? If so, could you add the JIRA ticket number to the commits :pray:

KevFan:
Bump ^^

martinhesko:
@KevFan fixed and ready to merge :+1:

PR #2287, opened 2021-11-22
ensure missing identities do not block the install

briangallagher:
# Issue link
[MGDAPI_3087](https://issues.redhat.com/browse/MGDAPI-3087)

# Verification steps
Install RHOAM based on this branch
Install the testing IDP with 10 users
Login to OpenShift as testuser01, testuser02, testuser03
Verify a User CR and Identity CR exist for each
For test_user01, delete the identity id on the user 
For test_user02, delete the identity cr 
Leave test_user03 as is
Ensure only test_user03 gets created as a Keycloak User CR in rhsso
After confirming delete the identity for test_user03
Verify that the Keycloak User CR also gets deleted

### DIFF
--- a/pkg/products/rhsso/reconciler.go
+++ b/pkg/products/rhsso/reconciler.go
@@ -386,7 +386,7 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, installation *inte
 	}
 
 	// Sync keycloak with openshift users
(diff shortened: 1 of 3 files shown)


MStokluska:
I have followed the verification steps and can verify that it works as described, code looks good as well. Good job.
/lgtm

PR #2291, opened 2021-11-24
add postgres serviceUpdates configuration

laurafitzgerald:
# Issue link
https://issues.redhat.com/browse/MGDAPI-3095

# What
Specify serviceUpdate configuration for Postgres
Related pr -> https://github.com/integr8ly/cloud-resource-operator/pull/235

# Verification steps
Install workloadwebapp in the cluster to verify downtime during the upgrade

Checkout the master branch - update products/installation.yaml cro reference to point to -> quay.io/laurafitzgerald/cro-index:v0.30.0
Run `INSTALLATION_TYPE=managed-api USE_CLUSTER_STORAGE=false make code/run`
Allow the installation to go completed

Log into the AWS account, verify that there are 3 RDS Postgres instances running Engine Version 10.15 (under Configuration Tab) and no Pending Maintenance is present (under Maintenance & Backup tab) for each one.

Complete an upgrade to this version of the code by:
checkout out this branch
Update update products/installation.yaml cro reference to point to -> quay.io/laurafitzgerald/cro-index:v0.31.0
Stop the operator and re-run using `INSTALLATION_TYPE=managed-api USE_CLUSTER_STORAGE=false make code/run`

During the upgrade Log into the AWS account and verify that
- each Postgres in the deployed is moved to EngineVersion 10.16
- each Postgres has applied any pending Service Updates present.
You should see each Postgres going into Upgrading and Maintenence Status

Verify that the config map `redhat-rhoam-installation-config` entry `cloud-resources` has a new entry for `STRATEGIES_CONFIG_MAP_NAME` with value `"cloud-resources-aws-strategies"`
Verify downtime by viewing the workload web app dashboard.

Verify that in the config map `"cloud-resources-aws-strategies` that there is now a serviceUpdate entry for Postgres. Before the upgrade there was no entry present.

PR #2329, opened 2021-12-08
MGDAPI-3192

laurafitzgerald:
fix to add additional reconciles on start up of the operator, allows for additional time on the upgrade
should be removed at a later date

# Issue link
https://issues.redhat.com/browse/MGDAPI-3192

# What
Add a hack to give cloud resources reconciler 5 reconciles before it reports that it's completed.

# Verification steps
Using CCS Cluster

For full upgrade verification which is recommended follow steps in -> #2291

### DIFF
--- a/pkg/config/cloudresources.go
+++ b/pkg/config/cloudresources.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"fmt"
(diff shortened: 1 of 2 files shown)


MStokluska:
/lgtm 
/approve
I've looked at the code and it looks ok. Verified with Laura on a cluster with OLM install and it works as expected. 
We seem to be giving enough time for sso to go in progress with 10 loops.

PR #2334, opened 2021-12-09
template for including CRO v0.32.1

makslion:
# Description
We still need to include new CRO in `go.mod` once the pipeline will; do the release. Everything elve is in place


MStokluska:
Verified on cluster in olm upgrade scenario.
Good job. 
/lgtm
/approve

PR #2347, opened 2021-12-15
scale down fuse deployments

briangallagher:
# Issue link
Scale down fuse pods and invert alerts such that they will fire if fuse pods are running. 

# Verification steps
Install 2.9.3
Verify Fuse pods running 
Install from this branch, verify that fuse pods are scaled down and no alerts are firing.

Stop the RHMI operator. 
Scale up the Fuse pods
Verify after a period alerts are firing

Start the operator again
Verify pods scale down to zero again
Verify no alerts firing


MStokluska:
I verified the functionality with Brian and it works as expected. All fuse dcs and fuse operator deployment are scaled-down, alerts triggered when any of the pods were brought back up and silenced as soon as the RHMI operator scaled them down again. 
Good job! 
There's still e2e and unit tests to be updated as of writing this.
Still verifying the code.

pkg/products/fuse/reconciler.go
```diff
@@ -237,6 +211,58 @@ func (r *Reconciler) createResource(ctx context.Context, resourceName string, se
...
+				Namespace: r.Config.GetNamespace(),
+			},
+		}
+		_, err := controllerutil.CreateOrUpdate(ctx, serverClient, dc, func() error {
+			dc.Spec.Replicas = replicas
+			return nil
+		})
+		if (k8serr.IsNotFound(err)) {
+			r.log.Infof("deployment not found", l.Fields{"deployment":dc.Name})
+		} else if (err != nil) {
```
MStokluska, line 260: Hey, do we not want to set Fuse in progress in case of failure in updating DCs, Deployment or removing alerts? I guess because the alerts have bsen updated to fire on presence of the pod we can assume all went well if alerts don't fire, this is more of a question rather than anything else.

PR #2363, opened 2021-12-21
bump CRO and add alert

briangallagher:
# Issue link
JIRA: https://issues.redhat.com/browse/MGDAPI-3234

# Verification steps
upgrade from 2.9.3 and verify CRO version is 0.33 and new alerts exists


pkg/products/cloudresources/reconciler.go
```diff
@@ -204,7 +204,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
-	phase, err = r.newAlertsReconciler(r.log, r.installation.Spec.Type).ReconcileAlerts(ctx, client)
+	alertsReconciler, err := r.newAlertsReconciler(ctx, client, r.log, r.installation.Spec.Type, r.installation.Namespace)
+	if err != nil {
+		events.HandleError(r.recorder, installation, phase, "Failed to get new alerts reconciler", err)
+		r.log.Error("Error getting cloud resources alerts reconciler", err)
+		return phase, err
```
MStokluska, line 211: hey, I'm a bit confused here. We are returning phase, err here but the output of newAlertsReconciler is the alertReconciler. Should the returning phase in this block be phaseFailed rather that phase returned from previous func

MStokluska:
The e2e passed before pushing final switch of phase inProgress > phaseFailed
Verified functionality on Brians cluster
/lgtm

PR #2367, opened 2021-12-22
Fix115basebranch

briangallagher:
# Issue link
Bring rhoam-release-v1.15 branch inline with rhoam-v1.15.0 tag (which is what was actually released)

# Verification steps
Compare commits to tag and ensure e2e tests pass


MStokluska:
/lgtm
The difference between 1.15 tag and the 1.15 release, all commits are present

PR #2382, opened 2021-12-24
MGDAPI-3237 don't send emails on info level alerts

makslion:
# Issue link
[JIRA](https://issues.redhat.com/browse/MGDAPI-3237)

# What
Added a `blackhole` receiver to dump alerts with `info` severity 

# Verification steps
1. Run the operator from this branch 
1.1 You might update [these](https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/threescale/prometheusRules.go#L351-L352) lines to decrease the "alert" threshold and timeout to speed up verification:
```
Expr:   intstr.FromString(fmt.Sprintf("sum by(container, pod) (container_memory_usage_bytes{container!='', container!='system-provider', namespace='%[1]v'}) / sum by(container, pod) (kube_pod_container_resource_limits{namespace='%[1]v',resource='memory'}) * 100 > 0", namespace)),
For:    "1m",
``` 
2. Open alertmanager (Networking -> Routes -> `alertmanager` in `redhat-rhoam-observability` namespace) and check the `Status` tab - you should see config with a new route:
```
- receiver: blackhole
    match:
      severity: info
```
and en empty receiver: 
```
- name: blackhole
```
3. Open prometheus route (Networking -> Routes -> `prometheus` in `redhat-rhoam-observability` namespace) and ensure that  `ThreeScaleContainerHighMemory` alert is firing 
4. Check logs of the `alertmanager` pod - you should not see anything there but the deadmanssnitch (given you have only `ThreeScaleContainerHighMemory` and `DeadMansSwitch` firing) that means that alertmanager correctly processed config and does not make attempts to send out notifications on info level alerts

### DIFF
--- a/templates/monitoring/alertmanager/alertmanager-application-monitoring.yaml
+++ b/templates/monitoring/alertmanager/alertmanager-application-monitoring.yaml
@@ -18,6 +18,9 @@ route:
         severity: critical
       receiver: critical
     - match:
+        alertname: ThreeScaleContainerHighMemory
+      receiver: blackhole
+    - match:
         alertname: DeadMansSwitch
       repeat_interval: 5m
       receiver: deadmansswitch
@@ -29,7 +32,7 @@ route:
       receiver: BUandCustomer
     - match:
         alertname: RHOAMApiUsageLevel3ThresholdExceeded
-      receiver: SRECustomerBU
+      receiver: BUandCustomer
     - match_re:
         alertname: RHOAMApiUsageSoftLimitReachedTier[0-9]+
       receiver: BU
@@ -38,8 +41,9 @@ route:
       receiver: BUandCustomer
     - match:
         alertname: RHOAMApiUsageRejectedRequestsMismatch
-      receiver: SRECustomerBU
+      receiver: BUandCustomer
 receivers:
+  - name: blackhole
   - name: default
     email_configs:
       - send_resolved: true


laurafitzgerald:
Hi @makslion I've worked through the verification steps and everything matches what's described.

I just wanted to note that I see that there are some other alerts which are marked severity `info` and wondered if this change would impact that. The alerts are three RHOAMApiUsageLevelXThresholdExceeded alerts and `RHOAMApiUsageRejectedRequestsMismatch`

The first three are referenced in the AlertManager config with the following

```
 - receiver: BUandCustomer
    match:
      alertname: RHOAMApiUsageLevel1ThresholdExceeded
    continue: false
  - receiver: BUandCustomer
    match:
      alertname: RHOAMApiUsageLevel2ThresholdExceeded
    continue: false
  - receiver: SRECustomerBU
    match:
      alertname: RHOAMApiUsageLevel3ThresholdExceeded
    continue: false
```

I think it's possible that this change would result in BU and Customer no longer receiving those alerts?  I believe this could be tested by following one of our manual tests which tests the funtionality of these alerts. 

Also when checking the AlertManger logs I see the following log repeated.
`level=warn ts=2022-01-06T11:08:35.823Z caller=notify.go:723 component=dispatcher receiver=deadmansswitch integration=webhook[0] msg="Notify attempt failed, will retry later" attempts=1 err="Post \"https://dms.example.com\": dial tcp: lookup dms.example.com on 172.30.0.10:53: no such host"
`
"This one I thiknk is unrelated but probably we can create a JIRA about it.

makslion:
Changed the work done according to the conversation with CS-SRE:
- Sending the `ThreeScaleContainerHighMemory` to the blackhole
- Removing SRE from an email list for `RHOAMApiUsageLevel3ThresholdExceeded` and `RHOAMApiUsageRejectedRequestsMismatch` as they are an info level alerts and require no action from the SRE side

laurafitzgerald:
Hi @makslion After installation of this branch I can see
```
- receiver: blackhole
    match:
      alertname: ThreeScaleContainerHighMemory
    continue: false
``` which I believe is expected.
As we've verified blackhole receiver works as expected, I think this is sufficient for verification
Other configuration changes are also present. 👍 
/approve
/lgtm

PR #2394, opened 2022-01-06
copy catalog source rhoam operator to 3scale to allow 3scale subscrip…

briangallagher:
# Issue link
[MGDAPI-3280](https://issues.redhat.com/browse/MGDAPI-3280)

# What
CPaaS specific changes. 
With CPaaS, 3Scale uses the same index image (Catalog Source) as the rhoam operator. The CS is created in the rhoam operator ns. This prevents the 3Scale Subscription from being reconciled. 
The change copies the CS to the 3Scale operator ns to allow the Subscription be reconciled. 

# Verification steps
- Scan code base for an obvious issues
- e2e tests passing

### DIFF
--- a/pkg/resources/marketplace/implicit_catalog_source_reconciler.go
+++ b/pkg/resources/marketplace/implicit_catalog_source_reconciler.go
@@ -2,11 +2,14 @@ package marketplace
 
 import (
 	"context"
-	"errors"
+	"github.com/pkg/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"strings"
 
 	"github.com/integr8ly/integreatly-operator/pkg/addon"
 	k8sresources "github.com/integr8ly/integreatly-operator/pkg/resources/k8s"
-	"github.com/integr8ly/integreatly-operator/pkg/resources/logger"
+	l "github.com/integr8ly/integreatly-operator/pkg/resources/logger"
 	rhmiresources "github.com/integr8ly/integreatly-operator/pkg/resources/rhmi"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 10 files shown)


pkg/resources/marketplace/implicit_catalog_source_reconciler.go
```diff
@@ -37,7 +40,7 @@ func NewImplicitCatalogSourceReconciler(log logger.Logger, client k8sclient.Clie
 // Reconcile finds the CatalogSource that provides the current installation,
 // returning an error if it fails to find it. Caches the found CatalogSource
 // to be used by r.CatalogSourceName() and r.CatalogSourceNamespace()
-func (r *ImplicitCatalogSourceReconciler) Reconcile(ctx context.Context) (reconcile.Result, error) {
+func (r *ImplicitCatalogSourceReconciler) Reconcile(ctx context.Context, subName string) (reconcile.Result, error) {
```
laurafitzgerald, line 43: @briangallagher I haven't code reviewed but with the addition of some logic to this function do you think it would be valuable to have a unit test for it now?

briangallagher: @laurafitzgerald Unit test added 👍

pkg/resources/marketplace/implicit_catalog_source_reconciler.go
```diff
@@ -48,6 +51,37 @@ func (r *ImplicitCatalogSourceReconciler) Reconcile(ctx context.Context) (reconc
...
+			ObjectMeta: v1.ObjectMeta{
+				Name:      catalogSource.Name,
+				Namespace: namespacePrefix + "3scale-operator",
+			},
+		}
+
+		controllerutil.CreateOrUpdate(ctx, r.Client, cs, func() error {
+			cs.Spec = catalogSource.DeepCopy().Spec
+			return nil
+		})
```
makslion, line 79: Hm, out of curiosity does this mean that we would have two identical CatalogSources? :thinking:  And won't the one in the `redhat-rhoam-3scale-operator` namespace be updated on each reconcile potentially rewriting any changes there might be to the copy?

briangallagher: We will have 2 but I don't think it will be overridden on each reconcile as we don't write the CS I believe. Unless you've seen otherwise @makslion? This is a temp solution until Cpaas guys implement a fix.

makslion:
Change looks good and tests are passing! :tada: 

/lgtm

PR #2395, opened 2022-01-06
MGDAPI-3099 don't delete subscription on uninstall

makslion:
# Issue link
[MGDAPI-3099](https://issues.redhat.com/browse/MGDAPI-3099)

# What
In the light of incoming changes to the reconciliation time of operators from the hive, we might bump into a problem when we delete the subscription during the uninstallation but it gets recreated immediately. To prevent this we want to delegate deletion of the subscription from the rhoam. 
This change makes sure that the subscription stays in place while allowing manual deletion in case of the dev work (in that case the subscription isn't managed by hive hence will never be deleted) by extending `make cluster/cleanup/olm`. 

# Verification steps 
(if for some weird reason you trust me or want to reuse [bundles](https://quay.io/mvavilov/rhoam-bundle:1.15.0) and [index](https://quay.io/repository/mvavilov/rhoam-index:1.15.0) I've generated you now have the links) 
1. Build an image from this PR
2. [Generate bundles ](https://github.com/RHCloudServices/integreatly-help/blob/master/guides/olm/install-upgrade-rhoam-bundle.md#1-generate-bundle) to do a fresh install and ensure they are using the freshly build image 
3. [Install](https://github.com/RHCloudServices/integreatly-help/blob/master/guides/olm/install-upgrade-rhoam-bundle.md#7-install-from-operatorhub) operator via OLM.
4. [Deploy](https://github.com/RHCloudServices/integreatly-help/blob/master/guides/olm/install-upgrade-rhoam-bundle.md#8-deploy-rhmi-cr) RHMI CR but we don't need a complete installation - just to ensure that deletion logic keeps the subscription - so you could [delete](https://github.com/makslion/integreatly-help/blob/MGDAPI-3099/guides/olm/install-upgrade-rhoam-bundle.md#uninstallation) RHMI to trigger the deletion logic. 
5. Make sure that the Subscription still in place by `oc get subscription integreatly -n redhat-rhoam-operator` once the rhoam pod reports `uninstallation complete`
6. Run `INSTALLATION_TYPE=managed-api make cluster/cleanup/olm` and watch subscription being deleted

### DIFF
--- a/pkg/addon/uninstall.go
+++ b/pkg/addon/uninstall.go
@@ -15,8 +15,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
-// UninstallOperator uninstalls the RHMI operator by deleting the subscription
-// and CSV. If the subscription is not found, it doesn't do anything, as the
+// UninstallOperator uninstalls the RHMI operator by deleting the CSV.
+// If the subscription is not found, it doesn't do anything, as the
 // operator might not be run through OLM
 func UninstallOperator(ctx context.Context, client k8sclient.Client, installation *integreatlyv1alpha1.RHMI) error {
 	// Get the operator subscription
@@ -31,13 +31,6 @@ func UninstallOperator(ctx context.Context, client k8sclient.Client, installatio
 		return nil
 	}
 
-	log.Infof("Deleting subscription", l.Fields{"name": subscription.Name})
-
-	// Declare the deleting subscription function
-	deleteSubscription := func() error {
-		return client.Delete(ctx, subscription)
-	}
-
 	// Retrieve the operator CSV
 	csv := &operatorsv1alpha1.ClusterServiceVersion{}
 	err = client.Get(ctx, k8sclient.ObjectKey{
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -406,11 +406,15 @@ cluster/cleanup/serviceaccount: kustomize
 	@-$(KUSTOMIZE) build config/rbac-$(INSTALLATION_SHORTHAND) | oc delete -f -
 
 .PHONY: cluster/cleanup/olm
-cluster/cleanup/olm: cluster/cleanup
+cluster/cleanup/olm: cluster/cleanup cluster/cleanup/subscription
```
KevFan, line 409: I wonder do we need to cleanup the subscription as the namespace is deleted in `cluster/cleanup` :thinking:

KevFan (commented):
Verified, looks good :+1: 
image

Just have one minor comment

Makefile
makslion: I've been wondering this myself and tried just running an old version of the `make cluster/cleanup/olm` subscription still was there :thinking: It sounds strange. Maybe I wasn't patient enough to wait for the subscription to be deleted. I will double-check this and will be back with more that my vague memories

makslion: Yep, the subscription is gone when the namespace is deleted. Removed changes to the makefile. Thanks

PR #2407, opened 2022-01-16
MGDAPI-3202 - RHOAM Multitenant on demand (crd based solution)

valerymo:
# Issue link
https://issues.redhat.com/browse/MGDAPI-3202
Make RHOAM Multitenant on demand

# What
- We want to make rhoam multitenant tenants created based on the presence of new CR of type RhoamTenant, 
where CR name to be same as username. Based on this user will be annotated as "tenant"
- RHMI controller will watch for change in user (no changes in rhmi controller)
- 3scale reconciler will get tenant users by annotation 
( userHelper.GetMultiTenantUsers() method changed to get tenant users by annotation instead of timestamp)

# Verification steps
1. Prepare Sandbox Rhoam cluster
1.1. Check that RhoamTenant CRD created after `make cluster/prepare/local`
1.2. Create IDP, and have an any user for testing (like customer-admin01, or test, any ...). Check that user have no "tenant" annotation before test
1.3. Create Instance of RhoamTenant crd. CR should have same names as user, like "customer-admin01".
1.4 Check that your test user annotated, as `Annotations:	tenant=yes`
1.5.Create one more CR, with name as non existing user. Save CR. Check CR status. You will see something like following: Last Error: error get user with name as current CR, users.user.openshift.io "customer1" not found

2. Prepare NON Sandbox Rhoam cluster
and repeat same steps. Nothing should happen, users will not be annotated, even if CR exists


controllers/tenant/rhoamtenant_controller.go
```diff
@@ -0,0 +1,96 @@
...
+			for _, user := range usersList.Items {
+				if user.Name == cr.Spec.Username {
+					_, err = controllerutil.CreateOrUpdate(context.TODO(), r.Client, &user, func() error {
+						if user.Annotations == nil {
+							user.Annotations = map[string]string{}
+						}
+						user.Annotations["tenant"] = "yes"
+						return nil
+					})
+					return nil
```
briangallagher, line 89: We need to check the error !=nil from CreateOrUpdate(). Return this error and add to CR status error field.

controllers/tenant/rhoamtenant_controller.go
```diff
@@ -0,0 +1,96 @@
...
+	Scheme *runtime.Scheme
+	mgr    manager.Manager
+	log    l.Logger
+}
+
+func (r *TenantReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
+	_ = context.Background()
+	err := r.addAnnotationToUser(request.Name)
+	if err != nil {
+		r.log.Error("error in addAnnotationToUser: %s ", err)
```
briangallagher, line 55: maybe add an error field to the status and add an error string to it. 
See rhmi_controller for example on how to update the status on a CR: `r.Status().Update(context.TODO(), installation)`

controllers/tenant/rhoamtenant_controller.go
```diff
@@ -0,0 +1,96 @@
...
+type TenantReconciler struct {
+	k8sclient.Client
+	Scheme *runtime.Scheme
+	mgr    manager.Manager
+	log    l.Logger
+}
+
+func (r *TenantReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
+	_ = context.Background()
+	err := r.addAnnotationToUser(request.Name)
```
briangallagher, line 53: Can we use request.Name and request.Namespace? Then in addAnnotationToUser we could get the CR, rather than listing all?

controllers/tenant/rhoamtenant_controller.go
```diff
@@ -0,0 +1,96 @@
...
+func (r *TenantReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.RhoamTenant{}).
+		Watches(&source.Kind{Type: &v1alpha1.RhoamTenant{}}, &handler.EnqueueRequestForObject{}).
+		Complete(r)
+}
+
+func (r *TenantReconciler) addAnnotationToUser(crName string) error {
+	crList := &v1alpha1.RhoamTenantList{}
+	err := r.Client.List(context.TODO(), crList)
```
briangallagher, line 69: See comment above, can we do a get instead of list here?

controllers/tenant/rhoamtenant_controller.go
```diff
@@ -0,0 +1,96 @@
...
+						}
+						user.Annotations["tenant"] = "yes"
+						return nil
+					})
+					return nil
+				}
+			}
+		}
+	}
+	log.Warning("User defined in CR was not found. Please check username in rhoamtenant CR: " + crName)
```
briangallagher, line 94: Maybe return this as an error and add it to the CR status field

controllers/rhmi/rhmi_controller.go
```diff
@@ -1464,3 +1466,16 @@ func getInstallation() (*rhmiv1alpha1.RHMI, error) {
 		},
 	}, nil
 }
+
+func userTenantAnnotationPredicate(installation *rhmiv1alpha1.RHMI) predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(m metav1.Object, _ runtime.Object) bool {
+		//if !rhmiv1alpha1.IsRHOAMMultitenant(rhmiv1alpha1.InstallationType(installation.Spec.Type)) {
```
briangallagher, line 1472: can we remove comments?

controllers/rhmi/rhmi_controller.go
```diff
@@ -1229,6 +1230,7 @@ func (r *RHMIReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	controller, err := ctrl.NewControllerManagedBy(mgr).
 		For(&rhmiv1alpha1.RHMI{}).
 		Watches(&source.Kind{Type: &usersv1.User{}}, enqueueAllInstallations).
+		Watches(&source.Kind{Type: &usersv1.User{}}, enqueueAllInstallations, builder.WithPredicates(userTenantAnnotationPredicate(installation))).
```
briangallagher, line 1233: Do we need this if we are already watching all Users from line: 1232?

config/crd-sandbox/bases/integreatly.org_rhoamtenants.yaml
```diff
@@ -0,0 +1,55 @@
...
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: RhoamTenantSpec defines the desired state of RhoamTenant
+            properties:
+              username:
```
briangallagher, line 39: Hey Valery, does it make sense that we could remove the username field from the spec and just use the name of the CR? So, users would create a RHOAMTenant CR with a name equal to their user name?

apis/v1alpha1/rhoamtenant_types.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const ()
+
+// RhoamTenantSpec defines the desired state of RhoamTenant
+type RhoamTenantSpec struct {
+	Username string `json:"username,omitempty"`
```
briangallagher, line 27: See comment below on just using the CR name

controllers/tenant/rhoamtenant_controller.go
valerymo: Done

controllers/tenant/rhoamtenant_controller.go
valerymo: Done

controllers/tenant/rhoamtenant_controller.go
valerymo: Updated. Not using listing now, but Get. 
Not using Namespace, as CR is cluster scope

controllers/tenant/rhoamtenant_controller.go
valerymo: Done

controllers/tenant/rhoamtenant_controller.go
valerymo: Done

controllers/rhmi/rhmi_controller.go
valerymo: Done. Also this method removed, not needed already , as this additional Watches was removed, as suggested in next comment

controllers/rhmi/rhmi_controller.go
valerymo: Done. Checked, this watch is not needed.

config/crd-sandbox/bases/integreatly.org_rhoamtenants.yaml
valerymo: Done. Username removed, and done as suggested; working fine. Thank you Brian.

apis/v1alpha1/rhoamtenant_types.go
valerymo: Done

PR #2412, opened 2022-01-19
MGDAPI-3084 ensure binaryData from CM in openshif-marketplace treated well

makslion:
# Issue link
[MGDAPI-3084](https://issues.redhat.com/browse/MGDAPI-3084)

# What
On the 4.9+ cluster, CSV stored in config map under the `openshift-marketplace` namespace resides within `binaryData` field instead of `data`. With the old code it meant that we could find CSV and was defaulting to the default (is service affecting). With this change, we would be able to fetch binary data to make the logic about auto-approval of non-service affecting upgrades work. 

# Verification steps
To verify this you would need to simulate an upgrade. I will provide you with indexes for installations. They all are using the same RHOAM image that was built from this PR with addition of the `managed-api-service` subscription [here](https://github.com/integr8ly/integreatly-operator/blob/master/controllers/subscription/subscription_controller.go#L41-L56) 
Indexes to be used: 
* 1.16.0 - fresh install `quay.io/mvavilov/rhoam-index:1.16.0`
* 1.17.0 - replaces 1.16.0 and **is not** service-affecting (will be approved automatically) `quay.io/mvavilov/rhoam-index:1.17.0`
* 1.18.0 - replaces 1.17.0 and **is** service-affecting (will not be approved automatically)  `quay.io/mvavilov/rhoam-index:1.18.0`

1. Obtain a 4.9+ cluster
2. Install RHOAM 1.16.0 via OLM
3. When `status.stage` == `complete` point `CatalgueSource` to 1.17.0
4. Watch an install plan being created and you could watch the RHOAM pod logs to see the moment it gets to the subscription reconciles and approves the new install plan 
5. After an upgrade point `CatalogueSource` to 1.18.0. RHMI pod should go  through the subscription reconciler and the install plan should not be approved 

> There is also a new unit test to cover added logic but if prow passes it all grand

### DIFF
--- a/controllers/subscription/csvlocator/csvlocator.go
+++ b/controllers/subscription/csvlocator/csvlocator.go
@@ -1,9 +1,13 @@
 package csvlocator
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"strings"
 
 	olmv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
@@ -67,7 +71,7 @@ func (l *ConfigMapCSVLocator) GetCSV(ctx context.Context, client k8sclient.Clien
 
 		// Get the ConfigMap
 		csvConfigMap := &corev1.ConfigMap{}
-		if err := client.Get(ctx, k8sclient.ObjectKey{
+		if err = client.Get(ctx, k8sclient.ObjectKey{
 			Name:      ref.Name,
 			Namespace: ref.Namespace,
 		}, csvConfigMap); err != nil {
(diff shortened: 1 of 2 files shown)


controllers/subscription/csvlocator/csvlocator.go
```diff
@@ -67,44 +71,87 @@ func (l *ConfigMapCSVLocator) GetCSV(ctx context.Context, client k8sclient.Clien
...
 			Name:      ref.Name,
 			Namespace: ref.Namespace,
 		}, csvConfigMap); err != nil {
 			return csv, fmt.Errorf("error retrieving ConfigMap %s/%s: %v", ref.Namespace, ref.Name, err)
 		}
 
 		// The ConfigMap may contain other manifests other than the CSV. Iterate
-		// through the content and skip the ones that have a kind other than
+		// through the data and skip the ones that have a kind other than
 		// ClusterSeerviceVersion
```
Boomatang, line 83: Can we add a note that this is required for openshift 4.8 and put a similar note for the 4.9 block

Boomatang (commented):
Generally the code looks ok. The update to comment strings is all I would like to see. Currently working on the verification steps.

controllers/subscription/csvlocator/csvlocator.go
makslion: done

Boomatang:
I have done the verification steps on both a 4.8 and 4.9 cluster. Service and Non Service affecting upgrades worked as expected.

/lgtm

PR #2416, opened 2022-01-20
add hive managed check

MStokluska:
# Issue link

# What

# Verification steps

### DIFF
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -89,6 +89,7 @@ const (
 	priorityClassNameEnvName         = "PRIORITY_CLASS_NAME"
 	managedServicePriorityClassName  = "rhoam-pod-priority"
 	routeRequestUrl                  = "/apis/route.openshift.io/v1"
+	hiveManagedLabel                 = "hive.openshift.io/managed"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


briangallagher:
@MStokluska I don't think we can use this anymore right: 

```
	installedViaOLM, err := addon.OperatorInstalledViaOLM(context.Background(), client, installation)
	installedViaAddon, err := addon.OperatorInstalledViaAddon(context.Background(), client, installation)
```

I think it might be enough to use: 
`isHiveManaged, err := r.isHiveLabelPresent(client)`

controllers/rhmi/rhmi_controller.go
```diff
@@ -1271,6 +1273,26 @@ func (r *RHMIReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	if err != nil {
+		return false, err
+	}
+
+	ns, err := resources.GetNS(context.TODO(), namespace, client)
+	if err != nil {
+		return false, fmt.Errorf("could not retrieve %s namespace:", err)
+	}
+
+	for label := range ns.ObjectMeta.Labels {
```
briangallagher, line 1287: Is it _ns.ObjectMeta.Labels_ or ns.Labels?
Or maybe doesn't matter?

controllers/rhmi/rhmi_controller.go
```diff
@@ -89,6 +89,7 @@ const (
 	priorityClassNameEnvName         = "PRIORITY_CLASS_NAME"
 	managedServicePriorityClassName  = "rhoam-pod-priority"
 	routeRequestUrl                  = "/apis/route.openshift.io/v1"
+	hiveManagedLabel                 = "hive.openshift.io/managed"
```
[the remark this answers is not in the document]
MStokluska: thanks Laura, good point but I've hit some dependency issues when trying to use hive package, taking the urgency of pushing this over the line I would leave it as const if that's ok.

controllers/rhmi/rhmi_controller.go
MStokluska: Thanks Brian, good point

MStokluska:
> @MStokluska I don't think we can use this anymore right:
> 
> ```
> 	installedViaOLM, err := addon.OperatorInstalledViaOLM(context.Background(), client, installation)
> 	installedViaAddon, err := addon.OperatorInstalledViaAddon(context.Background(), client, installation)
> ```
> 
> I think it might be enough to use: `isHiveManaged, err := r.isHiveLabelPresent(client)`

Hey @briangallagher we have to use installedViaOLM to avoid creation of rhmi CR in olm installs.
Since our OperatorInstalledViaOLM is checking subscription.Name and and the name of subscription for CPaaS is prefixed with addon- this function thinks that it's dealing with addon installation so we should be ok.
I removed the OperatorInstalledViaAddon check, thanks!

PR #2431, opened 2022-01-27
MGDAPI-3340 test cases for 1.16.0

psturc:
# Issue link
https://issues.redhat.com/browse/MGDAPI-3340


trepel:
Test case updates look ok, move from "Optional" to "Normal" too. I am not quite sure about that transition ID change, but I do believe you :)

/lgtm

PR #2432, opened 2022-01-27
MGDAPI-3256 update GO 1.16 -> 1.17

makslion:
# Issue link
[MGDAPI-3256](https://issues.redhat.com/browse/MGDAPI-1355)

# What
Updating go. There was a change of how GO handles modules - there is why we now would have two `require` fields on the `go.mod`. 

# Verification steps
Succeded tests here should be enough


makslion:
was it too soon to ask for a `/retest`? :thinking:

makslion:
Those should be fine after [this](https://github.com/openshift/release/pull/25824) gets through

makslion:
reopened a PR here https://github.com/integr8ly/integreatly-operator/pull/2444

PR #2444, opened 2022-02-01
MGDAPI-3256 update GO 1.16 -> 1.17

makslion:
# Issue link
[MGDAPI-3256](https://issues.redhat.com/browse/MGDAPI-3256)

# What
Updating go. There was a change of how GO handles modules - there is why we now would have two `required` fields on the `go.mod`.

# Verification steps
Succeded tests here should be enough


makslion:
The bot now should update [this](https://github.com/openshift/release/blob/master/ci-operator/config/integr8ly/integreatly-operator/integr8ly-integreatly-operator-MGDAPI-3256-working.yaml) file with the go1.17 base image :thinking:

PR #2447, opened 2022-02-01
MGDAPI-3021

Patryk-Stefanski:
# Issue link
https://issues.redhat.com/browse/MGDAPI-3021

# What
Provides dummy values for templateUtil which allows for RHOAM to be installed even when the SMTP secret is missing. Also includes unit tests for new get functions.

# Verification steps
1. Run the RHOAM operator locally  by running the following commands : 
```sh
make cluster/prepare/local
make code/run 
```
2. In a separate terminal remove the SMTP secret and check that its been removed by running the following command 
```sh
oc delete secret redhat-rhoam-smtp -n redhat-rhoam-operator
oc get secrets -n redhat-rhoam-operator
```
3. Check that the missingSmtp alert is firing in  observability -> Prometheus
4. Verify that RHOAM has been installed successfully.

### DIFF
--- a/pkg/products/monitoringcommon/alertmanagerConfig_test.go
+++ b/pkg/products/monitoringcommon/alertmanagerConfig_test.go
@@ -247,6 +247,14 @@ func TestReconciler_reconcileAlertManagerSecrets(t *testing.T) {
 		wantErr      string
 	}{
 		{
+			name: "succeeds when smtp secret cannot be found",
+			serverClient: func() k8sclient.Client {
+				return fakeclient.NewFakeClientWithScheme(basicScheme, pagerdutySecret, dmsSecret, alertmanagerRoute, clusterInfra, clusterVersion, clusterRoute)
+			},
+			wantErr: "",
+			want:    integreatlyv1alpha1.PhaseCompleted,
+		},
+		{
 			name: "fails when pager duty secret cannot be found",
 			serverClient: func() k8sclient.Client {
 				return fakeclient.NewFakeClientWithScheme(basicScheme, smtpSecret, alertmanagerRoute)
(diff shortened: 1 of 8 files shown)


pkg/products/monitoringcommon/alertmanagerConfig_test.go
```diff
@@ -611,3 +611,267 @@ func TestReconciler_getDMSSecret(t *testing.T) {
 		})
 	}
 }
+
```
laurafitzgerald, line 614: Great to see the addition of the tests here. Wdyt about also adding a test case to the `ReconcileAlertManagerSecrets` test which covers the case of no secret present. It could catch any change in future that would cause the installer to get stuck on this secret.

laurafitzgerald:
I didn't do full review so haven't assigned myself on the JIRA but just left a small suggestion inline.

pkg/products/monitoringcommon/alertmanagerConfig_test.go
Patryk-Stefanski: Good idea. New test case added.

psturc:
👀

psturc (commented):
LGTM, just one suggestion and question: should we update [`cluster/prepare` make target](https://github.com/integr8ly/integreatly-operator/blob/105ab3fdf7a381b9fd9868d8ecc034f384c20141/Makefile#L311-L312) to remove smtp secret creation, so this scenario gets tested in our pipelines by default?

psturc:
/lgtm

1. Verified the install with/without the SMTP secret being present
2. Changed smtp secret values and verified they got propagated to alert manager
3. Ran functional test suite (with smtp secret being present) and all tests passed

PR #2456, opened 2022-02-03
bump to newer version

laurafitzgerald:
# What
Bump SSO versions to latest available.

# Verification steps
I'm preparing clusters to do an install/upgrade, will make those available when they are ready
@psturc is there any additional tests that should be run here? I'll trigger e2e tests on the pr.


psturc:
@laurafitzgerald I think it'd be great to also set up SSO IDP and run [IDP and scalability tests](https://github.com/integr8ly/integreatly-operator/blob/7e9346ff955b990847a94df1f5b5c04cf19a6c24/test/common/tests.go#L85-L138).

makslion:
`failed installation of rhsso: secrets \"credential-rhsso\" not found\n"` is there a conspiracy of secrets? :thinking:

PR #2464, opened 2022-02-07
MGDAPI-1575

bryankeane0:
# Issue link
https://issues.redhat.com/browse/MGDAPI-1575

# What
Disabled user-monitoring label on namespaces so unwanted Thanos alerts no longer show on cluster UI.

# Verification steps
1. Provision a cluster **without** these changes and run RHOAM with the following commands:
```
make cluster/prepare/local
make code/run 
```
2. Verify that namespaces have the label `openshift.io/user-monitoring = true`
3. Stop the operator with `ctrl + c` and start it again with `make code/run`
4. Wait for it to do one full reconcile
5. Check that namespaces have the label `openshift.io/user-monitoring = false`


pkg/resources/reconciler_test.go
```diff
@@ -579,3 +607,20 @@ func TestReconciler_ReconcileNamespace(t *testing.T) {
 		})
 	}
 }
+
```
laurafitzgerald, line 610: We can remove this if it won't be added.

bryankeane0: Sorted below, thanks

pkg/resources/reconciler.go
```diff
@@ -322,11 +326,14 @@ func retryInstallation(ctx context.Context, client k8sclient.Client, log l.Logge
 	return integreatlyv1alpha1.PhaseAwaitingOperator, nil
 }
 
-func PrepareObject(ns metav1.Object, install *integreatlyv1alpha1.RHMI, addRHMIMonitoringLabels bool, addClusterMonitoringLabel bool) {
-	labels := ns.GetLabels()
+func PrepareObjectLabels(object metav1.Object, install *integreatlyv1alpha1.RHMI, addRHMIMonitoringLabels, addClusterMonitoringLabel, addUserMonitoringLabel bool) {
```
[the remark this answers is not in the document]
bryankeane0: I would agree with you actually. I'll make that change, thanks

den-rgb:
Hey Bryan,
I noticed the redhat-rhoam-operator doesn't contain the label like the rest of them do, is this expected?
image

bryankeane0:
@den-rgb Yeah so the alerts firing were related to Marin3r and AWS DBs so that shouldn't be a problem.

pkg/resources/reconciler_test.go
```diff
@@ -559,6 +563,30 @@ func TestReconciler_ReconcileNamespace(t *testing.T) {
...
+				Spec: integreatlyv1alpha1.RHMISpec{
+					PullSecret: integreatlyv1alpha1.PullSecretSpec{
+						Name:      "test",
+						Namespace: "test",
+					},
+				},
+			},
+			ExpectedStatus: integreatlyv1alpha1.PhaseFailed,
+			ExpectErr:      true,
+		},
```
MStokluska, line 589: hey @bryankeane0 
I am a bit confused, wonder if I'm missing something.
This unit test seems to cover a logic that's 2 years old. That's not bad but I feel like we could test here the following scenarios:
1. Is the label added to a ns that doesn't have it
2. Is the label updated if an ns has it set to true

bryankeane0:
Hey @MStokluska, So I was adding the third case to test that if during a namespace reconcile, if the given namespace doesn't already exist it should be created and tagged correctly. I ran into an issue with this as the method creates the namespace with a project request. I chatted to Laura about it and we concluded that the fake client provided for the test case doesn't have enough functionality to know to add/create the namespace. The only way we could get it to work is to pass the namespace into the client which defeats the purpose of the test.

pkg/resources/reconciler_test.go
```diff
@@ -576,6 +637,36 @@ func TestReconciler_ReconcileNamespace(t *testing.T) {
 			if tc.ExpectedStatus != phase {
 				t.Fatal("expected phase ", tc.ExpectedStatus, " but got ", phase)
 			}
+			labelExists, labelIsTrue, err := verifyNSLabelExistsAndIsTrue(tc.client)
+			if err != nil {
+				t.Fatal("error when verifying namespace label exists")
+			}
+			if !labelExists && tc.ExpectLabel {
```
MStokluska, line 644: if we have a tc.ExpectLabel bool and some of the tests cases use it as false, we should also have a condition here to check it, something like if labelExists and !tc.ExpectedLabel then fail

MStokluska: other than that, these unit tests look good to me :) :+1:

Patryk-Stefanski:
Code looks good and can verify that the `openshift.io/user-monitoring=false` label was added to a namespace (redhat-rhoam-marin3r) after the reconciliation with new changes.
/lgtm

Boomatang:
One question. Has any one check the UI to ensure these alerts are not showing up  for the user? Adding the labels should be enough but it would still be nice to confirm.

bryankeane0:
> One question. Has any one check the UI to ensure these alerts are not showing up for the user? Adding the labels should be enough but it would still be nice to confirm.

@Boomatang I did check initially once I verified it and no alerts showed. Might be good to have @Patryk-Stefanski leave his cluster running and double check

Boomatang:
Leaving lgtm on behalf of @Patryk-Stefanski 
Have done the last check and confirm the RHOAM alerts are not be collected by Thanos.
/lgtm

PR #2473, opened 2022-02-09
MGDAPI-3457 - uninstall stages performed in order

adam-cattermole:
# Issue link
Jira: [MGDAPI-3457](https://issues.redhat.com/browse/MGDAPI-3457)

# What
One of the nightly pipelines outlined an issue where the uninstall process could move onto later stages without removing all products from current stage. This was due to `pendingUninstalls` being overwritten on each iteration - if the last product in the products list returns true then we move onto the next stage, even if every other product is still pending uninstall.

- Added `productPending` boolean and only update `pendingUninstalls` to true if it hasn't already been set
- Whilst not directly related to this ticket, also spotted controller-runtime was imported twice so removed the extra reference
- In rhssocommon there is the possibility that the CR has been updated by delete before we set the finalizers and so refresh the realm object before update

# Verification steps

### DIFF
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -22,7 +22,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"net/http"
 	"os"
 	"reflect"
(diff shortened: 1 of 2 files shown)


makslion:
The change looks good. Not much to verify - it's a rare bug. Tests are passing and the installation is still functional - we just forcing a particular order in the uninstall 

/lgtm

PR #2479, opened 2022-02-15
MGDAPI-3475 skip only AWS tests

psturc:
# Issue link
https://issues.redhat.com/browse/MGDAPI-3475

# What
When running functional test suite, by default skip only AWS tests in case RHOAM is installed using cluster storage
(`useClusterStorage: true`)

# Verification steps

```bash
```

**On a cluster with RHOAM installed using cluster storage (`useClusterStorage: true`):**

Test should pass
```bash
make test/e2e/single TEST="A01"
```

(AWS) Test should be skipped, because we're using cluster storage
```bash
make test/e2e/single TEST="F01"
```

(AWS) Test should fail, because we're using cluster storage, but it's overridden with `BYPASS_STORAGE_TYPE_CHECK`
```bash
BYPASS_STORAGE_TYPE_CHECK=true make test/e2e/single TEST="F01"
```

**On a cluster with RHOAM installed using AWS storage (`useClusterStorage: false`):**

All commands below should pass (i.e. all tests should be run and should succeed)

```bash
make test/e2e/single TEST="A01"

make test/e2e/single TEST="F01"

BYPASS_STORAGE_TYPE_CHECK=true make test/e2e/single TEST="F01"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,8 +197,8 @@ make test/unit
 
 ### E2E tests
 
-A `BYPASS_STORAGE_TYPE_CHECK=true` flag is used to allow test to run when the operator is installed using cluster storage.
-This may cause side effects related to the cloud resources test.
+If you have RHOAM operator installed using cluster storage (`useClusterStorage: true`), all [AWS tests](https://github.com/integr8ly/integreatly-operator/blob/27c4a8c4fdf3461247fad2bb20fe958d4b709a99/test/functional/tests.go#L6-L12) are being skipped (because all AWS tests would fail).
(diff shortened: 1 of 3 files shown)


makslion:
On the cluster with `useClusterStorage: true`:
* A01 passed
* F01 skipped 
* F01 with `BYPASS_STORAGE_TYPE_CHECK` failed with an expected error 
On the cluster with `useClusterStorage: true` (to be updated):
* A01
* F01 
* F01 with `BYPASS_STORAGE_TYPE_CHECK` (there is no need to bypass it but still)

PR #2484, opened 2022-02-16
[MGDAPI-3454] B09 verify dedicated admin users are synced with User SSO

cecobask:
# Issue link
https://issues.redhat.com/browse/MGDAPI-3454

# What
https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/authorization/b09-verify-dedicated-admin-users-are-synced-with-user-sso.md

# Verification steps
```INSTALLATION_TYPE=managed-api TEST=B09 BYPASS_STORAGE_TYPE_CHECK=true make test/e2e/single```

### DIFF
--- a/test/common/user_dedicated_admin_synced_sso.go
+++ b/test/common/user_dedicated_admin_synced_sso.go
@@ -9,7 +9,6 @@ import (
 	userv1 "github.com/openshift/api/user/v1"
 	"io/ioutil"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -59,6 +58,9 @@ func TestDedicatedAdminUsersSyncedSSO(t TestingTB, tc *TestingContext) {
 
 	// Authenticate with Keycloak
 	hostKeycloakUserSSO, err := getHostKeycloak(ctx, tc.Client, RHSSOUserProductNamespace)
+	if err != nil {
+		t.Fatalf("%v", err)
+	}
 	credentialsKeycloak, err := getCredentialsRHSSOUser(ctx, tc.Client)
 	if err != nil {
 		t.Fatalf("%v", err)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: test/common/user_dedicated_admin_synced_sso.go [hunk 1]
```diff
@@ -0,0 +1,431 @@
...
+	}
+	t.Logf("added user %s to group %s", testUserName, groupNameDedicatedAdmins)
+
+	// Wait for RHOAM to reconcile the KeycloakUser CR in user-sso namespace and verify their group membership
+	if err := pollGeneratedKeycloakUserCR(ctx, tc.Client); err != nil {
+		t.Fatalf("%v", err)
+	}
+
+	// Wait for group memberships to propagate in Keycloak
+	time.Sleep(90 * time.Second)
```
makslion, line 103: I had 3/5 runs of this test failing with: 
```
 Expected user groups [dedicated-admins realm-managers rhmi-developers], got [rhmi-developers]
```
It might be because of the `pollGeneratedKeycloakUserCR` and/or this timeout :thinking:

cecobask: Yes, it's definitely a timeout issue in the highlighted snippet. I couldn't figure out a better solution than using time.Sleep().
I can increase the timer by another 30 seconds and this should be sufficient to consistently pass that test.
What's your advice on this?

makslion: Hm, I used to believe that the good timeout is twice the expected time. I'm not familiar with the Keycloak propagation time. Is there a way to wait for it with a timeout for something like 5 minutes? The first thing that comes into my mind is recording a timestamp here and then when we compare expected vs got we enter the loop that breaks either when things match or then we reach a timeout :thinking: Something like [this](https://github.com/cecobask/integreatly-operator/blob/MGDAPI-3454/test/common/quota.go#L60-L69) for example :thinking: 
Oh, and about that `poll` thing - If I'm not wrong it isn't working at all. Here is an [example](https://github.com/cecobask/integreatly-operator/blob/MGDAPI-3454/pkg/webhooks/operator_webhooks.go#L274-L284) I've managed to find. As far as I know the `Poll` breaks if we return `true` or an error, so to repeat it we should return `false` and no errors, right? :thinking:

cecobask: I've replaced `time.Sleep()` with `wait.PollImmediate()`. Tested on my cluster and it works as expected. The timeout I set is 5 minutes, as suggested. Thank you for the feedback :)

makslion: Looks nice!

makslion:
5 runs - 5 successes. The code looks nice as well. And I've intentionally compromised a couple of test runs - follow-ups were successful (the test failed and then cleaned up its resources). 

/lgtm

PR #2496, opened 2022-02-28
MGDAPI-3496 move to oo 3.0.9

laurafitzgerald:
# Issue link
https://issues.redhat.com/browse/MGDAPI-3496

# What

# Verification steps
This was verified as part of a RHOAM upgrade with @pb82 and @vmanley. 

We installed RHOAM pointing to 3.0.8 of oo on OSD 4.10
grafana-datasources-v2 secret was copied to grafana-datasources so that 3.0.8 wouldn't fail
We carried out an upgrade to 3.0.9 which pointed to -> https://github.com/redhat-developer/observability-operator/pull/107

We verified
- upgrade completed
- deleted the secret and the operator stayed in good state
- that the pvc name stayed the same and was 50Gi - no errors in the logs about this.
- that the observability cr did not update so frequently.

vendor check will fail until the oo release is done.


products/installation.yaml
```diff
@@ -105,7 +105,7 @@ products:
     channel: alpha
     package: observability-operator
     installFrom: "index"
-    index: "quay.io/rhoas/observability-operator-index:v3.0.8"
+    index: "quay.io/rhoas/observability-operator-index:v3.0.9"
```
austincunningham, line 108: Think you need to update the cpaas file as well https://github.com/integr8ly/integreatly-operator/blob/master/products/installation-cpaas.yaml

laurafitzgerald: updated. 👍

PR #2507, opened 2022-03-03
MGDAPI-3552 bump version of CRO in RHOAM to v0.37.1

austincunningham:
# Issue link

MGDAPI-3552

# What

bump CRO version , contains postgres version bump 10.18

# Verification steps

- install RHOAM on CCS with useclusterstorage set to 'false'
- check the Postgres CR's status for the version it should be `10.18`
- check the operatorhub installed operator CRO version should be v0.37.1
- check the rhmi CR the version of CRO should be v0.37.1


laurafitzgerald:
I think we were considering not including the manifests in the project any longer as we're using the index image but I think it's fine this time.
/test rhoam-e2e

austincunningham:
> I think we were considering not including the manifests in the project any longer as we're using the index image but I think it's fine this time.

would need to remove the `ci/prow/manifest` job first

MStokluska:
Verified it on cluster, cro got bumped and version of RDS updated.
Good job! 
/lgtm

PR #2532, opened 2022-03-11
WIP MGDAPI-3377 webhook block tenant apimanger

austincunningham:
# Issue link

MGDAPI-3377

# What
WIP PR investigate weather webhooks can be used to stop tenant CR creation and apimanger CR creation

They can

No need to review as not going to use this method to restrict CR creation in dev sandbox

### DIFF
--- a/main.go
+++ b/main.go
@@ -201,6 +201,36 @@ func setupWebhooks(mgr ctrl.Manager) error {
 		},
 	})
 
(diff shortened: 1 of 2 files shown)


briangallagher:
@austincunningham multitenant-rhoam-e2e is broke at the moment so you can ignore that for now

PR #2534, opened 2022-03-14
MGDAPI-3419 possible fix and more logs to be collected

makslion:
# Issue link
[MGDAPI-3419](https://issues.redhat.com/browse/MGDAPI-3419)

# What
check the comment in the JIRA 

# Verification steps
To fully verify you would need an installation with RHOAM pod on the cluster. Plus you would need to modify the code - right now the added portion of a code will be executed only when the error is encountered 5 times in a row. You could reuse the image I've been using: `quay.io/mvavilov/managed-api-service:mgdapi-3419`. It includes the following modifications:
1. The `FailureCounter` init value is set to 5
2. `connectToPostgres` and `restartPod` are called outside of an error catch statement on each reconcile loop. 

Verification:
This should be tested against a CCS cluster as the issue was spotted only on `useClusterStorage=false` installations. 
1. Install RHOAM by `USE_CLUSTER_STORAGE=false INSTALLATION_TYPE=managed-api IMAGE_FORMAT=<the image of your choice> DEV_QUOTA=10 make cluster/deploy` (to make the life easier, but the verification longer you might reach the `complete` stage using the `quay.io/integreatly/managed-api-service:master` and then flip the image) 
2. Open RHMI pod logs and Keycloak pods in the `rhsso` namespace
3. Watch for the Info message `Successfully connected` followed by the warning `Reached error counter limit of 5, Restarting redhat-rhoam-rhsso pods`
4. Keycloak pods should be deleted and recreated 
5. Before the info message you should see details of the Postgres instance - they should match the `rhsso-postgres-rhoam` in the redhat-rhoam-operator namespace.


makslion:
I'm not covering the code via unit tests as it is intended to be a temporary fix. If it will be decided to keep it as a solution - I'll provide a unit test to cover it

makslion:
rebased to fix failing images 

/test rhoam-e2e

Patryk-Stefanski:
Verified that keycloak pods get deleted and recreated after seeing the specified info message, code looks good.
/lgtm

makslion:
rebased

makslion:
For the one to approve it: could I ask to override code coverage test? It is failing because I've added a logic but, as discussed at the backlog refinement, there will be a follow-up task that will introduce permanent fix and would include unit tests.

makslion:
/hold

Michal has planted a seed of doubt about this PR. Need to sanity-check one thing

makslion:
I'm not insane. It works, hence

/unhold

makslion:
I've updated the logic to not get involved for customers. Since this is intended as a temporary fix and more logs collection we still will have one of the pipelines executing newly added logic, but we will keep away from pings to the client's databases during the installation. 
The coverage still should fail as I'm not providing the unit test for a new logic, so I would ask an approver to override it. Re-running the e2e test just in case.

Patryk-Stefanski:
Code change looks good and makes sense.
/lgtm

makslion:
seems like there was an issue creating or connecting to the cluster 

/retest

carlkyrillos:
/lgtm
All required tests are passing and the code coverage is because of a rebase

PR #2535, opened 2022-03-14
MGDAPI-1728 - Consume addon parameter for the STS role in RHOAM 

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-1728

# What

Consume sts role addon parameter and create the secret required by CRO for sts enabled clusters

**Note**: 
- this currently uses own custom index to install CRO from the STS feature branch
- while 3scale installation will complete, 3scale could not provide access the S3 bucket provisioned by CRO as the STS implemenation does not provide any access keys to the provisioned buckets so some functionality that uses the S3 bucket is not expected to work at this stage. We're expecting this to be provided via an ARN by the user as part of the RHOAM installation and 3scale will support sts and can be configured to use this (https://issues.redhat.com/browse/THREESCALE-7132)

# Verification steps

* Provision or borrow an STS Rosa cluster
* Create the RHOAM role on this account if needed
  * https://github.com/integr8ly/delorean/blob/master/make/ocm.mk#L78-L88 
* oc login to the rosa cluster
* Prepare the cluster with addon secret with the `ROLE_ARN` to install locally 
```
# ROLE_ARN could be different if using a different account or role name
ROLE_ARN="arn:aws:iam::408612754352:role/rhoam_role" make cluster/prepare/local
# Use aws resources for installation
USE_CLUSTER_STORAGE=false make deploy/integreatly-rhmi-cr.yml
```
* Run the operator
```
make code/run
```
* Verify installation of RHOAM completes successfully
* Delete the RHMI CR
* Verifiy RHOAM can be uninstalled successfully

### DIFF
--- a/pkg/resources/sts/sts.go
+++ b/pkg/resources/sts/sts.go
@@ -1,0 +1,49 @@
+package sts
+
+import (
+	"context"
+	"fmt"
+	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/apis/v1alpha1"
+	"github.com/integr8ly/integreatly-operator/pkg/addon"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/logger"
+	cloudcredentialv1 "github.com/openshift/api/operator/v1"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	clusterCloudCredentialName = "cluster"
+	AddonStsArnParameterName   = "sts-role-arn"
+)
+
+func IsClusterSTS(ctx context.Context, client k8sclient.Client, log logger.Logger) (bool, error) {
+	cloudCredential := &cloudcredentialv1.CloudCredential{}
+	if err := client.Get(ctx, k8sclient.ObjectKey{Name: clusterCloudCredentialName}, cloudCredential); err != nil {
+		log.Error("failed to get cloudCredential whle checking if STS mode", err)
+		return false, err
+	}
+
+	if cloudCredential.Spec.CredentialsMode == cloudcredentialv1.CloudCredentialsModeManual {
+		log.Info("STS mode")
+		return true, nil
+	}
+	log.Info("non STS mode")
+	return false, nil
+}
+
+func GetStsRoleArn(ctx context.Context, client k8sclient.Client, namespace string) (string, error) {
+	stsRoleArn, stsFound, err := addon.GetStringParameterByInstallType(
+		ctx,
+		client,
+		integreatlyv1alpha1.InstallationTypeManagedApi,
+		namespace,
+		AddonStsArnParameterName,
+	)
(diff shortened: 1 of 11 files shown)


products/installation.yaml
```diff
@@ -80,7 +80,7 @@ products:
     channel: "rhmi"
     installFrom: "index"
     package: "rhmi-cloud-resources"
-    index: "quay.io/integreatly/cloud-resource-operator:index-v0.37.1"
+    index: "quay.io/kevfan/cloud-resource-operator:index-v0.37.1"
```
KevFan, line 83: Using own custom index for now that is built from the CRO STS feature branch - https://github.com/integr8ly/cloud-resource-operator/pull/315 

This should be updated when the CRO STS feature branch is merged and a CRO release is created

KevFan:
Rebased against master to resolve conflicts

cecobask:
Previously verified by me
/lgtm

KevFan:
Pushed new index of sts feature branch

pkg/resources/sts/sts.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+func IsClusterSTS(ctx context.Context, client k8sclient.Client, log logger.Logger) (bool, error) {
+	cloudCredential := &cloudcredentialv1.CloudCredential{}
+	if err := client.Get(ctx, k8sclient.ObjectKey{Name: clusterCloudCredentialName}, cloudCredential); err != nil {
+		log.Error("failed to get cloudCredential whle checking if STS mode", err)
+		return false, err
+	}
+
+	if cloudCredential.Spec.CredentialsMode == cloudcredentialv1.CloudCredentialsModeManual {
```
laurafitzgerald, line 31: Is there any chance that the ManualMode could be used for other cases aside from STS? Would it be safer to get the secret and see that it has an arn in it? I n the case of non sts I believe there will be no arn in the secret.

KevFan: Seems to be used for when the operator is in a disabled mode 
* https://github.com/openshift/cloud-credential-operator/search?p=1&q=CloudCredentialsModeManual&type=code 
* https://github.com/openshift/cloud-credential-operator/blob/master/docs/mode-manual-creds.md

Not sure, is there a way to detect that the cluster is running in an AWS STS configuration :shrug: 

But yeah, probably be better to check for an ARN in the secret :thinking:

laurafitzgerald: I came to the same conclusion from those docs. It's likely this is fine as it is and either method would give the same result.

KevFan: @laurafitzgerald Should I update this to use the secret or you good to keep it as is ? As you said I think either or would work :thinking:

cecobask (approved):
/lgtm

PR #2538, opened 2022-03-18
MGDAPI-3412-2 fix version

MStokluska:
# Issue link

# What
Bump to version.go and makefile

# Verification steps


MStokluska:
@briangallagher not sure why these were not updated but I think this should go in.

PR #2550, opened 2022-03-24
H22 bug fix

makslion:
# Issue link
N\A

# What
If the first run of the test is skipped it locks `overallSuccess` at `false` and the test fails. 

# Verification steps
You could modify the code to fail the first iteration and check that it is a success and then update the code to skip all and see it fail  :thinking:


adam-cattermole (approved):
The change makes sense, and the test still runs successfully. Nice spot.

/lgtm

PR #2564, opened 2022-03-30
MGDAPI-3704 add openshift.io/cluster-monitoring="true" label to makefile target

austincunningham:
# Issue link

MGDAPI-3704

# What

Update the `make cluster/prepare/local` to add `openshift.io/cluster-monitoring="true"` label to the rhoam operator namespace for local development

# Verification steps

- Run 
```
```
- confirm that the `openshift.io/cluster-monitoring="true"` label has been added to the redhat-rhoam-operator namespace either in the ui
image
or via the command line
```
oc describe ns redhat-rhoam-operator | grep openshift.io/cluster-monitoring=
```


laurafitzgerald:
@austincunningham I believe @den-rgb is also adding this change so depending on which merges first we may have no need for this pr. 👍

PR #2573, opened 2022-04-04
MGDAPI-3412 bump to 1.20

MStokluska:
# Issue link

# What

# Verification steps


bundles/managed-api-service/1.20.0/bundle.Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM scratch
+
+# Core bundle labels.
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=managed-api-service
+LABEL operators.operatorframework.io.bundle.channels.v1=stable
+LABEL operators.operatorframework.io.bundle.channel.default.v1=stable
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.11.0+git
```
laurafitzgerald, line 10: Is this expected version?

MStokluska: We should be using 1.12 as we are on 1.12, not sure where it gets the 1.11+git from but since this is same for 1.18.2 I would address this as part of another Jira.

laurafitzgerald: Thanks Michal. Happy to address this elsewhere as it's the same for previous releases. 👍

laurafitzgerald:
Aside from the images test this change looks fine. 👍

PR #2593, opened 2022-04-08
updating various OWNERS files

laurafitzgerald:
# Issue link

# What

# Verification steps

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -2,7 +2,6 @@
 
 approvers:
 - laurafitzgerald
-- maleck13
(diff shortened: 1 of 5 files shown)


pmccarthy:
@laurafitzgerald can probably remove Craig as an approver while we're at it: https://github.com/integr8ly/integreatly-operator/blob/43fbd7f6ee06db2ec19af6e5e3685c20ff1680b3/OWNERS#L5

Overall looks good though 👍

PR #2599, opened 2022-04-13
MGDAPI-3793 - Implemented gosec install in Dockerfile.tools and added new make command

bryankeane0:
# Issue link
[MGDAPI-3793](https://issues.redhat.com/browse/MGDAPI-3793)

# What
- Added new make command to exclude incomplete rules
- Added install command to `Dockerfile.tools`

# Verification steps
- Comment out all but the newly added command to the `openshift-ci/test/run` file
- Run the two commands from the [openshift-ci README](https://github.com/integr8ly/integreatly-operator/tree/master/openshift-ci#build-and-test) while in the `openshift-ci` folder
- Check the resulting output of the second command prints the version of gosec in the beginning (it may throw errors but that doesn't matter for verification of gosec installation)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -569,3 +569,7 @@ olm/bundle:
 .PHONY: coverage
 coverage:
 	hack/codecov.sh
+
+.PHONY: gosec/exclude
+gosec/exclude:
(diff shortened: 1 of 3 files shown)


Patryk-Stefanski:
Can confirm that the output prints the correct version of gosec.
/lgtm

PR #2611, opened 2022-04-19
MGDAPI-3828 - adding grafana 4.2.0 manifests

laurafitzgerald:
# Issue link
https://issues.redhat.com/browse/MGDAPI-3828

# What
Update customer grafan version to 4.2.0 
Steps taken
Copy manifests from v4.2.0 tag of grafana-operator here -> https://github.com/grafana-operator/grafana-operator/tree/v4.2.0/bundle/manifests
One manual change to the replaces to `replaces: grafana-operator.v3.10.4`

Some manual steps in order to allow the upgrade to complete succesfully. 

Also move to new version of observability operator which brings in new 3.10 version of grafana. 

# Verification steps

As well as installation and upgrade verification, please diff the two manifest directories to see if there is anything standout changed.

```
diff manifests/integreatly-grafana/3.10.4/grafana-operator.v3.10.4.clusterserviceversion.yaml manifests/integreatly-grafana/4.2.0/grafana-operator.clusterserviceversion.yaml
```

## Installation Verification

Checkout this branch. 
Log in to an OSD
Run `INSTALLATION_TYPE=managed-api make cluster/prepare/local`
Then `INSTALLATION_TYPE=managed-api make code/run`

Run the installation to completion
Verify that the version of grafana-operator installed into redhat-rhoam-customer-monitoring-operator namespace is v4.2.0
Verify that the version of grafanaplugin_init container redhat-rhoam-customer-monitoring-operator in is 0.0.6
Verify that the version of observability-operator is 3.0.10 in redhat-rhoam-observability-operator namespace
Verify that the version of grafana-operator installed in redhat-rhoam-observability namespace is 3.10.5
Verify that the version fo grafanaplugin_init container in redhat-rhoam-observability namespace is 0.0.6

Full suite of e2e tests is required as this is a major version bump of grafana. I'm running this on the pr but feel free to run it locally as well.

Check there is no errors inside grafana-operator or grafana for the CRDs that we create for this installation
At a minimum there is a grafana cr, a grafanaDatasource cr and grafanaDashboard cr.


carlkyrillos:
Reviewed the changed files, verified the installation ran to completion, and confirmed that all tests passed. Also verified that the version of the grafana-operator in the redhat-rhoam-customer-monitoring-operator namespace is `v4.2.0`. I did spot two potential issues. 

First, the Image version/tag for the grafana-plugins-init container is `0.0.5` and not `0.5.0` as is stated in the PR's verification steps; not sure if that's a typo in the verification steps or if it's an actual problem. 

I also spotted some errors in the `grafana-proxy` container in the grafana-deployment pod. The error appears to be a permissions issue:
```
E0419 19:09:18.303489 1 webhook.go:111] Failed to make webhook authenticator request: tokenreviews.authentication.k8s.io is forbidden: User "system:serviceaccount:redhat-rhoam-customer-monitoring-operator:grafana-serviceaccount" cannot create resource "tokenreviews" in API group "authentication.k8s.io" at the cluster scope
2022/04/19 19:09:18 oauthproxy.go:793: requestauth: 10.128.4.8:49716 tokenreviews.authentication.k8s.io is forbidden: User "system:serviceaccount:redhat-rhoam-customer-monitoring-operator:grafana-serviceaccount" cannot create resource "tokenreviews" in API group "authentication.k8s.io" at the cluster scope
```

makslion:
`grafana-plugin-init` container was using `quay.io/grafana-operator/grafana_plugins_init:0.0.6` image (matches the PR verification criteria) 
Grafana under the `redhat-rhoam-customer-monitoring-operator` namespace has 4.2.0 version 
the `grafana-proxy` seems to be able to authenticate: 
```
2022/04/20 10:48:33 provider.go:627: 200 GET https://172.30.0.1/apis/user.openshift.io/v1/users/~ {"kind":"User","apiVersion":"user.openshift.io/v1","metadata":{"name":"kube:admin","creationTimestamp":null},"groups":["system:authenticated","system:cluster-admins"]}
2022/04/20 10:48:33 provider.go:627: 201 POST https://172.30.0.1/apis/authorization.openshift.io/v1/subjectaccessreviews {"kind":"SubjectAccessReviewResponse","apiVersion":"authorization.openshift.io/v1","allowed":true,"reason":"RBAC: allowed by ClusterRoleBinding \"cluster-admins\" of ClusterRole \"cluster-admin\" to Group \"system:cluster-admins\""}
2022/04/20 10:48:33 oauthproxy.go:686: 10.129.4.16:41900 authentication complete Session{kube:admin@cluster.local token:true}
```
The fresh install seems to be ok. Will update this comment once will be able to verify the upgrade scenario

pkg/products/grafana/reconciler.go
```diff
@@ -267,34 +268,72 @@ func (r *Reconciler) reconcileGrafanaDashboards(ctx context.Context, serverClien
...
+			},
+		}
+		err := client.Get(ctx, k8sclient.ObjectKey{Name: deployment.Name, Namespace: deployment.Namespace}, deployment)
+		if err != nil {
+			return fmt.Errorf("failed to get DeploymentConfig %s in namespace %s with error: %s", deployment.Name, namespace, err)
+		}
+
+		deployment.Spec.Replicas = &scaleValue
+		err = client.Update(ctx, deployment)
+		return err
```
laurafitzgerald, line 331: only return here if if err != nil

carlkyrillos: @laurafitzgerald I could be wrong but I think this is the correct behavior because the err being returned here is only for the `RetryOnConflict` update function. If err == nil then it breaks out of the `RetryOnConflict` block but if err != nil then it backs off and tries again until it retries too many times. 

That being said, if we want to update it for clarity/readability I can definitely change it to something like:
```
if err != nil {
    return err
}
return nil
```

laurafitzgerald: Thanks Carl. No need for the update. I see that it's the end of the function. 👍 
Would it have been an option to use CreateOrUpdate function here? I've never seen retryoncnflict before. 🤔

makslion:
Done fresh local install - success and as expected. 
Done bundles upgrade master -> this branch. Success and as expected. 
Grafana dashboards had a couple of minutes of disruption but kept the data, 
Using the workload web-app ensured that after the upgrade cusomer-monitoring grafana dashboard is functional. Hence (great work! :tada: 

/lgtm

pkg/products/grafana/reconciler.go
carlkyrillos: I forget if I tried using CreateOrUpdate at one point but I ultimately went with RetryOnConflict because it was used elsewhere in the codebase when working with Deployments. I believe one of the benefits of RetryOnConflict is that it uses exponential backoff when trying to get the latest version of the deployment which helps prevent exhausting the apiserver. Not sure if CreateOrUpdate also uses this or if we need to worry about that for one call but it's worth mentioning.

PR #2620, opened 2022-04-22
add slav as reviewer

cecobask:
# Issue link

# What

# Verification steps


KevFan:
@cecobask Should we create a ticket that can be used as part of the commit and to track this ? :thinking:

cecobask:
> @cecobask Should we create a ticket that can be used as part of the commit and to track this ? 🤔

It looks like I'm able to add `lgmt` labels to pull requests here, so this change is probably not needed.

PR #2623, opened 2022-04-24
MGDAPI-3889 fix envoyconfigRevision removal

MStokluska:
# Issue link

# What

# Verification steps

### DIFF
--- a/pkg/products/marin3r/reconciler.go
+++ b/pkg/products/marin3r/reconciler.go
@@ -147,24 +147,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, nil
 	}
 
-	// Ensures that for an upgrade scenario we are removing V2 EnvoyConfigRevision before
-	// initializing an upgrade to marin3r 0.10.0
-	// TODO remove after the next release from here
-	listOptions := []k8sclient.ListOption{
-		k8sclient.MatchingLabels(map[string]string{
-			"marin3r.3scale.net/envoy-api": "v2",
-		}),
-		k8sclient.InNamespace(fmt.Sprintf("%s3scale", installation.Spec.NamespacePrefix)),
+	phase, err = reconcileEnvoyConfigRevisionsDeletion(ctx, client, *r.installation)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, fmt.Sprintf("Failed to reconcile the envoyConfigRevisionsV2"), err)
+		return phase, err
 	}
-	envoyConfigRevisions := &v1alpha1.EnvoyConfigRevisionList{}
-	_ = client.List(ctx, envoyConfigRevisions, listOptions...)
-	for _, envoyConfigRevision := range envoyConfigRevisions.Items {
-		err = client.Delete(ctx, &envoyConfigRevision)
-		if err != nil || !k8serr.IsNotFound(err) {
-			return phase, err
-		}
-	}
-	// to here
 
 	r.RateLimitConfig = productConfig.GetRateLimitConfig()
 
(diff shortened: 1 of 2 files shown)


pkg/products/marin3r/reconciler.go
```diff
@@ -147,24 +147,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
-		k8sclient.InNamespace(fmt.Sprintf("%s3scale", installation.Spec.NamespacePrefix)),
-	}
-	envoyConfigRevisions := &v1alpha1.EnvoyConfigRevisionList{}
-	_ = client.List(ctx, envoyConfigRevisions, listOptions...)
-	for _, envoyConfigRevision := range envoyConfigRevisions.Items {
-		err = client.Delete(ctx, &envoyConfigRevision)
-		if err != nil || !k8serr.IsNotFound(err) {
-			return phase, err
-		}
+	phase, err = reconcileEnvoyConfigRevisionsDeletion(ctx, client, *r.installation)
```
laurafitzgerald, line 150:
```suggestion
	phase, err = reconcileEnvoyConfigRevisionsDeletion(ctx, client, installation)
```
we usually pass the in memory copy of installation down in the functions rather than the one belonging to the reconciler instantiation.

pkg/products/marin3r/reconciler.go
```diff
@@ -537,3 +524,50 @@ func (r *Reconciler) reconcileServiceMonitor(ctx context.Context, client k8sclie
...
+
+		// Fetch object again before removing finalizer to ensure that configRevision has not been deleted already by client.Delete call above - this can happen if for any reason
+		// there was no Finalizer on envoyConfigRevision
+		err = client.Get(ctx, k8sclient.ObjectKey{Name: envoyConfigRevision.Name, Namespace: envoyConfigRevision.Namespace}, &v1alpha1.EnvoyConfigRevision{})
+		if k8serr.IsNotFound(err) {
+			// if configRevision is not found, we are good to go
+			continue
+		} else if err != nil {
+			return integreatlyv1alpha1.PhaseFailed, err
+		}
```
laurafitzgerald, line 559:
```suggestion
		if err != nil {
			if k8serr.IsNotFound(err) {
			     // if configRevision is not found, we are good to go
			     continue
			}
			return integreatlyv1alpha1.PhaseFailed, err
		}
```
slightly cleaner without the else. and there's possibility the err in the first check is nil.

pkg/products/marin3r/reconciler_test.go
```diff
@@ -233,3 +237,220 @@ func strPtr(str string) *string {
...
+							},
+							Spec:   v1alpha1.EnvoyConfigRevisionSpec{},
+							Status: v1alpha1.EnvoyConfigRevisionStatus{},
+						},
+					},
+				})
+			},
+			installation: &integreatlyv1alpha1.RHMI{
+				Spec: integreatlyv1alpha1.RHMISpec{
+					NamespacePrefix: "redhat-rhoam-",
```
laurafitzgerald, line 384: What do you think about adding a case for other namespace prefixes that we anticipate used with the operator to protect from a change on that line in the future?

pkg/products/marin3r/reconciler_test.go
```diff
@@ -233,3 +237,220 @@ func strPtr(str string) *string {
...
+		if envoyConfigRevision.Finalizers != nil {
+			return fmt.Errorf("finalizer has not been cleaned up properly")
+		}
+
+	}
+
+	return nil
+}
+
+func confirmThatConfigRevisionsV3areThereAndV2areGone(ctx context.Context, client k8sclient.Client, nsPrefix string) error {
```
laurafitzgerald, line 430:
```suggestion
func confirmThatConfigRevisionsV3areThereAndV2areGone(ctx context.Context, client k8sclient.Client) error {
```
parameter unused in the function

pkg/products/marin3r/reconciler_test.go
```diff
@@ -233,3 +237,220 @@ func strPtr(str string) *string {
...
+
+			err = tt.testFunction(context.TODO(), serverClient, tt.installation.Spec.NamespacePrefix)
+			if err != nil {
+				t.Errorf("ConfigRevisions not cleaned %s", err)
+			}
+		})
+	}
+}
+
+func confirmThatConfigRevisionsAreCleared(ctx context.Context, client k8sclient.Client, nsPrefix string) error {
```
laurafitzgerald, line 409:
```suggestion
func confirmThatConfigRevisionsAreCleared(ctx context.Context, client k8sclient.Client) error {
```
Parameter unused in the function.

laurafitzgerald (changes requested):
thanks for the change @MStokluska I believe this will solve the issue faced on upgrade previously.  The flow of the logic makes sense to me. Just left a few code review comments/suggestions inline.

pkg/products/marin3r/reconciler.go
MStokluska: Hey, thanks, good point, actually, looking at the code, going to change it to string rather than entire rhmi.

pkg/products/marin3r/reconciler.go
MStokluska: Thanks. Done!

pkg/products/marin3r/reconciler_test.go
MStokluska: Good catch, had this using the 3scale ns before refactor. Thanks

pkg/products/marin3r/reconciler_test.go
MStokluska: Same as above. Thanks

pkg/products/marin3r/reconciler_test.go
MStokluska: Done, for `rhoami` ns.

laurafitzgerald:
**Verification**
Witnessed @MStokluska taking the following steps.
Reproduction
- install 1.20.0 with a hardcoded namespace value which caused the first upgrade error.
- In the rhmi cr marin3r reports 1.10.0
- The EnvoyConfigRevisions don't get deleted because of the namespace erorr
- In the rhmi cr marin3r block loses the version and is marked completed - the cr does not go complete
- Update to 1.20.0 with the correct namespace value
- The EnvoyConfigRevisions get a deletion timestamp but do not get finalized so don't get deleted
- Move to this pr image
- the EnvoyConfigRevisions have their finalizers removed and are as such deleted.
- Marin3r opertoar goes back into awaiting opertor
- The catalog source is now in a bad state.
- Deleteing the CSV manually progresses the upgrade.

This code plus 1 manual step of deleting the CSV will resolve the existing upgrade cluster state.

In order to verify 1.19.0 -> 1.20.0 we'll go ahead and merge this pr to build the new container image
and install 1.19.0 via managed-tenants and do full upgrade to 1.20.0 with the v2 envoyconfig revisions in place.

laurafitzgerald:
@MStokluska left the verifcation steps above.
Just needs a `make code/fix` to pass the final prow check and we can merge then. 👍

MStokluska:
> @MStokluska left the verifcation steps above. Just needs a `make code/fix` to pass the final prow check and we can merge then. +1

Thanks @laurafitzgerald I've ran the make code/fix a couple of times but it seems that there's nothing to fix. Forced pushed just now again, lets see what it comes back with!

pkg/products/marin3r/reconciler_test.go
```diff
@@ -233,3 +237,253 @@ func strPtr(str string) *string {
...
+							ObjectMeta: v1.ObjectMeta{
+								Finalizers: []string{
+									"envoyconfigrevisions.marin3r.3scale.net",
+								},
+								Labels: map[string]string{
+									"marin3r.3scale.net/envoy-api": "v2",
+								},
+								Namespace:         "redhat-rhoam-3scale",
+								Name:              "apicast-ratelimit",
+								DeletionTimestamp: &v1.Time{time.Now()},
```
austincunningham, line 304: looks like this line is what `make code/fix` is complaining about

MStokluska: thanks, I've refactored, let's see how it looks now

austincunningham: looks like format passed

PR #2626, opened 2022-04-25
MGDAPI-3794: Remove G101 from makefile exclude list

jjkiely:
# Issue link
https://issues.redhat.com/browse/MGDAPI-3794

# What
Remove any occurences of hard coded credentials, and remove G101 from the gosec exclude list in the makefile.

# Verification steps
Verify that the gosec prow check passes on this PR, and that all lines of code marked as false positives are in fact false positives (ie they do not contain hard coded credentials).


vendor/github.com/integr8ly/cloud-resource-operator/pkg/providers/aws/credentials.go
```diff
@@ -24,7 +24,7 @@ const (
 	defaultProviderCredentialName = "cloud-resources-aws-credentials"
 
 	defaultCredentialsKeyIDName = "aws_access_key_id"
-	// #nosec G101
+	// #nosec G101 -- This is a false positive
```
KevFan, line 27: Running `make vendor/fix` seems to prefer without the comment, which I think would make sense since this is in the `vendor` package, so we're just pulling in the current CRO version dependency files :thinking:

jjkiely: @KevFan good spot thanks for the tip, test's passing now :slightly_smiling_face:

KevFan: Yup, I think we can add the comment to CRO here instead if we want to add it in https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/credentials.go#L27 :slightly_smiling_face:

KevFan:
Changes looks good to me :+1: 

/lgtm

jjkiely:
Thanks @KevFan! Didn't even have to ask :laughing:

PR #2629, opened 2022-04-25
[MGDAPI-3882] use observability namespace and add unit tests

cecobask:
# Issue link
https://issues.redhat.com/browse/MGDAPI-3882

# What
This function - https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/threescale/reconciler.go#L1095 looks for the secret in the observability operator ns while it should look for it under observability ns.

# Verification steps
Check unit tests are passing and the coverage of `GetExistingSMTPFromAddress` is sufficient

### DIFF
--- a/pkg/resources/smtp.go
+++ b/pkg/resources/smtp.go
@@ -28,12 +28,18 @@ func GetExistingSMTPFromAddress(ctx context.Context, client k8sclient.Client, ns
 	if err != nil {
 		return "", err
 	}
-	monitoring := amSecret.Data[alertManagerConfigSecretFileName]
+	monitoring, ok := amSecret.Data[alertManagerConfigSecretFileName]
+	if !ok {
+		return "", fmt.Errorf("failed to find %s in %s secret data", alertManagerConfigSecretFileName, alertManagerConfigSecretName)
+	}
 	var config alertManagerConfig
 	err = yaml.Unmarshal(monitoring, &config)
 	if err != nil {
 		return "", fmt.Errorf("failed to parse alert monitoring yaml: %w", err)
 	}
-
-	return config.Global["smtp_from"], nil
+	smtpFrom, ok := config.Global["smtp_from"]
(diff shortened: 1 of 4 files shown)


pkg/resources/smtp_test.go
```diff
@@ -0,0 +1,76 @@
...
+			WantErr: true,
+		},
+	}
+
+	for _, scenario := range scenarios {
+		t.Run(scenario.Name, func(t *testing.T) {
+			_, err := GetExistingSMTPFromAddress(context.TODO(), scenario.FakeClient, "test")
+			if !scenario.WantErr && err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
```
Boomatang, line 73: There is no checks to see if the function actual returns expected values if the function is successfully.

pkg/resources/smtp.go
```diff
@@ -19,17 +19,16 @@ type alertManagerConfig struct {
...
-
+	err := client.Get(ctx, types.NamespacedName{
+		Name:      alertManagerConfigSecretName,
+		Namespace: ns,
+	}, amSecret)
 	if err != nil {
 		return "", err
 	}
-
 	monitoring := amSecret.Data[alertManagerConfigSecretFileName]
```
Boomatang, line 31: There could possible be a blank value here. Can error checking be added to on the existence of the `alertManagerConfigSecretFileName`. A test case should be added to for the missing data key. 

As github wont allow me to add the comment I will put it here. The final return statement could also have an error if the `stmp_from` key is missing. We should be testing for this also.

Boomatang (changes requested):
This is almost there. A few things need to be added.

Your commit message does not start with the Jira number.

The happy path test does not check the returned value is as expected.

There is two test case that needs to be covered.
1. secret.data does not contain key "alertmanager.yaml"
2. "smtp_from" does not exist in config.Global

PR #2631, opened 2022-04-26
release PR for version rhoam-v1.20.1-rc1

fh-automatron:


config/manifests-rhoam/bases/managed-api-service.clusterserviceversion.yaml
```diff
@@ -12,7 +12,7 @@ metadata:
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v2
```
austincunningham, line 12: do we need the skip range here ?

MStokluska: not really, that's a base and is fine as it is

austincunningham: @MStokluska

PR #2634, opened 2022-04-27
MGDAPI-3878 - Added annotation to 3scale api-manager to skip apicast service port reconcile

kryanbeane:
# Issue link
[MGDAPI-3878](https://issues.redhat.com/browse/MGDAPI-3878) & [THREESCALE-8385](https://issues.redhat.com/browse/THREESCALE-8385)
# What
Added function in 3scale reconciler to reconcile annotation to allow rate limit to take over apicast service port configuration
# Verification steps
- Change the 3scale image in the CSV to `quay.io/bkeane0/3scale-operator:0.4.0` (my image from the related 3scale PR to skip service port reconcile) and install RHOAM
- You'll also need to implement the rules from this PR https://github.com/3scale/3scale-operator/commit/d56aaee25608988641e16cf92b332c886948523d into integreatly for testing purposes.
- Update the APIManager CRD to match the latest 3scale version
- There may be some issue with 3scale install / the CRD for the apimanager so give 3scale one reconcile and then add the `externalComponents` section from here https://github.com/3scale/3scale-operator/blob/d4b44f6d04398b5858659729c899304b8dc3ff8f/bundle/manifests/apps.3scale.net_apimanagers.yaml#L2598-L2617 to the CRD manually
- Check that the `APIManager` CR has the annotation `apps.3scale.net/disable-apicast-service-reconciler: 'true'` after you have fixed this problem with the CRD (I've also run into issues with `FILE_UPLOAD_STORAGE` being duplicated and stopping 3scale install but this is a separate issue
- Check the `apicast-production` and `apicast-staging` services that the `gateway` port is set to `8443`
- Make a change to either apicast `deploymentConfig` to trigger the apicast reconciler.
- Watch both services and ensure the gateway port does not change to `8080`


bryankeane0:
Current issue: 3scale reconciler isn't picking up the annotation on reconcile or on creation of apimanager

bryankeane0:
WIll add unit test to cover new logic

pkg/products/threescale/reconciler.go
```diff
@@ -3078,3 +3088,25 @@ func (r *Reconciler) addSSOReadyAnnotationToUser(ctx context.Context, client k8s
...
 	return nil
 }
+
+func (r *Reconciler) reconcileRatelimitPortAnnotation(ctx context.Context, client k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
+	apim := &k8sappsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "threescale-operator-controller-manager-v2",
+			Namespace: r.Config.GetNamespace(),
+		},
+	}
```
KevFan, line 3098: Should this be adding the annotation to the `ApiManager` CR instead ? Changes from https://github.com/3scale/3scale-operator/pull/739/files#diff-58577e8a0e7d8aa700f7595e8b67800c4eab26d20123298924dec34e5f26151bR112-R124 looks to suggest the annotation is added to the `ApiManager` CR rather than the operator deployment :thinking:

bryankeane0: This was the approach I had taken initially and couldn't get it working.. it would make more sense it being on the cr though, I'll commit my current changes and test this, thanks :)

pkg/products/threescale/reconciler.go
```diff
@@ -577,8 +585,9 @@ func (r *Reconciler) reconcileComponents(ctx context.Context, serverClient k8scl
 	resourceRequirements := r.installation.Spec.Type != string(integreatlyv1alpha1.InstallationTypeWorkshop)
 	apim := &threescalev1.APIManager{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      apiManagerName,
-			Namespace: r.Config.GetNamespace(),
+			Name:        apiManagerName,
+			Namespace:   r.Config.GetNamespace(),
+			Annotations: map[string]string{"apps.3scale.net/disable-apicast-service-reconciler": "true"},
```
laurafitzgerald, line 590: I see that we are adding annotations here and also in a sperate function. Do we need both? If we are using it here I would suggest using the flow below. check if the annotations are nil. If it's nil add the string map. the above will overwrite existing annotations that are added which we don't want to do in case a user wants to add any to the list.

pkg/products/threescale/reconciler.go
```diff
@@ -150,13 +150,14 @@ func NewReconciler(configManager config.ConfigReadWriter, installation *integrea
...
-	appsv1Client  appsv1Client.AppsV1Interface
-	oauthv1Client oauthClient.OauthV1Interface
+	ConfigManager        config.ConfigReadWriter
+	Config               *config.ThreeScale
+	mpm                  marketplace.MarketplaceInterface
+	installation         *integreatlyv1alpha1.RHMI
+	tsClient             ThreeScaleInterface
+	appsv1Client         appsv1Client.AppsV1Interface
+	oauthv1Client        oauthClient.OauthV1Interface
+	apiManagerDeployment k8sappsv1.Deployment
```
laurafitzgerald, line 160: I think this can be removed

laurafitzgerald (changes requested):
@bryankeane0 did a quick review there. I think you had some test cases added or updated to push as well?

pkg/products/threescale/reconciler.go
bryankeane0: Ah okay I didn't realise they overwrite existing ones. pushing this and the change below now

bryankeane0:
note: Moved the tracking branch from master to the autoscaling one so commits look a bit funky

den-rgb:
All worked as expected and port did not change back to port 8080.

pkg/products/threescale/reconciler.go
```diff
@@ -434,6 +433,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
+	// Ensure ratelimit annotation is ready before returning phase complete
+	phase, err = r.reconcileRatelimitPortAnnotation(ctx, serverClient)
```
MStokluska, line 437: hey @bryankeane0 any reason to have it at the end of reconcile loop instead of BEFORE creating/reconciling rate limiting?
Reason I say that is because it might turn out that the rate limiting will be created, dc's update, then update again because of 3scale operator rollback, then we are switching off the reconciliation and only then update the dcs again from rate limiting?

bryankeane0: Cool I'll implement it now thanks for the suggestion

MStokluska:
by-passing codecov

PR #2645, opened 2022-05-04
MGDAPI-3209 fix CSV image containterimage

austincunningham:
# Issue link

# What

fix CSV image containterImage


trepel:
Haven't tested but the change looks reasonable and these are the only occurrences of 'austincunningham' in the repo that need to be fixed.

/lgtm

PR #2651, opened 2022-05-05
MGDAPI-3875 / Added resource limits + requests

den-rgb:
# Issue Link
[MGDAPI-3875](https://issues.redhat.com/browse/MGDAPI-3875)

# What
Added cpu + memory usage limits + requests to threescale reconciler.

# Verification
1. Install rhoam off this branch.
2. In DeploymentConfigs (apicast-production, apicast-staging, backend-listener) make sure the correct annotations are set.
3. In those 3 DeplymentConfigs make sure the envoy-sidecar container contains the resource limits and requests.
image

### DIFF
--- a/pkg/products/threescale/reconciler.go
+++ b/pkg/products/threescale/reconciler.go
@@ -108,6 +108,12 @@ var (
 		"zync-database",
 		"zync-que",
 	}
+
+	AnnotationDeploymentConfigs = []string{
+		"apicast-production",
+		"apicast-staging",
+		"backend-listener",
+	}
 )
 
 func NewReconciler(configManager config.ConfigReadWriter, installation *integreatlyv1alpha1.RHMI, appsv1Client appsv1Client.AppsV1Interface, oauthv1Client oauthClient.OauthV1Interface, tsClient ThreeScaleInterface, mpm marketplace.MarketplaceInterface, recorder record.EventRecorder, logger l.Logger, productDeclaration *marketplace.ProductDeclaration) (*Reconciler, error) {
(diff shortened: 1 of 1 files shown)


pkg/products/threescale/reconciler.go
```diff
@@ -433,6 +439,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
+	err = r.addResourceThreshold(ctx, serverClient)
+	r.log.Infof("ensureDeploymentConfigsReadyForEnvoy-SidecarContainer", l.Fields{"phase": phase})
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
```
makslion, line 444: not sure we need to care about the `phase` here :thinking: Since `addResourceThreshold` isn't updating it

makslion:
and shouldn't this PR be against a feature branch? :thinking:

makslion:
Was able to verify that requests and limits are in place. 
I've discussed this with Denis, but also mention it here for other people. This PR collides with the [e2e test](https://github.com/integr8ly/integreatly-operator/pull/2635). When one of these PRs be merged, the second one would need to update the test.

pkg/products/threescale/reconciler.go
den-rgb: how would you suggest i write this instead?

makslion: hmm, just take it away? :thinking: 
```
	err = r.addResourceThreshold(ctx, serverClient)
	r.log.Info("ensureDeploymentConfigsReadyForEnvoy-SidecarContainer")
	if err != nil {
		events.HandleError(r.recorder, installation, phase, "Failed to ensure deployment configs for envoy-sidecar were ready", err)
		return phase, err
	}
```

den-rgb:
> and shouldn't this PR be against a feature branch? thinking

yes you're right I'll have to make a new pr for this in that case

makslion:
Looks good. Let the `approve` race begin! 

\lgtm

makslion:
Looks good let the `approve` race begin! 

/lgtm

makslion:
multitenant rhoam tests are failing now all around PRs and even on empty ones -> https://github.com/integr8ly/integreatly-operator/pull/2650 To me this allows to override failing mt test here

makslion:
mt test should be fine now 

/retest

makslion:
I doubt, but maybe I can do it :thinking: 

/override multitenant-rhoam-e2e

makslion:
yep, I can't and that a relief

PR #2653, opened 2022-05-05
[MGDAPI-3890] bump 3scale-operator to 0.8.3+0.1649688682.p

cecobask:
# Issue link
https://issues.redhat.com/browse/MGDAPI-3890

# What
Add 3scale-operator manifest for version `0.8.3+0.1649688682.p`

# Verification steps
- Install RHOAM via olm from this index image: `quay.io/tdimov0/managed-api-service-index:1.21.0`
- Once installed, setup workload web app on your cluster (https://github.com/integr8ly/workload-web-app#rhoam-clusters)
- Update catalog source to point to a new index: `quay.io/tdimov0/managed-api-service-index:1.21.1`
- Approve the upgrade
- Wait for the upgrade to complete and check the total downtime in 3scale (should not be longer than 2mins)
- Verify rhoam's version is `1.21.1` and 3scale-operator's version is `0.8.3`

### DIFF
--- a/pkg/resources/marketplace/manifestHelper.go
+++ b/pkg/resources/marketplace/manifestHelper.go
@@ -267,10 +267,12 @@ func ReadAndFormatManifestYamlFile(path string) (string, error) {
 	return formattedString.String(), nil
 }
 
-// Gets the version number from the package yaml string
+// GetCurrentCSVFromManifest Gets the version number from the package yaml string
+// Allowed formats include regular semantic versions, however recently ones of the
+// following format had to be accounted for: 3scale-operator.v0.8.3-0.1649688682.p
 func GetCurrentCSVFromManifest(packageYaml string) (string, error) {
 	format1 := `.(V|v|)([0-9]+)(.)([0-9]+)(.|-)([0-9]+)($|\n)`
-	format2 := `.(V|v|)([0-9]+)(.)([0-9]+)(.|-)([0-9]+)(-)([0-9]+)(.)([0-9]+)(.p)($|\n)`
+	format2 := `.(V|v|)([0-9]+)(.)([0-9]+)(.|-)([0-9]+)(-|\+)([0-9]+)(.)([0-9]+)(.p)($|\n)`
 	r, _ := regexp.Compile(format1 + "|" + format2)
 	matches := removeEmptyStrings(r.FindStringSubmatch(packageYaml))
 	if len(matches) < 5 {
--- a/pkg/resources/marketplace/manifestHelper.go
+++ b/pkg/resources/marketplace/manifestHelper.go
@@ -9,7 +9,6 @@ import (
 	"reflect"
 	"regexp"
 	"sort"
-	"strconv"
 	"strings"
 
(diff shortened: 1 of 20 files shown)


pkg/resources/marketplace/manifestHelper.go
```diff
@@ -270,17 +269,28 @@ func ReadAndFormatManifestYamlFile(path string) (string, error) {
 
 // Gets the version number from the package yaml string
```
laurafitzgerald, line 270: Could we add some detail to the comment here about the new version syntax we allowing for.

pkg/resources/marketplace/manifestHelper_test.go
```diff
@@ -70,17 +70,17 @@ func TestGetCurrentCSVFromManifest(t *testing.T) {
                                 channels:
                                 - name: integreatly
                                 currentCSV: 3scale-operator.v1.9-7`,
-			ExpectedVer: "1.9.7",
+			ExpectedVer: "1.9-7",
```
laurafitzgerald, line 73: What was the reason for this change? 🤔

scripts/prepare-release.sh
```diff
@@ -301,10 +301,10 @@ if [[ -n "$SERVICE_AFFECTING" ]]; then
  set_csv_service_affecting_field "$SERVICE_AFFECTING"
 fi
 
-# The following is disabled to unblock rc1 cut of 1.13.0 - it should be renabled before final release.
-if [[ "${OLM_TYPE}" == "managed-api-service" ]]; then
- set_related_images
-fi
+# The following is disabled to unblock 1.21.0 - it should be re-enabled after release.
```
laurafitzgerald, line 304: I think this will cause related_images to not be present in the new CSV. Although I think we have the manifests geneated for the next release so we could address this later.

pkg/resources/marketplace/manifestHelper_test.go
cecobask: The reason for this change was that originally the `-` sign was always being replaced by a `.` sign which was not what we expect with formats like `3scale-operator.v0.8.3-0.1649688682.p`. There was never a version like `1.9-7` in the history, but this function can be updated in the future to account for such cases.

makslion:
hey, when we are expecting 3scale 0.9.0 to be in available? :thinking: and was this tested in the CPaaS? :thinking:

cecobask:
> hey, when we are expecting 3scale 0.9.0 to be available? 🤔 and was this tested in the CPaaS? 🤔

Hi @makslion, I'm not sure when 0.9.0 will be available. This wasn't tested in CPaaS yet.

makslion:
> > hey, when we are expecting 3scale 0.9.0 to be available? thinking and was this tested in the CPaaS? thinking
> 
> Hi @makslion, I'm not sure when 0.9.0 will be available. This wasn't tested in CPaaS yet.

Was able to verify the upgrade. With this being tested in CPaaS - 

/gtm

makslion:
seriously, bot? 

/lgtm

cecobask:
Closing this PR in favour of switch to using indexes when 0.9.0 of 3scale-operator is made available

PR #2656, opened 2022-05-06
Mgdapi 3764

Boomatang:
# Issue link

https://issues.redhat.com/browse/MGDAPI-3764

# What

Add support to allow for custom smtp configuration for 3scale

# Verification steps

* install this branch.
* set up the testing IDP

## Full configured custom smtp
 * add all the required fields to the `addon-managed-api-service-parameters` secret. This can be done using the config file in this pr. `oc process -n redhat-rhoam-operator -f config/secrets/custom-smtp-secret.yaml | oc apply -f -` updates the secret as required.
 * Allow full reconcile loops to happen.
 * Expected:
   * A new secret is added in the rhoam name space called `custom-smtp`. This will have the values from the parameters secret
   * The `system-smtp` secret in the 3scale name space will have the new custom values.
   * Log into the 3scale UI. The outgoing email address should be changed
   * The rhmi cr status block will have `customSmtp.active = true`

## Partially configured custom smtp
* remove some of the "custom-smtp" values from the `addon-managed-api-service-parameters` secret.
* Allow full reconcile loops to happen
* Expected:
  * The `custom-smtp` secret in the rhoam name space is removed
  * The `system-smtp` secret in the 3scale name space has reverted back to the default values
  * Log into the 3scale UI and the outgoing email address has revert back to the default value
  * The rhmi cr status block will have `customSmtp.active = false`
  * The  `customSmtp.error` will state which values are missing.

## No custom smtp
* remove all the "custom-smtp" values from the `addon-managed-api-service-parameters` secret.
 * Allow full reconcile loops to happen
 * * Expected:
  * The `custom-smtp` secret in the rhoam name space is removed
  * The `system-smtp` secret in the 3scale name space has reverted back to the default values
  * Log into the 3scale UI and the outgoing email address has revert back to the default value
  * The rhmi cr status block will not have a `customSmtp` block


config/secrets/custom-smtp-secret.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: quota-secret
+objects:
+  - apiVersion: v1
+    kind: Secret
+    metadata:
+      name: addon-managed-api-service-parameters
```
laurafitzgerald, line 9: We have another dummy secret in this folder which we're using for QUOTA configuration. I think we should rename the yaml file to `addon-managed-api-sevices-paramaters` as it covers all ocm parameters to the installation. wdyt?

pkg/resources/custom-smtp/custom-smtp.go
```diff
@@ -0,0 +1,233 @@
...
+	OpensslVerifyMode string
+	Username          string
+}
+
+func GetCustomAddonValues(serverClient k8sclient.Client, namespace string) (*CustomSmtp, error) {
+
+	addonName := addon.GetName(v1alpha1.InstallationTypeManagedApi)
+	secretName := fmt.Sprintf("addon-%s-parameters", addonName)
+
+	secret := &corev1.Secret{}
```
laurafitzgerald, line 38: Could we add a function for getting the secret. It's used a few times in this file.

apis/v1alpha1/rhmi_types.go
```diff
@@ -210,6 +210,11 @@ type AlertingEmailAddresses struct {
 	CSSRE        string `json:"cssre"`
 }
 
+type CustomSmtpStatus struct {
+	Active bool   `json:"active"`
```
laurafitzgerald, line 214: Small nit -
we use `Enabled bool` for SMTPEnabled in the cr. 
Could we use Enabled here as well instead of Active?

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -762,6 +770,43 @@ func (r *Reconciler) processQuota(installation *rhmiv1alpha1.RHMI, namespace str
...
+	validation := cs.ParameterValidation(smtp)
+	if validation == cs.Valid {
+		phase, err := cs.CreateOrUpdateCustomSMTPSecret(ctx, serverClient, smtp, r.installation.Namespace)
+		if err != nil {
+			return phase, err
+		}
+		r.installation.Status.CustomSmtp = &rhmiv1alpha1.CustomSmtpStatus{}
+		r.installation.Status.CustomSmtp.Active = true
+
+	} else if validation == cs.Partial {
```
laurafitzgerald, line 789: It seems that Valid Partial and Blank are mutually exclusive so you don't need the else here or below.

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -762,6 +770,43 @@ func (r *Reconciler) processQuota(installation *rhmiv1alpha1.RHMI, namespace str
...
+		phase, err := cs.DeleteCustomSMTP(ctx, serverClient, r.installation.Namespace)
+		if err != nil {
+			return phase, err
+		}
+
+		errorString := cs.ParameterErrors(smtp)
+		r.installation.Status.CustomSmtp.Active = false
+		r.installation.Status.CustomSmtp.Error = fmt.Sprintf("Custom SMTP partial configured, missing fields: %s", errorString)
+
+	} else {
```
laurafitzgerald, line 799: I would rather make this an if on the validation rather than an else in case the validation return gets updated in the future. 
After the if you could have a return of error of unknown validation type.

laurafitzgerald (changes requested):
@Boomatang partial code review done. I will complete later on but dropping the comments I have currently for times sake.

pkg/products/threescale/reconciler.go
```diff
@@ -1092,12 +1104,24 @@ func (r *Reconciler) reconcileOutgoingEmailAddress(ctx context.Context, serverCl
...
+			if !k8serr.IsNotFound(err) {
+				r.log.Error("error getting smtp_from address from custom smtp secret", err)
+				return integreatlyv1alpha1.PhaseFailed, nil
+			}
+		}
+	} else {
+		existingSMTPFromAddress, err = resources.GetExistingSMTPFromAddress(ctx, serverClient, observabilityConfig.GetNamespace())
+
+		if err != nil {
+			if !k8serr.IsNotFound(err) {
```
briangallagher, line 1121: I know this is existing code but I'm wondering if the same principle applies here. IsNotFound error should return an error as the alertmanager-application-monitoring secret should always exist.
A warning log at a minimum would be good

pkg/products/threescale/reconciler.go
```diff
@@ -492,7 +493,18 @@ func (r *Reconciler) reconcileSMTPCredentials(ctx context.Context, serverClient
...
+		r.log.Info("configuring user smtp for 3scale notifications")
+		secretName = cs.CustomSecret
+	}
+
+	err := serverClient.Get(ctx, k8sclient.ObjectKey{Name: secretName, Namespace: r.installation.Namespace}, credSec)
+	if err != nil {
+		r.log.Warningf("could not obtain smtp credentials secret", l.Fields{"error": err})
+	}
+
 	if err != nil {
```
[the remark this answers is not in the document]
Boomatang: That came from the rebase, will remove

pkg/products/threescale/reconciler.go
Boomatang: I think in this case it can be expected that the secret many not be there. There is a check a bit later for the empty `existingSMTPFromAddress` and if empty pulls the value from the a envvar.

I will add the warning log but I don't think we should always be returning on this secret being missing.

pkg/resources/custom-smtp/custom-smtp.go
```diff
@@ -0,0 +1,216 @@
...
+
+		secret.Data["from_address"] = []byte(smtp.FromAddress)
+		secret.Data["host"] = []byte(smtp.Address)
+		secret.Data["password"] = []byte(smtp.Password)
+		secret.Data["port"] = []byte(smtp.Port)
+		secret.Data["username"] = []byte(smtp.Username)
+
+		return nil
+	}); err != nil {
+		return v1alpha1.PhaseInProgress, err
```
briangallagher, line 174: Should this be PhaseFailed or InProgress? Is there a scenarion where it could be inProgress and then recover?

pkg/resources/custom-smtp/custom-smtp.go
```diff
@@ -0,0 +1,216 @@
...
+
+func DeleteCustomSMTP(ctx context.Context, serverClient k8sclient.Client, namespace string) (v1alpha1.StatusPhase, error) {
+
+	secret := &corev1.Secret{}
+	err := serverClient.Get(ctx, k8sclient.ObjectKey{
+		Name:      CustomSecret,
+		Namespace: namespace,
+	}, secret)
+
+	if k8serr.IsNotFound(err) {
```
briangallagher, line 189: I think the precedent in the code base might be more like this:

	if err != nil {
		if k8serr.IsNotFound(err) {
			return v1alpha1.PhaseCompleted, nil
		}
		return v1alpha1.PhaseFailed, err
	}

controllers/rhmi/bootstrapReconciler.go
laurafitzgerald: A switch statement would make the logic more readable here so i'm in favour of that. 👍

briangallagher (commented):
@Boomatang Reviewed, added some comments

pkg/resources/custom-smtp/custom-smtp.go
```diff
@@ -0,0 +1,216 @@
...
+		customSmtp.Username = string(value)
+	}
+
+	return customSmtp, nil
+}
+
+func ParameterValidation(smtp *CustomSmtp) ValidationResponse {
+
+	valid, partial := false, false
+
```
MStokluska, line 80: Would be worth checking the validation more in-depth rather than just checking if it's empty or not?
I wonder what's going to happen if I enter some random string in the port value?

MStokluska: The reason I ask is that we are saying a config is valid, although from what it looks to me, it might not be valid, it's just that it has all the values set.

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -762,6 +770,43 @@ func (r *Reconciler) processQuota(installation *rhmiv1alpha1.RHMI, namespace str
...
+
+	} else if validation == cs.Partial {
+		phase, err := cs.DeleteCustomSMTP(ctx, serverClient, r.installation.Namespace)
+		if err != nil {
+			return phase, err
+		}
+
+		errorString := cs.ParameterErrors(smtp)
+		r.installation.Status.CustomSmtp.Active = false
+		r.installation.Status.CustomSmtp.Error = fmt.Sprintf("Custom SMTP partial configured, missing fields: %s", errorString)
```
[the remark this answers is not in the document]
Boomatang: Nice catch

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -762,6 +770,43 @@ func (r *Reconciler) processQuota(installation *rhmiv1alpha1.RHMI, namespace str
...
+
+	} else {
+		phase, err := cs.DeleteCustomSMTP(ctx, serverClient, r.installation.Namespace)
+		if err != nil {
```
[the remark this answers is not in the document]
Boomatang: I guess form a safety point we can set the value to false if there is an error given when trying to delete the custom smtp secret.

pkg/resources/custom-smtp/custom-smtp.go
Boomatang: While yes your points are valid, there are a few blockers to doing that. Currently we don't have a good way to validate the configuration short from sending emails with the configuration. The question about the port being a number is good. We are creating a follow up JIRA to improve our validate. I will make note of this in that ticket.

controllers/rhmi/bootstrapReconciler.go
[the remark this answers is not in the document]
MStokluska: EVEN BETTER - Lets go wild and make it partiaLLy

pkg/resources/custom-smtp/custom-smtp.go
```diff
@@ -0,0 +1,216 @@
...
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      CustomSecret,
+			Namespace: namespace,
+		},
+	}
+
+	if _, err := controllerutil.CreateOrUpdate(ctx, serverClient, secret, func() error {
+		if smtp == nil {
+			return fmt.Errorf("nill pointer passed for smtp details")
```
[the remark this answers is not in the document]
Boomatang: Yes the link you shared is checking for the nil value. Currently this function is only used in the one place. In the future if the function is reused else where the smtp value maybe nil, that is why its being checked.

pkg/resources/custom-smtp/custom-smtp.go
Boomatang: Yap that should be PhaseFailed

pkg/resources/custom-smtp/custom-smtp.go
Boomatang: While change as there is the precedent.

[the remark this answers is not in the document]
Boomatang: While change as there is the precedent.

controllers/rhmi/bootstrapReconciler.go
Boomatang: While change. The default will be to throw an error if there is an unknown validation type.

pkg/resources/custom-smtp/custom-smtp.go
laurafitzgerald: Yes, it's a code neatness item so wouldn't block the change for it.

Boomatang: I am going to be creating a few follow up tickets. This can be one and I think it could be a good first issue also.

controllers/rhmi/bootstrapReconciler.go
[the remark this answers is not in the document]
briangallagher: Sorry, I put my comment in the wrong place. I meant in that else {} block it should be set to false to cover the scenario where it is currently true but the customer has deleted all values in ocm, in order to default back to the RHOAM supplied SMTP creds. @Boomatang

pkg/resources/custom-smtp/custom-smtp.go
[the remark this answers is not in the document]
Boomatang: I guess failing fast is the better opinion. Moving the check

controllers/rhmi/bootstrapReconciler.go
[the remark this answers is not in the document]
Boomatang: When the customer removes all the values from ocm the custom smtp status block gets zero out in the rhmi cr. As it will not exist there is no need to set it to false.

akostadinov:
One thing to point out is using TLS which requires proper CA trust settings. Locally on Fedora I see that I have an automatically set `ca_file` setting. But on a server I had an issue that neither `ca_file` nor `ca_path` was set and that had to be set in smtp.yaml

I don't know how it would work in the container.

Also note that these two settings are **NOT** documented but they are necessary in case peer validation needs to work. And I assume it has to.

Also I think these settings shouldn't be exposed to customers for configuration. Rather need to be set by default to the default CA trust. In case customer needs a custom CA, they may be allowed to use a secret perhaps.

Boomatang:
@akostadinov Thanks for the note about the TLS. We have removed that form the configuration. It should now be using the default behaviour of 3scale.

config/secrets/custom-smtp-secret.yaml
briangallagher: @Boomatang Can you create a JIRA so to follow up on this

pkg/resources/custom-smtp/custom-smtp.go
Boomatang: Related Jira https://issues.redhat.com/browse/MGDAPI-3965

config/secrets/custom-smtp-secret.yaml
Boomatang: Related Jira: https://issues.redhat.com/browse/MGDAPI-3962

briangallagher:
Verified, Full, followed by Partial, No custom smtp and back to Full. 
All working as expected.

I did encounter 1 RBAC issue when deleting the secret, rhmi-operator did not have permission to delete the secret which I think is correct as there doesn't seem to be a secret/delete perm in rhmi_controller at the moment. 

I added the delete [here](https://github.com/Boomatang/integreatly-operator/blob/db84089e961ea6ea5a663c46dec9e536bfad331e/controllers/rhmi/rhmi_controller.go#L208) and it worked.

PR #2657, opened 2022-05-06
[WIP] MGDAPI-3452 C08B automated test

Stellar-9:
# Issue link
[MGDAPI-3452](https://issues.redhat.com/browse/MGDAPI-3452)

# What
create a new automated test and update the [test case](https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/alerts/c08b-verify-alert-links-to-sops.md) by adding an "automated" tag
Work in progress

# Verification steps


test/common/verify_alert_links_to_SPOs.go
```diff
@@ -0,0 +1,97 @@
...
+type prometheusAPIResponse struct {
+	Status    string                 `json:"status"`
+	Data      json.RawMessage        `json:"data"`
+	ErrorType prometheusv1.ErrorType `json:"errorType"`
+	Error     string                 `json:"error"`
+	Warnings  []string               `json:"warnings,omitempty"`
+}
+*/
+
+func TestVerifyAlertLinksInSPOSs(t TestingTB, ctx *TestingContext) {
```
trepel, line 22: typo, also similar typo in file name.

test/common/verify_alert_links_to_SPOs.go
```diff
@@ -0,0 +1,97 @@
...
+
+	var apiLinks []string
+
+	output, err := execToPod("wget -qO - localhost:9090/api/v1/rules",
+		"prometheus-prometheus-0",
+		ObservabilityProductNamespace,
+		"prometheus", ctx)
+	if err != nil {
+		t.Fatal("failed to exec to pod:", err)
```
trepel, line 31: No capital 'f' at the beginning. Does not really matter but later in the code you start with capital letters so for the sake of consistency use capital 'F' here

test/common/verify_alert_links_to_SPOs.go
```diff
@@ -0,0 +1,97 @@
...
+		t.Fatal("Failed to unmarshal json", err)
+	}
+
+	for _, group := range rulesResult.Groups {
+		for _, rule := range group.Rules {
+			switch v := rule.(type) {
+			case prometheusv1.RecordingRule:
+			case prometheusv1.AlertingRule:
+				for k, j := range v.Annotations {
+					if k == "sop_url" {
```
trepel, line 54: minor thing, but I don't like that 'v', 'j', 'k'. Could the names be more descriptive? Although it is used like that in other tests  as I can see.

test/common/verify_alert_links_to_SPOs.go
```diff
@@ -0,0 +1,97 @@
...
+						apiLinks = append(apiLinks, string(j))
+						url := modifyLinks(apiLinks)
+						client := &http.Client{}
+						req, err := http.NewRequest("GET", url, nil)
+						if err != nil {
+							t.Fatalf("Failed to create http request %s", err)
```
trepel, line 60: Again inconsistency, either use `t.Fatal` or `t.Fatalf` but not both in the same file.

test/common/verify_alert_links_to_SPOs.go
```diff
@@ -0,0 +1,97 @@
...
+						}
+
+						req.Header.Add("Accept", `application/json`)
+						req.Header.Add("Authorization", fmt.Sprintf("token %s", os.Getenv("GITHUB_TOKEN")))
```
trepel, line 64: What if the GITHUB_TOKEN is undefined?

test/common/verify_alert_links_to_SPOs.go
```diff
@@ -0,0 +1,97 @@
...
+						resp, err := client.Do(req)
+						if err != nil {
+							t.Fatalf("Failed to make http request %s", err)
+						}
+						if resp.StatusCode != 200 {
+							t.Fatalf("The status code is not 200 %s", err)
+						}
+						fmt.Printf("%s\n", url)
+						fmt.Printf("%d\n", resp.StatusCode)
```
trepel, line 73: Prints like this need to be removed or converted to log messages

test/common/verify_alert_links_to_SPOs.go
```diff
@@ -0,0 +1,97 @@
...
+						}
+
+						req.Header.Add("Accept", `application/json`)
+						req.Header.Add("Authorization", fmt.Sprintf("token %s", os.Getenv("GITHUB_TOKEN")))
+						resp, err := client.Do(req)
+						if err != nil {
+							t.Fatalf("Failed to make http request %s", err)
+						}
+						if resp.StatusCode != 200 {
+							t.Fatalf("The status code is not 200 %s", err)
```
trepel, line 70: If the GITHUB_TOKEN is invalid the test would fail there? Would it continue to verify the other links of will it just fail with the very first link it couldn't access?

I think it would be good to test the validity of the token first and fail early with an error message saying that the token is invalid. And then it should loop through all the links and print out the non-accessible ones. At this point it will be known that the token is OK and that there is an issue with the links.

test/common/verify_alert_links_to_SPOs.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+}
+
+func modifyLinks(rawLinks []string) (url string) {
+	for _, link := range rawLinks {
+		if strings.Contains(link, "blob") {
+			url = strings.ReplaceAll(link, "github.com/RHCloudServices/integreatly-help/blob/master", "api.github.com/repos/RHCloudServices/integreatly-help/contents")
+		} else if strings.Contains(link, "tree") {
+			url = strings.ReplaceAll(link, "github.com/RHCloudServices/integreatly-help/tree/master", "api.github.com/repos/RHCloudServices/integreatly-help/contents")
```
trepel, line 92: Given the fact that the `to-be-replaced-string`s are so similar and the replacement string (api.github.com/repos/RHCloudServices/integreatly-help/contents) is actually the same I think the replacement using regular expression would be better here. It would eliminate the need for the `if` statement.

test/common/verify_alert_links_to_SPOs.go
```diff
@@ -0,0 +1,97 @@
...
+				t.Log("Unknown rule type %w", v)
+
+			}
+		}
+	}
+
+}
+
+func modifyLinks(rawLinks []string) (url string) {
+	for _, link := range rawLinks {
```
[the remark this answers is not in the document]
trepel: Why would you do the replacement in all the links in `rawlinks` if you only return one of them in the end?

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,147 @@
...
+
+		if resp.StatusCode != 200 {
+			t.Log("Response status :", resp.Status)
+			t.Log(url)
+			countFailedLinks++
+		}
+
+		if countFailedLinks != 0 {
+			status = false
+		}
```
trepel, line 115: The whole `if` should be moved outside the `for` loop. Reason is that in `for` loop you just count the failed links but it is enough to check the `countFailedLinks` just once once the `for` loop is over.

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,147 @@
...
+	req.Header.Add("Accept", `application/json`)
+	req.Header.Add("Authorization", fmt.Sprintf("token %s", githubToken))
+	resp, err := client.Do(req)
+	if err != nil {
+		t.Log(err)
+	}
+
+	defer func(Body io.ReadCloser) {
+		err := Body.Close()
+		if err != nil {
```
trepel, line 140: I have to admit I don't really understand this whole `defer` piece but this empty `if` looks highly suspicious.

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,147 @@
...
+func modifyLink(rawLink string) (url string) {
+	r := strings.NewReplacer(
+		"github", "api.github",
+		"com/", "com/repos/",
+		"blob/master", "contents",
+		"tree/master", "contents",
+	)
+
+	url = r.Replace(rawLink)
+	return url
```
trepel, line 86: I think you need just `return` here. The reason is that you specified that you want to return `url` in the function declaration (line 77). Alternatively just use `func modifyLink(rawLink string) string {` and keep the `return url` as is. See e.g. https://www.geeksforgeeks.org/named-return-parameters-in-golang/?ref=lbp

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,147 @@
...
+	"encoding/json"
+	"fmt"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"io"
+	"net/http"
+	"os"
+	"strings"
+)
+
+var githubToken = os.Getenv("TOKEN_GITHUB")
```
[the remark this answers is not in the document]
Stellar-9: Sure no problem, but can you please explain further, why exactly it would be better? I got inspiration from another test cases, where credentials were declared as global variables

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,147 @@
...
+
+func TestVerifyAlertLinksInSOPs(t TestingTB, ctx *TestingContext) {
+
+	if githubToken == "" {
+		t.Skip("github token not present")
+	}
+
+	var apiLinks []string
+	testUrl := "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/README.md"
+	testResp := getSOPAlertLinkStatus(t, testUrl)
```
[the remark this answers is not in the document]
Stellar-9: -     alert link to SOP (file name)
      -  This is because test case is called like that here [](https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/alerts/c08b-verify-alert-links-to-sops.md). So I suppose it's right name for the file?
-     alert link in SOP (this function name)
      - Changed for "in" to "to"
-     api link (name of slice of strings)
      - Changed to sopUrls
-     sop Url (later in the code around line 60)
-     sop link (from function name validateSopLink)
      - Changed to validateSOPUrls
-     raw link (input param name to modifyLink)
      - Changed to sopUrl
-     url (if getting the link from slice later in the code)
      - Changed to sopUrl

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,147 @@
...
+	"os"
+	"strings"
+)
+
+var githubToken = os.Getenv("TOKEN_GITHUB")
+
+func TestVerifyAlertLinksInSOPs(t TestingTB, ctx *TestingContext) {
+
+	if githubToken == "" {
+		t.Skip("github token not present")
```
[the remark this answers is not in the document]
Stellar-9: So this is basically checking if inside enviroment variable is some value. So if it's not, then user cant use it?

test/common/verify_alert_links_to_SOPs.go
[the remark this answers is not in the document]
trepel: Fair question. In general the variable should only be visible where it is needed. If you need it in just one function then it should be declared in that function and not globally. If you need it in many functions then it is better to have it declared/defined globally so that you don't have to pass it into that many functions via input parameter. And then there is a grey zone in between.

In this particular case you need it in two functions, it is not that many so you can just declare/define it in one and pass via input parameter to the other. However, it is fairly ok as is too, it mostly depends on personal taste.

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,147 @@
...
+		if _, ok := inResult[str]; !ok {
+			inResult[str] = true
+			result = append(result, str)
+		}
+	}
+	return result
+}
+
+func validateSOPLinks(t TestingTB, s []string) bool {
+	status := true
```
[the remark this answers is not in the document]
Stellar-9: Good point. Thank you

test/common/verify_alert_links_to_SOPs.go
Stellar-9: Good point. Thank you

test/common/verify_alert_links_to_SOPs.go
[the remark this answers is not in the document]
trepel: Yes, but you want to tell user what to do so that this test is not skipped. In other words you need to tell user how to specify the token.

test/common/verify_alert_links_to_SOPs.go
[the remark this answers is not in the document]
trepel: There's no right name. If the name is descriptive enough then it is just fine. The key thing is to use just one name for one thing across the test case code base.

test/common/verify_alert_links_to_SOPs.go
trepel: Yes, something similar is used in other places across the code so it should be fine. But yes, add something into that if, empty `if` is definitely not a good practise.

test/common/verify_alert_links_to_SOPs.go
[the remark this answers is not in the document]
trepel: c08b-verify-alert-links-to-sops.md -> it does not mean that you have to call it alertLinkToSOP in this test. Feel free to rename it to `sopUrl` everywhere, including the file name and function name. For the sake of consistency.

test/common/tests.go
```diff
@@ -63,6 +63,7 @@ var (
 				/*FLAKY on RHMI*/ {"C01 - Verify Alerts are not pending or firing apart from DeadMansSwitch", TestIntegreatlyAlertsPendingOrFiring},
 				{"C04 - Verify Alerts exist", TestIntegreatlyAlertsExist},
 				{"C10B - Verify Prometheus blackbox targets", TestAdditionalBlackboxTargets},
+				{"C08B - Verify alert links to SOPs", TestVerifyAlertLinksToSOPs},
```
trepel, line 66:
```suggestion
				{"C08B - Verify alert links to SOPs", TestSopUrls},
```

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,146 @@
...
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"io"
+	"net/http"
+	"os"
+	"strings"
+)
+
+var githubToken = os.Getenv("TOKEN_GITHUB")
+
+func TestVerifyAlertLinksToSOPs(t TestingTB, ctx *TestingContext) {
```
trepel, line 15:
```suggestion
func TestSopUrls(t TestingTB, ctx *TestingContext) {
```

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,146 @@
...
+
+	if githubToken == "" {
+		t.Skip("github token not present, use TOKEN_GITHUB environment variable to specify it")
```
trepel, line 18:
```suggestion
		t.Skip("Github token not provided, use GITHUB_TOKEN environment variable to specify it")
```

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,146 @@
...
+
+			default:
+				fmt.Printf("Unknown rule type %s", v)
+
+			}
+		}
+	}
+
+	sopUrls = unique(sopUrls)
+	result := validateSOPUrls(t, sopUrls)
```
trepel, line 71:
```suggestion
	result := validateSopUrls(t, sopUrls)
```

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,146 @@
...
+
+func modifySOPUrl(sopUrl string) (apiSOPUrl string) {
+	r := strings.NewReplacer(
+		"github", "api.github",
+		"com/", "com/repos/",
+		"blob/master", "contents",
+		"tree/master", "contents",
+	)
+
+	apiSOPUrl = r.Replace(sopUrl)
```
trepel, line 83:
```suggestion
	apiSopUrl = r.Replace(sopUrl)
```

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,146 @@
...
+			}
+		}
+	}
+
+	sopUrls = unique(sopUrls)
+	result := validateSOPUrls(t, sopUrls)
+	fmt.Println(result)
+}
+
+func modifySOPUrl(sopUrl string) (apiSOPUrl string) {
```
trepel, line 75:
```suggestion
func modifySopUrl(sopUrl string) (apiSopUrl string) {
```

PR #2666, opened 2022-05-13
MGDAPI-3743 - Update kc-client

adam-cattermole:
# Issue link
[MGDAPI-3743](https://issues.redhat.com/browse/MGDAPI-3743)

# What
Update kc-client to v0.1.8

# Verification steps
It should be sufficient that e2e tests pass, possibly worth install / uninstall from this branch also.


MStokluska:
I've verified that the installation works, I do not think there's a need for running an upgrade scenario.
Code changes make sense too.
Good job!
/lgtm

KevFan:
@adam-cattermole Looks like a rebase is required :technologist:

adam-cattermole:
Thanks Kev didn't spot this

PR #2677, opened 2022-05-23
MGDAPI-3903 Autoscaling RHMI CR Flag

bryankeane0:
# Issue link
[MGDAPI-3903](https://issues.redhat.com/browse/MGDAPI-3903)

# What
- Added flag to the RHMI cr `AutoscalingEnabled`.
- Added logic to `bootstrapReconciler.go` to reconcile quota accordingly
- Added config map creation with HPA resource limits.

# Verification steps
1. Provision a cluster large enough for a 50 Million quota install
2. Quota can be set to 20 mil or 50 mil for testing but ensure the cluster is large enough to handle a 50 mil install
3. Prepare your cluster as normal and then set quota with `INSTALLATION_TYPE=managed-api DEV_QUOTA="500" make cluster/prepare/quota`
4. Run the rhoam install and wait for the stage `complete`
5. Add the flag to the RHMI CR `autoscalingEnabled: true`, wait for bootstrap to reconcile and wait for quota to change to `50 Million` and stage `complete`. Set the flag to `false` or remove it entirely and wait for the quote to be reverted.
6. Run the test with ` INSTALLATION_TYPE=managed-api TEST="F10" make test/e2e/single` and make sure it passes
7. Set the flag `autoscalingEnabled` to `true` and rerun the test ensuring it passes 
>The flag may not be present at all so you may have to manually enter it into the spec

### DIFF
--- a/apis/v1alpha1/rhmi_types.go
+++ b/apis/v1alpha1/rhmi_types.go
@@ -198,6 +198,7 @@ type RHMISpec struct {
 	//
 	// url
 	DeadMansSnitchSecret string `json:"deadMansSnitchSecret,omitempty"`
+	AutoscalingEnabled   bool   `json:"autoscalingEnabled,omitempty"`
 }
 
 type PullSecretSpec struct {
--- a/controllers/rhmi/bootstrapReconciler.go
+++ b/controllers/rhmi/bootstrapReconciler.go
@@ -196,6 +196,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		}
 		metrics.SetQuota(installation.Status.Quota, installation.Status.ToQuota)
 
+		// Create autoscaling config map
+		phase, err = r.createAutoscalingConfigMap(ctx, serverClient)
+		if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+			events.HandleError(r.recorder, installation, phase, "Failed to create autoscaling config map", err)
+			return phase, errors.Wrap(err, "Failed to create autoscaling config map")
+		}
+
 		// temp code for RHOAM, remove once all clusters are upgraded to 1.14
 		// Remove all prometheus rules under redhat/sandbox-rhoam/rhoami-operator
 		phase, err = r.removePrometheusRules(ctx, serverClient, installation.Spec.NamespacePrefix)
@@ -740,11 +747,14 @@ func generateSecret(length int) string {
 
 func (r *Reconciler) processQuota(installation *rhmiv1alpha1.RHMI, namespace string,
 	installationQuota *quota.Quota, serverClient k8sclient.Client) error {
+	var err error
+	quotaParam := "500"
 	isQuotaUpdated := false
 
-	quotaParam, err := getSecretQuotaParam(installation, serverClient, namespace)
-	if err != nil {
-		return err
+	if installation.Spec.AutoscalingEnabled != true {
+		if quotaParam, err = getSecretQuotaParam(installation, serverClient, namespace); err != nil {
+			return err
+		}
 	}
 
 	// get the quota config map from the cluster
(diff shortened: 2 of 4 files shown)


apis/v1alpha1/rhmi_types.go
```diff
@@ -198,6 +198,7 @@ type RHMISpec struct {
 	//
 	// url
 	DeadMansSnitchSecret string `json:"deadMansSnitchSecret,omitempty"`
+	AutoscalingEnabled   bool   `json:"autoscalingEnabled,omitempty"`
```
MStokluska, line 201: hey @bryankeane0 
When you are updating the types you will need to also run the:
```make pkg/apis/integreatly/v1alpha1/zz_generated.openapi.go```
in order to update the CRDs. Otherwise this field will not be supported in rhmi CR on the cluster.

Comment (bryankeane0) [reply apis/v1alpha1/rhmi_types.go:201]: Thank you, resolved this in latest commit

File: test/common/autoscaling.go [hunk 2]
```diff
@@ -0,0 +1,98 @@
...
+	quotaConfig, err := getQuotaConfig(t, ctx.Client)
+	if err != nil {
+		t.Fatalf("Error retrieving Quota: %v", err)
+	}
+
+	installation, err := GetRHMI(ctx.Client, true)
+	if err != nil {
+		t.Fatal("couldn't get RHMI cr for quota test")
+	}
+	if installation == nil {
```
MStokluska, line 27: hey @bryankeane0 I don't think we need this check. It should be covered by getRHMI

test/common/autoscaling.go
```diff
@@ -0,0 +1,98 @@
...
+		t.Fatalf("Got invalid rhmi CR: %v", installation)
+	}
+	initialHPAValue := installation.Spec.AutoscalingEnabled
+
+	// wait if stage no complete
+	startTime := time.Now()
+	endTime := startTime.Add(time.Minute * time.Duration(timeoutWaitingQuotachange))
+
+	for startTime.Before(endTime) {
```
MStokluska, line 36: This is not required. The check if installation is completed is already done here:
https://github.com/integr8ly/integreatly-operator/blob/master/test/functional/integreatly_test.go#L24 when it comes to functionals and here: https://github.com/integr8ly/integreatly-operator/blob/master/test/e2e/integretly_test.go#L24 when it comes to prow e2e

test/common/autoscaling.go
```diff
@@ -0,0 +1,98 @@
...
+			break
+		}
+	}
+
+	//verify that the TOQuota value is set and that Quota is not set
+	//assuming this is run after installation
+	if installation.Status.Quota == "" {
+		t.Fatal("Expected '.status.quota' to contain a quota name after installation, but got an empty string")
+	}
+	if installation.Status.ToQuota != "" {
```
MStokluska, line 48: I'm not sure why are we checking this. 
In general, before this test will run you are guaranteed to have the following:
1. Installation fully completed
2. Therefore, quota present and toQuota not present.

test/common/autoscaling.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// If hpa flag is false - set it to true
+	if initialHPAValue != true {
+		result, err := changeAutoscalingFlag(t, ctx.Client, installation, true)
+		if err != nil {
+			t.Fatalf("Error updating RHMI CR: %v with results of: %v", err, result)
+		}
+	}
+
+	// Check that quota has been changed
```
MStokluska, line 72: In general I think the best flow of this test would be to:
1. Get initial quota
2. Set the autoscaling flag to true and expect quota to change to 50mil
3. Set the autoscaling flag to false and expect quota to bump back to whatever it was

However, we have to keep in mind that this might be a bit too heavy test to be listed as the happy path test as the quota bump will fail on prow due to lack of resources.

I would possibly introduce a quick check at the start of the test "inProw" (you can re-use existing function) and if it is in prow, skip test, otherwise, continue with test. 
I would also consider renaming the test to something else as we are not testing autoscaling here but the autoscaling flag.

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -727,11 +727,15 @@ func generateSecret(length int) string {
 
 func (r *Reconciler) processQuota(installation *rhmiv1alpha1.RHMI, namespace string,
 	installationQuota *quota.Quota, serverClient k8sclient.Client) error {
+	var err error
```
MStokluska, line 730: As agreed I think we can update the logic here to also create the initial config map here that will list the components and their target utilization. 
Also, I think some unit tests around this logic and cm map creation/no reconciliation on 1+ runs would be really good.

bryankeane0: Added this in the recent commit :)

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -858,3 +868,39 @@ func getSecretQuotaParam(installation *rhmiv1alpha1.RHMI, serverClient k8sclient
...
+			autoscalingConfig.Data = map[string]string{}
+		}
+
+		autoscalingConfig.Data["backend-listener"] = "40"
+		autoscalingConfig.Data["backend-worker"] = "60"
+		autoscalingConfig.Data["apicast-production"] = "40"
+		autoscalingConfig.Data["ratelimit"] = "40"
+		autoscalingConfig.Data["usersso"] = "40"
+
+		err := serverClient.Create(ctx, autoscalingConfig)
```
Boomatang, line 897: We should be using a createOrUpdate function. I seen one test case is checking that the values don't get reconciled. What is the reason for wanting this behaviour?

controllers/rhmi/bootstrapReconciler_test.go
```diff
@@ -247,6 +247,121 @@ func TestReconciler_reconcilePrometheusRules(t *testing.T) {
...
+			Recorder: record.NewFakeRecorder(50),
+			InitialObjs: []runtime.Object{
+				&corev1.ConfigMap{},
+			},
+			Assertion: assertConfigMapCreated,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.Name, func(t *testing.T) {
+			reconciler, err := NewBootstrapReconciler(tt.FakeConfig, tt.Installation, tt.FakeMPM, tt.Recorder, l.NewLogger())
```
Boomatang, line 319: There should be a defer action here to set the cluster back to the state it was in before this test started. If not these test could have a side effect on other tests.

bryankeane0:
Prow tests will fail as the cluster for prow is too small right now, it won't fail when we implement autoscaling properly as it wont try to provision a full 50 Million sized cluster

Boomatang (approved):
From the conversations I have had with Bryan all my concerns have been addressed.

MStokluska:
As things changed slightly yesterday, this PR might need to be reverted in the future, however, it will simplify things on our feature branch so happy for now for this to go in. Thanks @bryankeane0 Good job!

PR #2678, opened 2022-05-23
Mgdapi 3964 - Custom SMTP E2E tests

Boomatang:
# Issue link

https://issues.redhat.com/browse/MGDAPI-3964

# What

Add e2e tests based around the Custom SMTP feature.
# Notes
* IDP test are not ran as part of `rhoam-e2e` in prow. Manual running of the test is required.
* As hive manages on of the secrets used in this suite these test are will run depending on the install base.
* H11 and H34 have some dependence on each other. One will fail if the other is ran. Issue is still being looked into. These two test will only auto run when installed via master install.
# Verification steps

## Master install testing
* install cluster with this branch (setting up testing IDP manual is optional but will speed the test runs up)
* Run the three test case independently
  *  `INSTALLATION_TYPE=managed-api TEST=H11 make test/e2e/single`
  * `INSTALLATION_TYPE=managed-api TEST=H34 make test/e2e/single`
  * `INSTALLATION_TYPE=managed-api TEST=H35 make test/e2e/single`
* Expected: all test should pass

## Addon install testing
* install the current addon without adding custom SMTP values and set up the testing IDP
### State 1: No custom SMTP configuration
* Run the three test case independently
  *  `INSTALLATION_TYPE=managed-api TEST=H11 make test/e2e/single`
  * `INSTALLATION_TYPE=managed-api TEST=H34 make test/e2e/single`
  * `INSTALLATION_TYPE=managed-api TEST=H35 make test/e2e/single`
 * Expected:
   * H11 should pass
   * H34 and H35 should be skipped with warning of invalid configuration
### State 2: Partial custom SMTP configuration
* In the OCM addon menu partial fill in the custom SMTP values, leave at least one field missing
* Run the three test case independently
  *  `INSTALLATION_TYPE=managed-api TEST=H11 make test/e2e/single`
  * `INSTALLATION_TYPE=managed-api TEST=H34 make test/e2e/single`
  * `INSTALLATION_TYPE=managed-api TEST=H35 make test/e2e/single`
 * Expected:
   * H35 should pass
   * H34 and H11 should be skipped with warning of invalid configuration
### State 2: Fully custom SMTP configuration
* In the OCM addon menu fill in the custom SMTP values, copy values from the existing rhoam-smtp secret to get valid values
* Run the three test case independently
  *  `INSTALLATION_TYPE=managed-api TEST=H11 make test/e2e/single`
  * `INSTALLATION_TYPE=managed-api TEST=H34 make test/e2e/single`
  * `INSTALLATION_TYPE=managed-api TEST=H35 make test/e2e/single`
 * Expected:
   * H34 should pass
   * H35 and H11 should be skipped with warning of invalid configuration

### DIFF
--- a/test/utils/utils.go
+++ b/test/utils/utils.go
@@ -2,13 +2,7 @@ package utils
 
 import (
 	"fmt"
-	"github.com/integr8ly/integreatly-operator/apis/v1alpha1"
-	"github.com/sirupsen/logrus"
-	"golang.org/x/net/context"
-	"k8s.io/api/core/v1"
-	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"os"
-	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // JUnitFileName Allow adding a prefix into the junit file name.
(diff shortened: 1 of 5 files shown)


test/utils/utils.go
```diff
@@ -24,3 +30,27 @@ func SpecDescription(spec string) string {
 	}
 	return spec
 }
+
+//OperatorIsHiveManaged Returns true if addon is manged by hive
+func OperatorIsHiveManaged(client k8sclient.Client, inst *v1alpha1.RHMI) (bool, error) {
```
MStokluska, line 35: hey @Boomatang can we use the existing function for this? https://github.com/integr8ly/integreatly-operator/blob/master/pkg/addon/operator_run.go#L126

Boomatang: While that is something that could be done it is not a good practices to depend on the code that is under test as part of a black box test. Testing code and production code should have a degree of separation.

MStokluska: While I would agree to some degree, I think this is different. 
The function is very simple and does not require to be modified for the purpose of this test, additionally, if we would ever have to change the way of figuring out if a cluster is hive managed or not, this change would require updates in 2 places. I think we should avoid duplication in our code as much as we can.

PR #2679, opened 2022-05-23
MGDAPI-4021 refactor addon parameters secret retrieval

makslion:
# Issue link
[MGDAPI-4021](https://issues.redhat.com/browse/MGDAPI-4021)

# What
In the case of the internal installation, that secret would have a different name. I'm refactoring the way we retrieve a secret. The operator will now be looking for a RHOAM subscription and will use its name as a name for the secret. In case subscription is not found, it would default to const `addons-managed-api-service-parameters`

Currently we have the following combinations of subscription / secret (they are all covered in unit tests):
1. addon-managed-api-service | addon-managed-api-service-parameters
2. managed-api-service |  addon-managed-api-service-parameters
3. managed-api-service-internal |  addon-managed-api-service-insternal-parameters
4. none |  addon-managed-api-service-parameters
# Verification steps
There are two cases to verify: 
### Addon installation \ OLM
The secret is not deployed by us and the operator namespace along with the RHOAM subscription. The only way to check it would be to trick the operator into thinking it was installed by an addon flow: 
1. Deploy RHOAM via OLM 
5. Deploy RHMI CR and ensure that the bootstrap stage passed. 

### The rest of the installations. 
E2E tests passing would cover it (prow installation will not have subscription).


pkg/addon/parameters.go
```diff
@@ -43,26 +44,31 @@ func GetParameterByInstallation(ctx context.Context, client k8sclient.Client, in
...
-
-	secret := &corev1.Secret{}
-	if err := client.Get(ctx, k8sclient.ObjectKey{
-		Name:      secretName,
-		Namespace: namespace,
-	}, secret); err != nil {
+	secrets := &corev1.SecretList{}
+	opts := []k8sclient.ListOption{
+		k8sclient.InNamespace(namespace),
+	}
```
Boomatang, line 53: We should add some filtering on labels here to reduce the number of secrets we need to loop over.

Boomatang, line 53: We should add some filtering on labels here to reduce the number of secrets we need to loop over.

pkg/addon/parameters.go
```diff
@@ -43,26 +44,31 @@ func GetParameterByInstallation(ctx context.Context, client k8sclient.Client, in
...
-
 		return nil, false, fmt.Errorf("failed to retrieve parameters secret: %v", err)
 	}
 
-	value, ok := secret.Data[parameter]
-	return value, ok, nil
+	for _, secret := range secrets.Items{
+		matched, err := regexp.MatchString("^addon-.*-parameters$", secret.Name)
+		if err != nil{
+			return nil, false, err
```
Boomatang, line 64: As we spoke about. The getting of the secret should be in its own function call. This function can be used with this function and replace the code [here](https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/custom-smtp/custom-smtp.go#L34-L41)

pkg/addon/parameters.go
```diff
@@ -43,26 +44,31 @@ func GetParameterByInstallation(ctx context.Context, client k8sclient.Client, in
...
 		if errors.IsNotFound(err) {
 			return nil, false, nil
 		}
-
 		return nil, false, fmt.Errorf("failed to retrieve parameters secret: %v", err)
 	}
 
-	value, ok := secret.Data[parameter]
-	return value, ok, nil
+	for _, secret := range secrets.Items{
```
Boomatang, line 61: This style of for loops will fail gosec checks at a later date. Issue is here the `secret` not a memory reference to the value in `secrets.Items`. The problem is more of an issue with go routines but as it is easy to fix now, why not do it. The issue caused is "Implicit memory aliasing in for loop"

pkg/addon/parameters.go
```diff
@@ -43,26 +44,31 @@ func GetParameterByInstallation(ctx context.Context, client k8sclient.Client, in
 	)
 }
 
-// GetParameterByInstallType retrieves the value for an addon parameter by
-// selecting the addon name for installationType
+// GetParameterByInstallType retrieves the value for an addon parameter from
+// the addon parameters secret
 func GetParameterByInstallType(ctx context.Context, client k8sclient.Client, installationType integreatlyv1alpha1.InstallationType, namespace, parameter string) ([]byte, bool, error) {
-	addonName := GetName(installationType)
```
Boomatang, line 49: I think this `GetName` function can now be removed from the code base. It does not provide a reliable way of getting the name of instance installed by the operator.

Boomatang (changes requested):
There are a few changes that I would like to see. I think the inclusion of some unit testing around the getSecret function that one of my comments is suggesting would be a good thing.

makslion:
Reworked according to @Boomatang comments and discussion with Laura. Should be fine now, I guess :thinking:

makslion:
fix for e2e tests and bundles install retesting 

/test rhoam-e2e

pkg/addon/parameters_test.go
```diff
@@ -61,21 +85,39 @@ func TestGetParameter(t *testing.T) {
...
+				ObjectMeta: v1.ObjectMeta{
+					Name:      addonName,
+					Namespace: testOperatorNamespace,
+				},
+			}
+			if scenario.SecretFromEnvar == false {
+				namespace.Labels = labels
+				subscription.Labels = labels
+			} else {
+				os.Setenv("DEFAULT_ADDON_PARAMETERS_SECRET_NAME", "addon-managed-api-service-parameters")
```
Boomatang, line 103: This worries me. In the test you are setting the a system envar. What happens when I have the operator running locally with a different envar? Also take into account I auto run the unit tests on every change I make with the code base.

Boomatang (commented):
I have one concern over setting a system env during the unit tests.

pkg/addon/parameters_test.go
```diff
@@ -61,21 +85,39 @@ func TestGetParameter(t *testing.T) {
...
+				ObjectMeta: v1.ObjectMeta{
+					Name: testOperatorNamespace,
+				},
+			}
+			subscription := &v1alpha1.Subscription{
+				ObjectMeta: v1.ObjectMeta{
+					Name:      addonName,
+					Namespace: testOperatorNamespace,
+				},
+			}
```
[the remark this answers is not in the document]
makslion: We could setup client and init objects before this moment, but It would make it awkward to modify those init objects depending on the scenario we want to test :thinking: 
As for the `rhoami` - the only difference would be an operator namespace here. As far as I know, we have Internal RHOAM only installed as an add-on. In this case, it will find the secret using the same logic as if it would be RHOAM, so there is no need for an additional scenario. WDYT?

pkg/addon/parameters_test.go
makslion: If you will be running the operator locally with a different envar it would be overridden. Override would be valid as long as you don't update [this](https://github.com/integr8ly/integreatly-operator/blob/master/config/secrets/quota-secret.yaml#L9) value. If you do update it and specify a different envar you would expect that this `os.Setenv` should be also updated :thinking: 
This test should be fine if you run units on every change as long as nothing related to the addon parameters secret gets changed. 
Would you say it might be better to check the existence of this envar before setting it?

pkg/addon/parameters_test.go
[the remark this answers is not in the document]
Boomatang: If it makes it awkward to to modify these objects when moved I think it's ok to leave them here.

As for a scenario that is testing rhoami. Yes, we should have one. This has caused issues before. It is more about proving works and that it keeps working.

pkg/addon/parameters_test.go
Boomatang: I am unsure to what I think the best approach is here. I think a second opinion on the matter is needed. The idea of a unit test make changes to my system does not sit well with me. When the use of envars don't seem right to me.

austincunningham: Agree that setting a local env var on local system is undesirable.

> Would you say it might be better to check the existence of this envar before setting it?
yes sounds good.

config/manager/manager.yaml
```diff
@@ -61,4 +61,6 @@ spec:
             value: "default@test.com"
           - name: QUOTA
             value: "200"
+          - name: DEFAULT_ADDON_PARAMETERS_SECRET_NAME
+            value: "addon-managed-api-service-parameters"
```
MStokluska, line 65: @makslion I wonder would it not be better to expose this envar via subscription config in MTR? 
By doing so, we would have good enough flexibility to set this up as `addon-managed-api-service-parameters` for rhoam and `addon-managed-api-service-internal-parameters` for edge evn.

makslion: this envar is needed for local and bundles install types - would we need it to be `addon-managed-api-service-internal-parameters`? :thinking: And by putting it in the MTR wouldn't we leave these instal types without a default value?

pkg/addon/parameters_test.go
makslion: added check for existence of envar

pkg/addon/parameters_test.go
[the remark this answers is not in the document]
makslion: now we will be testing for `redhat-rhoam-operator` namespace and I've added one extra case for `rhoami` namespace that would ask to extract parameters from the secret

makslion:
more releases to fix codecov 

/test rhoam-e2e

pkg/addon/parameters.go
```diff
@@ -5,74 +5,92 @@ import (
...
+		if len(subsciptions.Items) == 0 {
+			return nil, fmt.Errorf("haven't found subscription in %s namespace", namespace)
+		}
+		parametersSecretName = subsciptions.Items[0].Name + "-parameters"
+	}
+
+	return retrieveAddonParametersSecret(ctx, client, parametersSecretName, namespace)
+}
 
+func retrieveAddonParametersSecret(ctx context.Context, client k8sclient.Client, secretName, secretNamespace string) (*corev1.Secret, error) {
```
MStokluska, line 84: Hey Maks, I don't think we need to make this get a function on it's own.

pkg/addon/parameters.go
```diff
@@ -5,74 +5,92 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
 
-// GetParameterByInstallType retrieves the value for an addon parameter by
-// selecting the addon name for installationType
-func GetParameterByInstallType(ctx context.Context, client k8sclient.Client, installationType integreatlyv1alpha1.InstallationType, namespace, parameter string) ([]byte, bool, error) {
-	addonName := GetName(installationType)
-	secretName := fmt.Sprintf("addon-%s-parameters", addonName)
+	if isHiveManaged {
```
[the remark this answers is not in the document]
makslion: Following discussion. 
Now it would try to retrieve the subscription. If it finds one - use the name of the subscription as a name of a secret. If it fails to find the subscription it uses const `addon-managed-api-service-parameters`. 

Not, the fun with subs. We have 3 names of subscriptions: 
- `addon-managed-api-service` - addon. Secret name `addon-managed-api-service-parameters`
- `managed-api-service` - olm and sandbox. Secret name `addon-managed-api-service-parameters`
- `addon-managed-api-service-internal` - internal. Secret name `addon-managed-api-service-internal-parameters`
As you can see the code will now check for the `addon` at the beginning of the sub and if it doesn't have - the `addon` will be added. This approach also leaves room for the Sandbox people to update the function once they change the name for the MT subscription. 
This covers us for all installation types. I've also extended unit tests to properly simulate all 4 cases (3 subs and a local install)

makslion:
I need to investigate what is going on with those checks. Moving it back to `inProgress` for the time being :facepalm:

makslion:
Coverage fails because I'm reducing the size of the `parameter.go` and, this causes the reduction in a number of hits

cecobask (approved):
Code looks good. I installed RHOAM successfully via OLM after I created an image pointing to your code. The second scenario was matched: `managed-api-service | addon-managed-api-service-parameters`.
preflightStatus: successful
bootstrap: completed

/lgtm

pkg/addon/parameters.go
```diff
@@ -3,76 +3,72 @@ package addon
...
-	"k8s.io/apimachinery/pkg/api/errors"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
 	QuotaParamName      = "addon-managed-api-service"
 	TrialQuotaParamName = "trial-quota"
+	ManagedAPIService   = "managed-api-service"
+	RHMI                = "rhmi"
+	DefaultSecretName   = "addon-managed-api-service-parameters"
```
MStokluska, line 21: hey Maks, any reason for exporting this const?

pkg/addon/parameters.go
```diff
@@ -3,76 +3,72 @@ package addon
...
+
+	opts := &k8sclient.ListOptions{
+		Namespace: namespace,
+	}
+	subsciptions := &v1alpha1.SubscriptionList{}
+	if err := client.List(ctx, subsciptions, opts); err != nil {
+		return nil, err
+	}
+	if len(subsciptions.Items) > 1 {
+		return nil, fmt.Errorf("recieved %d subscriptions in %s namespace. Expected one", len(subsciptions.Items), namespace)
```
MStokluska, line 52:
```suggestion
		return nil, fmt.Errorf("received %d subscriptions in %s namespace. Expected one", len(subsciptions.Items), namespace)
```

MStokluska:
@makslion looks good to me, left a couple of comments but other than that it looks good! 
Good job!

pkg/addon/parameters.go
makslion: Right now no, but there is only one more place we are using those and if anywhere in `pkg` we would need to use them we will have an import cycle. Made them public for future usage. Could go with private if you have a strong opinion 🤔

MStokluska: If you decide to fix typos I would update this value too.

I understand that thinking of supporting future RHOAM versions might be good to have, but, have you ever heard 
`Don't cross the bridge till you come to it` sentence?

PR #2682, opened 2022-05-24
MGDAPI-4022 add a check to bootstrap stage for addon-managed-api-service-parameters secret

austincunningham:
# Issue link

# What

# Verification steps


pkg/config/manager.go
```diff
@@ -158,6 +159,10 @@ func (m *Manager) GetOauthClientsSecretName() string {
 	return "oauth-client-secrets"
 }
 
+func (m *Manager) GetAddonManagedApiServiceParametersSecretName() string {
+	return "addon-managed-api-service-parameters"
```
MStokluska, line 163: hey @austincunningham any ways of not making this hardcoded? 
on edge the name of the secret is addon-managed-api-service-internal-parameters

austincunningham: waiting for @makslion PR to merge https://github.com/integr8ly/integreatly-operator/pull/2679

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -588,6 +595,14 @@ func (r *Reconciler) reconcileOauthSecrets(ctx context.Context, serverClient k8s
 	return integreatlyv1alpha1.PhaseCompleted, nil
 }
 
+func (r *Reconciler) reconcileAddonManagedApiServiceParameters(ctx context.Context, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
+	_, err := addon.GetAddonParametersSecret(ctx, serverClient, r.ConfigManager.GetOperatorNamespace())
+	if k8serr.IsNotFound(err) && err != nil {
```
makslion, line 600: Is it just me or there is something weird here. So we want to fail when there is an error and this is a `not found` error. Why not just fail on `not found`? :thinking: or on any error?

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -129,6 +130,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, errors.Wrap(err, "failed to reconcile github oauth secrets")
 	}
 
+	phase, err = r.reconcileAddonManagedApiServiceParameters(ctx, serverClient)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, "Failed to reconcile addon-managed-api-service-parameters", err)
```
makslion, line 135: this message might be misleading imo - the secret could be also called `addon-managed-api-service-internal-parameters` and there is a chance that it will be also different for MT RHOAM in the future. Can we just say that we `failed to reconcile the addon parameters secret`? :thinking:

austincunningham: guess I want to cover the only use case that CS-SRE will care about customer clusters, the SOP covers if its a different addon name https://github.com/RHCloudServices/integreatly-help/pull/828/files#diff-a062bd65551cf4a6220379a4d0f187ace6396aacf1711372005cfcf277086a16R25 guess there is no harm in changing it.

makslion:
went through verification steps - all as expected. Code changes look good as well. Once tests will be happy I'll drop `lgtm`

makslion:
Let's try once again 

/lgtm

PR #2687, opened 2022-05-25
MGDAPI-3979 - add 3scale index build script

adam-cattermole:
# Issue link
[MGDAPI-3979](https://issues.redhat.com/browse/MGDAPI-3979)

# What
- Added new bundles.yaml containing a list of all the 3scale bundle images used to build an index
- New script create-3scale-index which uses the bundle file and parameters to create an index

# Verification steps
- Verify indexes can be successfully created by the script

note: I think it is best to verify the install / upgrade of these indices in the separate PR that changes the installation mode to use these indices.

~~[WIP] until referenced bundles exist.~~

### DIFF
--- a/scripts/create-3scale-index.sh
+++ b/scripts/create-3scale-index.sh
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+#
+# Prereq:
+# - opm
+# - yq (4.24.2+)
+#
+# Function:
+# Script creates 3scale index/indices for RHOAM 
+#
+# Usage:
+# REG=<REGISTRY> ORG=<QUAY ORG> IMAGE=<IMAGE NAME> BUILD_TOOL=<docker|podman>  VERSION=<VERSION TAG> make create/3scale/index
+# REG - registry of where to push the bundles and indices, defaults to quay.io
+# ORG - organization of where to push the bundles and indices
+# IMAGE - image name of the image to push
+# BUILD_TOOL - tool used for building the bundle and index, defaults to docker
+# VERSION - index version to build; this defines the max version and includes all bundles beneath it
+# TAG - version tag to be used for the image
+#
+# Example:
+# VERSION=0.8.3+0.1645735250.p ORG=acatterm make create/3scale/index
+
+
+REG="${REG:-quay.io}"
+ORG="${ORG:-integreatly}"
+IMAGE="${IMAGE:-3scale-index}"
+BUILD_TOOL="${BUILD_TOOL:-docker}"
+VERSION="${VERSION}"
+
+
+generate_index() {
+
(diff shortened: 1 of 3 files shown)


scripts/create-3scale-index.sh
```diff
@@ -0,0 +1,63 @@
...
+#
+# Example:
+# VERSION=0.8.1 ORG=acatterm ./scripts/create-3scale-index.sh
+
+
+REG="${REG:-quay.io}"
+ORG="${ORG:-integreatly}"
+IMAGE="${IMAGE:-3scale-index}"
+BUILD_TOOL="${BUILD_TOOL:-docker}"
+VERSION="${VERSION:-0.8.3+0.1649688682.p}"
```
MStokluska, line 27: hey @adam-cattermole I wonder would it make more sense to not default version to anything, and check if version is present - if not fail early? or is the plan to keep bumping the default version everytime there's new version available?

scripts/create-3scale-index.sh
```diff
@@ -0,0 +1,63 @@
...
+# REG=<REGISTRY> ORG=<QUAY ORG> IMAGE=<IMAGE NAME> BUILD_TOOL=<docker|podman>  VERSION=<VERSION TAG> ./scripts/create-3scale-index.sh
+# REG - registry of where to push the bundles and indexes, defaults to quay.io
+# ORG - organization of where to push the bundles and indexes
+# IMAGE - image name of the image to push
+# BUILD_TOOL - tool used for building the bundle and index, defaults to docker
+# VERSION - index version to build; this defines the max version and includes all bundles beneath it
+# TAG - version tag to be used for the image
+#
+# Example:
+# VERSION=0.8.1 ORG=acatterm ./scripts/create-3scale-index.sh
```
MStokluska, line 20: hey! Not a big deal but do you think we could add a make command?

scripts/create-3scale-index.sh
adam-cattermole: I did consider this, tbh given you've brought it up too I'll opt for that as it'll be extra work to update this

scripts/create-3scale-index.sh
adam-cattermole: Sure can do

MStokluska:
It works as expected, well done! I really like it. 
Left some comments, let me know wdyt.
Also, do you think it would make sense to remove the tmp dirs created? I know they would get eventually removed and since this is going to be mostly used by jenkins so no strong opinion on it.

MStokluska:
> I think that the tmp dirs are removed by the opm command if it completes succesfully - it's only if you ctrl-c the command you can be left with some tmp dirs, but we could maybe catch that and clean up if you think necessary.
> 
> I've addressed the prior comments.

Thanks a lot! ` if you ctrl-c the command` - I did just that :D 

/lgtm

PR #2688, opened 2022-05-26
MGDAPI-4034 bump limitador image

MStokluska:
# What
bump limitador image

# Verification steps
- install rhoam
- verify rate limiting works as expected


austincunningham:
/lgtm
verified on a call with @MStokluska 
rate limiting works as expected

PR #2690, opened 2022-05-27
MGDAPI-3979 - change install mode to use 3scale-index

adam-cattermole:
# Issue link
[MGDAPI-3979](https://issues.redhat.com/browse/MGDAPI-3979)

# What
Update installation yaml to use 3scale-index instead of manifests.

# Verification steps
- Install RHOAM is successful
- Upgrade completes successfully (maybe upgrade pipeline could be used?)

### DIFF
--- a/products/installation.yaml
+++ b/products/installation.yaml
@@ -57,9 +57,10 @@
 #
 products:
   3scale:
-    channel: "rhmi"
-    installFrom: "local"
-    manifestsDir: "integreatly-3scale"
+    channel: "threescale-2.11"
+    installFrom: "index"
+    package: "3scale-operator"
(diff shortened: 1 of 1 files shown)


MStokluska:
Verified on upgrade from 1.21 > 1.22:
- upgrade is not service affecting (no downtime recorded in wlwa)
- 3scale upgrade completes successfully 
- RHMI cr reports that version 1.22 upgrade has been completed, after which 3scale operator bump starts, 3scale reports in progress for another 10 minutes.
- after 3scale sucessful re-deployments of some of the dcs it reports completed

Fresh install works as excepted.

Given that the upgrade takes roughly 10mins without service interruptions I think we are ok.
Awesome @adam-cattermole ! Thanks 

/lgtm

PR #2714, opened 2022-06-07
MGDAPI-4039: Removing RHMI for the code base

Boomatang:
# Issue link
[https://issues.redhat.com/browse/MGDAPI-4039](https://issues.redhat.com/browse/MGDAPI-4039)

# What
This work is to remove the old RHMI product from the operator.
As RHMI has hit its EOL support date there is no need maintain or make business logic decisions based on RHMI.

# Verification steps

**NOTE:** Commits have rearranged, squashed and fixed up, checkout an individual commit does not mean the code will compile.

Open questions:
* Can the assets/icons folder be removed?

As there is a very large area to verify, the verification is broken out into the areas that will need to be reviewed.
The hope is to cover every area but assume that something has been missed from the list and keep an eye out for any kind of issue.

## Code Review
With the removal of the RHMI a lot of if-statements have being removed which was checking for the installation type.
Think about the effect the removal of the if-statements might have when looking at the sandbox deployments.

The rhmi-config has being removed.
CRO seemed to be the only area that had this dependency.
Check that this dependency was correctly refactored.

## RBAC
As the rhmiConfig CR has being removed from the installation, any RBAC rules relating to this should also have being removed.
Same goes for the products that RHMI installed.
Check for RBAC rules that are out of place with the current set of products being installed.
Our backlog does [MGDAPI-3974](https://issues.redhat.com/browse/MGDAPI-3974) which is about review the RBAC access levels.
In this PR we want to make sure that RBAC is removed for components that are no longer installed.

## Building Images
There has being a number of Dockerfiles that have had a change made to them.
Build all Dockerfiles found in the repo.
Use the make targets where possible. 

This testing should also include the building of indexes and bundles.

## unit tests
Refactoring of the unit test has happened.
Check that refactor test are still fit for propose.
Unit test should all pass.
This is possibly covered by the PROW check but no harm in a manual run being carried out.

## e2e tests
e2e test are ran by prow but this is not enough.
Prow uses a limited set of tests.
A jenkins pipeline needs to run the full suite.
The e2eosd is also going to be affected by this work, these test should also be checked.

## Releases
There has being updates done to the prepare release script. 
It would be good to validate the that bundles are generate correctly and bundles can be included in indexes.
Running the release pipeline would also be a good idea if the managed tenants repos can be mocked out.

### Local Development
Running the operator should work as normal.
The defaults for environment variables should now all be mapped to managed-api. 

Uninstall should work as normal.
The cleanup script should be un affected by the changes.

#### OLM Installations
For testing the OLM installations it should require the building of images for the operator, bundle and indexes.
This should be as close the addon flow that we can get. 

#### Installation
A fresh installation should be done.
Nothing is expected to be out of place. 
Configure the testing IDP and check that all routes still work as expected.
Configure a simple rated limited end point to ensure rate limiting is still functioning.

#### upgrade
Set up the cluster with the existing addon version of the operator.
Configure the testing IDP and rate limiting example. 
Upgrade should happen without any issue.

It would be a good idea to check the upgrade as both a service affecting and non-service affecting.
This would prove there has being no on wanted changes to the upgrade logic.

#### uninstall
For the uninstallation use the method that add the label to the config map. 
This method is as close to how hive will do this action.

## Dashboards, Alerts and Metrics
As all these are very much linked it would be good to review all at the same time.
Do an OLM installation as the metrics depend on having the operator running on the cluster.

Review all the dashboards, both internal and customer facing.
All panels in the dashboards should be working and not have any mention of RHMI

Same goes for the alerts.
There should be no alerts with the RHMI reference. 
Check metrics used in the queries to ensure these metrics are still exposed.

For the metrics, no metric should now have reference to RHMI.


pkg/products/monitoringcommon/blackbox.go
```diff
@@ -0,0 +1,66 @@
+package monitoringcommon
+
+import (
+	"context"
+	"fmt"
+	v1alpha12 "github.com/integr8ly/integreatly-operator/apis/v1alpha1"
```
adam-cattermole, line 6: nit: could this just be v1alpha1

pkg/products/rhsso/reconciler_test.go
```diff
@@ -803,6 +798,7 @@ func TestReconciler_fullReconcile(t *testing.T) {
 		ExpectedError         string
 		FakeConfig            *config.ConfigReadWriterMock
 		FakeClient            k8sclient.Client
+		FakeOauthClient       oauthClient.OauthV1Interface
```
adam-cattermole, line 832: nit: I think this `FakeOauthClient` isn't used

Fiona-Waters:
Looked at unit tests section. Tests running successfully.

apis/v1alpha1/installationTypesHelper.go
```diff
@@ -21,5 +17,5 @@ func isMultitenant(installType InstallationType) bool {
 }
 
 func IsManaged(installType InstallationType) bool {
-	return installType == InstallationTypeManaged || installType == InstallationTypeManagedApi || isMultitenant(installType)
+	return installType == InstallationTypeManagedApi || isMultitenant(installType)
```
MStokluska, line 20: I think we should make a clear difference between what managed is and what multitenant is. 
This function does make much sense to me anymore since managed in this scenario can be managed-api or multitenant-manged-api, since this is all what intly operator can be, how can this return false?

apis/v1alpha1/installationTypesHelper_test.go
```diff
@@ -128,14 +101,9 @@ func TestManagedInstallType(t *testing.T) {
...
 		},
-		{
-			name:            "test that installation is managed",
-			installType:     InstallationTypeManaged,
-			expectedOutcome: true,
-		},
 		{
 			name:            "test that installation is not managed",
-			installType:     InstallationTypeSelfManaged,
+			installType:     "Dummy Type",
```
MStokluska, line 106: since we are only leaving the multitenant and managed-api I'm not sure what is the benefit in testing if a type is not managed since it can't be anything else

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -82,31 +82,29 @@ func (r *Reconciler) GetPreflightObject(ns string) runtime.Object {
...
-			phase, err := resources.RemoveNamespace(ctx, installation, serverClient, observabilityConfig.GetNamespace(), r.log)
-			if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
-				return phase, err
-			}
-			return integreatlyv1alpha1.PhaseCompleted, nil
-		}, r.log)
+	phase, err := r.ReconcileFinalizer(ctx, serverClient, installation, string(observabilityConfig.GetProductName()), uninstall, func() (integreatlyv1alpha1.StatusPhase, error) {
+		phase, err := resources.RemoveNamespace(ctx, installation, serverClient, observabilityConfig.GetNamespace(), r.log)
 		if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
-			events.HandleError(r.recorder, installation, phase, "Failed to reconcile finalizer", err)
```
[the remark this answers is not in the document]
MStokluska: nvm, we are moving it not removing

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -179,45 +165,37 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
-			return phase, errors.Wrap(err, "Failed to remove existing prometheus rules from rhoam-operator namespace")
-		}
+	if err = r.processQuota(installation, request.Namespace, installationQuota, serverClient); err != nil {
+		events.HandleError(r.recorder, installation, integreatlyv1alpha1.PhaseFailed, "Error while processing the Quota", err)
+		installation.Status.LastError = err.Error()
+		return integreatlyv1alpha1.PhaseFailed, err
+	}
+	metrics.SetQuota(installation.Status.Quota, installation.Status.ToQuota)
 
+	// temp code for RHOAM, remove once all clusters are upgraded to 1.14
```
MStokluska, line 193: We don't need this code anymore.

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -319,25 +281,23 @@ func (r *Reconciler) setTenantMetrics(ctx context.Context, serverClient k8sclien
...
+		priorityClass.GlobalDefault = false
+		priorityClass.Description = "Priority Class for managed-api"
 
-			return nil
-		}); err != nil {
-			return integreatlyv1alpha1.PhaseInProgress, err
-		}
+		return nil
+	}); err != nil {
+		return integreatlyv1alpha1.PhaseInProgress, err
```
MStokluska, line 300: Would phaseFailed, err make more sense here?

controllers/rhmi/rhmi_controller.go
```diff
@@ -452,41 +452,34 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
...
-			removeOldStages(installation)
-			// end of removal
-			installation.Status.Quota = installationQuota.GetName()
-			installation.Status.ToQuota = ""
-		}
+		metrics.SetVersions(string(installation.Status.Stage), installation.Status.Version, installation.Status.ToVersion, string(externalClusterId), installation.CreationTimestamp.Unix())
+		installation.Status.Quota = installationQuota.GetName()
+		installation.Status.ToQuota = ""
+		// remove the code for next release
+		removeOldStages(installation)
```
MStokluska, line 459: this can go

controllers/rhmi/rhmi_controller.go
```diff
@@ -452,41 +452,34 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
...
+	metrics.SetStatus(installation)
 
-	if rhmiv1alpha1.IsRHOAM(rhmiv1alpha1.InstallationType(installation.Spec.Type)) {
-		if _, ok := installation.Status.Stages[rhmiv1alpha1.MonitoringStage]; ok {
-			log.Info("delete Monitoring stage from installation.Status")
-			delete(installation.Status.Stages, rhmiv1alpha1.MonitoringStage)
-		}
+	if _, ok := installation.Status.Stages[rhmiv1alpha1.MonitoringStage]; ok {
+		log.Info("delete Monitoring stage from installation.Status")
+		delete(installation.Status.Stages, rhmiv1alpha1.MonitoringStage)
```
MStokluska, line 482: do we need this?

controllers/rhmi/rhmi_controller.go
```diff
@@ -998,65 +979,63 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
...
 
-	if rhmiv1alpha1.IsRHOAM(rhmiv1alpha1.InstallationType(installation.Spec.Type)) {
-		// Check if the quota parameter is found from the add-on
-		okParam, err := addon.ExistsParameterByInstallation(context.TODO(), r.Client, installation, addon.QuotaParamName)
+	// Check if the quota parameter is found from the add-on
+	okParam, err := addon.ExistsParameterByInstallation(context.TODO(), r.Client, installation, addon.QuotaParamName)
+	if err != nil {
+		preflightMessage := fmt.Sprintf("failed to retrieve addon parameter %s: %v", addon.QuotaParamName, err)
+		log.Warning(preflightMessage)
+		return result, err
```
MStokluska, line 987: I'm confused on how is this to work with trialQuota. Without looking deeply into code, if quota param of `QuotaParamName` is not found, how can it ever attempt to check if `TrialQuotaParamName` is present if we are returning result,err here?

controllers/rhmi/rhmi_controller.go
```diff
@@ -998,65 +979,63 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
...
-				return result, err
-			}
-		}
+	// Check if the quota parameter is found from the environment variable
+	quotaEnv, envOk := os.LookupEnv(rhmiv1alpha1.EnvKeyQuota)
 
-		// Check if the quota parameter is found from the environment variable
-		quotaEnv, envOk := os.LookupEnv(rhmiv1alpha1.EnvKeyQuota)
+	// If the quota parameter is not found:
+	if !okParam {
```
MStokluska, line 1004: same as above, if non of the quota params are found, with the above returning result,err how are we ever going to reach this point?

controllers/rhmi/rhmi_controller.go
```diff
@@ -82,7 +82,7 @@ import (
 const (
 	deletionFinalizer                = "configmaps/finalizer"
 	previousDeletionFinalizer        = "finalizer/configmaps"
-	DefaultInstallationName          = "rhmi"
+	DefaultInstallationName          = "rhoam"
```
MStokluska, line 85: if rhoam is the default, why are specifying rhoam again in the line below? I think we should agree that rhoam is the default type without specifying it again

products/products.yaml
```diff
@@ -12,12 +12,6 @@ products:
     installType: "rhmi/rhoam"
```
MStokluska, line 12: rhoam only

MStokluska:
I've only partially reviewed so far. So far a couple of comments that need to be addressed, also, I might have missed it but I can't see rhmi config CR deletion logic? Are we sure that none of our customers have the RHMI config CR created? Also, should if they do, should rhmi config be removed prior to removing the CRD?

carlkyrillos:
I completed all steps in the Deployments section including local install/upgrade and olm install/upgrade both service affecting and non-service affecting. I also set up the testing idp and rate limiting. Lastly, I tested the uninstall using both `oc delete rhmi rhoam` and by adding the `api.openshift.com/addon-managed-api-service-delete=true` label to the `managed-api-service` ConfigMap

MStokluska:
Issue with creating a new release: 
```
Adding related images to the CSV
./test/common/selfmanaged_apicast.go:1:1: expected 'package', found 'EOF'
make: *** [Makefile:444: release/prepare] Error 2
```

MStokluska:
Tested upgrade from prod > this branch, apart from the issue with remaining rhmi config CR it looks ok

pkg/products/threescale/prometheusRules.go
```diff
@@ -17,30 +16,23 @@ import (
...
-		}
-
-		namespace = observabilityConfig.GetNamespace()
-		operatorNamespace = observabilityConfig.GetNamespace()
-		alertNamePrefix = "3scale-"
-		operatorAlertNamePrefix = "3scale-operator-"
+	observabilityConfig, err := r.ConfigManager.ReadObservability()
+	if err != nil {
+		logger.Warning("failed to get observability config")
+		return nil, nil
```
KevFan, line 28:
```suggestion
		return nil, err
```

Shoudl we be returning `err` here ?

Makefile
```diff
@@ -205,21 +205,6 @@ image/build/push: image/build image/push
 
 
 ############ E2E TEST COMMANDS ############
-.PHONY: test/e2e/prow
```
MStokluska, line 208: would it make more sense to make rhoam a default e2e so that the above line would trigger rhoam tests instead e2e-rhoam?

apis/v1alpha1/installationTypesHelper_test.go
Boomatang: Currently it can't be anything else. Unit test help during development and future refactors. It is for future refactor that this test case is addressing. As there is no maintenance cost to having this test case, I don't see the need to remove it.

controllers/rhmi/bootstrapReconciler.go
Boomatang: This is code related to RHOAM not RHMI. There is enough change happening with the RHMI removal. The removal of this is out side the scope of this work.

controllers/rhmi/rhmi_controller.go
Boomatang: This is code related to RHOAM not RHMI. There is enough change happening with the RHMI removal. The removal of this is out side the scope of this work.

controllers/rhmi/rhmi_controller.go
Boomatang: This is part of the move to OO. While I am guess it is no longer needed, it is also related to RHOAM and not RHMI. The removal of RHOAM code is outside the scope of this work.

pkg/products/threescale/prometheusRules.go
Boomatang: Possibly you are correct but this code is related to RHOAM and not RHMI. I feel this change would be out side the scope of this work.

Makefile
Boomatang: If we only had RHOAM e2e test I would agree with you but as there is also the multitenant-rhoam I think it is best to keep it as is in the Makefile. It is clearer what the make targets do. However it would be possible to update the openshift/release configuration. To have that to be called e2e instant of e2e-rhoam. That is even something that could be done now without this PR being merged.

controllers/rhmi/bootstrapReconciler.go
MStokluska: As a general rule that we have agreed on during one of our sessions, it is recommended to make code improvements when we can while working on other jiras. There is no harm in deleting a couple of lines that are not needed. Let's tidy up our code since we are looking line by line!

pkg/products/threescale/prometheusRules.go
KevFan: Sure, I'm good with addressing this on a later ticket

controllers/rhmi/rhmi_controller.go
MStokluska: Same as above - since we are doing these massive efforts of going literally through every file and cleaning up our repo to me this needs to go as well.

controllers/rhmi/rhmi_controller.go
MStokluska: same as above..

apis/v1alpha1/installationTypesHelper.go
Boomatang: After doing a bit more digging into this function and where it is used. The function its self is only used once in the preflight checks when checking for the Pager Duty secret. I am in agreement now, this function can be fully removed.

apis/v1alpha1/installationTypesHelper_test.go
Boomatang: On second pass the function under test was removable. This unit test has being fully removed.

controllers/rhmi/bootstrapReconciler.go
Boomatang: As this was the existing behaviour I am reluctant to change it now. There are other functions that give the same `PhaseInProgress` return value on error. If one is to be changed they should all be changed. There are also no unit tests checking this function. The removal of RHMI is not driving the need for this change.

controllers/rhmi/rhmi_controller.go
Boomatang: As this is not a typical PR I do not agree with code changes that are not related to the task at hand. To resolve this I have opened a second PR which has the code removal you are looking for. https://github.com/integr8ly/integreatly-operator/pull/2920

controllers/rhmi/rhmi_controller.go
Boomatang: As this is not a typical PR I do not agree with code changes that are not related to the task at hand. To resolve this I have opened a second PR which has the code removal you are looking for. https://github.com/integr8ly/integreatly-operator/pull/2920

controllers/rhmi/rhmi_controller.go
Boomatang: This took bit of digging. The `okParam` is a boolean and is false if the parameter is not found in the addon secret. The `err` is nil if the parameter is not found in the secret. An error is only raised when there is a problem during the function call. Not finding the addon secret would raise an error for instances. This is ok and working as expected.

controllers/rhmi/rhmi_controller.go
Boomatang: This took bit of digging. The `okParam` is a boolean and is false if the parameter is not found in the addon secret. The `err` is nil if the parameter is not found in the secret. An error is only raised when there is a problem during the function call. Not finding the addon secret would raise an error for instances. This is ok and working as expected.

apis/v1alpha1/installationTypesHelper_test.go
MStokluska: Thanks

controllers/rhmi/bootstrapReconciler.go
MStokluska: Thanks

controllers/rhmi/bootstrapReconciler.go
MStokluska: No strong opinion, but maybe a follow up Jira to double check if our returns of phases make sense would be good here

controllers/rhmi/rhmi_controller.go
MStokluska: Thanks

controllers/rhmi/rhmi_controller.go
MStokluska: Thanks

Makefile
MStokluska: No strong opinion on this so whatever you think is best.

controllers/rhmi/rhmi_controller.go
MStokluska: Thanks for explanation, looks good

controllers/rhmi/rhmi_controller.go
MStokluska: Thanks for explanation, looks good

controllers/rhmi/rhmi_controller.go
Boomatang: Did some looking into this. Turns out both constants here are only used in the one non-export helper function which would return one or the other. Have removed the helper function and both constants.

pkg/products/monitoringcommon/blackbox.go
Boomatang: Turns out the function in this file is not used anywhere. File can be removed.

MStokluska:
@Boomatang I have finished my review, with the supporting PRs that you have created everything looks good here apart from the removal logic for RHMIConfig and RHMIConfig CRDs removal, not sure if you want this as part of another PR or this (imo it should be introduced with this PR) but ideally it would need to go in the same RHOAM release.

Boomatang:
@MStokluska I agree the RHMIconfig CRD is a problem with this PR and its removal should be included in this. Currently think about what the best way to achieve this is. Thanks for the review so far.

MStokluska:
Force merged to avoid having to rebase this pr again after lifting code freeze

PR #2592, opened 2022-04-08
Single-Stage RHOAM demo

makslion:
# Issue link
[MGDAPI-3747](https://issues.redhat.com/browse/MGDAPI-3747)

# What
Demo of the Single-Stage RHOAM. This PR isn't intended to be merged at this stage of development. 

# Verification steps
Not applicable

PR #2729, opened 2022-06-13
MGDAPI-4150 squash installation stages into one

makslion:
# Issue link
[MGDAPI-4150](https://issues.redhat.com/browse/MGDAPI-4150)

# What
Cleaned up version of #2592 
I'm squashing installation stages (authentication, observability, cloud resources and products) into one stage. 

# Verification steps
Verification was done before, however, it wouldn't hurt to be extra safe. 
1. Local installs could be done of this PR 
2. "Bundles-like" install will be done by E2E tests on this PR 
3. An upgrade scenario could be tested using `quay.io/mvavilov/rhoam-index:2.0.0`->`quay.io/mvavilov/rhoam-index:3.0.0` path. Where 2.0.0 is a master from a couple of days ago and 3.0.0 is the current brunch. Or you could build your own bundles using [this](https://github.com/RHCloudServices/integreatly-help/blob/master/guides/olm/installing-rhmi-bundle-format.md) guide.


adam-cattermole (commented):
I think I mentioned this in the original ticket, but I think the uninstall stages should be reverted to multi-stage. This is because the ordering of uninstall matters - we have seen before that if CRO is removed before SSO we can get stuck and fail to uninstall. The performance benefit from moving to single stage uninstall was also minimal iirc?

makslion:
/test rhoam-e2e multitenant-rhoam-e2e

> I think I mentioned this in the original ticket, but I think the uninstall stages should be reverted to multi-stage. This is because the ordering of uninstall matters - we have seen before that if CRO is removed before SSO we can get stuck and fail to uninstall. The performance benefit from moving to single stage uninstall was also minimal iirc?

Right, must've forgotten about it. Reverted uninstallation.

Boomatang:
What work is being done to address the upgrades. When I tried an upgrade the rhoam CR keeps the old stages as well as adding the `installation` stage. I think we need to clean out the old stages as part of this work. From the point SOPs having the old stages there might lead to miss information.

makslion:
adressed @Boomatang point - cleaning up CR after the upgrade. There is also a strange behaviour of e2e tests: third time in a row rhoam-e2e container gets killed. Investigating, but just in case

/test rhoam-e2e multitenant-rhoam-e2e

makslion:
two out of two runs of `IMAGE_FORMAT=quay.io/mvavilov/managed-api-service:rhoam-v3.0.0 INSTALLATION_TYPE=managed-api make test/e2e/rhoam/prow` succeeded. Not sure what is up with prow checks :thinking:

Boomatang:
@makslion From what I was looking for this looks good. The upgrade cleans the rhmi CR up. For @adam-cattermole point I'll let him confirm that one. 

Question: Did you check to see if there is any alerts depending on the old stage names?

Side Note: It is so nice to see this install in less than 20 minutes and as far as I know all my custom watch scripts are still working. Which is nice.

makslion:
hey @Boomatang no mention of stages in the Prometheus rules but I have a list of SOPs to update. I will link PR to the JIRA once it will be ready

adam-cattermole:
Thanks @makslion for reverting the uninstall stuff. Glad the tests are running properly now and passing! Code changes make sense to me

makslion:
Update `make cluster/deploy/integreatly-rhmi-cr.yml` target with proper wait command. Couldn't find where it is used, but just in case as Jim pointed. 

/test rhoam-e2e multitenant-rhoam-e2e

KevFan:
* Verified local installation and uninstallation
* Verified upgrade and uninstallation via config map 

Changes looks good to me :+1: 

/lgtm

makslion:
rebased to account for go 1.18

makslion:
the coverage diff is a removal of old stages after an upgrade. Temporary code that will be gone for a next release

MStokluska:
bypassing codecov failure

PR #2760, opened 2022-06-27
MGDAPI-3452-C08B-automated-test

Stellar-9:
# Issue link
[https://issues.redhat.com/browse/MGDAPI-3452](url)

# What
https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/alerts/c08b-verify-alert-links-to-sops.md

# Verification steps
`INSTALLATION_TYPE=managed-api TEST=C08B BYPASS_STORAGE_TYPE_CHECK=true make test/e2e/single`
Test implemented and passing when ran locally and in nightly pipelines

### DIFF
--- a/test/common/verify_alert_links_to_SOPs.go
+++ b/test/common/verify_alert_links_to_SOPs.go
@@ -1,0 +1,179 @@
+package common
+
+import (
+	"encoding/json"
+	"fmt"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"io"
+	"net/http"
+	"os"
+	"strings"
+	"sync"
+)
+
+var githubToken = os.Getenv("TOKEN_GITHUB")
+var failedSOPurls = make(chan string)
+var wg sync.WaitGroup
+var countFailedLinks int
+
+func TestSOPUrls(t TestingTB, ctx *TestingContext) {
+
+	if githubToken == "" {
+		t.Skip("Github token not provided, use GITHUB_TOKEN environment variable to specify it")
+	}
+
+	var sopUrls []string
+
+	// test connection to Github API, with single url
+
+	testUrl := "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/README.md"
+	apiUrl := modifyLink(testUrl)
+	client := &http.Client{}
+	req, err := http.NewRequest("GET", apiUrl, nil)
+	if err != nil {
+		log.Fatal("%s", err)
+	}
+
+	req.Header.Add("Accept", `application/json`)
+	req.Header.Add("Authorization", fmt.Sprintf("token %s", githubToken))
+	testResp, err := client.Do(req)
+	if err != nil {
+		t.Log(err)
+	}
+
+	defer func(Body io.ReadCloser) {
+		err := Body.Close()
+		if err != nil {
+
+		}
+	}(testResp.Body)
+
+	if testResp.StatusCode != 200 {
+		t.Fatal("Response status: ", testUrl, testResp.Status, "Given token does not allow access to SOP URLs")
+	}
+
+	output, err := execToPod("wget -qO - localhost:9090/api/v1/rules",
+		"prometheus-prometheus-0",
+		ObservabilityProductNamespace,
+		"prometheus", ctx)
+	if err != nil {
+		t.Fatal("failed to exec to pod:", err)
+	}
+
+	var ApiOutput prometheusAPIResponse
+
+	err = json.Unmarshal([]byte(output), &ApiOutput)
+	if err != nil {
+		t.Fatal("failed to unmarshal json", err)
+	}
+
+	var rulesResult prometheusv1.RulesResult
+
+	err = json.Unmarshal([]byte(ApiOutput.Data), &rulesResult)
+	if err != nil {
+		t.Fatal("failed to unmarshal json", err)
+	}
+
+	for _, group := range rulesResult.Groups {
+		for _, rule := range group.Rules {
+			switch v := rule.(type) {
+			case prometheusv1.RecordingRule:
+			case prometheusv1.AlertingRule:
+				for annotation, sopUrl := range v.Annotations {
+					if annotation == "sop_url" {
+						sopUrls = append(sopUrls, string(sopUrl))
+					}
+				}
+
+			default:
+				fmt.Printf("Unknown rule type %s", v)
+
+			}
+		}
+	}
+	sopUrls = unique(sopUrls)
+	validateSOPLinks(t, sopUrls)
+}
+
+// modify raw link to Github API verison
+func modifyLink(rawLink string) (apiSOPUrl string) {
+	r := strings.NewReplacer(
+		"github", "api.github",
+		"com/", "com/repos/",
+		"blob/master", "contents",
+		"tree/master", "contents",
+	)
+
+	apiSOPUrl = r.Replace(rawLink)
+	return
+}
+
+// remove duplicate links
+func unique(s []string) []string {
+	inResult := make(map[string]bool)
+	var result []string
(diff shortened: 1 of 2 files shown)


test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,179 @@
...
+				}
+
+			default:
+				fmt.Printf("Unknown rule type %s", v)
+
+			}
+		}
+	}
+	sopUrls = unique(sopUrls)
+	validateSOPLinks(t, sopUrls)
```
trepel, line 95: Minor thing but you are using `sopUrl` and `sopLink` randomly across the code. I would stick with just one name since it is the same thing. Eg here the the function name is `validateSOPLlinks` but the parameter name is `sopUrls`. I would expect either `validateSopUrls` as function name or `sopLinks` as argument name. Just for the sake of consistency.

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// modify raw link to Github API verison
+func modifyLink(rawLink string) (apiSOPUrl string) {
```
trepel, line 99: `modifyLInk` is not very descriptive. I think better name would be something like `convertToGithubApiLink`. And even better would probably be `convertToGithubApiUrl` and using `sopUrl` instead of generic `rawLink`. But again this is fairly minor.

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,179 @@
...
+		go getSOPAlertLinkStatus(t, url, failedSOPurls)
+
+	}
+
+	go func() {
+		wg.Wait()
+		close(failedSOPurls)
+	}()
+
+	for failedSOPUrl := range failedSOPurls {
```
trepel, line 138: I think this `for` loop needs to be moved into the `if` below. There's no point in looping through the `failedSOPurls` if we know that `countFailedLinks` equals to zero.

trepel (changes requested):
Also, I would update the Verification steps. I think one has to run the test three times, once with no gh token specified, 2nd time with invalid token specified, 3rd time with valid token. Of course you can/should do that yourself and can just share logs, that would make the reviewer's job simpler.

test/common/verify_alert_links_to_SOPs.go
Stellar-9: I'm thinking about removing boolean logic that function? I think there is maybe no need for it? Because when there will be failed url, it will be displayed to the user

Could that be better solution?

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,179 @@
...
+		"prometheus-prometheus-0",
+		ObservabilityProductNamespace,
+		"prometheus", ctx)
+	if err != nil {
+		t.Fatal("failed to exec to pod:", err)
+	}
+
+	var ApiOutput prometheusAPIResponse
+
+	err = json.Unmarshal([]byte(output), &ApiOutput)
```
makslion, line 65: is this a redundant type conversion? :thinking: just a "cosmetic" comment

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,179 @@
...
+	for failedSOPUrl := range failedSOPurls {
+		t.Log("failed to connect to url: ", failedSOPUrl)
+
+	}
+
+	if countFailedLinks != 0 {
+		status = false
+	}
+
+	return status
```
makslion, line 147: That's a silly question, but how this test would fail in case I've provided the wrong link in the SOP? I would imagine this bool when returning would have a check and would `t.Fatal` if `false` :thinking:

makslion:
Do I understand correctly that this test fully relies on the token envar? I was included in the `HAPPY_PATH` suite so I was expecting it to run and pass every time we are running e2e tests. But it seems like (because we will be missing that envar) that it will be skipped every time :thinking: 

Questions: 
* do we want to consider the introduction of  this envar in our prow checks and/or pipelines? 
* Do we want to move it to another test suite? 
:thinking:

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if githubToken == "" {
+		t.Skip("Github token not provided, use GITHUB_TOKEN environment variable to specify it")
+	}
+
+	var sopUrls []string
+
+	// test connection to Github API, with single url
+
+	testUrl := "https://github.com/RHCloudServices/integreatly-help/blob/master/sops/README.md"
```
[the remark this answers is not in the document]
Stellar-9: Done in latest commit

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,179 @@
...
+			case prometheusv1.RecordingRule:
+			case prometheusv1.AlertingRule:
+				for annotation, sopUrl := range v.Annotations {
+					if annotation == "sop_url" {
+						sopUrls = append(sopUrls, string(sopUrl))
+					}
+				}
+
+			default:
+				fmt.Printf("Unknown rule type %s", v)
```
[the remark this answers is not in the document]
Stellar-9: Done in latest commit

test/common/verify_alert_links_to_SOPs.go
Stellar-9: Done in latest commit

test/common/verify_alert_links_to_SOPs.go
Stellar-9: Done in latest commit

test/common/verify_alert_links_to_SOPs.go
```diff
@@ -0,0 +1,179 @@
...
+	"fmt"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"io"
+	"net/http"
+	"os"
+	"strings"
+	"sync"
+)
+
+var githubToken = os.Getenv("TOKEN_GITHUB")
```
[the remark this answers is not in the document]
Stellar-9: Done in latest commit

test/common/verify_alert_links_to_SOPs.go
Stellar-9: I'm not sure. I have an idea where your comment comming from, but can you please elaborate in more details or suggest me better way?

test/common/verify_alert_links_to_SOPs.go
Stellar-9: Done in latest commit

trepel:
@makslion My idea was to have it in HAPPY_PATH and yes, it will be skipped if the GH token in not provided. My thinking was that it is just fine. As a next step we would look into how to define that env var in various environments
- pipelines
- prow checks
- osde2e

test/common/verify_alert_links_to_SOPs.go
trepel: I am not sure myself. But I assume @Stellar-9 just copy-pasted it from e.g. https://github.com/integr8ly/integreatly-operator/blob/master/test/common/alerts_exist.go#L1088-L1098 so maybe we can keep it there for the sake of consistency?
Of course you can try without that (I assume @makslion meant that `byte`) conversion and if it indeed works you can get rid of it.

trepel: I don't mind either way

makslion:
Cool. The work to add GH tokens will be done under this JIRA or there will be a follow-up?

trepel:
I think this is ok too. I would just squash the commits and ask for approval

/lgtm

PR #2766, opened 2022-06-29
MGDAPI-3097 - add commit check function

adam-cattermole:
# Issue link
[MGDAPI-3097](https://issues.redhat.com/browse/MGDAPI-3097)

# What
- Added a script and associated make target to check that all commits begin with a MGDAPI- ticket.

# Verification steps
- Checkout PR
- Add some commits that fit or break the name structure
- `make commits/check`

### DIFF
--- a/scripts/commits-check.sh
+++ b/scripts/commits-check.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+#
+# This script retrieves the list of commits difference from master
+# and ensures that each commit starts with a MGDAPI- numbered ticket
+# See https://issues.redhat.com/browse/MGDAPI-3097
+
+merge_base=$(git merge-base master HEAD)
+commits=$(git rev-list --no-merges $merge_base..HEAD)
+invalidCommits=()
+
+if [ -z "$commits" ]; then
+    echo "No commits to scan"
+    exit 0
+fi
+
+for sha in $commits; do
+    msg=$(git log --format=%B -n 1 $sha)
+    if [[ ! $msg =~ ^(MGDAPI-[0-9]+)[[:blank:]].*$ ]]; then
+        invalidCommits+=( "Commit $sha:\n$msg\n" )
+    fi
+done
+
+if [ ${#invalidCommits[@]} -gt 0 ]; then
+    printf "Not all commits start with a \"MGDAPI-\" ticket\n\n"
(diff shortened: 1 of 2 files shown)


scripts/commits-check.sh
```diff
@@ -0,0 +1,31 @@
...
+
+for sha in $commits; do
+    msg=$(git log --format=%B -n 1 $sha)
+    if [[ ! $msg =~ ^(MGDAPI-[0-9]+)[[:blank:]].*$ ]]; then
+        invalidCommits+=( "Commit $sha:\n$msg\n" )
+    fi
+done
+
+if [ ${#invalidCommits[@]} -gt 0 ]; then
+    printf "Not all commits start wtih a \"MGDAPI-\" ticket\n\n"
```
carlkyrillos, line 24:
```suggestion
    printf "Not all commits start with a \"MGDAPI-\" ticket\n\n"
```
Minor typo in the printed string.

scripts/commits-check.sh
```diff
@@ -0,0 +1,31 @@
...
+    for i in "${invalidCommits[@]}"; do
+        printf "$i"
+    done
+    exit 1
```
carlkyrillos, line 28: Does it make sense to exit with an error code of `1` because the script technically hasn't failed at this point? Or is the `exit 1` meant to highlight to the user that there is an error with their commits?

carlkyrillos (commented):
Added two small comments but overall everything looks good. I verified that `make commits/check` passes when all commits are formatted properly and outputs which commits are formatted incorrectly if any exist.

scripts/commits-check.sh
adam-cattermole: I made the assumption that the script has to exit with a non-zero exit code for the prow job to detect that this has not passed. I just chose 1 as it's the general error type. Open to suggestions if you think there's a neat alternative though

scripts/commits-check.sh
adam-cattermole: woops - fixed this

scripts/commits-check.sh
carlkyrillos: Oh I didn't realize that was for prow to detect it wasn't passing. I think 1 makes sense for the error type so no action is needed on this.

carlkyrillos:
Reviewed latest changes and all comments have been resolved.
/lgtm

PR #2767, opened 2022-06-29
MGDAPI-3805 - gosec: address G601

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-3805
# What

Address gosec G601 - https://stackoverflow.com/questions/62446118/implicit-memory-aliasing-in-for-loop

# Verification steps

* Passing gosec prow check 
* Passing e2e should be sufficient

### DIFF
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -412,7 +412,9 @@ func (r *RHMIReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 	}
 
 	installationQuota := &quota.Quota{}
-	for _, stage := range installType.GetInstallStages() {
+	installStages := installType.GetInstallStages()
+	for i := range installStages {
+		stage := installStages[i]
 		var err error
 		var stagePhase rhmiv1alpha1.StatusPhase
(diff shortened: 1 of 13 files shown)


controllers/rhmi/rhmi_controller.go
```diff
@@ -1180,24 +1182,24 @@ func (r *RHMIReconciler) processStage(installation *rhmiv1alpha1.RHMI, stage *St
 		}
 
 		// Verify that watches for this productStatus CRDs have been created
-		config, err := configManager.ReadProduct(productStatus.Name)
+		productConfig, err := configManager.ReadProduct(productName)
```
adam-cattermole, line 1185: minor nit - if we change the config read to use `productName` we should probably adjust these err messages below to use the key also, or we should change this back to retrieving from `productStatus.Name`

KevFan: Good catch !  Will revert this back to `productStatus.Name` :+1:

adam-cattermole:
Thanks for the change

/lgtm

KevFan:
/test rhoam-e2e

@adam-cattermole Rebased against master to resolve conflicts if you can have a look again

adam-cattermole:
Sure, all still looks great

/lgtm

PR #2775, opened 2022-07-05
[MGDAPI-4147] add installation_controller_reconcile_duration_seconds metric and InstallationControllerReconcileLoopDelayed alert

cecobask:
# Issue link
https://issues.redhat.com/browse/MGDAPI-4147

# What
Create a metric and warning alert for when `rhmi_controller` Reconcile() takes longer than 12 mins

# Verification steps
1. Create a fresh cluster and ocm log in to it
2. `INSTALLATION_TYPE=managed-api make cluster/prepare/local`
3. Create a catalog source pointing to my index where I added the following:
    1. `time.Sleep(time.Minute * 3)` to this line to simulate an install delay https://github.com/integr8ly/integreatly-operator/blob/master/controllers/rhmi/rhmi_controller.go#L472
    2. Tweaked the `RHOAMInstallationControllerReconcileLoopDelayed` alert to fire after 2 minutes, instead of 12, to save time
    3. The rest of the code is the same as in the PR
```
cat << EOF | oc create -f -
kind: CatalogSource
metadata:
  name: rhmi-operators
spec:
  sourceType: grpc
  image: quay.io/tdimov0/managed-api-service-index:1.24.0
EOF
```
4. Install RHOAM from the operator hub (source=rhmi-operators)
5. Deploy an RHMI CR `INSTALLATION_TYPE=managed-api make deploy/integreatly-rhmi-cr.yml`
6. Verify that:
    1. During the installation of RHOAM, the `RHOAMInstallationControllerReconcileLoopDelayed` alert didn't fire
    2. When the RHMI CR's status stage changes to `complete`, the `RHOAMInstallationControllerReconcileLoopDelayed` should be triggered and should keep firing indefinitely
7. Simulate an upgrade scenario by changing the catalog source image to `quay.io/tdimov0/managed-api-service-index:1.24.1` and upgrading the RHOAM operator through OCM
8. Verify that:
    1. During the upgrade of RHOAM, the `RHOAMInstallationControllerReconcileLoopDelayed` alert didn't fire
    2. When the RHOAM upgrade completes, the `RHOAMInstallationControllerReconcileLoopDelayed` should be triggered and should keep firing indefinitely
9. After this PR has been merged, the same logic will apply. However, the alert will only fire in the case that `rhmi_controller` single reconciliation is taking more than 12 minutes instead of 2 minutes

### DIFF
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"github.com/integr8ly/integreatly-operator/pkg/resources/k8s"
 	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/prometheus/client_golang/prometheus"
 	"io"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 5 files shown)


Patryk-Stefanski:
The code looks good, the alert fired after the specified time and stopped firing after RHOAM finished installing.
/lgtm

MStokluska:
Force merging, unit tests around rhmi controller will be covered by a different jira

PR #2776, opened 2022-07-05
MGDAPI-4262 fie E2E tests for MT RHOAM

makslion:
# Issue link
[MGDAPI-4262](https://issues.redhat.com/browse/MGDAPI-4262)

# What
The following tests failed on the MT upgrade pipeline. 
A26 - couldn't reproduce
C04 - couldn't reproduce
M01 - couldn't reproduce

A16 - excluded from MT suite
H03 - excluded from MT suite
H11 - excluded from MT suite
B04 - excluded from MT suite
B09 - excluded from MT suite
B03 - excluded from MT suite
H07 - excluded from MT suite

H24 - fixed 
A34 - fixed 

# Verification steps
1. Install MT RHOAM 
2. Run fixed e2e tests from this PR (they are not included in the prow suite):
```
INSTALLATION_TYPE=managed-api TEST=A34 BYPASS_STORAGE_TYPE_CHECK=true make test/e2e/single
INSTALLATION_TYPE=managed-api TEST=H24 BYPASS_STORAGE_TYPE_CHECK=true make test/e2e/single
```
3. Ensure tests are passing

### DIFF
--- a/test/common/3scale_smtp.go
+++ b/test/common/3scale_smtp.go
@@ -99,7 +99,7 @@ func Test3ScaleSMTPConfig(t TestingTB, ctx *TestingContext) {
 			t.Log(err)
 		}
 		if rhmiv1alpha1.IsRHOAMMultitenant(rhmiv1alpha1.InstallationType(inst.Spec.Type)) {
-			_, err = rest3ScaleSecret(ctx, t)
+			_, err = reset3ScaleSecret(ctx, t)
 			if err != nil {
 				t.Log(err)
 			}
@@ -493,7 +493,7 @@ func sendTestEmail(ctx *TestingContext, t TestingTB) {
 	}
 }
 
-func rest3ScaleSecret(ctx *TestingContext, t TestingTB) (string, error) {
+func reset3ScaleSecret(ctx *TestingContext, t TestingTB) (string, error) {
 	t.Log("Resetting 3Scale secret")
 	secret := v1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
--- a/test/common/3scale_smtp.go
+++ b/test/common/3scale_smtp.go
@@ -49,6 +49,10 @@ var (
 	originalPassword              = ""
 	originalPort                  = ""
 	originalUsername              = ""
+	original3ScalePassword        = ""
+	original3ScalePort            = ""
+	original3ScaleHost            = ""
+	original3ScaleUsername        = ""
 )
 
 //Test3ScaleSMTPConfig to confirm 3scale can send an email
@@ -81,12 +85,24 @@ func Test3ScaleSMTPConfig(t TestingTB, ctx *TestingContext) {
 	if err != nil {
 		t.Log(err)
 	}
+	if rhmiv1alpha1.IsRHOAMMultitenant(rhmiv1alpha1.InstallationType(inst.Spec.Type)) {
+		_, err = patch3ScaleSecret(ctx, t)
+		if err != nil {
+			t.Log(err)
+		}
+	}
 
 	defer func(t TestingTB, ctx *TestingContext, isCreated bool) {
 		t.Log("Reset SMTP details")
 		_, err = resetSecret(ctx, t, isCreated)
 		if err != nil {
 			t.Log(err)
+		}
+		if rhmiv1alpha1.IsRHOAMMultitenant(rhmiv1alpha1.InstallationType(inst.Spec.Type)) {
+			_, err = rest3ScaleSecret(ctx, t)
+			if err != nil {
+				t.Log(err)
+			}
 		}
 	}(t, ctx, isCreated)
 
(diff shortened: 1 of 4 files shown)


test/common/3scale_smtp.go
```diff
@@ -477,6 +493,27 @@ func sendTestEmail(ctx *TestingContext, t TestingTB) {
 	}
 }
 
+func rest3ScaleSecret(ctx *TestingContext, t TestingTB) (string, error) {
```
Patryk-Stefanski, line 496:
```suggestion
func reset3ScaleSecret(ctx *TestingContext, t TestingTB) (string, error) {
```

test/common/3scale_smtp.go
```diff
@@ -81,13 +85,25 @@ func Test3ScaleSMTPConfig(t TestingTB, ctx *TestingContext) {
...
+	}
 
 	defer func(t TestingTB, ctx *TestingContext, isCreated bool) {
 		t.Log("Reset SMTP details")
 		_, err = resetSecret(ctx, t, isCreated)
 		if err != nil {
 			t.Log(err)
 		}
+		if rhmiv1alpha1.IsRHOAMMultitenant(rhmiv1alpha1.InstallationType(inst.Spec.Type)) {
+			_, err = rest3ScaleSecret(ctx, t)
```
Patryk-Stefanski, line 102:
```suggestion
			_, err = reset3ScaleSecret(ctx, t)
```

test/common/quota.go
```diff
@@ -27,20 +27,23 @@ import (
...
-	prometheusRule2           = "marin3r-api-usage-alert-level2"
+	prometheusRule2           = "api-usage-alert-level2"
 	prometheusRule2Desc       = "per minute over 2 hours"
-	prometheusRule3           = "marin3r-api-usage-alert-level3"
+	prometheusRule3           = "api-usage-alert-level3"
 	prometheusRule3Desc       = "per minute over 30 minutes"
 	higherQuotaParam          = "200"
 	lowerQuotaParam           = "1"
 	higherQuotaName           = "20 Million"
+	higherQuotaParamMT        = "10"
```
Patryk-Stefanski, line 39: I think 100 is the value we use to set the quota to 10million
```suggestion
	higherQuotaParamMT        = "100"
```

test/common/3scale_smtp.go
makslion: is there an extension to IDE to check for spelling? :facepalm: Not a single PR without a typo. huh?

test/common/quota.go
makslion: adressed. Changed the name of the quota to 1 Million. Nicely spotted!

makslion:
Confident that is it not needed (affected tests aren't in prow suite), but since this is not a time sensitive 

/test rhoam-e2e multitenant-rhoam-e2e

makslion:
rebased to account for the `codecov` failure

test/common/3scale_smtp.go
```diff
@@ -477,6 +493,27 @@ func sendTestEmail(ctx *TestingContext, t TestingTB) {
...
+		},
+		Data: map[string][]byte{},
+	}
+	secret.Data["address"] = []byte(original3ScaleHost)
+	secret.Data["password"] = []byte(original3ScalePassword)
+	secret.Data["port"] = []byte(original3ScalePort)
+	secret.Data["username"] = []byte(original3ScaleUsername)
+
+	if err := ctx.Client.Update(goctx.TODO(), secret.DeepCopy(), &k8sclient.UpdateOptions{}); err != nil {
+		return secret.APIVersion, err
```
MStokluska, line 511:
```suggestion
		"", err
```
we don't seem to be using the secret at all, why return it at all?

test/common/3scale_smtp.go
```diff
@@ -477,6 +493,27 @@ func sendTestEmail(ctx *TestingContext, t TestingTB) {
 	}
 }
 
+func reset3ScaleSecret(ctx *TestingContext, t TestingTB) (string, error) {
```
MStokluska, line 496:
```suggestion
func reset3ScaleSecret(ctx *TestingContext, t TestingTB) error {
```

test/common/3scale_smtp.go
```diff
@@ -49,6 +49,10 @@ var (
 	originalPassword              = ""
 	originalPort                  = ""
 	originalUsername              = ""
+	original3ScalePassword        = ""
```
MStokluska, line 52: hey, not sure if it's valid, but is there any chance these values will not be "" at the time when the test starts? and we are going to change them to "" possibly making this test destructive?

test/common/3scale_smtp.go
```diff
@@ -513,6 +550,39 @@ func resetSecret(ctx *TestingContext, t TestingTB, isCreated bool) (string, erro
 	return "", nil
 }
 
+func patch3ScaleSecret(ctx *TestingContext, t TestingTB) (string, error) {
```
MStokluska, line 553:
```suggestion
func patch3ScaleSecret(ctx *TestingContext, t TestingTB) error {
```

test/common/3scale_smtp.go
makslion: Fair point. Removed the secret from returns

test/common/3scale_smtp.go
makslion: We will always first path 3scale secret (this will populate these) and then reset it (using those to update the secret). To make it destructive we could just add an `if` statement switching off storing original details. I could hardcode them and skip the step of retrieving the original data, but if we will decide to change the default values in that secret - this test will modify the installation.

makslion:
retesting

/test rhoam-e2e multitenant-rhoam-e2e

test/common/3scale_smtp.go
MStokluska: Thanks Maks

PR #2779, opened 2022-07-06
MGDAPI-4310 - Fix commit check

adam-cattermole:
# Issue link
[MGDAPI-4310](https://issues.redhat.com/browse/MGDAPI-4310)

# What
Issue with commit check function - the git repo created by prow does not have remotes and performs a merge, making `git merge-base` return the current PR HEAD. This means the check is effectively skipped on every PR and will pass.
- Add environment variables that use the commit SHA's provided by prow
- Use these references to determine which commits belong to the PR

# Verification steps

### DIFF
--- a/scripts/commits-check.sh
+++ b/scripts/commits-check.sh
@@ -3,9 +3,12 @@
 # This script retrieves the list of commits difference from master
 # and ensures that each commit starts with a MGDAPI- numbered ticket
 # See https://issues.redhat.com/browse/MGDAPI-3097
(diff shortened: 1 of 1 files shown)


adam-cattermole:
Added invalid commit message - check should fail initially

adam-cattermole:
Updated commit message to now be valid - tests should pass

kryanbeane:
Looks good, ci/prow/commits running & passing after the initial fail should be enough for verification
/lgtm

PR #2789, opened 2022-07-13
MGDAPI-4348 update H33 step 3 for correct navigation to ACCOUNTS

austincunningham:
# Issue link

MGDAPI-4348 
# What

MGDAPI-4348 update H33 step 3 for correct navigation to ACCOUNTS
# Verification steps

Screenshot from 2022-07-13 10-40-24


austincunningham:
@trepel minor update to test case

KevFan:
No need to check for code coverage for test case update (failing because PR is behind master) so overriding

/override codecov/project

PR #2790, opened 2022-07-13
[MGDAPI-4354] improve reconcile delay tracking logic

cecobask:
# Issue link
https://issues.redhat.com/browse/MGDAPI-4354

# What
The alert that tracks if the rhmi controller reconcile function takes more than 12 minutes to complete was firing continuously which is not what we want. The issue was caused by not giving enough time to Prometheus to detect if the metric has a non-zero value. To address that I refactored the logic to not track the duration, but only track whether the function has completed in the specified time. The way this works is by using a goroutine which will execute in 12 minutes if the reconcile hasn't completed. The goroutine will be killed if the function completes before the timer runs out, so there's no possibility of false alerts.

# Verification steps

### DIFF
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"github.com/integr8ly/integreatly-operator/pkg/resources/k8s"
 	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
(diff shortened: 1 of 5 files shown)


makslion:
verified that the alert is functional. The fresh install (confirmed by @MStokluska ) and upgrade scenario (confirmed by me) has shown that the alert isn't firing constantly

makslion:
turning a blind eye on the `codecov`: 

/lgtm

PR #2792, opened 2022-07-14
MGDAPI-4357 - Fix script #3 in Test E04B - Verify SLO dashboard - userr-sso statefulset not scaling down

valerymo:
# Issue link
https://issues.redhat.com/browse/MGDAPI-4357
Fix script # 3 in Test E04B - Verify SLO dashboard - user-sso statefulset not scaling down

# What
Issue was in doc (e04b-verify-slo-dashboard.md) of Manual test  `E04B - Verify SLO dashboard`, in script, `Code # 3 `
User-sso statefulset did not scale down, because of incorrect expected replicas number (`3`).
The fix was - change the number of expected replicas to `2` (`grep 2`  instead of `grep 3`).

# Verification steps
It's a Manual test.
All test details are described in doc (file e04b-verify-slo-dashboard.md, in PR)
- To check that User-sso statefulset was scaled to zero:
$ oc get statefulset -n redhat-rhoam-user-sso
NAME READY AGE
keycloak 0/0 ....
-  Check Grafana Dashboard `Critical SLO summary`, see that user-sso panel shows the number of firing alerts.
(all panels should show number of firing alerts except Marin3r)

### DIFF
--- a/test-cases/tests/dashboards/e04b-verify-slo-dashboard.md
+++ b/test-cases/tests/dashboards/e04b-verify-slo-dashboard.md
@@ -84,8 +84,12 @@ while true; do   if oc get deployment keycloak-operator -n redhat-rhoam-rhsso-op
 ### Code #3
 
 ```bash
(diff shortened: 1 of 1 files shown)


trepel:
@valerymo thinking about this a little, this might actually depends on quota (ie 50M quota results in 3 user sso pods while 100k quota results in 2 user sso pods. I agree that using `2` instead of `3` is fine since we typically use small quotas for testing but it might be good to drop a note to test case that the number of replicas depends on quota and might need to be manually adjusted.

valerymo:
@trepel , Thank you for your comment. It's done. Note is added to doc to code #3 script, as in your comment.

Patryk-Stefanski:
@valerymo, Could you also mention that the pods increase from 2 to 3 when the quota is 5 million or higher?  Running `make test-cases/fix` will format your test cases.

trepel:
Thanks guys!

/approve

MStokluska:
by-passing code-cov

PR #2794, opened 2022-07-15
MGDAPI-4309 token retrieval for 4.11 clusters

makslion:
# Issue link
[MGDAPI-4309 ](https://issues.redhat.com/browse/MGDAPI-4309)

# What
Support for login into OSD 4.11 clusters. 

# Verification steps
You would need to verify that it is possible to login into 4.11 clusters as well as 4.10 clusters. The 4.11 oc client is required and can be downloaded from [here](https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.11.0-rc.2/)
## 4.10 cluster
1. Provision a cluster (current default for dev environment in 4.10.20 - will do nicely). 
2. Ensure you have 4.11 oc cli by `oc version`
3. Initiate installation 
4. Once reconcile loops have started - you were able to log into the cluster. 
## 4.11 Cluster
1. Provision cluster. I've used elorean CLI for this. You could get the list of versions by `ocm get "/api/clusters_mgmt/v1/versions?size=2000" | jq '.items[] | select(.id | contains("nightly") | not).id | select (. | contains("v4.11"))'`. Select the latest RC candidate. 
2. Ensure you have 4.11 oc cli by `oc version`
3. Initiate installation 
4. Once reconcile loops have started - you were able to log into the cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,7 +117,7 @@ setup/moq:
 
 .PHONY: setup/service_account/oc_login
 setup/service_account/oc_login:
-	@oc login --token=$(shell sh -c "oc serviceaccounts get-token rhmi-operator -n ${NAMESPACE}") --server=$(shell sh -c "oc whoami --show-server") --kubeconfig=TMP_SA_KUBECONFIG --insecure-skip-tls-verify=true
+	@oc login --token=$(shell sh -c "oc create token rhmi-operator -n ${NAMESPACE} --duration=24h") --server=$(shell sh -c "oc whoami --show-server") --kubeconfig=TMP_SA_KUBECONFIG --insecure-skip-tls-verify=true
 
 .PHONY: setup/service_account
 setup/service_account: kustomize


adam-cattermole:
Going to run e2e - should still be using 4.10 so should be fine
/test rhoam-e2e

Makefile
```diff
@@ -117,7 +117,7 @@ setup/moq:
 
 .PHONY: setup/service_account/oc_login
 setup/service_account/oc_login:
-	@oc login --token=$(shell sh -c "oc serviceaccounts get-token rhmi-operator -n ${NAMESPACE}") --server=$(shell sh -c "oc whoami --show-server") --kubeconfig=TMP_SA_KUBECONFIG --insecure-skip-tls-verify=true
+	@oc login --token=$(shell sh -c "oc create token rhmi-operator") --server=$(shell sh -c "oc whoami --show-server") --kubeconfig=TMP_SA_KUBECONFIG --insecure-skip-tls-verify=true
```
adam-cattermole, line 120: These new tokens have a fixed duration of 3600s by default, so the operator stops working after an hour - this could be a little inconvenient, perhaps setting 24h is preferable, or perhaps even longer? I'm unsure what / if there is a downside to a longer-lasting token.

I also I don't see any harm in specifying the namespace explicitly given the SA is in that namespace instead of assuming we're in it

```suggestion
	@oc login --token=$(shell sh -c "oc create token rhmi-operator -n ${NAMESPACE} --duration=24h") --server=$(shell sh -c "oc whoami --show-server") --kubeconfig=TMP_SA_KUBECONFIG --insecure-skip-tls-verify=true
```

makslion: Hmm, excellent points. 
There is no harm in restricting token to only operator namespace and we should, indeed, allow reconciler to run for more than an hour. Since this token will be only for a local running reconciler and each restart of the reconcile loop will regenerate it - 24 hours is more than enough. Can't imagine someone leaving it running over a night, but I'll drop the question in the general chat - maybe I'm missing something there.

adam-cattermole: > There is no harm in restricting token to only operator namespace

It's not actually restricting the token - the create token command is creating a token for the already existing service account named rhmi-operator so if you switch to a different project after running make cluster/prepare/local this command will fail until you switch back to the correct project.

Yep I think 24h should be fine. I've sometimes left it running overnight but not for any particular reason. The only thing that comes to mind is memory profiling but I'm unsure we do that and I guess that could be done via OLM deploy.

makslion:
addressed comments and accepted the proposed 24h timeout. Running e2e tests 

/test rhoam-e2e

adam-cattermole:
Thanks

/lgtm

MStokluska:
by-passing code-cov

PR #2798, opened 2022-07-18
MGDAPI-4272 / Removed code

den-rgb:
# Issue link
[MGDAPI-4272](https://issues.redhat.com/browse/MGDAPI-4272)

# What
Remove redundant code.

# Verification steps
Make sure prow checks passed

### DIFF
--- a/bundles/managed-api-service/1.24.0/manifests/managed-api-service.clusterserviceversion.yaml
+++ b/bundles/managed-api-service/1.24.0/manifests/managed-api-service.clusterserviceversion.yaml
@@ -51,7 +51,7 @@ metadata:
     capabilities: Basic Install
     categories: Integration & Delivery
     certified: "false"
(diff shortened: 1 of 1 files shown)


pkg/resources/prometheusRules.go
```diff
@@ -68,10 +65,6 @@ func (r *AlertReconcilerImpl) ReconcileAlerts(ctx context.Context, client k8scli
 		}
 	}
 
-	if err := r.deleteAlerts(ctx, client, r.RemovedAlerts); err != nil {
-		return integreatlyv1alpha1.PhaseFailed, err
-	}
-
```
KevFan, line 74: I'd be more in favor of keeping this code and the relevant tests as this provides a way to remove alerts in the future if alerts are renamed / required to be removed :thinking:

den-rgb: sure I could leave it in for the future rather then having to write it again later.

PR #2800, opened 2022-07-18
[MGDAPI-4378] defer stop of reconcileDelayedMetric timer

cecobask:
# Issue link
https://issues.redhat.com/browse/MGDAPI-4378

# What
The second parameter (func) of the AfterFunc runs in its own goroutine and has to be killed before we exit the reconcile function. This can be done by using a defer call to https://pkg.go.dev/time#Timer.Stop after the timer has been setup.

# Verification steps


adam-cattermole:
Verified this change upgrading from current 1.24.0 -> 1.24.1 (this PR)

On 1.24.0 can see the metric flipping between 1 and 0, however once upgraded it stays at 0 as expected and the alert is no longer pending / firing.

reconcile-delayed

The only thing to mention is coverage failing due to 1 additional line of untested code - I would override in my opinion given it should be covered by [MGDAPI-4286](https://issues.redhat.com/browse/MGDAPI-4286).

/lgtm

PR #2841, opened 2022-08-05
MGDAPI-4422 Updating command to use cluster name instead of display_name

Patryk-Stefanski:
# Issue link
[MGDAPI-4422](https://issues.redhat.com/browse/MGDAPI-4422)

# What
While working on MGDAPI-4422  the command to retrieve the cluster-ID didn't work , this was because the display name was slightly different from the name. Seen as the openshift URL uses the actual name I think we should retrieve the cluster id by it name rather than the display  name

# Verification steps
make sure the new command works
```
CLUSTER_ID=$(ocm get clusters --parameter search="name like '<CLUSTER_NAME>'" | jq -r '.items[0].id')
echo $CLUSTER_ID
```

Should return ID of  desired cluster

### DIFF
--- a/test-cases/tests/installation/a34-verify-quota-values.md
+++ b/test-cases/tests/installation/a34-verify-quota-values.md
@@ -121,7 +121,7 @@ CLUSTER_NAME="<CLUSTER_NAME>"
 10. Get cluster id and assign it to a variable
 
 ```bash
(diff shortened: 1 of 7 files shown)


trepel:
Thanks. No other `display_name` is used in test-cases

/lgtm
/approve

PR #2802, opened 2022-07-19
[wip]MGDAPI-3929 refactor mirroing alerts logic

MStokluska:
# Issue link

# What

# Verification steps

PR #2842, opened 2022-08-05
MGDAPI-4375 / Added CriticalMetricsMissing Alert

den-rgb:
# Issue link
[MGDAPI-4375](https://issues.redhat.com/browse/MGDAPI-4375)

# What
Added an alert as well as a SOP to check if Critical metrics are missing.

# Verification steps
1. Checkout this branch.
2. Install rhoam through OLM.
3. Make sure alert is not triggered during install.
4. Once installed is finished disable rhmi-operator pod and wait for alert to trigger.
5. Enable the pod again.
6. Upgrade the version.
7. Make sure alert doesn't trigger.

- You can use images: 
quay.io/dmoskale/managed-api-service:v1.26.3
quay.io/dmoskale/managed-api-service:v1.26.4

- index: 
quay.io/dmoskale/managed-api-service-index:1.26.3
quay.io/dmoskale/managed-api-service-index:1.26.4

### DIFF
--- a/pkg/products/observability/prometheusRules.go
+++ b/pkg/products/observability/prometheusRules.go
@@ -352,7 +352,7 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
 						"sop_url": resources.SopUrlCriticalMetricsMissing,
 						"message": fmt.Sprintf("%s metrics have been missing for more than 5 minutes", strings.ToUpper(installationName)),
 					},
-					Expr:   intstr.FromString(`(` + installationName + `_version{version=~".*", to_version=~""} > 0) * on(pod) group_left(` + installationName + `_custom_domain)(absent(` + installationName + `_custom_domain)) or absent(` + installationName + `_version)`),
+					Expr:   intstr.FromString(fmt.Sprintf(`(%[1]s_version{version=~".*", to_version=~""} > 0) * on(pod) group_left(%[1]s_custom_domain)(absent(%[1]s_custom_domain)) or absent(%[1]s_version)`, installationName)),
 					For:    "5m",
 					Labels: map[string]string{"severity": "critical"},
 				},
--- a/pkg/products/observability/prometheusRules.go
+++ b/pkg/products/observability/prometheusRules.go
@@ -341,6 +341,23 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
 				},
 			},
 		},
+		{
+			AlertName: fmt.Sprintf("%s-missing-metrics", installationName),
+			Namespace: namespace,
+			GroupName: fmt.Sprintf("%s-general.rules", installationName),
+			Rules: []monitoringv1.Rule{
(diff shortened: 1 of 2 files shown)


kryanbeane:
/lgtm
Alert fired when RHMI operaotr pod was scaled down, and didn't fire during upgrade or install

pkg/products/observability/prometheusRules.go
```diff
@@ -341,6 +341,23 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
...
+			Namespace: namespace,
+			GroupName: fmt.Sprintf("%s-general.rules", installationName),
+			Rules: []monitoringv1.Rule{
+				{
+					Alert: fmt.Sprintf("%sCriticalMetricsMissing", strings.ToUpper(installationName)),
+					Annotations: map[string]string{
+						"sop_url": resources.SopUrlCriticalMetricsMissing,
+						"message": fmt.Sprintf("%s metrics have been missing for more than 5 minutes", strings.ToUpper(installationName)),
+					},
+					Expr:   intstr.FromString(`(` + installationName + `_version{version=~".*", to_version=~""} > 0) * on(pod) group_left(` + installationName + `_custom_domain)(absent(` + installationName + `_custom_domain)) or absent(` + installationName + `_version)`),
```
MStokluska, line 355: is the empty space before the installationName necessary?

MStokluska: nvm! :man_facepalming:

MStokluska: Although I would keep it the same as all other alerts; using the fmt.Sprintf if you can like here:
https://github.com/integr8ly/integreatly-operator/pull/2802/files#diff-7139745d8ee64821c1d2176216e15673c888f7cd064df149a05527f34027efd4R81 to keep consistent

den-rgb: I don't mind changing it but most of the alerts above it don't use the fmt.Sprintf in the experession

MStokluska: @den-rgb most of them (if not all) do use fmt.Sprintf if we need to pass in a variable to the query. I'd like this alert to stay consistent with others if possible.

MStokluska: also, looking at only the file you are changing, notice how the other alerts are using it; it is not used when we don't need to take in a variable

MStokluska: and just to re-iterate; this is not a big deal and the only reason I've picked that up is that I made a mistake when reading it for the first time, due to it being constructed in a different way than others.

kryanbeane:
@den-rgb will leave lgtm again since the change was minor and doesn't need rereview, but can you fix the new commit message to have `MGDAPI-4375`

kryanbeane:
/lgtm
since last change was minor formatting edit, shouldn't require reverification

PR #2851, opened 2022-08-11
[MGDAPI-4458] skip custom domain checks if 3scale is installed

cecobask:
# Issue link
https://issues.redhat.com/browse/MGDAPI-4458

# What
It was noticed that alerts were firing and the rhmi cr was not reconciling in the following scenario: adding a custom domain parameter to the addon-params secret after 3scale has been installed.
In this scenario RHOAM should ignore the custom domain. It should fully reconcile RHOAM and no alerts should be firing. 

# Verification steps
1. Provision/borrow an OSD cluster
2. `make cluster/prepare` from the master branch
3. Create a CatalogSource CR and install RHOAM from it:
```
oc apply -f - <<EOF
---
kind: CatalogSource
metadata:
  name: rhmi-operators
spec:
  sourceType: grpc
  image: quay.io/tdimov0/managed-api-service-index:1.25.0
EOF
```
4. Wait for the RHOAM installation to complete or at least until the 3scale-operator namespace is created
5. Patch the addon params secret with a custom domain value
```
echo -n "apps.rhmi.me" | base64 | xargs -I {} \
oc patch secret addon-managed-api-service-parameters \
-n redhat-rhoam-operator --type=json \
-p='[{"op": "replace", "path": /data/custom-domain_domain, "value": "{}"}]'
```
6. Observe no changes to the RHMI CR relating to custom domain (`customDomain` property under status block) after a successful reconcile.
7. No alerts firing after addon param patch

### DIFF
--- a/controllers/rhmi/bootstrapReconciler.go
+++ b/controllers/rhmi/bootstrapReconciler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	customDomain "github.com/integr8ly/integreatly-operator/pkg/resources/custom-domain"
+	"github.com/integr8ly/integreatly-operator/test/common"
 	v1 "github.com/openshift/api/config/v1"
 	"math/rand"
 	"os"
@@ -612,7 +613,6 @@ func (r *Reconciler) reconcileAddonManagedApiServiceParameters(ctx context.Conte
 }
 
 func (r *Reconciler) retrieveConsoleURLAndSubdomain(ctx context.Context, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
-
 	consoleRouteCR, err := getConsoleRouteCR(ctx, serverClient)
 	if err != nil {
 		if k8serr.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


controllers/rhmi/bootstrapReconciler.go
```diff
@@ -623,27 +623,34 @@ func (r *Reconciler) retrieveConsoleURLAndSubdomain(ctx context.Context, serverC
...
+					r.installation.Status.LastError = err.Error()
+				}
+			} else {
+				r.installation.Spec.RoutingSubdomain = strings.TrimPrefix(consoleRouteCR.Status.Ingress[0].RouterCanonicalHostname, "router-default.")
+			}
+			return integreatlyv1alpha1.PhaseCompleted, nil
 		}
-	} else {
-		r.installation.Spec.RoutingSubdomain = strings.TrimPrefix(consoleRouteCR.Status.Ingress[0].RouterCanonicalHostname, "router-default.")
+		log.Warning(err.Error())
```
laurafitzgerald, line 650: If the original Get on the threescale namespace is something other than IsNotFound should we return because otherwise we may end up resetting the RoutingSubdomain

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -623,27 +623,34 @@ func (r *Reconciler) retrieveConsoleURLAndSubdomain(ctx context.Context, serverC
...
+				}
+				r.installation.Status.CustomDomain.Enabled = true
+				if err != nil {
+					r.installation.Status.CustomDomain.Error = err.Error()
+					r.installation.Status.LastError = err.Error()
+				}
+			} else {
+				r.installation.Spec.RoutingSubdomain = strings.TrimPrefix(consoleRouteCR.Status.Ingress[0].RouterCanonicalHostname, "router-default.")
+			}
+			return integreatlyv1alpha1.PhaseCompleted, nil
```
laurafitzgerald, line 648: A code neatness suggestion. Not so important to fix.

```suggestion
			found, domain, err := customDomain.GetDomain(ctx, serverClient, r.installation)
			if err != nil {
				log.Warning(err.Error())
				return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("customDomain.GetDomain() failure: %w", err)
			}
			if found {
				r.installation.Spec.RoutingSubdomain = domain
				if r.installation.Status.CustomDomain == nil {
					r.installation.Status.CustomDomain = &integreatlyv1alpha1.CustomDomainStatus{}
				}
				r.installation.Status.CustomDomain.Enabled = true
				if err != nil {
					r.installation.Status.CustomDomain.Error = err.Error()
					r.installation.Status.LastError = err.Error()
				}
			if !found {
				r.installation.Spec.RoutingSubdomain = strings.TrimPrefix(consoleRouteCR.Status.Ingress[0].RouterCanonicalHostname, "router-default.")
			}
			return integreatlyv1alpha1.PhaseCompleted, nil
```

PR #2865, opened 2022-08-18
MGDAPI-1729 fix H34 functional test and bring sts branch up to date with master

cecobask:
# Issue link
https://issues.redhat.com/browse/MGDAPI-1729

# What
The H34 functional test (verify 3scale custom SMTP full config) is failing on ROSA STS clusters installed using Jenkins pipeline [var installRhmi](https://gitlab.cee.redhat.com/integreatly-qe/ci-cd/-/blob/master/vars/installRhmi.groovy). The failure reason is that is still using the old rhmi stages structure, before single stage installation stage was introduced.

This PR also brings in latest commits from master.

# Verification steps


controllers/rhmi/bootstrapReconciler.go
```diff
@@ -612,26 +612,25 @@ func (r *Reconciler) reconcileAddonManagedApiServiceParameters(ctx context.Conte
 }
 
 func (r *Reconciler) retrieveConsoleURLAndSubdomain(ctx context.Context, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
-
+	if r.installation.Spec.RoutingSubdomain != "" {
+		log.Info("routing subdomain already set, exiting retrieveConsoleURLAndSubdomain() early")
+		return integreatlyv1alpha1.PhaseCompleted, nil
+	}
```
KevFan, line 618: I'm not too sure about this :thinking:  Would this not prevent using the custom domain work if the customer decides to use that at a later stage after the intial install ? :thinking:

cecobask: This was brought from rebase with master. It was decided that custom domain can only be specified before the RHOAM installation has begun

KevFan: Oh cool, thanks :+1:

PR #2874, opened 2022-08-22
MGDAPI-4478 - Increase UpgradeExpectedDuration - update alert RHOAMUp…

valerymo:
# Issue link
https://issues.redhat.com/browse/MGDAPI-4478
Increase UpgradeExpectedDuration - update alert RHOAMUpgradeExpectedDuration30minExceeded

# What
- change RHOAMUpgradeExpectedDuration30minExceeded Alert, to increase Rhoam Upgrade Expected Duration time from 30 min to 60 min.
- rename alert to RHOAMUpgradeExpectedDuration60minExceeded
- delete RHOAMUpgradeExpectedDuration30minExceeded alert that remained after upgrade from previous version.

## What was tested
- Pipeline https://master-jenkins-csb-intly.apps.ocp-c1.prod.psi.redhat.com/job/Playground/job/valerym/job/managed-api-upgrade-4390/30/   (builds 29 (install) and 30 upgrade)
- new alert name and properties
- old alert with name *30min* not present  (deleted) after upgrade

# Verification steps
- install rhoam (could be locally) and check that Alert name changed (30->60 min), and properties of alert changed - waiting time and description - 60 min. Note - please see alert in Prometheus in `openshift-monitoring` namespace
- (optional) run pipeline `managed-api-upgrade` and check upgrade time and alert behavior

### DIFF
--- a/pkg/resources/prometheusRules.go
+++ b/pkg/resources/prometheusRules.go
@@ -68,6 +68,13 @@ func (r *AlertReconcilerImpl) ReconcileAlerts(ctx context.Context, client k8scli
 		}
 	}
 
+	//removing RHOAMUpgradeExpectedDuration30minExceeded alert as it was renamed (30min -> 60min)
+	alertToDelete := AlertConfiguration{
+		AlertName: "RHOAMUpgradeExpectedDuration30minExceeded",
+		Namespace: "openshift-monitoring",
+	}
+	r.RemovedAlerts = append(r.RemovedAlerts, alertToDelete)
+
 	if err := r.deleteAlerts(ctx, client, r.RemovedAlerts); err != nil {
 		return integreatlyv1alpha1.PhaseFailed, err
 	}
--- a/controllers/rhmi/prometheusRules.go
+++ b/controllers/rhmi/prometheusRules.go
@@ -68,13 +68,13 @@ func (r *RHMIReconciler) newAlertsReconciler(installation *integreatlyv1alpha1.R
 					Labels: map[string]string{"severity": "warning", "product": installationName, "addon": getAddonName(installation), "namespace": "openshift-monitoring"},
 				},
 				{
-					Alert: fmt.Sprintf("%sUpgradeExpectedDuration30minExceeded", strings.ToUpper(installationName)),
+					Alert: fmt.Sprintf("%sUpgradeExpectedDuration60minExceeded", strings.ToUpper(installationName)),
 					Annotations: map[string]string{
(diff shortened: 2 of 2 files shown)


pat-cremin (approved):
Approved.

pkg/resources/prometheusRules.go
```diff
@@ -50,7 +50,13 @@ type AlertConfiguration struct {
 func (r *AlertReconcilerImpl) ReconcileAlerts(ctx context.Context, client k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
 	// If the installation was marked for deletion, delete the alerts
 	if r.Installation.DeletionTimestamp != nil {
-		allAlerts := append(r.Alerts, r.RemovedAlerts...)
+		//removing RHOAMUpgradeExpectedDuration30minExceeded alert as it was renamed (30min -> 60min)
+		alertToDelete := AlertConfiguration{
```
MStokluska, line 54: @valerymo this will only work with when the deletionTimestamp is present. Therefore the alert to be deleted will only be added to the r.RemovedAlerts will only be updated after the deletion of RHOAM is triggered.

valerymo: @MStokluska Fixed, hope better now. Thank you for comment and review

pkg/resources/prometheusRules.go
```diff
@@ -68,6 +68,13 @@ func (r *AlertReconcilerImpl) ReconcileAlerts(ctx context.Context, client k8scli
 		}
 	}
 
+	//removing RHOAMUpgradeExpectedDuration30minExceeded alert as it was renamed (30min -> 60min)
+	alertToDelete := AlertConfiguration{
+		AlertName: "RHOAMUpgradeExpectedDuration30minExceeded",
+		Namespace: "openshift-monitoring",
```
MStokluska, line 74: Sorry Valery just noticed now..can we use here observability.OpenshiftMonitoringNamespace instead of having a string

valerymo: Thanks @MStokluska , I actually tried do it initially, but have some very strange error, so used hardcoded. I will try again, and if not work, I will share the error

MStokluska: Actually, that would be an import cycle; since this is a temporary code let's go with it

MStokluska:
Good job Valery, verified on upgrade from 1.25.0 to 1.26, old alert is gone and new one is in as expected.

/lgtm

MStokluska:
by-passing codecov until we progress the prom unit testing

PR #2901, opened 2022-09-01
MGDAPI-4530 refactoring tests regarding deletion of users

Fiona-Waters:
# Issue link
[MGDAPI-4530
](https://issues.redhat.com/browse/MGDAPI-4530)

# What
cleanUpTestDedicatedAdminUsersSyncedSSO function has been refactored to poll the keycloak user cr to ensure it has been deleted as expected.

# Verification steps
 - This can be verified by successfully running the test suite(or specifically B09 and C19) on a cluster with RHOAM installed.

### DIFF
--- a/test/common/user_dedicated_admin_synced_sso.go
+++ b/test/common/user_dedicated_admin_synced_sso.go
@@ -18,6 +18,7 @@ import (
 	userv1 "github.com/openshift/api/user/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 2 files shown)


test/common/user_dedicated_admin_synced_sso.go
```diff
@@ -431,15 +431,18 @@ func cleanUpTestDedicatedAdminUsersSyncedSSO(ctx context.Context, t TestingTB, c
...
-			Namespace: RHSSOProductNamespace,
-		},
+	// Ensure KeycloakUser CR has been deleted within 2 minutes
+	kcUser := &keycloak.KeycloakUser{}
+	err = wait.Poll(time.Second*15, time.Minute*2, func() (done bool, err error) {
+		err = c.Get(ctx, types.NamespacedName{Name: fmt.Sprintf("%s-%s", TestingIDPRealm, testUserName),
+			Namespace: RHSSOProductNamespace}, kcUser)
+		if err == nil {
+			return false, nil
+		}
```
MStokluska, line 441: I think we should continue only if kc user cr is gone, so only continue if the err != nil && error is IsNotFound

MStokluska: Reason for it is that the error could be for example for another reason, like interrupted connection for example

MStokluska:
Thanks Fiona,
Verified with Fiona on her cluster.
/lgtm

KevFan:
Code coverage not required as it is an e2e test update - overriding

/override codecov/project

PR #2909, opened 2022-09-06
MGDAPI-4565 add permissions to update subscriptions in other namespaces

austincunningham:
# Issue link

MGDAPI-4565

# What

hive now creates the CRO namespace on stage so rhmi-operator service account no longer has access to create subscription to namespaces it doesn't own, adding permissions to allow subscription updates 

# Verification steps

### DIFF
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -175,6 +175,7 @@ func New(mgr ctrl.Manager) *RHMIReconciler {
 // +kubebuilder:rbac:groups=operators.coreos.com,resources=catalogsources;operatorgroups,verbs=create;list;get
 // +kubebuilder:rbac:groups=operators.coreos.com,resources=catalogsources,verbs=update,resourceNames=rhmi-registry-cs
 // +kubebuilder:rbac:groups=operators.coreos.com,resources=installplans,verbs=update
+// +kubebuilder:rbac:groups=operators.coreos.com,resources=subscriptions,verbs=update
 
 // Monitoring resources not covered by namespace "admin" permissions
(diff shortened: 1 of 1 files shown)


bundles/managed-api-service/1.27.0/manifests/managed-api-service.clusterserviceversion.yaml
```diff
@@ -57,692 +57,694 @@ metadata:
...
-    serviceAffecting: "true"
```
KevFan, line 63: Would this prevent the CVE pipeline from working correctly ? :thinking:

austincunningham: possibly can add it back in, its all generated from the make bundle command but I have had to make a number of changes after it was run

KevFan: Oh okay, I think it's the `make release` command that adds the container images annotation :thinking: 
```
 OLM_TYPE=managed-api-service SEMVER=1.27.0 IMAGE_TAG=master SERVICE_AFFECTING=true make release/prepare 
```

austincunningham: added them back in manually, not sure why github is not showing this as outdated as its gone as a change in the files view.

KevFan:
Verified addition `Subscription` permission is added to the operator `ClusterRole` :+1: 

/lgtm

austincunningham:
code cov fails on the rhmi controller as the file was touched to add the rbac permissions via comment. No new code in the rhmi controller so will override codcov
/override codecov/project

PR #2916, opened 2022-09-08
MGDAPI-4498 - fix: prevent using custom domain when addon param is empty string

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-4498

# What

In the bootstrap reconciler, there is no check for the domain parameter to be an empty string. Because of this, if an empty string is in the domain field in the addon parameters secret, the bootstrap reconciler will set the installation as a custom domain installation. 

This will prevent the installation from completing due to being an invalid domain.

What should happen is that if there custom domain is empty on installation, the default cluster route should be used instead

# Verification steps

* Passing prow checks with e2e passing should be sufficient

However if you want to do a manual verification:
* Checkout this branch
* Install RHOAM
```
make cluster/prepare/local
make code/run
```
* Verifiy installation completes successfully

### DIFF
--- a/controllers/rhmi/bootstrapReconciler.go
+++ b/controllers/rhmi/bootstrapReconciler.go
@@ -621,11 +621,12 @@ func (r *Reconciler) retrieveConsoleURLAndSubdomain(ctx context.Context, serverC
 	}
 	r.installation.Spec.MasterURL = consoleRouteCR.Status.Ingress[0].Host
 	ok, domain, err := customDomain.GetDomain(ctx, serverClient, r.installation)
+	// Only fail when unable to get custom domain parameter from the addon secret to allow for installation of monitoring stack
 	if err != nil && !ok {
 		log.Warning(err.Error())
 		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("customDomain.GetDomain() failure: %w", err)
 	}
-	if ok {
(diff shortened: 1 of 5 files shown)


controllers/rhmi/bootstrapReconciler.go
```diff
@@ -621,11 +621,12 @@ func (r *Reconciler) retrieveConsoleURLAndSubdomain(ctx context.Context, serverC
 	}
 	r.installation.Spec.MasterURL = consoleRouteCR.Status.Ingress[0].Host
 	ok, domain, err := customDomain.GetDomain(ctx, serverClient, r.installation)
+	// Only fail when unable to get custom domain parameter from the addon secret to allow for installation of monitoring stack
 	if err != nil && !ok {
 		log.Warning(err.Error())
 		return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("customDomain.GetDomain() failure: %w", err)
 	}
-	if ok {
+	if ok && domain != "" {
```
austincunningham, line 629: Yep that looks like it will fix it

Boomatang:
/lgtm
Was able to reproduce the issue and the pr fixed it.

austincunningham:
/approve
override codcov looks like one line that's variable assignment not covered 
/override codcov/project

PR #2931, opened 2022-09-27
MGDAPI-4560 Enhanced APIManagementTenantCRFailed alert

carlkyrillos:
# Issue link
JIRA: https://issues.redhat.com/browse/MGDAPI-4560

# What
This PR removes the old tenant metrics and replaces them with one that is able to return labels that provide the alert with more information about the failing CR(s).

# Verification steps
Passing the e2e tests should ensure that this change doesn't break any existing functionality but it can be also verified using the following steps:

1. Checkout this PR
2. Build the operator, bundle, and index images using `INSTALLATION_TYPE=multitenant-managed-api`. Before pushing the images to quay, confirm that the operator image is for `quay.io/<username>/managed-api-service` and the bundle/index images are for `quay.io/<username>/multitenant-managed-api-service`
3. Prepare the cluster for a multitenant installation: `INSTALLATION_TYPE=multitenant-managed-api LOCAL=false make cluster/prepare/local`
4. Install Multitenant RHOAM using a CatalogSource and installing through the OperatorHub in the `sandbox-rhoam-operator` Namespace.
5. Once the installation is completed, create a Namespace called `bad-namespace`
6. Create a APIManagementTenant CR in that Namespace:
```yaml
kind: APIManagementTenant
metadata:
  name: example
  namespace: sandbox-rhoam-observability
spec: {}
```
7. Navigate to the Prometheus route in the `sandbox-rhoam-observability` Namespace
8. Verify that the `APIManagementTenantCRFailed` alert is firing and that the alert contains a number of labels with relevant information about the failing CR. Note that it might take a minute or so for the alert to recognize the new CR.
9. Navigate the the Grafana route in the `sandbox-rhoam-observability` Namespace and verify that the `Multitenant Detailed` dashboard is showing data.


controllers/rhmi/bootstrapReconciler.go
```diff
@@ -299,22 +299,14 @@ func (r *Reconciler) deleteObsoleteService(ctx context.Context, serverClient k8s
...
-	if err != nil {
-		return err
-	}
-	reconciled, err := userHelper.GetReconciledAPIManagementTenantsCount(ctx, serverClient)
-	if err != nil {
-		return err
-	}
-	failed, err := userHelper.GetFailedAPIManagementTenantsCount(ctx, serverClient)
+	tenants := &integreatlyv1alpha1.APIManagementTenantList{}
+	err := serverClient.List(ctx, tenants)
```
MStokluska, line 303: I wonder, would this be better in apimanagement tenant controller? What is the reason for it being in the bootstrap controller?

MStokluska:
/lgtm
verified as per verification steps, all looks good!

controllers/rhmi/bootstrapReconciler.go
carlkyrillos: After discussing this offline we decided to keep the implementation as is and we can move it the ApiMangementTenant controller later if needed.

PR #2934, opened 2022-09-28
MGDAPI-1727 remove arn checks from feature branch for sts

austincunningham:
# Issue link

[MGDAPI-1727](https://issues.redhat.com/browse/MGDAPI-1727)

# What

Remove arn checks as the secret is now created by a credentialsRequest in managed tenants 

# Verification steps
The steps taken to install the addon with the feature branch

1. install the addon on an sts cluster with the rosa cli
```bash
rosa install addon --cluster <cluster_name> managed-api-service -y --cidr-range 10.1.0.0/26 --notification-email <your_email>@redhat.com --addon-managed-api-service 10 --addon-resource-required true --s3-access-key-id <aws_access_key> --s3-secret-access-key <aws_access_secret>
```
2. Patch the rhoam operator CSV with the following rbac (due to the bundle in the old addon not having these rbac permissions)
```yaml
- apiGroups: 
  - operator.openshift.io
  resources: 
  - cloudcredentials
  verbs: 
  - get
  - list
  - watch
- apiGroups:
  - operators.coreos.com
  resources:
  - subscriptions
  verbs:
  - update
 ```
3. patch the managed-api-service cluster role with the same rbac 

4. build a operator image off the dev branch
```bash
export ORG=austincunningham
OLM_TYPE=managed-api-service INSTALLATION_TYPE=managed-api make image/build
OLM_TYPE=managed-api-service INSTALLATION_TYPE=managed-api make image/push
```

5. Run the patch-image-csv.sh from the /scripts directory to patch the operator image

```bash
# e.g.
export VERSION=v1.27.0
export ADDON_VERSION=v1.26.0
export USE_CLUSTER_STORAGE=false
./scripts/patch-image-csv.sh
```
6. Confirm that the operator installs as expected

### DIFF
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -1094,12 +1094,7 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
 		return result, err
 	}
 	if isSTS {
-		log.Info("validation of STS role ARN parameter ")
-		validArn, err := checkStsCredentialsPresent(r.Client, installation.Spec.NamespacePrefix+"-cloud-resources-operator")
-		if err != nil || !validArn {
-			log.Error("STS role ARN parameter pattern validation failed", err)
-			return result, err
-		}
+		log.Info("STS mode enabled for cluster")
 	}
 
 	installation.Status.PreflightStatus = rhmiv1alpha1.PreflightSuccess
@@ -1682,18 +1677,6 @@ func getInstallation() (*rhmiv1alpha1.RHMI, error) {
 			Type: installType,
 		},
 	}, nil
-}
-
-// function is checking if STS secret exists
-func checkStsCredentialsPresent(client k8sclient.Client, namespace string) (bool, error) {
-	stsCredentials := &corev1.Secret{}
-	err := client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: sts.CredsSecretName}, stsCredentials)
-
-	if err != nil {
-		return false, fmt.Errorf("failed to get %s secret", sts.CredsSecretName)
-	}
-
-	return true, nil
 }
 
 func formatAlerts(alerts []prometheusv1.Alert) (critical resources.AlertMetrics, warning resources.AlertMetrics) {
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -1095,7 +1095,7 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
 	}
 	if isSTS {
 		log.Info("validation of STS role ARN parameter ")
-		validArn, err := sts.ValidateAddOnStsRoleArnParameterPattern(r.Client, installation.Namespace)
+		validArn, err := checkStsCredentialsPresent(r.Client, installation.Spec.NamespacePrefix+"-cloud-resources-operator")
 		if err != nil || !validArn {
 			log.Error("STS role ARN parameter pattern validation failed", err)
 			return result, err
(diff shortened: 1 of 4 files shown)


controllers/rhmi/rhmi_controller.go
```diff
@@ -1095,7 +1095,7 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
 	}
 	if isSTS {
 		log.Info("validation of STS role ARN parameter ")
-		validArn, err := sts.ValidateAddOnStsRoleArnParameterPattern(r.Client, installation.Namespace)
+		validArn, err := checkStsCredentialsPresent(r.Client, installation.Spec.NamespacePrefix+"-cloud-resources-operator")
```
laurafitzgerald, line 1098:
```suggestion
		validArn, err := checkStsCredentialsPresent(r.Client, installation.Spec.NamespacePrefix+"cloud-resources-operator")
```
I think the prefix usually contains the `-`

pkg/products/cloudresources/reconciler.go
```diff
@@ -616,3 +616,13 @@ func (r *Reconciler) reconcileCIDRValue(ctx context.Context, client k8sclient.Cl
 
 	return client.Patch(ctx, cfgMap, k8sclient.Merge)
 }
+
+// createSTSARNSecret create the STS arn secret - should be already validated in preflight checks
+func (r *Reconciler) copySTSSecret(ctx context.Context, client k8sclient.Client, operatorNamespace string) (integreatlyv1alpha1.StatusPhase, error) {
```
laurafitzgerald, line 621: any benefit to putting this in the STS package?

controllers/rhmi/rhmi_controller.go
laurafitzgerald: Are we losing the validate function here?

pkg/products/cloudresources/reconciler.go
```diff
@@ -616,3 +616,13 @@ func (r *Reconciler) reconcileCIDRValue(ctx context.Context, client k8sclient.Cl
...
+	err := resources.CopySecret(ctx, client, sts.CredsSecretName, operatorNamespace, sts.CredsSecretName, r.Config.GetOperatorNamespace())
```
KevFan, line 622:
```suggestion
	err := resources.CopySecret(ctx, client, sts.CredsSecretName, operatorNamespace, sts.CredsSecretName, r.ConfigManager.GetOperatorNamespace())
```
Should this be using the `ConfigManager` instead for the rhoam operator namespace ? 🤔

controllers/rhmi/rhmi_controller.go
austincunningham: yea have it change locally

I don't think so this check was originally for when we were passing in an arn rather that letting managed tenants create the arn.

pkg/products/cloudresources/reconciler.go
austincunningham: no objection to changing it, but at the end of the day the result is the same we get the operator namespace.

pkg/products/cloudresources/reconciler.go
austincunningham: actually wondering if there is any benefit to the function as a whole, if hive is creating the secret in the CRO ns then we don't need this move. Its hanging around since before we came up with the idea to have managed tenants create the secret directly in CRO.

austincunningham: will know more after testing

pkg/products/cloudresources/reconciler.go
austincunningham: may be removing this function

austincunningham: actually will refactor to check for the secret instead

controllers/rhmi/rhmi_controller.go
austincunningham: going to remove this functionally as it relied on CRO ns being created by managed tenants/hive

pkg/products/cloudresources/reconciler.go
austincunningham: actually will refactor to check for the secret instead

KevFan:
* Verified installation completes
* Verified uninstallation
  * However addon was stuck on uninstallation, subscription was in an upgrading state, might be related to the selector syncset migration
Screenshot 2022-09-29 at 19 17 34

/lgtm

PR #2940, opened 2022-10-03
MGDAPI-4710 Refactored create bundle script to support arg to not overwrite images

carlkyrillos:
# Issue link
JIRA: https://issues.redhat.com/browse/MGDAPI-4710

# What
Adds support for new command line arg, `OVERWRITE`, which if set to `false` will prevent the script from rebuilding and pushing images that already exist in quay. This parameter is set to false by default.

# Verification steps

1. Checkout this PR
2. Run make command to build and push bundle/index images for first time and verify the images are pushed to your quay org: 
`ORG=<your-quay-org> INSTALLATION_TYPE=managed-api UPGRADE=false BUNDLE_VERSIONS=1.27.0 make create/olm/bundle`
3. Run the command again but set `OVERWRITE=false` or don't set it at all and verify that the existing images in your quay org are not overwritten
4. Run the command one last time but this time either set `OVERWRITE=true` and verify that the existing images in your quay org _are_ overwritten

### DIFF
--- a/scripts/bundle-rhmi-operators.sh
+++ b/scripts/bundle-rhmi-operators.sh
@@ -15,6 +15,7 @@
 # REG - registry of where to push the bundles and indexes, defaults to quay.io
 # BUILD_TOOL - tool used for building the bundle and index, defaults to docker
 # OC_INSTALL - set to true if you want the catalogue source to be created pointing to the "oldest" version within the versions specified (version must have no replaces field)(must be oc logged in)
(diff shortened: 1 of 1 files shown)


Fiona-Waters:
Images are not built or pushed when DONT_OVERWRITE=true 
Images are built/pushed and ones already in repo are overwritten when DONT_OVERWRITE=false or variable not set.
/lgtm

PR #2947, opened 2022-10-10
MGDAPI-4727 Added LimitRanges to each Namespace managed by RHOAM

carlkyrillos:
# Issue link
JIRA: https://issues.redhat.com/browse/MGDAPI-4727

# What
In order to support RHOAM on autoscaled clusters, we need to provide a way to assign default CPU/memory request values for every pod/container that doesn't explicitly set them. This is required by the OpenShift autoscaler so it can accurately calculate each node's resource requirements which is a prerequisite for the autoscaler to scale up/down nodes as needed.

We can enforce default CPU/memory request values by creating a LimitRange object in each namespace that RHOAM manages. Note that the LimitRanges are intentionally set to only request a small amount of resources for each container (5m of CPU and 10 Mi of memory) . This is because certain containers need significantly less resources than others and the LimitRange enforces defaults for every container in the namespace. Therefore if the LimitRanges assigned default values to cater to the most resource intensive container for each namespace, most containers will be assigned unnecessarily large CPU/memory request values. This will make it seem like the RHOAM product requires far more resources than it actually requires and may even break installations as the scheduler will reserve so many resources for containers/pods that don't need them, it wont be able to schedule all the necessary pods.

# Verification steps

1. Checkout this [delorean PR](https://github.com/integr8ly/delorean/pull/284) and provision a ROSA cluster with autoscale enabled:
`CLUSTER_NAME=<cluster-name> ENABLE_AUTOSCALING=true STS_ENABLED=false make ocm/rosa/cluster/create`
2. Once the cluster is ready, install RHOAM via CatalogSource using this index image (you can also build your own images if you'd like just make sure to checkout this PR): `quay.io/ckyrillo/managed-api-service-index:1.28.0`
3. Verify that the installation completes and that a LimitRange object was created in each namespace (there should be 13 in all): `oc get limitranges -A`
4. Lastly, verify that the LimitRanges are correctly assigning default requests to containers that don't explictly set them. This can be done by finding a container on eng-long-lived that doesn't have set requests, for example [this one](https://console-openshift-console.apps.eng-long-lived.9p7x.s1.devshift.org/k8s/ns/redhat-rhoami-observability/pods/prometheus-prometheus-0/containers/blackbox-exporter), and comparing it to the same container on your cluster. The container on your cluster should have these values for its Resource requests: `cpu: 5m, memory: 10Mi`

### DIFF
--- a/controllers/rhmi/bootstrapReconciler.go
+++ b/controllers/rhmi/bootstrapReconciler.go
@@ -113,6 +113,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 	}
 	// MGDAPI-4641 block end
 
+	phase, err = resources.ReconcileLimitRange(ctx, serverClient, r.installation.Namespace, resources.DefaultLimitRangeParams)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", r.installation.Namespace), err)
+		return phase, err
+	}
+
 	phase, err = r.reconcileOauthSecrets(ctx, serverClient)
 	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
 		events.HandleError(r.recorder, installation, phase, "Failed to reconcile oauth secrets", err)
--- a/pkg/products/cloudresources/reconciler.go
+++ b/pkg/products/cloudresources/reconciler.go
@@ -164,6 +164,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
+	phase, err = resources.ReconcileLimitRange(ctx, client, operatorNamespace, resources.DefaultLimitRangeParams)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", operatorNamespace), err)
+		return phase, err
+	}
+
 	// Check if STS Cluster, get STS role ARN addon parameter and pass ARN to Secret in CRO namespace
 	isSTS, err := sts.IsClusterSTS(ctx, client, r.log)
 	if err != nil {
--- a/pkg/products/marin3r/reconciler.go
+++ b/pkg/products/marin3r/reconciler.go
@@ -165,6 +165,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
+	phase, err = resources.ReconcileLimitRange(ctx, client, operatorNamespace, resources.DefaultLimitRangeParams)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", operatorNamespace), err)
+		return phase, err
+	}
+
+	phase, err = resources.ReconcileLimitRange(ctx, client, productNamespace, resources.DefaultLimitRangeParams)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", productNamespace), err)
+		return phase, err
+	}
+
 	phase, err = r.reconcileSubscription(ctx, client, operatorNamespace, operatorNamespace)
 	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
 		events.HandleError(r.recorder, installation, phase, fmt.Sprintf("Failed to reconcile %s subscription", constants.CloudResourceSubscriptionName), err)
--- a/pkg/products/rhsso/reconciler.go
+++ b/pkg/products/rhsso/reconciler.go
@@ -145,6 +145,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
+	phase, err = resources.ReconcileLimitRange(ctx, serverClient, operatorNamespace, resources.DefaultLimitRangeParams)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.Recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", operatorNamespace), err)
+		return phase, err
+	}
+
+	phase, err = resources.ReconcileLimitRange(ctx, serverClient, productNamespace, resources.DefaultLimitRangeParams)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.Recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", productNamespace), err)
+		return phase, err
+	}
+
 	phase, err = resources.ReconcileSecretToProductNamespace(ctx, serverClient, r.ConfigManager, adminCredentialSecretName, productNamespace, r.Log)
 	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
 		events.HandleError(r.Recorder, installation, phase, "Failed to reconcile admin credentials secret", err)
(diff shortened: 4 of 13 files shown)


pkg/products/threescale/reconciler.go
```diff
@@ -260,6 +260,28 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+
+	productLimitRangeParams := resources.LimitRangeParams{
+		CpuRequest:    "5m",
+		MemoryRequest: "10Mi",
+		ResourceType:  corev1.LimitTypeContainer,
+	}
+	phase, err = resources.ReconcileLimitRange(ctx, serverClient, productNamespace, productLimitRangeParams)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", operatorNamespace), err)
+		return phase, err
```
KevFan, line 282:
```suggestion
	phase, err = resources.ReconcileLimitRange(ctx, serverClient, productNamespace, productLimitRangeParams)
	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
		events.HandleError(r.recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", productNamespace), err)
		return phase, err
```

pkg/products/rhssouser/reconciler.go
```diff
@@ -181,6 +182,28 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+	productLimitRangeParams := resources.LimitRangeParams{
+		CpuRequest:    "5m",
+		MemoryRequest: "10Mi",
+		ResourceType:  corev1.LimitTypeContainer,
+	}
+	phase, err = resources.ReconcileLimitRange(ctx, serverClient, productNamespace, productLimitRangeParams)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.Recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", operatorNamespace), err)
+		return phase, err
+	}
```
KevFan, line 205:
```suggestion
	phase, err = resources.ReconcileLimitRange(ctx, serverClient, productNamespace, productLimitRangeParams)
	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
		events.HandleError(r.Recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", productNamespace), err)
		return phase, err
	}
```

pkg/products/rhsso/reconciler.go
```diff
@@ -145,6 +145,28 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+	productLimitRangeParams := resources.LimitRangeParams{
+		CpuRequest:    "5m",
+		MemoryRequest: "10Mi",
+		ResourceType:  corev1.LimitTypeContainer,
+	}
+	phase, err = resources.ReconcileLimitRange(ctx, serverClient, productNamespace, productLimitRangeParams)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.Recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", operatorNamespace), err)
+		return phase, err
+	}
```
KevFan, line 168:
```suggestion
	phase, err = resources.ReconcileLimitRange(ctx, serverClient, productNamespace, productLimitRangeParams)
	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
		events.HandleError(r.Recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", productNamespace), err)
		return phase, err
	}
```

pkg/products/marin3r/reconciler.go
```diff
@@ -165,6 +165,28 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+	productLimitRangeParams := resources.LimitRangeParams{
+		CpuRequest:    "5m",
+		MemoryRequest: "10Mi",
+		ResourceType:  corev1.LimitTypeContainer,
+	}
+	phase, err = resources.ReconcileLimitRange(ctx, client, productNamespace, productLimitRangeParams)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", operatorNamespace), err)
+		return phase, err
+	}
```
KevFan, line 188:
```suggestion
	phase, err = resources.ReconcileLimitRange(ctx, client, productNamespace, productLimitRangeParams)
	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
		events.HandleError(r.recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", productNamespace), err)
		return phase, err
	}
```

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -113,6 +113,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 	}
 	// MGDAPI-4641 block end
 
+	operatorLimitRangeParams := resources.LimitRangeParams{
+		CpuRequest:    "5m",
+		MemoryRequest: "10Mi",
+		ResourceType:  corev1.LimitTypeContainer,
+	}
```
KevFan, line 120: Since these values are used throughout each reconciler and acts as a default, any thoughts on just externalising this to a `var` called `defaultLimitRangeParams` like in https://github.com/carlkyrillos/integreatly-operator/blob/e7ccc0c79e56ab343b6eb27be33727f58ad7fd4e/pkg/resources/limitRange.go#L25 and used as an exported parameter in the reconcile function 🤔

controllers/rhmi/bootstrapReconciler.go
```diff
@@ -113,6 +113,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+	phase, err = resources.ReconcileLimitRange(ctx, serverClient, r.installation.Namespace, operatorLimitRangeParams)
```
KevFan, line 121:
```suggestion
	phase, err = resources.ReconcileLimitRange(ctx, serverClient, r.installation.Namespace, resources.DefaultLimitRangeParams)
```
Example if you adapt to use the exported parameter 🤔

pkg/products/cloudresources/reconciler.go
```diff
@@ -164,6 +164,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+	operatorLimitRangeParams := resources.LimitRangeParams{
+		CpuRequest:    "5m",
+		MemoryRequest: "10Mi",
+		ResourceType:  corev1.LimitTypeContainer,
+	}
+	phase, err = resources.ReconcileLimitRange(ctx, client, operatorNamespace, operatorLimitRangeParams)
+	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+		events.HandleError(r.recorder, installation, phase, fmt.Sprintf("Failed to reconcile LimitRange for Namespace %s", operatorNamespace), err)
+		return phase, err
+	}
```
KevFan, line 176: Just a thought, but since this should be run on every namespace created, would it be better to move it to the `ReconcileNamespace` function instead ? 🤔 
* https://github.com/carlkyrillos/integreatly-operator/blob/e7ccc0c79e56ab343b6eb27be33727f58ad7fd4e/pkg/resources/reconciler.go#L122

Downside would be that it would be harder to find the logic / execution flow since it would be within another reconcile function, but the plus side is that this would be applied to another other namespace created in the future as long as the same function is used, assuming this is what we want 🤔

KevFan (commented):
Looks good ! Left some minor comments / questions 🧑‍💻

pkg/products/cloudresources/reconciler.go
carlkyrillos: @KevFan It's a good idea and I initially had the same thought but I ultimately opted to keep it separate from `ReconcileNamespace` because not every namespace is created through it (for example `redhat-rhoam-operator` on non-addon installs) and also for readability/traceability. So I think I'd prefer to keep them separate but I don't feel too strongly about it and can move it to `ReconcileNamespace` if you think that makes more sense.

KevFan: Cool, keeping it as is sounds good to me 👍

KevFan:
Changed looks good to me 👍 

Verified:
* Installation completing successfully on autoscaled rosa cluster
image

* `LimitRanges` created in each namespace
image
* Default Limits are applied to containers without specified limits
image

/lgtm

PR #2949, opened 2022-10-11
Mgdapi 4741 - Removal of temporary code

Boomatang:
# Issue link

https://issues.redhat.com/browse/MGDAPI-4741

# What

This code was found during the removal of rhmi in https://github.com/integr8ly/integreatly-operator/pull/2714 
The code being removed is related to the adding of OO and was meant to be removed post 1.14 of RHOAM

# Verification steps

- install the current version of rhoam via olm
- upgrade to this version
- expected: upgraded should happen as normal

### DIFF
--- a/controllers/rhmi/bootstrapReconciler.go
+++ b/controllers/rhmi/bootstrapReconciler.go
@@ -202,8 +202,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 	}
 	metrics.SetQuota(installation.Status.Quota, installation.Status.ToQuota)
 
-	}
-
 	phase, err = r.reconcileCustomSMTP(ctx, serverClient)
 	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
 		events.HandleError(r.recorder, installation, phase, "Reconciling custom SMTP has failed ", err)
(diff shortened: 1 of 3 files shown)


controllers/rhmi/bootstrapReconciler.go
```diff
@@ -202,14 +202,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 	}
 	metrics.SetQuota(installation.Status.Quota, installation.Status.ToQuota)
 
-	// temp code for RHOAM, remove once all clusters are upgraded to 1.14
-	// Remove all prometheus rules under redhat/sandbox-rhoam/rhoami-operator
-	phase, err = r.removePrometheusRules(ctx, serverClient, installation.Spec.NamespacePrefix)
```
KevFan, line 207: Should the relevant function also be removed now that it's unused ? 🤔 
- https://github.com/integr8ly/integreatly-operator/blob/master/controllers/rhmi/bootstrapReconciler.go#L228

Boomatang: Yes it should have been :) I have cleaned up the test around the this function and removed any used functions because of it. Also did a clean up of most of the error golangci-lint would have shown.

KevFan:
* Verified upgrade from 1.27.0 to this branch 
/lgtm

PR #2957, opened 2022-10-13
MGDAPI-4122 - Support setting container platform in docker build commands

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-4122

# What

**ARM64** machines such as M1 MacBooks by default build images in this architecture. When these images are deployed to Openshift which used **AMD64** architecture, pods will fail to run due to mismatch of architectures, with an error similar to:
```
exec /bin/opm: exec format error fix
```

Docker / Podman allows specifying to container build platform for this. By allowing a build flag for this in the image build commands, a consistent image build / architecture would be used across machines for deploying the images onto clusters. This would allow specifying the container build platform to another architecture easier in the future also if there is a use case

* https://docs.docker.com/build/building/multi-platform/#building-multi-platform-images
* https://docs.podman.io/en/latest/markdown/podman-build.1.html#platform-os-arch-variant

The changes were required to install via OLM successfully using an M1 MacBook

# Verification steps

* Verify images are built successfully
```
export ORG=<yourOrg>                                  
```
```
* Verify images are built successfully
# Build Operator Image
OLM_TYPE=managed-api-service INSTALLATION_TYPE=managed-api make image/build
# Build index images
OC_INSTALL=true ./scripts/bundle-rhmi-operators.sh
# Build Test image
make image/functional/build
# Build OSDE2e image
make image/osde2e/build
# Build scorecard image
make scorecard/build
```
* Verify the index pod deployed is successfully running
```
oc get pods -l "olm.catalogSource"="rhmi-operators" -n openshift-marketplace -o jsonpath="{.items[0].status.phase}"
```


scripts/bundle-rhmi-operators.sh
```diff
@@ -134,33 +136,37 @@ generate_index() {
...
     fi
   fi
 
   opm index add \
-      --bundles "$bundles" \
-      --container-tool "$BUILD_TOOL" \
-      --tag "$REG/$ORG/${OLM_TYPE}-index:$NEWEST_VERSION"
+    --bundles "$bundles" \
+    --container-tool "$BUILD_TOOL" \
+    --generate
```
Boomatang, line 167: I am interested to what the `--generate` does. I guess it is creating the index.Dockerfile mentioned on L169. This raises two questions.
1. Are you sure the creation of the file while always be in the same location?
2. When running this script locally is it going to be up the user to clean up this file after completion?

Makefile
```diff
@@ -35,6 +35,7 @@ S3_SECRET_ACCESS_KEY ?= ''
 TYPE_OF_MANIFEST ?= master
 
 CONTAINER_ENGINE ?= docker
+CONTAINER_PLATFORM ?= linux/amd64
```
Boomatang, line 38: This should also added to the notes here: https://github.com/integr8ly/integreatly-operator/blob/master/openshift-ci/README.md

Boomatang (commented):
Looking forward to seen this working.

scripts/bundle-rhmi-operators.sh
KevFan: 1. Yeah, I believe it would always be in the same location. The docs mentions it is generated in the directory where `opm` is ran. From what I have seen / checked, it is created in the temp bundle directory 🤔 
  * https://github.com/operator-framework/operator-registry/blob/master/docs/design/opm-tooling.md#add-1 
  * https://github.com/integr8ly/integreatly-operator/blob/master/scripts/bundle-rhmi-operators.sh#L65-L69
2. No, it would be cleaned up by the script. So as above it is created in the work area temp directory and this directory is cleaned up at the end of the script
  * https://github.com/integr8ly/integreatly-operator/blob/master/scripts/bundle-rhmi-operators.sh#L180-L183

PR #2964, opened 2022-10-19
MGDAPI-4595 add golangci-lint checks

cecobask:
# Issue link
https://issues.redhat.com/browse/MGDAPI-4595

# What
Configured the installation of `golangci-lint` binary.
Added linter presets and disabled all failing linters (temporary).
In subsequent tasks, the disabled linters will have to be re-enabled as we fix the issues that cause the failures.

# Verification steps
- Run the lint make command:
```
make test/lint
```
- It should install the `golangci-lint` binary to the `bin` folder of your project if it doesn't already exist there
- The output of running the linting should be empty. If it isn't, there's a problem...

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,29 @@
+output:
+  format: tab
+  sort-results: true
+
+linters:
+  disable:
+    - errcheck
+    - ineffassign
+    - unused
+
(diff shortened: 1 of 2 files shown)


austincunningham:
Testing goland

PR #2967, opened 2022-10-20
Mgdapi 4720 remove kc dependency

MStokluska:
# Issue link
https://issues.redhat.com/browse/MGDAPI-4720

# What
Removing kc dependency by its dependencies to keycloak client instead

# Verification steps

## Local install
Install locally
Create IDP
Log in as test user to 3scale
Uninstall

# OLM Install
No need for upgrade verification

## Tests
Verify that the tests on this PR pass


Boomatang:
/lgtm
Log in of both admins and users both worked when the operator was running locally and installed via OLM. Uninstall also works.

PR #2969, opened 2022-10-23
MGDAPI-4814 - Race condition setting CRO network strategy

valerymo:
# Issue link
https://issues.redhat.com/browse/MGDAPI-4814

# What
There is a race condition in RHOAM addon install where the CRO network strategy might not be updated until after the installation has already began. We saw this in the [pipeline run here](https://master-jenkins-csb-intly.apps.ocp-c1.prod.psi.redhat.com/job/ManagedAPI/job/managed-api-install-addon-flow/862/console).

Please see more details in Jira: https://issues.redhat.com/browse/MGDAPI-4814

# HOW
Move the CIDR block update to after the create / update of the strategy map
Ensure that if the addon parameter CIDR range is set, that we do not create the CRO subscription until the strategy is present with the CIDR range set

# Verification steps
- Create ccs cluster
- Create private Pipeline in playground, as copy from managed-api-install-addon-flow pipeline.
- Check that Pipeline passed successfully


pkg/products/cloudresources/reconciler.go
```diff
@@ -439,6 +446,12 @@ func (r *Reconciler) cleanupResources(ctx context.Context, installation *integre
 }
 
 func (r *Reconciler) reconcileSubscription(ctx context.Context, serverClient k8sclient.Client, inst *integreatlyv1alpha1.RHMI, productNamespace string, operatorNamespace string) (integreatlyv1alpha1.StatusPhase, error) {
+
+	_, err := r.waitAddonCidrPresentInStrategy(ctx, serverClient)
```
adam-cattermole, line 450: I think we can maybe do this using the existing code to set the CIDR block - this code is mostly duplicated from there and the boolean return is unused

adam-cattermole (commented):
@valerymo instead of the new function, if we removed the check for the strategy existing before setting it, would this be sufficient to stop the reconcile until it's present?

These lines here -> https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/cloudresources/reconciler.go#L558-L560

adam-cattermole:
Thanks for the updates, I think that given these changes are very minimal and look correct I'm happy to lgtm.

/lgtm

PR #2972, opened 2022-10-25
MGDAPI-4639 - Add resource requests to Envoy sidecars and new e2e test

carlkyrillos:
# Issue link
JIRA: https://issues.redhat.com/browse/MGDAPI-4639

# What
This PR includes 3 discrete changes which can be identified by their different commits:

1. Adds a function called `addResourceRequests()` to the 3scale reconciler that annotates the `apicast-production`, `apicast-staging`, and `backend-listener` DeploymentConfigs in order to set CPU/memory requests.
2. Adds a new e2e test called `ValidateResourceRequirements` which verifies that every container created in a RHOAM namespace has explicitly set values for CPU/memory requests.
3. Refactors the subscription_controller reconciler to **not** overwrite the `Resources` field in the Subscription's `Config` block with an empty struct `{}`

# Verification steps
## Verify Envoy containers have properly set CPU/memory requests

1. Checkout this PR
2. Provision an OSD cluster
3. Prepare the cluster by running: `INSTALLATION_TYPE=managed-api LOCAL=false make cluster/prepare/local`
4. Create the CatalogSource:
```yaml
kind: CatalogSource
metadata:
  name: rhoam-operators
spec:
  sourceType: grpc
  image: quay.io/ckyrillo/managed-api-service-index:1.29.0
```
5. Install RHOAM via the OperatorHub in the `redhat-rhoam-operator` Namespace - make sure to check `Automatic` for the `Update approval` (this is needed to force the subscription controller to trigger its reconciler)
6. Create the RHMI CR by running: `INSTALLATION_TYPE=managed-api USE_CLUSTER_STORAGE=true LOCAL=false IN_PROW=true make deploy/integreatly-rhmi-cr.yml`
7. Wait for the installation to complete
8. Verify that all of the `envoy-sidecar` Containers in each of the `apicast-production`, `apicast-staging`, `backend-listener` Pods from the `redhat-rhoam-3scale` Namespace have the proper values for CPU/memory requests (`cpu: 190m, memory: 90Mi`)

## Verify new e2e test, `ValidateResourceRequirements`, is passing

1. Nothing to do here as long as prow e2e tests are passing.

## Verify refactored subscription controller is working as expected

1. Verify that the `spec.config.resources` block in the `managed-api-service` Subscription from the `redhat-rhoam-operator` Namespace has the same request values as the `managed-api-service.v1.29.0` [ClusterServiceVersion](https://github.com/integr8ly/integreatly-operator/blob/master/bundles/managed-api-service/1.29.0/manifests/managed-api-service.clusterserviceversion.yaml#L547-L553)
    * Note: Sometimes the subscription's YAML can't be rendered in the UI for some reason. If this happens, run the following command to view the Subscription's contents: `oc get subscription managed-api-service -n redhat-rhoam-operator -o yaml`

### DIFF
--- a/controllers/subscription/subscription_controller.go
+++ b/controllers/subscription/subscription_controller.go
@@ -131,6 +131,39 @@ func (r *SubscriptionReconciler) Reconcile(request ctrl.Request) (ctrl.Result, e
 
 	if subscription.Spec.InstallPlanApproval != operatorsv1alpha1.ApprovalManual {
 		subscription.Spec.InstallPlanApproval = operatorsv1alpha1.ApprovalManual
+
+		// We need to get the latest InstallPlan to get the CSV in order to set the subscription's Config.Resources
+		// The Config.Resources field needs to be explicitly set because otherwise r.Update will silently set the Config to {} if it's not already set
+		latestInstallPlan := &olmv1alpha1.InstallPlan{}
+		err := wait.Poll(time.Second*5, time.Minute*3, func() (done bool, err error) {
+			if subscription.Status.InstallPlanRef == nil {
+				log.Info("InstallPlanRef from Subscription is nil, trying again...")
+				return false, nil
+			}
+			err = r.Client.Get(context.TODO(), k8sclient.ObjectKey{Name: subscription.Status.InstallPlanRef.Name, Namespace: subscription.Status.InstallPlanRef.Namespace}, latestInstallPlan)
+			if err != nil {
+				log.Infof("Failed to get InstallPlan, trying again...", l.Fields{"Error": err})
+				return false, nil
+			}
+			return true, nil
+		})
+		if err != nil {
+			log.Infof("Failed to get the latest InstallPlan", l.Fields{"Error": err})
+			return ctrl.Result{}, err
+		}
+		csv, err := r.csvLocator.GetCSV(context.TODO(), r.Client, latestInstallPlan)
+		if err != nil {
+			log.Infof("Failed to get CSV from the latest InstallPlan", l.Fields{"Error": err})
+			return ctrl.Result{}, err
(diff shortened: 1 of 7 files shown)


pkg/products/threescale/reconciler.go
```diff
@@ -2911,6 +2924,30 @@ func (r *Reconciler) ensureDeploymentConfigsReady(ctx context.Context, serverCli
...
+			return nil
+		})
+		if err != nil {
+			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("failed to add resource annotation to DeploymentConfig %s: %v", deploymentConfig.Name, err)
+		}
+	}
+
+	return integreatlyv1alpha1.PhaseCompleted, nil
+}
+
```
KevFan, line 2950: Would adding these annotations cause a subsequent pod rollout ?

If so, I'm wondering would it be better to add the annotations into https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/ratelimit/envoy.go#L100-L104 instead ? Downside is readability though 🤔 

I think if we do these seperately, and if they cause muliple pod rollouts, the installation would be generally longer.

Alternatively maybe move the function up closer to after`r.reconcileRatelimitingTo3scaleComponents` instead of near the end of the reconcile loop 🤔

carlkyrillos: I didn't notice the installation taking longer or see multiple rollouts but I also wasn't looking for them. I think adding the annotations to the [patchDeploymentConfig](https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/ratelimit/envoy.go#L100-L104) function like you suggested makes more sense. I'll get started on testing the change.

PR #2974, opened 2022-10-26
WIP: Mgdapi 4807

Boomatang:
# Issue link

https://issues.redhat.com/browse/MGDAPI-4807

# What

# Verification steps


pkg/metrics/metrics.go
```diff
@@ -361,3 +394,85 @@ func GetRhoamState(cr *integreatlyv1alpha1.RHMI) (RhoamState, error) {
...
+	}
+
+	resultValue := result.(model.Vector)
+
+	if len(resultValue) == 1 {
+		value := float32(resultValue[0].Value)
+		return value, warnings, nil
+	} else {
+		return 0, warnings, fmt.Errorf("vector models length is unexpected: Expected 1; Got %v", len(resultValue))
+	}
```
KevFan, line 460:
```suggestion
	} 
	
	return 0, warnings, fmt.Errorf("vector models length is unexpected: Expected 1; Got %v", len(resultValue))
```
nit: reduntant `else`

pkg/metrics/metrics.go
```diff
@@ -7,10 +7,14 @@ import (
 	"github.com/integr8ly/integreatly-operator/pkg/resources"
 	l "github.com/integr8ly/integreatly-operator/pkg/resources/logger"
 	"github.com/integr8ly/integreatly-operator/version"
+	"github.com/prometheus/client_golang/api"
 	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/config"
+	"github.com/prometheus/common/model"
```
KevFan, line 14:
```suggestion
	prometheusApi "github.com/prometheus/client_golang/api"
	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
	"github.com/prometheus/client_golang/prometheus"
	prometheusConfig "github.com/prometheus/common/config"
	prometheusModel "github.com/prometheus/common/model"
```

nit: maybe use more descriptive imports ? 🤔 

If not, I think we should at least keep imports consistent with `prometheusConfig` as this import name is used in `rhmi_controller.go`

KevFan:
* Verified `rhoam_7d_slo_percentile ` value is `1`
image
* Verified `rhoam_7d_slo_remaining_error_budget ` value is `604800 `
image
* Verified `rhoam_7d_slo_percentile:max ` value is `1`
image
* Verified `rhoam_7d_slo_remaining_error_budget:max` value is `604800 `
image

pkg/metrics/metrics.go
Boomatang: I am going to except this but I am in two minds about it. One of the items I address in the commit "MGDAPI-4807 : golangci-lint refactor" was caused by try not to have the redundant else. The use of the else seem to be more explicit in what should happen next while not using it gives the impression that more might be allowed to happen afterwards. At some point I would like to open a discussion with the wider group on this pattern. For now I am happy to except it.

PR #3006, opened 2022-11-15
MGDAPI-4894-test-cases-updates-for-1.29.0-release

Stellar-9:
# Issue link
[https://issues.redhat.com/browse/MGDAPI-4894](url)

# What
Targets updates, based on performed tests

# Verification steps
Eye review


test-cases/tests/alerts/c05b-verify-alerts-are-exposed-and-no-critical-alerts-have-fired.md
```diff
@@ -6,6 +6,8 @@ products:
     environments:
       - osd-fresh-install
       - osd-post-upgrade
+    targets:
+    - 1.29.0
```
trepel, line 10: All modifications like this (adding the whole `target` flag) need to be reverted. There's already another tag ( `per-build` in this particular case) that causes that the test case is generated when needed - for each rc in this particular case. No need to use `target`.

test-cases/tests/performance/k04-run-performance-test-against-rhoam-3scale-user-sso.md
```diff
@@ -10,6 +10,7 @@ products:
       - 1.11.0
       - 1.13.0
       - 1.16.0
+      - 1.29.0
```
trepel, line 13: This was not done as part of 1.29.0 testing so this need to be removed

trepel (changes requested):
Address the two comments. Then look here to address the linting issue: https://github.com/integr8ly/integreatly-operator/tree/master/test-cases#how-to-create-a-test-case

Also, the commit messages could be nicer :) But more important is that each commit has to start with jira number. Your commits starts like that so I suspect that there has to be space (not `-`) after jira ticket number to make that ci/prow/commit check pass

trepel:
@Stellar-9 looks good to me now. Could you update the PR to only have one commit message? Having four commits for such small change really pollutes the commit history heavily. Feel free just to force push

PR #3009, opened 2022-11-16
MGDAPI-4706 - increase permissions for when ns is created by hive

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-4706

# What

We want to have the CRO and 3scale namespaces to be created by Hive to speed up the sync sts secrets and subsequently the install itself.

However, the permission scope needs to be expanded when the namespace is externally created rather than by the Operator.

# Verification steps

* Checkout this branch and install operator
```
make cluster/prepare/local
make code/run
```
* Verify installation is successful
* Delete the RHMI CR and verify uninstall is successfull
* Prow e2e test for RHOAM and Multitenant is successfull

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -326,6 +326,10 @@ create/rhsso-operator/index:
 
 .PHONY: cluster/prepare/project
 cluster/prepare/project:
+	@ - oc new-project $(NAMESPACE_PREFIX)cloud-resources-operator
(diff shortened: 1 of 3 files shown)


cecobask (approved):
Used this branch to run the RHOAM operator locally against a regular OSD cluster.
RHOAM was installed and uninstalled successfully 👍🏻 

/lgtm

PR #3013, opened 2022-11-22
MGDAPI-4598 : Enable ineffassign

Boomatang:
# Issue link

https://issues.redhat.com/browse/MGDAPI-4598

# What

Enabled insffassign in golangci-lint. This checks for poorly assigned variables.

# Verification steps

Pass of prow checks should be enough.


pkg/products/threescale/reconciler.go
```diff
@@ -864,6 +864,10 @@ func (r *Reconciler) resyncRoutes(ctx context.Context, client k8sclient.Client)
 		k8sclient.MatchingLabels(map[string]string{"deploymentConfig": "system-sidekiq"}),
 	}
 	err := client.List(ctx, pods, listOpts...)
+	if err != nil {
+		r.log.Error("Error getting list of pods", err)
```
MStokluska, line 868:
```suggestion
		r.log.Error("error getting list of pods", err)
```

It seems error logs in other files will need editing as well

PR #3015, opened 2022-11-24
MGDAPI-4763 Ensure a tenant can only create 1 APIManagementTenant CR

Stellar-9:
# Issue link
[https://issues.redhat.com/browse/MGDAPI-4763](url)

# What
Add verification to the controller to ensure that the user does not have an APIManagementTenant CR already in either the dev or stg namespace. If they do add a warning to the second CR informing them that they can not have more than 1 APIManagementTenant CR

# Verification steps

1.  Create cluster
2. Run Operator locally
2.1 `INSTALLATION_TYPE=multitenant-managed-api LOCAL=false make cluster/prepare/local`
2.2 `INSTALLATION_TYPE=multitenant-managed-api LOCAL=false make code/run`
2.3 Command for check RHOAM installation status `oc get rhmi rhoam -n sandbox-rhoam-operator -o yaml`
3. Setup IDP `./scripts/setup-sso-idp.sh`
4. You need to create APIManagementTenant CR with test-user or customer-admin. You need to setup permissions for them with command: `oc adm policy add-cluster-role-to-user cluster-admin customer-admin01`
5. Login with one of the users into cluster 
6. Try create 2 APIManagementTenant CRs in -dev or -stage namespace
7. Ensure that you will get won't provision status if there is user present already and that alert not fire, if this status is present

### DIFF
--- a/controllers/tenant/apimanagementtenant_controller.go
+++ b/controllers/tenant/apimanagementtenant_controller.go
@@ -136,9 +136,11 @@ func (r *TenantReconciler) getAPIManagementTenant(crName string, crNamespace str
 	return tenant, nil
 }
 
+// The purpose of this method is to verify if particular APIManagementTenant is created or in progress of creation. If yes then skip, if no then proceed with checking if there is ANOTHER APIManagementTenant already created or in progress of creation
+
 func (r *TenantReconciler) verifyAPIManagementTenant(tenant *v1alpha1.APIManagementTenant) (bool, string, error) {
 	// Skip verification if the tenant is already reconciled
-	if tenant.Status.ProvisioningStatus != v1alpha1.ThreeScaleAccountReady {
+	if tenant.Status.ProvisioningStatus != v1alpha1.ThreeScaleAccountReady && tenant.Status.ProvisioningStatus != v1alpha1.ThreeScaleAccountRequested && tenant.Status.ProvisioningStatus != v1alpha1.UserAnnotated {
 		// Fails if APIManagementTenant isn't from a namespace ending in -dev or -stage
 		if !strings.HasSuffix(tenant.Namespace, "-dev") && !strings.HasSuffix(tenant.Namespace, "-stage") {
 			return false, "tenant not created in a namespace ending in {USERNAME}-dev or {USERNAME}-stage", nil
@@ -166,9 +168,10 @@ func (r *TenantReconciler) verifyAPIManagementTenant(tenant *v1alpha1.APIManagem
 				return false, "an error occurred while trying to check if another reconciled APIManagementTenant CR already exists", err
 			}
 			for _, t := range tenants.Items {
-				if t.Status.ProvisioningStatus == v1alpha1.ThreeScaleAccountReady {
-					return false, "a reconciled APIManagementTenant CR already exists", nil
+				if t.Status.ProvisioningStatus == v1alpha1.ThreeScaleAccountReady || t.Status.ProvisioningStatus == v1alpha1.ThreeScaleAccountRequested || t.Status.ProvisioningStatus == v1alpha1.UserAnnotated {
+					return false, "can't create more than 1 APIManagementTenant CR in -dev or -stage namespace", nil
 				}
+
 			}
 		}
 	}
--- a/pkg/products/observability/prometheusRules.go
+++ b/pkg/products/observability/prometheusRules.go
@@ -355,7 +355,7 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
 						"sop_url": resources.SopApiManagementTenantCRFailed,
 						"message": "An APIManagementTenant CR has failed to reconcile. See the labels for details.",
 					},
-					Expr:   intstr.FromString(`tenants_summary{provisioningStatus!="3scale account ready"}`),
+					Expr:   intstr.FromString(`tenants_summary{provisioningStatus!="3scale account ready" && provisioningStatus != "won't provision"}`),
 					For:    "10m",
 					Labels: map[string]string{"severity": "critical", "product": installationName},
 				},
--- a/pkg/products/observability/prometheusRules.go
+++ b/pkg/products/observability/prometheusRules.go
@@ -355,7 +355,7 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
 						"sop_url": resources.SopApiManagementTenantCRFailed,
 						"message": "An APIManagementTenant CR has failed to reconcile. See the labels for details.",
 					},
-					Expr:   intstr.FromString(`tenants_summary{provisioningStatus!="3scale account ready"} or tenants_summary{provisioningStatus != "won't provision"}`),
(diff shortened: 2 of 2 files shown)


pkg/products/observability/prometheusRules.go
```diff
@@ -355,7 +355,7 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
 						"sop_url": resources.SopApiManagementTenantCRFailed,
 						"message": "An APIManagementTenant CR has failed to reconcile. See the labels for details.",
 					},
-					Expr:   intstr.FromString(`tenants_summary{provisioningStatus!="3scale account ready"}`),
+					Expr:   intstr.FromString(`tenants_summary{provisioningStatus!="3scale account ready" && provisioningStatus != "won't provision"}`),
```
carlkyrillos, line 358: I think this suggestion will make the `ci/prow/multitenant-rhoam-e2e` test pass. CC: @Stellar-9 
```suggestion
					Expr:   intstr.FromString(`tenants_summary{provisioningStatus!="3scale account ready"} and tenants_summary{provisioningStatus != "won't provision"}`),
```

pkg/products/observability/prometheusRules.go
```diff
@@ -355,7 +355,7 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
 						"sop_url": resources.SopApiManagementTenantCRFailed,
 						"message": "An APIManagementTenant CR has failed to reconcile. See the labels for details.",
 					},
-					Expr:   intstr.FromString(`tenants_summary{provisioningStatus!="3scale account ready"}`),
+					Expr:   intstr.FromString(`tenants_summary{provisioningStatus!="3scale account ready"} or tenants_summary{provisioningStatus != "won't provision"}`),
```
carlkyrillos, line 358: @Stellar-9 Sorry for the confusion but the conditional should actually be `and` instead of `or`. That's my bad, I gave you the wrong advice in the previous suggestion. This fix should enable the changes to work as expected.
```suggestion
					Expr:   intstr.FromString(`tenants_summary{provisioningStatus!="3scale account ready"} and tenants_summary{provisioningStatus != "won't provision"}`),
```

carlkyrillos:
Reviewed the changed files and used OLM installation to verify that the APIManagementTenant reconciliation logic and its corresponding PrometheusRule work as expected.
/lgtm

controllers/tenant/apimanagementtenant_controller.go
```diff
@@ -136,9 +136,11 @@ func (r *TenantReconciler) getAPIManagementTenant(crName string, crNamespace str
 	return tenant, nil
 }
 
+// The purpose of this method is to verify if particular APIManagementTenant is created or in progress of creation. If yes then skip, if no then proceed with checking if there is ANOTHER APIManagementTenant already created or in progress of creation
+
 func (r *TenantReconciler) verifyAPIManagementTenant(tenant *v1alpha1.APIManagementTenant) (bool, string, error) {
 	// Skip verification if the tenant is already reconciled
-	if tenant.Status.ProvisioningStatus != v1alpha1.ThreeScaleAccountReady {
+	if tenant.Status.ProvisioningStatus != v1alpha1.ThreeScaleAccountReady && tenant.Status.ProvisioningStatus != v1alpha1.ThreeScaleAccountRequested && tenant.Status.ProvisioningStatus != v1alpha1.UserAnnotated {
```
trepel, line 143: Given that there are only four options here: https://github.com/integr8ly/integreatly-operator/blob/master/apis/v1alpha1/apimanagementtenant_types.go#L25-L30 would not it be much simpler to change the condition to `tenant.Status.ProvisioningStatus == v1alpha1.WontProvisionTenant`

However, it does not make sense to me. I might be missing something here probably

carlkyrillos:
> @carlkyrillos @Stellar-9 Do you guys know whether `tenant.Status` can be nil ?

@trepel Yes the `tenant.ProvisioningStatus` starts as an empty string.

controllers/tenant/apimanagementtenant_controller.go
trepel: Probably not good enough since a situation when `ProvisioningStatus` was not yet populated needs to be addressed as well?

trepel: Thanks @carlkyrillos so `tenant.Status.ProvisioningStatus == v1alpha1.WontProvisionTenant || tenant.Status.ProvisioningStatus == ""` might potentially work too but that's probably not worth the effort to change and retest.

carlkyrillos: @Stellar-9 @trepel I actually think this line should be reverted to only skip verifying an APIManagementTenant CR if it's already been reconciled (and therefore already verified). I don't think there's any harm in continuously verifying all other CRs.
```suggestion
	if tenant.Status.ProvisioningStatus != v1alpha1.ThreeScaleAccountReady{
```

trepel: I am not very familiar with the logic to comment on that. It was just that long condition that caught my eye

Stellar-9: @trepel  I think my if statement is more conscise and easier to read. But yours is more explicit and less error-prone maybe. But I think it's matter of taste mostly

@carlkyrillos So I would be happy if you can correct me if I'm wrong. But I think that might be not enough for what we want to achieve based on that JIRA. If user will create CR and provision status will not be 3scale account ready yet, but some of the other ones. So still in process of creation and user will try to create another CR, then in that case it will proceed, which we don't want to allow them. I mean in testing enviroment, CR creation is pretty quick and almost imidiately in status 3scale account ready. But based on that JIRA was created, I guess in production are actually this conflict situations happening. And from my perspectie I don't see any cons to have this statement there, just in case. But I don't know. if I'm wrong here, that I would be happy to learn more about it

carlkyrillos: @Stellar-9 Let's chat offline to discuss further but I think changing [this line](https://github.com/integr8ly/integreatly-operator/blob/c431f386775a5f8a7d5c828d6c787d26c210e5f6/controllers/tenant/apimanagementtenant_controller.go#L143) to:
`if tenant.Status.ProvisioningStatus == ""`
and changing [this line](https://github.com/integr8ly/integreatly-operator/blob/c431f386775a5f8a7d5c828d6c787d26c210e5f6/controllers/tenant/apimanagementtenant_controller.go#L171) to:
`if t != tenant && t.Status.ProvisioningStatus != v1alpha1.WontProvisionTenant`
will prevent the conflict and simplify the `if` statements but I will need to test to confirm.

carlkyrillos: @Stellar-9 @trepel Sorry for all of my flip-flopping but after testing/verifying a few different options, I think Ales' original implementation is best and the PR can be merged as is.

trepel: Thanks for clarifications, guys! I am fine with this as is. It's approved now anyway.

PR #3024, opened 2022-12-06
MGDAPI-4721 update OO to 4.0.0

MStokluska:
# Issue link
https://issues.redhat.com/browse/MGDAPI-4721

# What
Updated OO to official 4.0.0 image

# Verification steps
- Install 1.29.0 from index (you can use quay.io/mstoklus/managed-api-service-index:1.29.0)
- Create index for 1.30.0 using this branch
- Upgrade from 1.29.0 to 1.30.0
- Verify upgrade was successful
- Verify all required e2e tests passed on this PR

- navigate to customer monitoring grafana and confirm that the rate limiting dashboard works as expected.


KevFan:
Should we be also removing the temporary replace and bumping the dependency version in `go.mod` ? 🤔 
* https://github.com/integr8ly/integreatly-operator/blob/a6db086f5ae456e8e3e63ebfd73f2e5fba88436b/go.mod#L212-L213
* https://github.com/integr8ly/integreatly-operator/blob/a6db086f5ae456e8e3e63ebfd73f2e5fba88436b/go.mod#L38

MStokluska:
> Should we be also removing the temporary replace and bumping the dependency version in `go.mod` ? thinking
> 
>     * https://github.com/integr8ly/integreatly-operator/blob/a6db086f5ae456e8e3e63ebfd73f2e5fba88436b/go.mod#L212-L213
> 
>     * https://github.com/integr8ly/integreatly-operator/blob/a6db086f5ae456e8e3e63ebfd73f2e5fba88436b/go.mod#L38

I don't think we can unfortunately. 
OO got released and tagged with 4.0.0; however the go.mod package definition still has /v3, go doesn't like it :/
but happy to accept a solution if you find one!

KevFan:
> > Should we be also removing the temporary replace and bumping the dependency version in `go.mod` ? thinking
> > ```
> > * https://github.com/integr8ly/integreatly-operator/blob/a6db086f5ae456e8e3e63ebfd73f2e5fba88436b/go.mod#L212-L213
> > 
> > * https://github.com/integr8ly/integreatly-operator/blob/a6db086f5ae456e8e3e63ebfd73f2e5fba88436b/go.mod#L38
> > ```
> 
> I don't think we can unfortunately. OO got released and tagged with 4.0.0; however the go.mod package definition still has /v3, go doesn't like it :/ but happy to accept a solution if you find one!

Oh right, in that case, I wonder can we just target the specific release commit instead ? 🤔 Something like:
```
go get github.com/redhat-developer/observability-operator/v3@e335ad3a1a56881129530c69bce15ad72c3d06ed
```
to bump the dependencies

MStokluska:
> > > Should we be also removing the temporary replace and bumping the dependency version in `go.mod` ? thinking
> > > ```
> > > * https://github.com/integr8ly/integreatly-operator/blob/a6db086f5ae456e8e3e63ebfd73f2e5fba88436b/go.mod#L212-L213
> > > 
> > > * https://github.com/integr8ly/integreatly-operator/blob/a6db086f5ae456e8e3e63ebfd73f2e5fba88436b/go.mod#L38
> > > ```
> > 
> > 
> > I don't think we can unfortunately. OO got released and tagged with 4.0.0; however the go.mod package definition still has /v3, go doesn't like it :/ but happy to accept a solution if you find one!
> 
> Oh right, in that case, I wonder can we just target the specific release commit instead ? thinking Something like:
> 
> ```
> go get github.com/redhat-developer/observability-operator/v3@e335ad3a1a56881129530c69bce15ad72c3d06ed
> ```
> 
> to bump the dependencies

Sorry we can't neither as there's a bit of dependency mismatch between RHOAM and OO. We need to use replaces here.
I can update the commit in the replaces if that makes it any better!

KevFan:
> Sorry we can't neither as there's a bit of dependency mismatch between RHOAM and OO. We need to use replaces here. I can update the commit in the replaces if that makes it any better!

Oh right, I think it's fine as is so, not worth the hassle 👍

KevFan:
* Verified upgrade from 1.29.0 -> 1.30.0 from this branch on OSD 4.11
* Verified fresh install from index from this branch on OSD 4.12
* Verified rate limit dashboard looks normal
  * OSD 4.11 cluster - no workload webapp deployed
  
image

  * OSD 4.12 Cluster - Workload webapp deployed
image

/lgtm

PR #3031, opened 2022-12-09
MGDAPI-4549 : Add filter for active tenants to dns bypass metrics

Boomatang:
# Issue link

https://issues.redhat.com/browse/MGDAPI-4549

# What

This add a filter to ensure that the routes used for the dns bypass alerts are from an active tenant. 

# Verification steps

## Install master version
* Install the current master of the operator via OLM. Index: quay.io/jfitzpat/managed-api-service-index:1.30.0-4549
* Create muitly tenants in 3scale.
* Check the routes in the 3scale namespace. If the system-provider and system-developer routes for the created tenants are listed above the default routes. In 3scale UI delete the tenants belonging to the routes.
* Expected: alerts for DnsBypassThreeScaleDeveloperUI or DnsBypassThreeScaleSystemAdminUI shoud start firing.
* Allow alert to go into a firing state before moving onto the upgrade verification

## Upgrade to these changes
* Use index quay.io/jfitzpat/managed-api-service-index:1.31.0-4549
* Allow upgrade to finish
* Expected: dns bypass alerts will stop firing.

### DIFF
--- a/pkg/products/threescale/filters.go
+++ b/pkg/products/threescale/filters.go
@@ -7,10 +7,10 @@ import (
 
 type TenantAccountsFilter struct {
 	providers  []AccountDetail
-	developers []developerAccounts
+	developers []developerRoute
 }
 
-type developerAccounts struct {
+type developerRoute struct {
 	Url   string
 	State string
 }
@@ -18,7 +18,7 @@ type developerAccounts struct {
 func NewTenantAccountsFilter(accounts []AccountDetail) TenantAccountsFilter {
 	f := TenantAccountsFilter{}
 	f.setProviders(accounts)
-	f.generateDeveloperAccounts(accounts)
+	f.generateDeveloperRoutes(accounts)
 
 	return f
 }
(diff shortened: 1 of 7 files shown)


pkg/products/threescale/filters.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/openshift/api/route/v1"
+	"strings"
+)
+
+type TenantAccountsFilter struct {
+	providers  []AccountDetail
+	developers []developerAccounts
+}
+
+type developerAccounts struct {
```
cecobask, line 13: As we discussed earlier, this struct's name should probably be singular. To more accurately represent its data it should be called `developerRoute` or something similar

pkg/products/threescale/filters.go
```diff
@@ -0,0 +1,61 @@
...
+func (f *TenantAccountsFilter) Developer(r v1.Route) bool {
+	for _, account := range f.developers {
+		if strings.HasSuffix(account.Url, r.Spec.Host) {
+			return account.State == "approved"
+		}
+	}
+	return false
+}
+
+func (f *TenantAccountsFilter) generateDeveloperAccounts(accounts []AccountDetail) {
```
cecobask, line 44: Similar naming convention should be used here too. Something close to `generateDeveloperRoutes` makes sense to me

cecobask (changes requested):

cecobask (approved):
I followed the verification instructions closely and I can confirm that everything works as described. The `DnsBypass***` alerts stop firing false positives after the upgrade is completed.
My comments were addressed and I'm happy to approve.

/lgtm

PR #3033, opened 2022-12-12
MGDAPI-4729 - fix: F03 - only check for managed tag for sts clusters

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-4729

# What

Only checked for managed tag for elasticache instances for sts clusters

# Verification steps

* Ensure F03 passes on non sts ccs install
```
LOCAL=false NAMESPACE=redhat-rhoam-operator TEST=F03 make test/e2e/single
```
image


trepel:
`LOCAL=false INSTALLATION_TYPE=managed-api TEST=F03 make test/e2e/single` against non STS cluster was OK.

Running on cluster in container against STS cluster was also OK (I did it this way to avoid permission issues with `sts:AssumeRole`. Passed as well.

Code change looks ok

/lgtm

PR #3048, opened 2023-01-04
[WIP] Update oo api dependencies

laurafitzgerald:
# Issue link
TBD

# What
Dependency update dependent on this commit https://github.com/laurafitzgerald/observability-operator/commit/f1e38a9832b4fef98980f06ee2a37f279205d4b6 which currently just lives in my fork. 

If we are happy with the changes here I'd like to submit the above commit via pr to the observability team.

The main goal here is to begin to suggest to operator teams which we include as dependencies in integreatly-operator to separate their api code as submodules inside their repos to allow for less dependency conflicts when depending on them so we can move our dependencies faster. 

# Verification steps
current verification would suffice as end to end and other pr tests passing.


go.mod
```diff
@@ -163,34 +166,33 @@ require (
...
 	google.golang.org/appengine v1.6.7 // indirect
 	google.golang.org/genproto v0.0.0-20220920201722-2b89144ce006 // indirect
 	google.golang.org/grpc v1.49.0 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/warnings.v0 v0.1.2 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-	k8s.io/apiserver v0.24.3 // indirect
-	k8s.io/component-base v0.24.3 // indirect
+	k8s.io/apiserver v0.26.0 // indirect
+	k8s.io/component-base v0.26.0 // indirect
```
laurafitzgerald, line 195: This bump causes a test to fail on the following 
```vendor/k8s.io/component-base/metrics/registry.go:35:29: undefined: atomic.Bool
note: module requires Go 1.19
```

laurafitzgerald: I've fixed the offending packages to 0.24.3 for the moment. I think it can be solved by moving other dependent packages to submodule their apis. Next one that seems sensible is marin3r.

go.mod
```diff
@@ -206,17 +207,23 @@ replace github.com/openshift/api => github.com/openshift/api v0.0.0-202108310919
...
 
+replace k8s.io/apimachinery => k8s.io/apimachinery v0.24.3
+
+replace k8s.io/client-go => k8s.io/client-go v0.24.3
+
+replace k8s.io/component-base => k8s.io/component-base v0.24.3
+
+replace k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.24.3
+
+replace k8s.io/apiserver => k8s.io/apiserver v0.24.3
```
Boomatang, line 218: I know this is a wip and would be subject to change when the correct OO version is being used. But we agree that any `replace` needs to be commented to why its being done.

Boomatang (commented):
Apart from the one comment I think this looks good to me.

go.mod
laurafitzgerald: Agreed. I mentioned it in my comment above. I believe marin3r is now the offender given that OO is moved out to submodule. Can leave a comment in the code for that for sure.

go.mod
```diff
@@ -25,36 +25,36 @@ require (
...
 	google.golang.org/protobuf v1.28.1
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.24.3
-	k8s.io/apiextensions-apiserver v0.24.3
-	k8s.io/apimachinery v0.24.3
-	k8s.io/client-go v12.0.0+incompatible
-	k8s.io/kube-openapi v0.0.0-20220803164354-a70c9af30aea // indirect
-	k8s.io/utils v0.0.0-20220922133306-665eaaec4324
-	sigs.k8s.io/controller-runtime v0.12.3
+	k8s.io/api v0.26.0
```
MStokluska, line 45: I think I'm more in favor of either removing deps from other operators in one PR and bumping RHOAMs k8s.io to latest, or bump in other operators first and then in RHOAM to avoid introducing mass replaces.
WDYT?

PR #3051, opened 2023-01-05
MGDAPI-5062 - refactor of user code and tests

laurafitzgerald:
# Issue link
https://issues.redhat.com/browse/MGDAPI-5062

# What
Some code/test refactoring in the user reconciler.

# Verification steps
Test should pass.
There should be no logic changes here but have no objection to this being run and a few users created to validate nothing is broken.

### DIFF
--- a/controllers/user/user_controller.go
+++ b/controllers/user/user_controller.go
@@ -2,9 +2,9 @@ package controllers
 
 import (
 	"context"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"time"
 
-	rhmiv1alpha1 "github.com/integr8ly/integreatly-operator/apis/v1alpha1"
 	l "github.com/integr8ly/integreatly-operator/pkg/resources/logger"
 
 	userHelper "github.com/integr8ly/integreatly-operator/pkg/resources/user"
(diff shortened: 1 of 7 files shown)


controllers/user/user_controller.go
```diff
@@ -18,48 +18,59 @@ import (
...
 	restConfig.Timeout = time.Second * 10
-	scheme := runtime.NewScheme()
-	err := rhmiv1alpha1.AddToSchemes.AddToScheme(scheme)
+
+	client, err := k8sclient.New(restConfig, k8sclient.Options{
+		Scheme: mgr.GetScheme(),
+	})
 	if err != nil {
-		return ctrl.Result{}, err
+		panic("could not setup k8s client for user controller")
```
Boomatang, line 42: My understanding was a panic will crash the application, which in our case would cause a pod restart. Is this not what happens?

Boomatang (commented):
I only have the one question around the use of panic. Most say I like (and can't belive I didn't think of it) how you handle the unused method signature variable in the `GetPreflightObject`.Have seen a few of them and never knew what I was going to do with them.

controllers/user/user_controller.go
laurafitzgerald: Good question, i questioned it's use as well. On review the New Functions when they return an error that is followed by an os.exit, which is effectively the same. the container will bump and retry. Not returning the error here makes this more consistent with other uses of the New function and the resulting error handling is effectively the same result as returning.

Boomatang: Ok that makes sense. I do think that calling os.exit is the cleaner way to exit. In this case i don't think it matters but if a panic is called will that still allow the calling of any defer methods?

Boomatang:
\lgtm

PR #3055, opened 2023-01-06
MGDAPI-5059 skip cloud-resources-operator and 3scale namespace removal when rhoam operator is hive managed

cecobask:
# Issue link
https://issues.redhat.com/browse/MGDAPI-5059

# What
The following namespaces can be managed by Hive: https://gitlab.cee.redhat.com/service/managed-tenants/-/blob/a2810640575475dd72e2a7eb25599c450f9e7194/addons/rhoams/metadata/stage/addon.yaml#L42-44

RHOAM uninstallation was getting stuck in a loop of RHOAM removing the above namespaces and Hive recreating them.
To address that I've added a check to determine if the RHOAM operator is managed by Hive and based on that, skip the namespace removal if `true`.

# Verification steps


trepel:
I installed and uninstalled from this PR (via cluster/deploy), the namespaces were removed as expected. Then I added a new function called `OperatorIsHiveManagedAlwaysTrue` and used this one instead of `OperatorIsHiveManaged` in the two reconcilers to simulate Hive managed scenario. The two namespaces were not removed in this case as expected. What I am a little confused with is the fact that both 3scale and OO namespaces still contain workloads (pods, deployments etc.)

Code looks good to me.

pkg/products/cloudresources/reconciler.go
```diff
@@ -139,10 +138,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+			// the ns can be managed by hive and removing it here can cause rhoam uninstall to get stuck
+			isHiveManaged, err := addon.OperatorIsHiveManaged(ctx, client, installation)
+			if err != nil {
+				return integreatlyv1alpha1.PhaseFailed, err
+			}
+			if !isHiveManaged {
+				phase, err = resources.RemoveNamespace(ctx, installation, client, operatorNamespace, r.log)
+				if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+					return phase, err
+				}
```
Boomatang, line 150: Would it not be better to remove this logic altogether. Being hive managed is the expected work flow. Its only when we install via OLM or local while this be needed. In my opinion it would be better to update the cleanup make targets to remove the namespece. WTY?

cecobask: I'm okay with this approach 👍🏻 

When deleting a RHOAM installation, I always relied on simply deleting the RHMI CR and then waiting for the reconcile to complete. If there are other people using this approach, there might be some confusion why the namespaces/resources in them are still existing after a successful RHOAM uninstall.

This means we have to update the README to make it obvious how to properly uninstall RHOAM.

trepel:
One thing came to my mind. What if somebody triggers a 2nd RHOAM install while the 3scale and CRO namespaces (with all the pods etc) are still there because Hive did not manage to remove the namespaces yet?

pkg/products/cloudresources/reconciler.go
MStokluska: I think I'd prefer this logic as it is, the reason why is because it gives us an extra check during our nightly runs (which are often OLM/deployment). 
If for any reason there's anything left behind the ns would not be successfully deleted which would be highlighted in the pipelines. If we ignore removing the 3scale and cro ns, then we are hoping that everything related to these namespaces got correctly deleted or, we will need to provide an additional stage on our pipeline to run cleanup commands and have them successfully complete.

MStokluska:
> One thing came to my mind. What if somebody triggers a 2nd RHOAM install while the 3scale and CRO namespaces (with all the pods etc) are still there because Hive did not manage to remove the namespaces yet?

Good point although hive will remove the NS, not the resources on the NS. Resources on the ns should be removed by RHOAM. 

So in your scenario the flow would be (IMO):
- RHOAM gets removed
- RHOAM cleans up the resources under 3scale and cro ns
- Hive doesn't reconcile the ns deletion yet so they are present
- Antoher RHOAM install kicks in - actually, this is unlikely since hive needs to reconcile the state of the addon before you can install it again, meaning, you can't install addon if addon reports "deleting" - but if let's say its not saying "deleting" and namespaces are present?
This to me would mean that ns has not been marked for deletion yet, so when a new install takes over, before hive reconciles ns, then I guess the ns won't be removed since new addon install needs them.

pkg/products/threescale/reconciler.go
```diff
@@ -200,9 +201,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+			return integreatlyv1alpha1.PhaseFailed, err
+		}
+		if !isHiveManaged {
+			phase, err = resources.RemoveNamespace(ctx, installation, serverClient, productNamespace, r.log)
+			if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
+				return phase, err
+			}
 		}
 
 		phase, err = resources.RemoveNamespace(ctx, installation, serverClient, operatorNamespace, r.log)
```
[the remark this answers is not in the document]
trepel: Don't ask me why but these namespaces are Hive managed: https://gitlab.cee.redhat.com/service/managed-tenants/-/blob/main/addons/rhoams/metadata/stage/addon.yaml#L41-44

pkg/products/cloudresources/reconciler.go
Boomatang: This wont give as clear a signal to resources being left behind as an extra stage in the pipelines would. I do not agree with putting development only logic into the operator as it bring it further away from production configurations. That said I am willing to let this go.

pkg/products/threescale/reconciler.go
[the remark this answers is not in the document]
Boomatang: That's so odd. Ok I am happy with that

cecobask:
@MStokluska @trepel I added checks in the final uninstall function of threescale that will ensure it removes the api manager and envoy configs before proceeding with the threescale namespaces deletion.
The CRO uninstallation already handles cleanup of all its resources and I didn't experience any problems with the uninstallation order.
Let me know if something else needs to be addressed. Thanks!

cecobask:
Verified the changes work as expected by installing RHOAM via addon flow and patching the csv to point at [my development image](quay.io/tdimov0/managed-api-service:v1.66.0) and finally triggering addon uninstall:
```
ORG=tdimov0 TAG=1.66.0 make image/build/push
ORG=tdimov0 VERSION=v1.66.0 ADDON_VERSION=v1.30.0 ./scripts/patch-image-csv.sh
```
RHOAM deleted all required resources, skipped removing the threescale product and cloud resource operator namespaces. Hive cleaned up the remaining namespaces.

trepel:
I verified using both "master" and "addon flow" uninstallations. Fine in both cases. For "master" the namespaces were removed. For "addon flow" the namespaces were not removed and Hive removed them in the end.

trepel:
After more thorough test it seems that there are still pods left in CRO ns. The CRO pod is gone but two other pods remained. I think that it is OK (similarly the two such pods remained for RHOAM operator itself). For the record:
```
[trepel@trepel integreatly-operator]$ oc get ns | grep rhoam
redhat-rhoam-3scale                                               Active
redhat-rhoam-cloud-resources-operator                             Active
redhat-rhoam-operator                                             Active
[trepel@trepel integreatly-operator]$ oc get pods -n redhat-rhoam-cloud-resources-operator
NAME                                                              READY   STATUS      RESTARTS   AGE
bd8e25289a50cf45add0bb74edaa43d9843694700c7248c4eb640e808e68zgj   0/1     Completed   0          155m
rhmi-registry-cs-kfrgj                                            1/1     Running     0          156m
[trepel@trepel integreatly-operator]$ oc get pods -n redhat-rhoam-operator
NAME                                                              READY   STATUS      RESTARTS   AGE
addon-managed-api-service-catalog-b44xk                           1/1     Running     0          162m
d6d667b0c05f726043feb35610e2055f80959f789616a2aaad500183c2p9qgh   0/1     Completed   0          162m
```

cecobask:
@trepel, I believe these are okay to be left for removal when Hive deletes the remaining namespaces.
Thank you for reviewing the pull request!

PR #3071, opened 2023-01-20
MGDAPI-5078 - feat: use larger backend node size for 100M quota

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-5078

# What

* CRO v.0.43.0
* Use backend node size for 100M quota option

# Verification steps


KevFan:
# Fresh Installation at 100M 
* Backend redis node size created was `cache.m5.large`
Screenshot 2023-01-20 at 10 02 47
* Installation completed successfully
Screenshot 2023-01-20 at 10 46 32

products/products.yaml
```diff
@@ -13,7 +13,7 @@ products:
     manifestsDir: "integreatly-3scale"
     quayScan: true
   - name: cloud-resource-operator
-    version: v0.42.0
+    version: v0.43.0
```
austincunningham, line 16: we need to bump in `installation-cpaas.yaml` as well
https://github.com/integr8ly/integreatly-operator/blob/master/products/installation-cpaas.yaml#L68

KevFan:
# Upgrade from 1.30.0
* Verified with @adam-cattermole that postrges engine version is upgraded

image

* 340s downtime due to upgrade

Screenshot 2023-01-20 at 14 40 55

KevFan:
/hold 

Seems like the croUtil function only allows for modification before creation and doesn't allow for updating, so redis changes are not applied on quota changes 
- https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/client/util.go#L107-L114

austincunningham:
verified on call with Kevin 
code review again 
/lgtm
/approve

KevFan:
* Verified with @austincunningham that quota changes with backend redis size change is applied immediately
image

image

PR #3080, opened 2023-01-24
MGDAPI-4456 Fix installation stage watches for J03

Boomatang:
# Issue link

https://issues.redhat.com/browse/MGDAPI-4456

# What

Fix watches on installation stages

# Verification steps

Test on both managed-api and multitenant-managed-api.

* Install the operator for the installation type
* Follow the test case for [J03B](https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/backup-restore/j03b-verify-that-namespaces-get-recreated-by-the-integreatlyoper.md)
* Expected result: test case should pass for insatallation type.

Repeated above steps for each installation type.

### DIFF
--- a/test/common/namespace_restoration.go
+++ b/test/common/namespace_restoration.go
@@ -3,6 +3,7 @@ package common
 import (
 	goctx "context"
 	marin3rv1alpha1 "github.com/3scale-ops/marin3r/apis/marin3r/v1alpha1"
+	marin3rOperatorv1alpha1 "github.com/3scale-ops/marin3r/apis/operator.marin3r/v1alpha1"
 	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/apis/v1alpha1"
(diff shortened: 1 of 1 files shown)


KevFan:
Looks like there is a finalizer on `DiscoveryService` in the `redhat-rhoam-3scale` namespace that is preventing it from being deleted

image

It's deleted once the marin3r operator is restored but probably good to manually delete the finalizer which might make the test complete slightly faster since the 3scale namespace can be restored sooner 🤔

Boomatang:
This is a bit out of scope for the Jira, but I can see the improvement that it cause and you found the underlining resource with the issue. So I will do some updates here.

Boomatang:
The termination of the 3scale namespace is happening much faster with the finalizers being removed on the discovery service. It has not made much of a speed improvement but what it has done is made the logs line up more with what is actually happening to the namespaces.

PR #3085, opened 2023-01-26
MGDAPI-4998 add upgrades cm

MStokluska:
# Issue link
https://issues.redhat.com/browse/MGDAPI-4998

# What
Add cm that will block/allow upgrades
note. utests to be done.

# Verification steps

## Test 1 
The goal of this test is to confirm that the cm logic works as expected.
The cm can have updatesAllowed to true or false, or there can be no cm at all (at which point serviceAffecting is the guard in front of upgrades).

There have been 7 bundles + indexes created, each upgrade path is described in table below.

Create catalog source pointing to `quay.io/mstoklus/managed-api-service-index:1.31.0` and let RHOAM install.
Once installed, keep bumping the catalog source according to table and expect manual / automatic updates based on the data provided in the table. 

Create config map
```
oc create configmap rhoam-upgrades --from-literal=allowRhoamUpdates=false -n redhat-rhoam-operator
```
| Version from | Version to | Upgrade type | CM state | Manual Approval Expected | Completed? |
|--------------|------------|--------------|----------|--------------------------|------------|
|1.31.0|1.32.0|Non service affecting|False| Yes | <ul><li>[ ] </li></ul> |
|1.32.0|1.33.0|Service affecting| False | Yes | <ul><li>[ ] </li></ul> |

Edit config map value to true

| Version from | Version to | Upgrade type | CM state | Manual Approval Expected | Completed? |
|--------------|------------|--------------|----------|--------------------------|------------|
|1.33.0|1.34.0|Non service affecting|True| No | <ul><li>[ ] </li></ul> |
|1.34.0|1.35.0|Service affecting|True| Yes | <ul><li>[ ] </li></ul> |

Remove config map

| Version from | Version to | Upgrade type | CM state | Manual Approval Expected | Completed? |
|--------------|------------|--------------|----------|--------------------------|------------|
|1.35.0|1.36.0|Non service affecting|N/A| No | <ul><li>[ ] </li></ul> |
|1.36.0|1.37.0|Service affecting|N/A| Yes | <ul><li>[ ] </li></ul> |

## Test 2
Install RHOAM 1.31.0 from the index above and create config map:
```
oc create configmap rhoam-upgrades --from-literal=allowRhoamUpdates=false -n redhat-rhoam-operator
```
Bump the catalog source to 1.32.0 and wait for the upgrade available to appear in the operator hub
Bump catalog source to 1.33.0 without approving the 1.32 upgrade. 
Bump catalog source to 1.34.0 without approving the 1.32 upgrade.
Bump catalog source to 1.35 without approving the 1.32 upgrade. 
Bump catalog source to 1.36 and approve 1.32 after a minute.

The excepted outcome is that after approving 1.32 you will be asked to approve 1.33 upgrade. Do so, after approval of 1.33 the 1.34 upgrade should kick in, approve that one too, then before approving 1.35, remove CM and approve 1.35. 
1.36 should approve automatically.

### DIFF
--- a/controllers/subscription/subscription_controller.go
+++ b/controllers/subscription/subscription_controller.go
@@ -10,6 +10,9 @@ import (
 	"github.com/integr8ly/integreatly-operator/pkg/resources/k8s"
 	"github.com/integr8ly/integreatly-operator/pkg/resources/rhmi"
 	"github.com/integr8ly/integreatly-operator/version"
+	corev1 "k8s.io/api/core/v1"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 
 	l "github.com/integr8ly/integreatly-operator/pkg/resources/logger"
 	"github.com/sirupsen/logrus"
@@ -42,6 +45,7 @@ const (
 	ManagedAPIAddonSubscription     = "addon-managed-api-service"
 	ManagedAPIAddonSubscriptionEdge = "addon-managed-api-service-internal"
 	ManagedAPIolmSubscription       = "managed-api-service"
+	RHOAMUpgradeConfigMap           = "rhoam-upgrades"
 )
 
 var subscriptionsToReconcile []string = []string{
@@ -248,7 +252,15 @@ func (r *SubscriptionReconciler) HandleUpgrades(ctx context.Context, rhmiSubscri
 		return ctrl.Result{}, nil
 	}
 
-	if !isServiceAffecting && !latestInstallPlan.Spec.Approved {
+	upgradeAllowed, err := r.upgradeAllowed()
+	if err != nil {
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: 10 * time.Second,
+		}, nil
+	}
+
+	if !isServiceAffecting && !latestInstallPlan.Spec.Approved && upgradeAllowed {
 		eventRecorder := r.mgr.GetEventRecorderFor("Operator Upgrade")
 		logrus.Infof("Approving install plan %s ", latestInstallPlan.Name)
 		err = rhmiConfigs.ApproveUpgrade(ctx, r.Client, installation, latestInstallPlan, eventRecorder)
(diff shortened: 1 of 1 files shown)


controllers/subscription/subscription_controller.go
```diff
@@ -248,7 +252,15 @@ func (r *SubscriptionReconciler) HandleUpgrades(ctx context.Context, rhmiSubscri
 		return ctrl.Result{}, nil
 	}
 
-	if !isServiceAffecting && !latestInstallPlan.Spec.Approved {
+	upgradeAllowed, err := r.upgradeAllowed()
+	if err != nil {
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: 10 * time.Second,
```
laurafitzgerald, line 259: In the case of blocking the upgrade, do we want to Requeue it this frequently? Will this run ever 10 seconds until the config map is removed?

controllers/subscription/subscription_controller.go
```diff
@@ -269,6 +281,23 @@ func (r *SubscriptionReconciler) HandleUpgrades(ctx context.Context, rhmiSubscri
...
+func (r *SubscriptionReconciler) upgradeAllowed() (bool, error) {
+	upgradeConfigMap := &corev1.ConfigMap{}
+	err := r.Get(context.TODO(), types.NamespacedName{Namespace: r.operatorNamespace, Name: RHOAMUpgradeConfigMap}, upgradeConfigMap)
+	if err != nil {
+		if !k8serr.IsNotFound(err) {
+			return false, err
+		}
+	}
+
+	if upgradeConfigMap.Data["allowRhoamUpdates"] == "false" {
```
laurafitzgerald, line 293: Do we need to have any data in it or is it enough that the config map is present?

controllers/subscription/subscription_controller.go
MStokluska: no, controller will be re-queued only if there's an error fetching the cm which is not isNotFound error, for example, temporary connection error

controllers/subscription/subscription_controller.go
MStokluska: not really that we need, I had a discussion about it with Brian and we both agreed that having an entry in the cm doesn't harm us and brings extra clarity on what is this supposed to do.

PR #3096, opened 2023-02-02
[wip] MDGAPI-5126 - fix: go permission issue in checks

KevFan:
# Issue link

# What

# Verification steps


carlkyrillos:
Verified that [these tests](https://github.com/integr8ly/integreatly-operator/blob/master/openshift-ci/test/run#L22-L25) pass with the new Dockerfile (I attached output for the successful vendor check). It seems risky to merge the changes when none of the PR checks are passing but I think we have to in this case (and we can always revert if needed).

/lgtm

MGDAPI-5126-success

PR #3099, opened 2023-02-07
MGDAPI-5193 Unskip alert tests

trepel:
# Issue link

https://issues.redhat.com/browse/MGDAPI-5193

# What

The https://issues.redhat.com/browse/MGDAPI-1404 no longer causes alerts to fire - at least not right after the RHOAM installation so it should be safe to run these tests.

However, the UIBBT alerts are firing in Prow (possibly a limitation of Prow environment). This PR skips the alert tests if running in Prow. Ideal solution would be to ignore the UIBBT alerts if in Prow not skipping the whole tests but this is better than having these tests skipped in non-Prow environment as it is the case now.

For even more context read https://chat.google.com/room/AAAAssKWhTI/h5DeT583XH4

# Verification steps

Eye review only, passing rhoam-e2e should suffice.

### DIFF
--- a/test/common/alerts_firing.go
+++ b/test/common/alerts_firing.go
@@ -115,8 +115,15 @@ func TestIntegreatlyAlertsFiring(t TestingTB, ctx *TestingContext) {
 	//fail immediately if one or more alerts have fired
 	if err := getFiringAlerts(t, ctx); err != nil {
 		podLogs(t, ctx)
-		t.Skipf("Skipping due to known issue : %v, reported in Jira: https://issues.redhat.com/browse/MGDAPI-1404", err.Error())
-		//t.Fatal(err.Error())
+		isInProw, e := isInProw(ctx)
+		if e != nil {
+			t.Logf("error getting in_prow annotation: %s", e)
+		}
+		if isInProw {
+			t.Skipf("Skipping due to known issue in Prow, UIBBT alests fire there: %v, reported in Jira: https://issues.redhat.com/browse/MGDAPI-5193", err.Error())
+		} else {
+			t.Fatal(err.Error())
+		}
 	}
 
 }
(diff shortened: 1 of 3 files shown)


trepel:
rhoam-e2e failed on the two alerts that have been unskipped:
```
[38;5;9m[1mSummarizing 2 Failures:[0m
  [38;5;9m[FAIL][0m [0mintegreatly [38;5;243mmanaged-api HAPPY PATH [0m[38;5;9m[1m[It] C01 - Verify Alerts are not pending or firing apart from DeadMansSwitch[0m[0m
  [38;5;243m/go/src/github.com/integr8ly/integreatly-operator/vendor/github.com/onsi/ginkgo/v2/internal/suite.go:605[0m
  [38;5;9m[FAIL][0m [0mintegreatly [38;5;243mmanaged-api HAPPY PATH [0m[38;5;9m[1m[It] Verify Alerts are not firing during or after installation apart from DeadMansSwitch[0m[0m
  [38;5;243m/go/src/github.com/integr8ly/integreatly-operator/vendor/github.com/onsi/ginkgo/v2/internal/suite.go:605[0m
```
Reason:
```
  The following alerts were fired:
  	alert: ThreeScaleAdminUIBBT
  	alert: ThreeScaleDeveloperUIBBT
  	alert: ThreeScaleSystemAdminUIBBT
```

test/common/alerts_firing.go
```diff
@@ -115,8 +115,15 @@ func TestIntegreatlyAlertsFiring(t TestingTB, ctx *TestingContext) {
...
-		t.Skipf("Skipping due to known issue : %v, reported in Jira: https://issues.redhat.com/browse/MGDAPI-1404", err.Error())
-		//t.Fatal(err.Error())
+		isInProw, e := isInProw(ctx)
+		if e != nil {
+			t.Logf("error getting in_prow annotation: %s", e)
+		}
+		if isInProw {
+			t.Skipf("Skipping due to known issue in Prow, UIBBT alests fire there: %v, reported in Jira: https://issues.redhat.com/browse/MGDAPI-5193", err.Error())
+		} else {
+			t.Fatal(err.Error())
```
MStokluska, line 125: do we want this test to run at as part of functionals? If so, it looks to me like we are going to error out on non-prow runs

trepel: Yes, we do. I can validate this via pipeline to be 100% sure but `err` is only populated via `getFiringAlerts` (line 116) and if it is non-null then we should error because under normal circumstances (non-prow) there should be no alerts firing. If there is an error when calling `isInProw` then `e` is populated (not `err`) so that should not affect anything

trepel: Hey @MStokluska I executed both tests (C01 and "Verify Alerts are not firing during or after installation apart from DeadMansSwitch") against the cluster (using test/e2e/single make target) and both passed fine. I tried with `in_prow` set to false and also with `in_prow` missing completely.

PR #3101, opened 2023-02-08
refactor of creation of missing metrics alerts

laurafitzgerald:
# Issue link
https://issues.redhat.com/browse/MGDAPI-4964

#Changes

- openshift-monitoring - rhoam missing metrics alert with metrics used in critical alerts in that stack for rhoam with appropriate time 
- rhoam absent metrics monitoring - rhoam missing metrics alerts with metrics used in crtical alerts in that stack for rhoam with appropriate time 
- threescale absent metrics monitoring - threescale missing metrics created at the same time as the metrics critical to those alerts.
- isinreconciling error should no longer use absent.

- moving custom domain alerts to threescale reconciler but should be present even when custom domain isn't configure

# Verification steps

Run in a ROSA STS enabled cluster.
Verify that there is new alert rhoam missing metrics in openshift-monitoring.
Verify that on installation the alert doesn't fire. reach out to Tomas 

Trigger the metrics missing alert for both openshift-monitoring and observability promethues by scaling down RHOAM.

Non custom domain
Ensure that the observability rhoam missing metric alert does not include rhoam_custom_domain metric in the query

Custom domain

Configure Custom domain - probably will need an uninstall and install again 
Ensure that the observabiity rhoam missing metric alert does contain the rhoam_custom_domain metric in the query.


controllers/rhmi/prometheusRules.go
```diff
@@ -76,6 +76,23 @@ func (r *RHMIReconciler) newAlertsReconciler(installation *integreatlyv1alpha1.R
...
+			Namespace: observability.OpenshiftMonitoringNamespace,
+			GroupName: fmt.Sprintf("%s-general.rules", installationName),
+			Rules: []monitoringv1.Rule{
+				{
+					Alert: fmt.Sprintf("%sCriticalMetricsMissing", strings.ToUpper(installationName)),
+					Annotations: map[string]string{
+						"sop_url": resources.SopUrlCriticalMetricsMissing,
+						"message": "one or more critical metrics relating to RHOAM installation/upgrade have been missing for 30+ minutes",
+					},
+					Expr:   intstr.FromString(fmt.Sprintf(`absent(%s_version) == 1`, installationName)),
```
laurafitzgerald, line 90: only inlcuding the rhoam_version here as this is the only metric used in critical alerts for rhoam in the openshift-monitoring stack

controllers/rhmi/rhmi_controller.go
```diff
@@ -271,6 +272,18 @@ func (r *RHMIReconciler) Reconcile(ctx context.Context, request ctrl.Request) (c
 		return ctrl.Result{}, err
 	}
 
+	alertsClient, err := k8sclient.New(r.mgr.GetConfig(), k8sclient.Options{
```
laurafitzgerald, line 275: moving the creation of the alerts as early as possible in the reconciler

pkg/products/cloudresources/prometheusRules.go
```diff
@@ -130,9 +130,9 @@ func addElasticCacheSnapshotNotFoundAlert(ctx context.Context, client k8sclient.
 	return &alertsReconciler, nil
 }
 
-func getRedisCRsNames(ctx context.Context, client k8sclient.Client, logger l.Logger, ns string) ([]string, error) {
+func getRedisCRsNames(ctx context.Context, client k8sclient.Client, ns string) ([]string, error) {
```
laurafitzgerald, line 133: these changes could be removed. just noticed them during development.

pkg/products/observability/prometheusRules.go
```diff
@@ -325,7 +325,7 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
 						"sop_url": resources.SopUrlRHOAMIsInReconcilingErrorState,
 						"message": fmt.Sprintf("%s operator has finished installing, but has been in a error state while reconciling for last 10 minutes", strings.ToUpper(installationName)),
 					},
-					Expr:   intstr.FromString(fmt.Sprintf(`(%s_status{stage!="complete"} > 0) * on(pod) group_left(to_version, version) (%[1]s_version{to_version="",version=~".+"} > 0) OR absent(%[1]s_status) OR absent(%[1]s_version)`, installationName)),
+					Expr:   intstr.FromString(fmt.Sprintf(`(%s_status{stage!="complete"} > 0) * on(pod) group_left(to_version, version) (%[1]s_version{to_version="",version=~".+"} > 0)`, installationName)),
```
laurafitzgerald, line 328: removing the use of absent as it's covered by the missing metrics alert now

pkg/products/observability/prometheusRules.go
```diff
@@ -341,23 +341,6 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
 				},
 			},
 		},
-		{
```
laurafitzgerald, line 344: this is created in the rhmi controller now.

pkg/products/observability/prometheusRules.go
```diff
@@ -341,23 +341,6 @@ func (r *Reconciler) newAlertsReconciler(logger l.Logger, installType string) re
...
-			Namespace: namespace,
-			GroupName: fmt.Sprintf("%s-general.rules", installationName),
-			Rules: []monitoringv1.Rule{
-				{
-					Alert: fmt.Sprintf("%sCriticalMetricsMissing", strings.ToUpper(installationName)),
-					Annotations: map[string]string{
-						"sop_url": resources.SopUrlCriticalMetricsMissing,
-						"message": "one or more critical metrics have been missing for 10+ minutes",
-					},
-					Expr:   intstr.FromString(`(absent(kube_endpoint_address_available) or absent(kube_pod_container_status_ready) or absent(kube_pod_labels) or absent(kube_pod_status_phase) or absent(kube_pod_status_ready) or absent(kube_secret_info) or absent(rhoam_version) or absent(threescale_portals)) == 1`),
```
laurafitzgerald, line 355: removed the check on the kube state metrics as they will be present if any pod produces them. and the the alerts using them are 3scale specific.

pkg/products/threescale/prometheusRules.go
```diff
@@ -146,11 +152,10 @@ func (r *Reconciler) newAlertReconciler(logger l.Logger, installType string, ctx
 					},
 				},
 			},
-
 			{
 				AlertName: operatorAlertNamePrefix + "ksm-endpoint-alerts",
 				GroupName: " 3scale-operator-endpoint.rules",
-				Namespace: operatorNamespace,
+				Namespace: namespace,
```
laurafitzgerald, line 158: i'm not sure why this was like this. the prometheus rule was in the observability namespace not the observability operator namespace. that's where it should be.

pkg/products/threescale/prometheusRules.go
```diff
@@ -329,8 +333,78 @@ func (r *Reconciler) newAlertReconciler(logger l.Logger, installType string, ctx
 					},
 				},
 			},
+			{
+				AlertName: fmt.Sprintf("%s-missing-metrics", installationName),
```
laurafitzgerald, line 337: new alert here to cover metrics used in critical alerts that are created by the threescale reconciler. we have a similar pattern in cro as agreed. this should now cover the missing metrics at the right time.

pkg/products/threescale/prometheusRules.go
```diff
@@ -329,8 +333,78 @@ func (r *Reconciler) newAlertReconciler(logger l.Logger, installType string, ctx
...
+						For:    "15m",
+						Labels: map[string]string{"severity": "critical"},
+					},
+				},
+			},
 		},
-	}, nil
+	}
+
+	if customDomain.IsCustomDomain(r.installation) {
```
laurafitzgerald, line 356: only add this alert if custom domain is used

pkg/products/threescale/reconciler.go
```diff
@@ -269,12 +269,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, nil
 	}
 
-	phase, err = r.reconcileCustomDomainAlerts(ctx, serverClient)
```
laurafitzgerald, line 272: created as part of the regular 3scael prometheus rule reconciliation

pkg/products/threescale/reconciler.go
```diff
@@ -538,15 +538,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
-	alertsReconciler, err = r.newAlertReconciler(r.log, r.installation.Spec.Type, ctx, serverClient)
```
laurafitzgerald, line 541: moved earlier in the reconcile

pkg/resources/custom-domain/prometheusRules.go
```diff
@@ -1,81 +0,0 @@
-package custom_domain
```
laurafitzgerald, line 1: covered by threescale prometheus rule reconciliation.

pkg/products/threescale/prometheusRules.go
```diff
@@ -329,8 +333,78 @@ func (r *Reconciler) newAlertReconciler(logger l.Logger, installType string, ctx
...
+					{
+						Alert: "DnsBypassThreeScaleSystemAdminUI",
+						Annotations: map[string]string{
+							"sop_url": resources.SopUrlDnsBypassThreeScaleSystemAdminUI,
+							"message": "3Scale System Admin UI, bypassing DNS: If this console is unavailable, the client is unable to perform Account Management, Analytics or Billing.",
+						},
+						Expr:   intstr.FromString("threescale_portals{system_provider='false'} > 0"),
+						For:    "15m",
+						Labels: map[string]string{"severity": "critical", "product": installationName},
+					},
```
Boomatang, line 402: The DNS Bypass alerts also apply to clusters that are on a private network without public routes. These need to be part of every install.

Boomatang (changes requested):
From the initial code review the only problem I see is with the DNS bypass alerts. They are general alerts not linked to only custom domain alerts

pkg/products/threescale/prometheusRules.go
laurafitzgerald: ack. I'll update that. 👍

carlkyrillos (commented):
I gave the changed files an eye-review and the changes look good to me other than the [issue](https://github.com/integr8ly/integreatly-operator/pull/3101#discussion_r1100379806) that Jim flagged.

@laurafitzgerald Let me know when you're ready and I can do the verification steps

carlkyrillos:
The failing `ci/prow/commits` check can be fixed by changing the start of the commit message from:
```
MGDAPI-4964
refactor of creation of missing metrics alerts
```
to
```
MGDAPI-4964 refactor of creation of missing metrics alerts
```

controllers/rhmi/rhmi_controller.go
MStokluska: Just out of interest, what's the exact benefit of moving it as early as possible? 
If RHOAM fails due to pre-flight checks, do we want the alerts to fire, if so, maybe it's worth updating sop to include steps to take to include checking if pre-flights passed.

laurafitzgerald: Yes, if the installation fails for any reason from the beginning of the code we should have some mechanism for that to be flagged with SRE. Agree we should have SOP cases which cover pre-flight failures as well.

pkg/products/threescale/prometheusRules.go
laurafitzgerald: updated

laurafitzgerald:
Verification steps carried out with @carlkyrillos 

Before Scaling down The rhoam pod to trigger a missing metrics scenario - missing metrics alert in the openshift-monitoring stack.
Screenshot 2023-02-09 at 18 46 17
New Threescale missing metric alert in rhoam observabiity stack
Screenshot 2023-02-09 at 18 47 06
Verifying DNS alerts are still created even when custom domain isn't configured
Screenshot 2023-02-09 at 18 47 58
Scaled down RHOAM pod to produce a missing metrics scenario

Threescale missing metrics alert fired
Screenshot 2023-02-09 at 18 52 02
This metric missing previously would have triggered the rhoam missing metrics. it's not now. this alert only gets created at the time of threescale reconciler now so it shouldn't trigger on installation. @trepel are you fine to verify the cases this occured at rc testing? aka with ROSA STS enabled cluster

rhoam critical metrics missing firing.
Screenshot 2023-02-09 at 18 54 08

No other alerts that had previously used absent are firing in the case of rhoam being scaled down. aka it's metrics aren't available. in this case the installation alert firing doesn't make sense because we are not in failed installation, we are in missing metric scenario.

trepel:
yes, I am fine with retesting during rc1 installation.

carlkyrillos:
Verified changes with @laurafitzgerald and all tests are passing
/lgtm

PR #3109, opened 2023-02-15
MGDAPI-5222 : Fix custom domain status block not being reconciled.

Boomatang:
# Issue link

https://issues.redhat.com/browse/MGDAPI-5222

# What

If any error happen during the initial reconcile loop the status block for the custom domain may not get set.

Changes here allow the custom domain status block to be reconciled will not reconciling the sub domain route. Understand the linked SOP for why the sub domain route is not reconciled. 

# Verification steps

* Install this branch with a custom domain enabled
  * Guide: https://integreatly-operator.readthedocs.io/en/latest/products/custom_domain/
* Expected: Status block for custom domain is added to the rhmi CR. 
  * may take multiply reconcile loops to happen

PR #3111, opened 2023-02-16
MGDAPI-5232 : Fix route selection based on route prefix to route labels

Boomatang:
# Issue link

https://issues.redhat.com/browse/MGDAPI-5232

# What

Selection of master route was based on the URL containing `apps`. This may not be the case when using a custom domain.

# Verification steps

* Install this branch using a custom domain that **DOES NOT** start with `apps.`
  * custom domain guide: https://integreatly-operator.readthedocs.io/en/latest/products/custom_domain/
* Expected: 
  * In the rhmi CR the `routingSubdomain` will equal the custom domain
  * Install will go to a complete state
  * The custom domain status block will be missing from the rhmi CR. This PR does not include the changes from https://github.com/integr8ly/integreatly-operator/pull/3109

# Images
**operator:** quay.io/jfitzpat/managed-api-service:rhoam-v1.33.0-5232
 **index:** quay.io/jfitzpat/managed-api-service-index:1.33.0-5232

### DIFF
--- a/pkg/products/threescale/reconciler.go
+++ b/pkg/products/threescale/reconciler.go
@@ -3724,14 +3724,16 @@ func (r *Reconciler) reconcileTenantOutgoingEmailAddress(ctx context.Context, se
 		return fmt.Errorf("failed to get 3scale route list during portaClient creation, error: %v", err)
 	}
 	for _, route := range routes.Items {
-		if strings.Contains(route.Spec.Host, "master.apps") {
+		routeLabels := route.GetLabels()
+		value, exists := routeLabels["zync.3scale.net/route-to"]
+		if exists && value == "system-master" {
 			admRoute = route
 			found = true
 			break
 		}
 	}
 	if !found {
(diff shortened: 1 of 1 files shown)


pkg/products/threescale/reconciler.go
```diff
@@ -3724,7 +3724,9 @@ func (r *Reconciler) reconcileTenantOutgoingEmailAddress(ctx context.Context, se
 		return fmt.Errorf("failed to get 3scale route list during portaClient creation, error: %v", err)
 	}
 	for _, route := range routes.Items {
-		if strings.Contains(route.Spec.Host, "master.apps") {
+		routeLabels := route.GetLabels()
+		value, exists := routeLabels["zync.3scale.net/route-to"]
+		if exists && value == "system-master" {
```
MStokluska, line 3729: I would also update the "err" value on 3736 as the error is not reflecting the true error of the potential issue.

PR #3113, opened 2023-02-16
remove unused dependency

laurafitzgerald:
# Issue link

# What

# Verification steps
e2e tests passing


austincunningham:
e2e failed
```bash
  A08 - Verify all products routes are created
  /go/src/github.com/integr8ly/integreatly-operator/test/e2e/integretly_test.go:69
E0216 17:11:27.885439   20231 runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
```

laurafitzgerald:
I think the failure is caused by the unhandled error here -> https://github.com/integr8ly/integreatly-operator/blob/master/test/common/shared_functions.go#L452 or potentially not nil checking the rhmi object before checking the stage. Probably something we should look into fixing but I would say unrelated to this change. I've started the test again. I'll log a JIRA about the err not being handled.

laurafitzgerald:
@austincunningham e2e tests are passing now. are you fine with overriding codecov?

PR #3162, opened 2023-03-23
MGDAPI-5136 fixed maintenance window spam bug on fresh install

martinhesko:
# Issue link
[MGDAPI-5136](https://issues.redhat.com/browse/MGDAPI-5136)

# What
Fixes a bug with maintenanceWindow flag cycling on fresh installs of service affecting RHOAM versions. Further investigation still needed to cover upgrade scenarios. 

One problem I stumbled upon was `threescale-backend-redis-rhoam` not going back to 1 after the change like all the 5 other CR's, but going from ~200 to 55 `generation`s in the same time span, the number wasn't increasing with time as I checked on it so it's likely just 3scale reconciling.

# Verification steps
Fresh install:

- Follow [RHOAM OLM Install guide](https://integreatly-operator.readthedocs.io/en/latest/installation_guides/olm_installation) to create a service affecting release based on one of our existing releases (Include the tag in the manifest just like [here](https://github.com/integr8ly/integreatly-operator/blob/master/bundles/managed-api-service/1.32.0/manifests/managed-api-service.clusterserviceversion.yaml#L105)
- Check on the redis and postgres CR's, generation should be a high number ( > 1 )
- Modify the release by including changes from my PR and install it on a fresh cluster
- After RHOAM fully installs check redis and postgres CR's
- `generation` field should be `1` on all of them (except the one mentioned above)

### DIFF
--- a/controllers/subscription/subscription_controller.go
+++ b/controllers/subscription/subscription_controller.go
@@ -278,7 +278,7 @@ func (r *SubscriptionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func (r *SubscriptionReconciler) allowDatabaseUpdates(ctx context.Context, installation *integreatlyv1alpha1.RHMI, isServiceAffecting bool) error {
-	if installation.Status.ToVersion != "" && isServiceAffecting {
+	if installation.Status.Version != "" && installation.Status.ToVersion != "" && isServiceAffecting {
 		log.Info("Service affecting and upgrading, setting maintenanceWindow to true")
 		postgresInstances := &crov1alpha1.PostgresList{}
 		if err := r.Client.List(ctx, postgresInstances); err != nil {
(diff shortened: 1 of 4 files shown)


KevFan:
@martinhesko There are some failing unit tests. Looks like some needs to be updated 🤔

KevFan:
Changes looks good 👍 

I think the reason for why it's still affecting `threescale-backend-redis-rhoam` is due to the following does not account for fresh installations where `r.installation.Status.Quota` is `""`

https://github.com/martinhesko/integreatly-operator/blob/MGDAPI-5136/pkg/products/threescale/reconciler.go#L1281-L1288

Something like the following would probably work to fix this (would probably be good to have this as a function that can be unit tested):
```
	// If there is a quota change, the quota on the installation spec would not be set to the active quota yet
	isQuotaChange := func() bool {
		// During fresh installation, quota is not set until installation completes
		if r.installation.Status.Quota == "" {
			return false
		}
		
		return r.installation.Status.Quota != activeQuota
	}
```

PR #3180, opened 2023-04-04
MGDAPI-5300 Consume GCP maintenance addon parameter in RHOAM

Fiona-Waters:
# Issue link
[MGDAPI-5300](https://issues.redhat.com/browse/MGDAPI-5300)

# What
Consume new addon parameters from RHOAM that specify maintenance date and time. This includes maintenance day and maintenance hour parameters that will be added as part of [MGDAPI-5301](https://issues.redhat.com/browse/MGDAPI-5301). 

# Verification steps
Code review for this PR but can be tested properly as part of [MGDAPI-5301](https://issues.redhat.com/browse/MGDAPI-5301).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,8 @@ SMTP_ADDRESS ?= ''
 SMTP_PASS ?= ''
 SMTP_PORT ?= ''
 SMTP_FROM ?= ''
+MAINTENANCE_DAY ?= ''
+MAINTENANCE_HOUR ?= ''
 CRO_ROLE_ARN ?= 'arn:aws:iam::123456789012:role/example'
 THREESCALE_ROLE_ARN ?= 'arn:aws:iam::123456789012:role/example'
 
@@ -394,7 +396,7 @@ cluster/prepare/dms:
 .PHONY: cluster/prepare/addon-params
 cluster/prepare/addon-params:
 	@-oc process -n $(NAMESPACE) QUOTA=$(DEV_QUOTA) DOMAIN=$(CUSTOM_DOMAIN) \
- 		USERNAME=$(SMTP_USER) HOST=$(SMTP_ADDRESS) PASSWORD=$(SMTP_PASS) PORT=$(SMTP_PORT) FROM=$(SMTP_FROM) -f config/secrets/addon-params-secret.yaml | oc apply -f -
+ 		USERNAME=$(SMTP_USER) HOST=$(SMTP_ADDRESS) PASSWORD=$(SMTP_PASS) PORT=$(SMTP_PORT) FROM=$(SMTP_FROM) DAY=$(MAINTENANCE_DAY) HOUR=$(MAINTENANCE_HOUR) -f config/secrets/addon-params-secret.yaml | oc apply -f -
 
 .PHONY: cluster/prepare/sts
 cluster/prepare/sts:
--- a/pkg/products/cloudresources/reconciler.go
+++ b/pkg/products/cloudresources/reconciler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
@@ -50,6 +51,10 @@ import (
 
 const (
 	defaultInstallationNamespace = "cloud-resources"
+	MaintenanceDay               = "maintenance-day"
+	MaintenanceHour              = "maintenance-hour"
+	DefaultMaintenanceDay        = time.Thursday
+	DefaultMaintenanceHour       = 2
 )
 
 var redisServiceUpdatesToInstall = []string{"elasticache-20210615-002", "elasticache-redis-6-2-6-update-20230109", "elasticache-20230315-001"}
@@ -212,7 +217,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		}
 	}
 
-	phase, err = r.reconcileCloudResourceStrategies(client)
+	maintenanceDay, _, err := addon.GetStringParameter(ctx, client, operatorNamespace, MaintenanceDay)
+	if err != nil {
+		return phase, err
+	}
+
+	maintenanceHour, _, err := addon.GetStringParameter(ctx, client, operatorNamespace, MaintenanceHour)
+	if err != nil {
+		return phase, err
+	}
+
+	phase, err = r.reconcileCloudResourceStrategies(client, maintenanceDay, maintenanceHour)
 	if err != nil {
 		events.HandleError(r.recorder, installation, phase, "Failed to reconcile Cloud Resource strategies", err)
 		return phase, err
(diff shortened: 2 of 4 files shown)


pkg/products/cloudresources/reconciler.go
```diff
@@ -212,7 +217,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+	if err != nil {
+		return phase, err
+	}
+
+	maintenanceHour, _, err := addon.GetStringParameter(ctx, client, operatorNamespace, MaintenanceHour)
+	if err != nil {
+		return phase, err
+	}
+
+	phase, err = r.reconcileCloudResourceStrategies(client, maintenanceDay, maintenanceHour)
```
KevFan, line 230: I think this would be applied to AWS cluster also, do we want this or should this be only applied to GCP clusters ? 🤔

Fiona-Waters: I think we only want it for GCP clusters. As far as I know maintenance date is set in another way for AWS - these defaults were set prior to this change and will be used when no params are provided. Certainly adding the drop down in OCM is being explicitly asked for just for GCP. https://issues.redhat.com/browse/MGDAPI-5301 
Hope I understand correctly.

pkg/products/cloudresources/reconciler.go
```diff
@@ -212,7 +217,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
-	phase, err = r.reconcileCloudResourceStrategies(client)
+	maintenanceDay, _, err := addon.GetStringParameter(ctx, client, operatorNamespace, MaintenanceDay)
+	if err != nil {
+		return phase, err
+	}
+
+	maintenanceHour, _, err := addon.GetStringParameter(ctx, client, operatorNamespace, MaintenanceHour)
+	if err != nil {
+		return phase, err
+	}
```
KevFan, line 228:
```suggestion
	maintenanceDay, _, err := addon.GetStringParameter(ctx, client, r.ConfigManager.GetOperatorNamespace(), MaintenanceDay)
	if err != nil {
		return phase, err
	}

	maintenanceHour, _, err := addon.GetStringParameter(ctx, client, r.ConfigManager.GetOperatorNamespace(), MaintenanceHour)
	if err != nil {
		return phase, err
	}
```
Think this should be the RHOAM operator namespace where the addon parameter secret would be 🤔

Fiona-Waters: This variable has been set at the start of the reconcile function `operatorNamespace := r.Config.GetOperatorNamespace()` so I am using that here.

KevFan: Nope, these are different. `operatorNamespace := r.Config.GetOperatorNamespace()` is the CRO operator namespace ("redhat-rhoam-cloud-resource-operator"). It is the `ConfigManager` rather than the `Config` if we want to use the RHOAM operator namespace ("redhat-rhoam-operator") where the addon parameters secret would be 🤔

Fiona-Waters: Ok no problem, I will update it. Thank you for explaining.

pkg/products/cloudresources/reconciler.go
```diff
@@ -212,7 +217,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
 
-	phase, err = r.reconcileCloudResourceStrategies(client)
+	maintenanceDay, _, err := addon.GetStringParameter(ctx, client, r.ConfigManager.GetOperatorNamespace(), MaintenanceDay)
+	if err != nil {
+		return phase, err
+	}
+
+	maintenanceHour, _, err := addon.GetStringParameter(ctx, client, r.ConfigManager.GetOperatorNamespace(), MaintenanceHour)
+	if err != nil {
+		return phase, err
```
KevFan, line 227:
```suggestion
		return integreatlyv1alpha1.PhaseFailed, err
	}

	maintenanceHour, _, err := addon.GetStringParameter(ctx, client, r.ConfigManager.GetOperatorNamespace(), MaintenanceHour)
	if err != nil {
		return integreatlyv1alpha1.PhaseFailed, err
```

Let's return `integreatlyv1alpha1.PhaseFailed` here. If we are using the pre-exisiting `phase` from previous function calls, it can be `integreatlyv1alpha1.PhaseCompleted` which would be inaccurate.

Also any thoughts on logging the error ? Alternatively we can move these functions into `reconcileCloudResourceStrategies` instead so that any errors would emit an error event instead 🤔

Fiona-Waters: Yes I think it makes sense to move these into `reconcileCloudResourceStrategies` and return errors there where necessary. Pushing changes.

pkg/products/cloudresources/reconciler_test.go
```diff
@@ -415,3 +415,215 @@ func TestReconciler_setPlatformStrategyName(t *testing.T) {
...
+			r := &Reconciler{
+				Config:        tt.fields.Config,
+				ConfigManager: tt.fields.ConfigManager,
+				installation:  tt.fields.installation,
+				mpm:           tt.fields.mpm,
+				log:           tt.fields.log,
+				Reconciler:    tt.fields.Reconciler,
+				recorder:      tt.fields.recorder,
+			}
+			got, err := r.reconcileCloudResourceStrategies(tt.args.client, tt.args.maintenanceDay, tt.args.maintenanceHour)
```
KevFan, line 619: Unit tests likely need to be updated now that we've moved getting the addon parameters to this function 🤔

Fiona-Waters: Yes of course, apologies, will update.

trepel:
@Fiona-Waters @KevFan What do you think about updating Makefile as part of this PR? It would be possible to validated this via local installation (or via make cluster/deploy)
https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L397

Fiona-Waters:
@trepel @KevFan I have updated the makefile. In order to verify I have installed RHOAM locally and can confirm that the strategy config map is updated when the secret is manually updated. The make command works as expected also. Please let me know if any more changes or verification is required. Thanks.
Screenshot from 2023-04-12 14-39-28

Makefile
```diff
@@ -28,6 +28,8 @@ SMTP_ADDRESS ?= ''
 SMTP_PASS ?= ''
 SMTP_PORT ?= ''
 SMTP_FROM ?= ''
+MAINTENANCE_DAY ?= ''
+MAINTENANCE_HOUR ?= ''
```
trepel, line 32: Minor thing but would not it be better if there are some defaults here? Looking at the code it seems that Thursday 2-3 AM are defaults if nothing is specified so not a big deal, just for the sake of consistency. In OCM these params are required so user will have to select some option anyway.

Out of curiosity, why these defaults, I would expect Sunday? :)

trepel:
@Fiona-Waters thanks for the Makefile updates. I leave the final say to @KevFan but the verification steps you did are sufficient imo.

Makefile
Fiona-Waters: Thanks Tomas. I didn't add defaults as none of the other params had them and as you said we have defaults set elsewhere. But if you think it would be better I can add them.
@adam-cattermole might be able to clarify why the default is set to Thursday at 2am?

adam-cattermole: Hey, so these defaults are what have been used since 1.28.0 from what I can tell (from before the GCP branch was merged) and were set as the default once we removed RHMIConfig - https://github.com/integr8ly/integreatly-operator/commit/f6ef2a954e4b4f5a845b62bdf55b70d166e2244e

trepel: Thanks for all the replies. I have no objections for this staying as is

KevFan:
Changes looks good to me ! 👍 

/lgtm

PR #3184, opened 2023-04-05
MGDAPI-5512 - refactor: remove conversion of cro_postgres_max_memory metric to bytes

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-5512

# What

There is a discrepency is metric value for `cro_postgres_max_memory` which causes an alert to be fired for only GCP installs. This is due to the AWS provider uses MiB whereas GCP provider uses bytes for this metric.

https://github.com/integr8ly/cloud-resource-operator/pull/677 handles the conversion to bytes for the `cro_postgres_max_memory` metric for the AWS provider. This will use bytes as the standard for this metric.

With this change, there is no longer a need to handle conversion of this metric in alerts or dashboards

# Verification steps

* Code review and passing prow checks should be sufficient

### DIFF
--- a/pkg/products/monitoringcommon/dashboards/croResources.go
+++ b/pkg/products/monitoringcommon/dashboards/croResources.go
@@ -397,7 +397,7 @@ const MonitoringGrafanaDBCROResourcesJSON = `{
           "pluginVersion": "7.1.1",
           "targets": [
             {
-              "expr": "cro_postgres_max_memory*1048576",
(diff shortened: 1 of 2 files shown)


MStokluska:
/lgtm

Let's wait for cro fix to be in place and all the tests to complete successfully before mering. Thanks Kevin!

KevFan:
Rebased with master with new CRO release

/hold cancel

PR #3203, opened 2023-04-20
MGDAPI-5501 - feat: test module

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-5501

# What

* Go module for test directory to remove test dependencies from main module

# Verification steps

* Passing prow checks should be sufficient


Makefile
```diff
@@ -503,6 +503,10 @@ vendor/check/prow:
 vendor/fix:
 	go mod tidy
 	go mod vendor
+	cd test
+	go mod tidy
```
laurafitzgerald, line 507: I'm interestd to get thoughts on whether we should vendor the test folders dependencies. The goal of vendoring is about rebuilding older versions. W

KevFan: I've no real preference on this. Probably better to not vendor the test directory as it would otherwise not reduce the amount of files committed 🤔

MStokluska:
applying lgtm again - was lost due to rebase
/lgtm

PR #3208, opened 2023-04-27
MGDAPI-5688 reduce resources required for mcg

MStokluska:
# Issue link
https://issues.redhat.com/browse/MGDAPI-5688

# What
MCG had been using default values for memory and cpu requirements. RHOAM does not need these to be so high therefore they were reduced.

# Verification steps
- Install as fresh install on cluster and confirm that the resources allocated to nooba-core, nooba db resources and endpoints are as low as they can be.

### DIFF
--- a/pkg/products/mcg/reconciler.go
+++ b/pkg/products/mcg/reconciler.go
@@ -24,6 +24,7 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
+	k8sresource "k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/record"
 	k8spointer "k8s.io/utils/pointer"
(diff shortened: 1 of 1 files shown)


pkg/products/mcg/reconciler.go
```diff
@@ -24,6 +24,7 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
+	k8sresource "k8s.io/apimachinery/pkg/api/resource"
```
adam-cattermole, line 27: Minor but if we are opting for naming `k8sresource` could we rename original uses of `resource` instead of importing twice.

adam-cattermole:
The changes look good, verified that they are as expected on cluster and I tested this with some new values we have picked for storage - I think it makes sense to add these changes as part of this ticket. The new values for [the constants](https://github.com/integr8ly/integreatly-operator/blob/master/pkg/products/mcg/reconciler.go#L38-L39) should be these if you don't mind updating:

```go
	pvpoolStorageSize             = "16Gi"
	dbStorageSize                 = "1Gi"
```

pkg/products/mcg/reconciler.go
MStokluska: Thanks Adam, going to leave as `resource`.

adam-cattermole:
Prow seems very flaky today
/test scripts

Changes look good and verified resources were as intended

/lgtm

PR #3238, opened 2023-05-24
MGDAPI-5728 refactor alert manager secret creation

MStokluska:
# Issue link
https://issues.redhat.com/browse/MGDAPI-5728

# What
- update to the reconciliation of alert manager secret. Secret has been moved to rhoam ns and reconciliation is no longer triggered by OO but the bootstrap stage.

# Verification steps
- Install and uninstall RHOAM from this branch and confirm it works as expected. 
(ensure to run make cluster/prepare/local first to create monitoring stack)


MStokluska:
/retest

Failed to install/or watch cro (rhoam) and 3scale operator (mt).

pkg/products/monitoringcommon/alertmanagerConfig_test.go
```diff
@@ -116,14 +116,11 @@ func TestReconciler_reconcileAlertManagerSecrets(t *testing.T) {
 		},
 		Type: corev1.SecretTypeOpaque,
 	}
-	alertmanagerRoute := &v1.Route{
+	alertmanagerService := &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      alertManagerRouteName,
+			Name:      alertManagerServiceName,
```
carlkyrillos, line 121: Should the name here be `mockAlertManagerServiceName`? If not we should remove that [const](https://github.com/integr8ly/integreatly-operator/pull/3238/files#diff-a0492f0c6f3f90d001065aab6bef7c3084e210616030e84ed72f7c1a3a14e608R33) because it's not being used otherwise. 
```suggestion
			Name:      mockAlertManagerServiceName,
```

pkg/products/monitoringcommon/alertmanagerConfig.go
```diff
@@ -190,7 +192,7 @@ func ReconcileAlertManagerSecrets(ctx context.Context, serverClient k8sclient.Cl
 	configSecret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      config.AlertManagerConfigSecretName,
```
carlkyrillos, line 194: The name of the Secret needs to be changed to `alertmanager-<alertmanager-name>` (which in our case is `alertmanager-rhoam`) because we are unable to override the Alertmanager's .spec.configSecret with OBO. See the API [docs](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.AlertmanagerSpec) for more info about the default configSecret name.

Additionally, we should probably move all the alertmanager [config values](https://github.com/integr8ly/integreatly-operator/blob/rhoam-v1.36.0/pkg/config/observability.go#L14-L18) out of `pkg/config/observability.go` and put them in `pkg/config/monitoring.go` instead.

carlkyrillos (commented):
Verified that the alertmanager Secret gets properly consumed when it's name is changed to `alertmanager-rhoam`. Also verified that the contents of the Secret are correct. Just need to fix the name and then this is good to merge.

pkg/products/monitoringcommon/alertmanagerConfig.go
carlkyrillos: Once #3249 is merged, the constant `config.AlertManagerConfigSecretName` can be moved to [pkg/config/obo.go](https://github.com/integr8ly/integreatly-operator/pull/3249/files#diff-984212e2b097f414b48d126918cc96f447df4496c7d70f1219e54d53830d8f43R3-R5) and changed to `"alertmanager-rhoam"`. Also the namespace will need to be changed to `redhat-rhoam-operator-observability`, this [helper function](https://github.com/integr8ly/integreatly-operator/pull/3249/files#diff-984212e2b097f414b48d126918cc96f447df4496c7d70f1219e54d53830d8f43R7-R9) can also be used once #3249 is merged.
```suggestion
			Name:      config.AlertManagerConfigSecretName,
			Namespace: config.GetOboNamespace(installation.Namespace),
```

CC @MStokluska

pkg/products/monitoringcommon/alertmanagerConfig_test.go
MStokluska: Done, I've removed the const

pkg/products/monitoringcommon/alertmanagerConfig.go
MStokluska: Thanks Carl, so following this, the name will be "alertmanager-rhoam-alertmanager" ?

MStokluska: I've pushed an update, let me know what you think.

carlkyrillos:
@MStokluska I reviewed the latest changes and verified the install works as expected. However, the uninstall fails and hangs indefinitely. When uninstalling RHOAM, the rhmi_controller creates silences [here](https://github.com/integr8ly/integreatly-operator/blob/mgdapi-5727-obo/controllers/rhmi/rhmi_controller.go#L525-L534) that are still relying on the alertmanager route which doesn't exist and this blocks the uninstall. So we will need to refactor the silencing logic to work without the alertmanager route.

carlkyrillos:
Installation has been verified and all prow checks are passing. The uninstallation issues will be investigated/resolved in a followup ticket.

/lgtm

PR #3247, opened 2023-05-31
MGDAPI-5758 : set subscription resources to be empty

Boomatang:
# Issue link

https://issues.redhat.com/browse/MGDAPI-5758

# What

Set resources in the subscription so that the resources values are not taken form the CSV's

# Verification steps

* install RHOAM 1.36
* check resources on operators in, resources should be set
  * 3scale
  * marin3r
  * observability
* upgrade to this branch
* recheck the resources on the operators
* EXPECTED: resources set to the namespace default which does not include values for the limits.

### DIFF
--- a/pkg/resources/marketplace/manager.go
+++ b/pkg/resources/marketplace/manager.go
@@ -3,6 +3,7 @@ package marketplace
 import (
 	"context"
 	"fmt"
+	corev1 "k8s.io/api/core/v1"
 
 	l "github.com/integr8ly/integreatly-operator/pkg/resources/logger"
 	v1 "github.com/operator-framework/api/pkg/operators/v1"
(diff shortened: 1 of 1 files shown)


adam-cattermole:
Tested this on cluster and can see that the resource config is set on the deployment before the change, and it gets removed after:

1.36.0 deployment:
```
containers:
  - resources:
      limits:
        cpu: 100m
        memory: 300Mi
      requests:
        cpu: 100m
        memory: 300Mi
```

This PR deployment:
```
containers:
  - resources: {}
```

1.36.0 subscription:
```
spec:
  channel: threescale-mas
  installPlanApproval: Manual
  name: 3scale-operator
  source: rhmi-registry-cs
  sourceNamespace: redhat-rhoam-3scale-operator
```

this PR subscription:
```
spec:
  channel: threescale-mas
  config:
    resources: {}
  installPlanApproval: Manual
  name: 3scale-operator
  source: rhmi-registry-cs
  sourceNamespace: redhat-rhoam-3scale-operator
```

Can see that the limits are removed across the board post upgrade:
image

/lgtm

PR #2699, opened 2022-06-01
MGDAPI-3923

Patryk-Stefanski:
# Issue link
[MGDAPI-3923](https://issues.redhat.com/browse/MGDAPI-3923)

# What
Addition of a new `LOCAL` variable. If set to true will change the namespace name to local-rhoam-operator
Changed default  `INSTALLATION_TYPE` value to managed-api 
 
# Verification steps

Verify that you can install RHOAM locally and via OLM and that it works as expected (All products install successfully)

1. Local install 
     1.1 Checkout this branch 
     1.2 run `make cluster/prepare/local`
     1.3 run `make code/run`
     1.4 make sure that RHOAM installs as expected

2. OLM install guide
       2.1 install RHOAM via OLM - [guide](https://github.com/RHCloudServices/integreatly-help/blob/master/guides/olm/installing-rhmi-bundle-format.md#installing-rhmi-through-olm-with-bundle-format)
       2.2 make sure that RHOAM installs as expected

PR #3257, opened 2023-06-06
MGDAPI-5764 - removed all instances of local-rhoam-operator

den-rgb:
# Issue link
https://issues.redhat.com/browse/MGDAPI-5764

# What
As part of the migration from OO to OBO, RHOAM will leverage MTBR's PnP functionality to deploy/reoncile ServiceMonitors and Probes (and probably other CRs) when installing RHOAM. Due to some limitations with the PnP's templating, hardcoded namespace for these CRs are needed. There is no workaround for "local-rhoam-operator" as there is with other namespaces and so it was determined that removing it from the codebase was the best path forward.

# Verification steps
Make sure all instances of local-rhoam-operator have been removed from the integreatly-operator codebase.


README.md
```diff
@@ -133,7 +134,7 @@ Once the installation completed the command wil result in following output:
 ### Local and OLM installation type
 If you installed RHOAM locally or via OLM then you can uninstall one of two ways:
 
-- for local installation use the `local-rhaom-operator Namespace`
+- for local installation use the namespace specified in the RHMI CR's spec.namespaceprefix value.
```
carlkyrillos, line 137: Not a big deal but this may be easier to read for people who aren't as familiar with kubernetes/openshift

```suggestion
- for local installation use the namespace that the RHMI CR was created in.
```

carlkyrillos:
I reviewed the changes and it looks like everything is cleaned up as intended but @laurafitzgerald would you be able to give this a quick look since you're more familiar with some of the original `local-rhoam-` work?

laurafitzgerald:
There's three references to the LOCAL variable in the Makefile that will need to be removed as well. You could also refer to the changes here -> https://github.com/integr8ly/integreatly-operator/pull/2699/files

carlkyrillos:
I think this is ready to be merged now that the Makefile references have been removed, CC @laurafitzgerald, @den-rgb 
/lgtm

carlkyrillos:
e2e prow checks are passing after rebase
/lgtm

PR #3235, opened 2023-05-23
MGDAPI-5732 Update prometheus alerts groups and location

MStokluska:
# Issue link
https://issues.redhat.com/browse/MGDAPI-5732
https://issues.redhat.com/browse/MGDAPI-5731

TODO:
- tidy up imports order
- e2e once monitoring stack is in place

# What
- Moving Observability namespace alerts to redhat-rhoam-operator namespace
- Editing the alert creation logic so that alerts can be created using monitoringv1 and monitoring.rhobs
- Openshift monitoring stack remain unchanged
- Copying KCs alerts from KC namespace over to RHOAM and updating it's group to monitoring.rhobs
- Refactoring unit tests
- Updating go mod to be compatible with provided version of obo prometheus (also to be updated to most recent one in another jira as it requires rhoam deps to be updated accordingly)
- Updating probes to change the apis to monitoring.rhobs and namespace along with some unit tests refactor (the MGDAPI-5731 has been completed as part of this PR as well due to the amount of changes that would need to be replicated over both PRs)

For now, leaving along e2e tests since we don't have obo monitoring stack in place yet. 
Leaving generated-deadmansswitch alert in OO ns for now as this will need to be configured outside of RHOAM repo.

# Verification steps
- Install from this branch and confirm that alerts are created under rhoam ns - please note, OO prometheus is not a valid verification.
- Potentially wait for OBO monitoring stack before reviewing this PR
- Confirm that probes have been created under rhoam ns along with correct api group

PR #3262, opened 2023-06-07
MGDAPI-5783 refactored some of the e2e tests for OBO

carlkyrillos:
# Issue link
JIRA: [MGDAPI-5783](https://issues.redhat.com/browse/MGDAPI-5783)

# What
The e2e prow checks([example](https://github.com/integr8ly/integreatly-operator/pull/3235)) are failing when run on the OBO [feature branch](https://github.com/integr8ly/integreatly-operator/tree/mgdapi-5727-obo) because the prow checks are run on OCP clusters and OCP clusters don't have the package-operator (or OBO) installed by default. In order to get around this, we need to move any tests that rely on any observability components into a new test suite. We will then only run that test suite when the IN_PROW [environment variable](https://github.com/integr8ly/integreatly-operator/blob/mgdapi-5727-obo/Makefile#L17) is set to false. We also want to update the Makefile to only run [cluster/prepare/rhoam-config](https://github.com/integr8ly/integreatly-operator/blob/mgdapi-5727-obo/Makefile#L427-L429) when IN_PROW is set to false.

**NOTE:** the scope of this ticket is not to refactor all tests to be compatible with OBO or remove any/all references to OO in the test suite. The scope of this ticket is just to get majority of the e2e tests working while development on the OBO feature branch continues. A complete refactor of the e2e test suite that removes references to OO will be completed in a followup ticket.

# Verification steps
Assuming the e2e PROW checks pass, the only other verification this PR needs is an eye review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -226,7 +226,7 @@ test/e2e/rhoam/prow: export NAMESPACE:= $(NAMESPACE_PREFIX)operator
 test/e2e/rhoam/prow: export INSTALLATION_PREFIX := redhat-rhoam
 test/e2e/rhoam/prow: export INSTALLATION_NAME := rhoam
 test/e2e/rhoam/prow: export INSTALLATION_SHORTHAND := rhoam
-test/e2e/rhoam/prow: IN_PROW = "true"
+test/e2e/rhoam/prow: export IN_PROW = "true"
 test/e2e/rhoam/prow: test/e2e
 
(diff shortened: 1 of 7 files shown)


laurafitzgerald:
Hi @carlkyrillos the changes here look good and I can see the test are passing now. wdyt about going ahead and merging this so that we can ensure that for other prs there isn't regression for the rest of the test as we work through the observability changes.

PR #3219, opened 2023-05-05
MGDAPI-5377 Updates to alerts check script to be more flexible

trepel:
# Issue link

https://issues.redhat.com/browse/MGDAPI-5377

# What

Changes to make the script for checking the alerts more flexible
- parametrized sleep time
- added timestamp to generated csv file names
- added alert severity do the generated csv files
- updated test cases that mentioned the script

# Verification steps

Run the script
./alerts-check.sh
./alerts-check.sh 30
./alerts-check.sh 30 3scale

and see the results (generated csv files)

PR #3268, opened 2023-06-12
MGDAPI-5706 - updated access to prometheus pod and changed namespace

den-rgb:
# Issue link
https://issues.redhat.com/browse/MGDAPI-5706

# What
- Changed scripts to access prometheus api endpoints without using a route.
- Modified scripts to access `redhat-rhoam-operator-observability` namespace instead of the depreciated `redhat-rhoam-observability` namespace.
 

# Verification steps
- Create a cluster
- Once cluster is created sign into it and run these commands: 
1. `INSTALLATION_TYPE=managed-api LOCAL=false make cluster/prepare/local`

## Running *alert-check.sh*
- run the script `scripts/alert-check.sh `
- `INSTALLATION_TYPE=managed-api LOCAL=false  make code/run  ` , wait for install to complete.
- THe script should be waiting and only complete once rhoam install is finished.
- Scale down a resource causing an alert to fire eg. *3scale*
- Make sure it generates all the necessary csv files corresponding to the firing/pending alerts
- It is also possible to check individual products by running eg. `scripts/alert-check.sh  3scale`

This should be the output along with a csv file listing all the alerts:
image

## Running *capture-resource-metrics.sh*
- Create necessary files following the instructions on the top of the file.
- run `scripts/capture_resource_metrics.sh`

The output should be a list of metrics generated in the terminal

**Note** if you get an error such as:
`wget: server returned error: HTTP/1.1 400 Bad Request`
1. Make sure your cluster login hasn't timed out.
2. Try removing the timestamp files and creating them again making sure you create the start time first.

## Running *monitoringupgradetest.sh*
- Create an OLM upgrade scenario
- Before triggering the upgrade run the script `scripts/monitoringupgradetest.sh`
- Make sure there are no errors

### DIFF
--- a/scripts/alerts-check.sh
+++ b/scripts/alerts-check.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 # USAGE
-# ./alerts-check.sh <optional sleep in seconds> <optional product-name>
+# ./alerts-check.sh <optional product-name>
 # ^C to break
 # Generates two files one for firing and one for pending alerts
 #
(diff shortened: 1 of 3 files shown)


scripts/alerts-during-perf-testing.sh
```diff
@@ -11,83 +11,149 @@
 NAMESPACE_PREFIX="${NAMESPACE_PREFIX:-$(oc get RHMIs --all-namespaces -o jsonpath='{.items[0].spec.namespacePrefix}')}"
 RHSSO="rhsso"
 USER_SSO="user-sso"
+ROUTE="http://localhost:9090/api/v1/alerts"
 THREESCALE="3scale"
 TOKEN=$(oc whoami --show-token)
-MONITORING_ROUTE=$(echo "https://$(oc get route prometheus -n ${NAMESPACE_PREFIX}observability -o=jsonpath='{.spec.host}')")/api/v1/alerts
+OPENSHIFT_MONITORING=$(oc exec -n openshift-monitoring prometheus-k8s-0 -- curl $ROUTE -s -H "Authorization: Bearer $TOKEN")
+RHOAM_MONITORING=$(oc exec -n redhat-rhoam-operator-observability prometheus-rhoam-0 -- wget -qO- --header='Accept: application/json' --header="Authorization: Bearer $TOKEN" --no-check-certificate $ROUTE)
```
trepel, line 18: Use NAMESPACE_PREFIX here instead of hard-coded ns name

trepel:
@den-rgb The "alert" script has been changed as part of https://github.com/integr8ly/integreatly-operator/pull/3219/files so it will cause troubles when merging OBO branch into master. Not sure if it makes sense to try to deal with this in advance (I think that it would be nice though)

Not strictly related to your ticket ticket but could you please implement some waits so that the script can be triggered before RHOAM installation without failing? It would allow for triggering the script before RHOAM installation and then initiating RHOAM installation and once finished and script stopped it would tell what alerts were firing/pending during RHOAM installation - which is quite useful. Something along the lines of https://github.com/trepel/integreatly-operator/commit/b0db01fe353abd37ff4fe048463b0b0064462b08 (not tested yet)

scripts/alerts-during-perf-testing.sh
```diff
@@ -11,83 +11,149 @@
...
 }
 
-# If no args are passed in then run for all products
-if (( $# == 0 )); then
-  while :; do
-    # generate a report
-    curl -s -H "Authorization: Bearer $TOKEN" $MONITORING_ROUTE \
-    | jq -r '.data.alerts[]| select(.state=="pending") | [.labels.alertname, .state, .activeAt ] | @csv'>> tmp-alert-pending-during-perf-testing-report.csv
+# Define an array of products to report on
+declare -a products=("3scale" "user-sso" "rhsso" "marin3r")
```
Patryk-Stefanski, line 93: I think the definitions should be grouped together at the top.

scripts/alerts-during-perf-testing.sh
Patryk-Stefanski, line 18: can we use `NAMESPACE_PREFIX`  in here?

scripts/alerts-during-perf-testing.sh
```diff
@@ -11,83 +11,149 @@
...
-      echo "============================================================================"
-    fi
+function CHECK_NO_ALERTS() {
+  # Extract firing alerts from OpenShift monitoring
+  openshift_alerts=$(echo "$OPENSHIFT_MONITORING" | jq -r '.data.alerts[] | select(.state == "firing") | [.labels.alertname, .state, .activeAt] | @csv')
+
+  # Extract firing alerts from RHOAM monitoring
+  rhoam_alerts=$(echo "$RHOAM_MONITORING" | jq -r '.data.alerts[] | select(.state == "firing") | [.labels.alertname, .state, .activeAt] | @csv')
+
+  # Extract firing alerts from OpenShift monitoring
```
Patryk-Stefanski, line 39:
```suggestion
  # Extract pending alerts from OpenShift monitoring
```

scripts/alerts-during-perf-testing.sh
```diff
@@ -11,83 +11,149 @@
...
+  openshift_alerts_pending=$(echo "$OPENSHIFT_MONITORING" | jq -r '.data.alerts[] | select(.state == "pending") | [.labels.alertname, .state, .activeAt] | @csv')
+
+  # Extract firing alerts from RHOAM monitoring
```
Patryk-Stefanski, line 42:
```suggestion
  # Extract pending alerts from RHOAM monitoring
```

den-rgb:
@trepel I've made the suggested changes, Is there anything else that could cause possible issues?

scripts/alert-check.sh
```diff
@@ -0,0 +1,181 @@
...
+do
+    echo "Waiting for Prometheus route in ${NAMESPACE_PREFIX}observability namespace to appear on cluster. Next check to be in 1 minute."
+    sleep 60
+done
+
+SLEEP_TIME="${1-5}"
+RHSSO="rhsso"
+USER_SSO="user-sso"
+ROUTE="http://localhost:9090/api/v1/alerts"
+THREESCALE="3scale"
```
Patryk-Stefanski, line 32: Dont think we need these.

```suggestion
ROUTE="http://localhost:9090/api/v1/alerts"
```

Patryk-Stefanski:
Is there a reason why we changed form `alerts-check.sh` -> `alert-check.sh`  ?

Patryk-Stefanski:
code looks good and verification was successful.  /lgtm

carlkyrillos:
Not sure why but it looks like @Patryk-Stefanski's lgtm wasn't added.
/lgtm

PR #3276, opened 2023-06-16
MGDAPI-4895 - resolve go.opentelemetry.io/proto/otlp replaces

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-4895

# What

* Resolve go.opentelemetry.io/proto/otlp replaces
  * Remove use of `github.com/operator-framework/operator-registry/pkg/lib/bundle` as it causing a protobuf namespace conflict issue until it is resolved
      *  https://github.com/operator-framework/operator-registry/issues/1114
* Update gosec to v2.16.0
# Verification steps

* Passing prow check should be sufficient to ensure no breaking changes except for `gosec`
* `make gosec` is failing due to the image was previously using v2.11.0 with the following error:
```
[gosec] 2023/06/20 14:11:43 Checking file: /Users/chfan/dev/integreatly-operator/pkg/resources/constants/subscriptionConstants.go
2023/06/20 14:11:45 Unexpected package creation during export data loading
make: *** [gosec] Error 1
```
Updating to v2.16.0 will resolve this but this PR needs to be merged first since the base image is still using the old version. You can install the updated version locally and running gosec locally to ensure there is no issues: 
```
# Might need to delete current version if hitting the same issue
go install github.com/securego/gosec/v2/cmd/gosec@v2.16.0
make vendor/fix
make gosec
```


KevFan:
\cc @austincunningham for awareness as I believe you were also facing the same issue

KevFan:
Go 1.19 has been factored out to it's own PR so that the dependency changes can be fully prow e2e tested 
* https://github.com/integr8ly/integreatly-operator/pull/3283

pkg/products/monitoringspec/reconciler.go
```diff
@@ -380,7 +380,7 @@ func (r *Reconciler) reconcileRoleBinding(ctx context.Context,
 		}
 		roleBinding.RoleRef = rbac.RoleRef{
 			APIGroup: roleRefAPIGroup,
-			Kind:     bundle.RoleKind,
+			Kind:     "Role",
```
adam-cattermole, line 383: I see these constants still exist in the latest version of the operator-framework package, any reason for opting for manually defined string instead? https://pkg.go.dev/github.com/operator-framework/operator-registry@v1.28.0/pkg/lib/bundle#pkg-constants

KevFan: Yes, this is a workaround to prevent - https://github.com/operator-framework/operator-registry/issues/1114

adam-cattermole: Aha gotcha thanks

test/go.mod
```diff
@@ -20,28 +20,27 @@ require (
...
 	golang.org/x/net v0.8.0
 	golang.org/x/term v0.6.0
 	golang.org/x/text v0.8.0
 	google.golang.org/api v0.114.0
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.25.0
-	k8s.io/apiextensions-apiserver v0.25.0
-	k8s.io/apimachinery v0.25.0
+	k8s.io/api v0.26.3
+	k8s.io/apiextensions-apiserver v0.26.2
```
adam-cattermole, line 39: Should we opt for using the same version of the k8s.io package here as the others (v0.26.3)?

KevFan: Good catch, yes I think we should. I'll update 👍

adam-cattermole:
Changes look good, thanks. Also verified running `gosec` locally succeeds.

/lgtm

PR #3283, opened 2023-06-19
MGDAPI-5854 - go: v1.19

KevFan:
# Issue link

https://issues.redhat.com/browse/MGDAPI-5854

# What

Update to use go1.19. Passing prow checks is sufficient.

`vendor` prow check will fail until this gets merged. Once merged, the master prow pipelines will get auto updated to use go1.19 as the base image

# Verification steps

* Passing prow checks except for vendor

### DIFF
--- a/pkg/resources/marketplace/manifestHelper.go
+++ b/pkg/resources/marketplace/manifestHelper.go
@@ -3,7 +3,6 @@ package marketplace
 import (
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -120,7 +119,7 @@ func GetCRDsFromManifestAsStringList(dir string, regex string, packageYaml strin
 			} else {
 				// iterate all other versions look for crds that don't exist in the current version
 				// and add them to the stringlist for building the config map
-				files, err := ioutil.ReadDir(currentFolder)
+				files, err := os.ReadDir(currentFolder)
 				if err == nil {
 					for _, f := range files {
 						// iterate through all fils in the folder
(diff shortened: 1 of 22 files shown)


pkg/resources/marketplace/manifestHelper.go
```diff
@@ -179,8 +178,8 @@ func CheckFoldersForMatch(dir string, folders []*semver.Version, currentFolder s
 }
 
 // GetCrdDetails reads the crd file
-func GetCrdDetails(crdConfig *apiextensionv1.CustomResourceDefinition, currentFolder string, f os.FileInfo) {
-	yamlFile, err := ioutil.ReadFile(filepath.Clean(currentFolder + string(os.PathSeparator) + f.Name()))
+func GetCrdDetails(crdConfig *apiextensionv1.CustomResourceDefinition, currentFolder string, f os.DirEntry) {
```
laurafitzgerald, line 181: Hi Kev. Do you know if this is covered by e2e testing? I don't see a related unit test.

laurafitzgerald:
@KevFan one question above but overall the changes look good.

pkg/resources/marketplace/manifestHelper.go
KevFan: @laurafitzgerald Yes, I think this is generally covered by an installation using a config map catalog source (I believe marin3r installation is still using this instead of using an index). So I think as long as the e2e succeeds, this is technically covered. Installation will not complete if this change had any issues 🤔

laurafitzgerald: Thanks for clarification. In that case I'm happy to approve lgtm this if you are ready.

KevFan: Thanks ! Yeah, I think this PR is ready 👍

PR #3284, opened 2023-06-22
[WIP]MGDAPI-5304 - feat: addon instance status reporting

KevFan:
# Issue link

# What

# Verification steps


apis/v1alpha1/rhmi_conditions.go
```diff
@@ -0,0 +1,99 @@
...
+	)
+}
+
+func (i *RHMI) HealthyCondition() metav1.Condition {
+	return newRHMICondition(HealthyConditionType, metav1.ConditionTrue, "Healthy", "Core components healthy")
+}
+
+func (i *RHMI) UnHealthyCondition() metav1.Condition {
+	return newRHMICondition(HealthyConditionType, metav1.ConditionFalse, "Healthy", "One or more core components unhealthy")
+}
```
KevFan, line 74: Since we can't use the Healthy condition type provided by addon addon instance, I would like people's thought on whether we want our own `Healthy` conditions for core compoents since it's technically covered by the `Degraded` condition in terms of reporting. 

The only benefit currently to having this is to signify that core compoents is in a healthy state (phase complete). So we can have a status with `Healthy == True` and `Degraded == True` if for example, only Observability is down.

apis/v1alpha1/rhmi_types.go
```diff
@@ -248,6 +248,65 @@ func (i *RHMI) GetPullSecretSpec() *PullSecretSpec {
...
+// IsUninstallBlocked when IsUninstalling and deletion timestamp is over 2 hours old
+func (i *RHMI) IsUninstallBlocked() bool {
+	return i.IsUninstalling() && i.DeletionTimestamp.Add(2*time.Hour).Before(time.Now())
+}
+
+// IsCoreComponentsHealthy Helper for checking if core components affecting SLO are healthy
+func (i *RHMI) IsCoreComponentsHealthy() bool {
+	return i.IsProductInInstallStagePhaseComplete(ProductCloudResources) &&
+		i.IsProductInInstallStagePhaseComplete(ProductRHSSOUser) &&
+		i.IsProductInInstallStagePhaseComplete(Product3Scale)
```
KevFan, line 307: Are we happy with just these as our core componts which affect SLO ?

main.go
```diff
@@ -176,6 +190,43 @@ func main() {
...
+
+	// Check is addon operator installed
+	addonOperatorInstalled, err := status.IsAddonOperatorInstalled(client)
+	if err != nil {
+		setupLog.Error(err, "unable to check is addon operator installed", "controller", "AddonInstance")
+		os.Exit(1)
+	}
+
+	// Only start controller if addon operator is installed and is not a sandbox install
+	if addonOperatorInstalled && !isSandbox {
```
KevFan, line 215: I assumed we do not want to start the controller on sandbox installs

carlkyrillos: @KevFan I think this is probably safe. RHOAM on the current stg and prod sandbox clusters is already installed by CatalogSource anyways.

Makefile
```diff
@@ -403,6 +403,11 @@ cluster/prepare/sts:
 	@-oc process -n $(NAMESPACE_PREFIX)cloud-resources-operator NAME=sts-credentials NAMESPACE=$(NAMESPACE_PREFIX)cloud-resources-operator ROLE_ARN=$(CRO_ROLE_ARN) -f config/secrets/sts-secret.yaml | oc apply -f -
 	@-oc process -n $(NAMESPACE_PREFIX)3scale NAME=sts-s3-credentials NAMESPACE=$(NAMESPACE_PREFIX)3scale ROLE_ARN=$(THREESCALE_ROLE_ARN) -f config/secrets/sts-secret.yaml | oc apply -f -
 
+
+.PHONY: cluster/prepare/addon-instance
+cluster/prepare/addon-instance:
+	@-oc apply -f config/samples/addoninstance_v1alpha1.yaml
```
carlkyrillos, line 409: Do we need to update the `make cluster/cleanup` target to remove this as well?

carlkyrillos (commented):
Completed verification steps, left a few comments, and reviewed the changed files
/lgtm

Makefile
KevFan: No, the addon instance will be deleted when the namespace is deleted in `make cluster/cleanup`

apis/v1alpha1/rhmi_conditions.go
Patryk-Stefanski: I'm happy with this approach, think the condition descriptions are clear enough to distinguish the difference between healthy and degraded statuses.

Patryk-Stefanski:
Should Teardown complete report true once the RHMI CR is removed ?

KevFan:
> Should Teardown complete report true once the RHMI CR is removed ?

No. I think when RHMI CR is removed, the status type`addons.managed.openshift.io/Installed` is `False` because the reason is `Teardown complete` when it's been fully uninstalled unless I've been using the status conditions wrong 🤔

Patryk-Stefanski:
Ah okay,  I was just a bit confused with the layout of the conditions and reading it wrong.

The code looks good and verification was successful.  /lgtm

KevFan:
Hold for now with new conversations in EPIC brief
/hold

PR #3297, opened 2023-07-03
MGDAPI-5857 - return non-empty result to requeue rhmi controller

kramaranya:
# Issue link
[MGDAPI-5857](https://issues.redhat.com/browse/MGDAPI-5857)

# What

- Fix a [bug](https://github.com/integr8ly/integreatly-operator/pull/3297/commits/e6ac2c59057596332bb43232f05a8dd75aad20de) when rhoam install stucks if pagerduty secret does not exist.
- Update a logic by removing extra `ctrl.Result`s in preflightChecks and handleUninstall functions as we already have `ctrl.Result` in Reconcile function.

# Verification steps

- Eye review of the code changes should be sufficient, but if you want you can debug and see if rhoam install does not stuck if pagerduty secret does not exist - don't forget to remove `redhat-rhoam-pagerduty` secret before debugging

### DIFF
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -357,7 +357,7 @@ func (r *RHMIReconciler) Reconcile(ctx context.Context, request ctrl.Request) (c
 		if resources.Contains(installation.GetFinalizers(), previousDeletionFinalizer) {
 			installation.SetFinalizers(resources.Replace(installation.GetFinalizers(), previousDeletionFinalizer, deletionFinalizer))
 			if err = r.Update(context.TODO(), installation); err != nil {
-				return ctrl.Result{}, err
+				return retryRequeue, nil
 			}
 		} else {
 			installation.SetFinalizers(append(installation.GetFinalizers(), deletionFinalizer))
@@ -371,7 +371,11 @@ func (r *RHMIReconciler) Reconcile(ctx context.Context, request ctrl.Request) (c
 	// either not checked, or rechecking preflight checks
 	if installation.Status.PreflightStatus == rhmiv1alpha1.PreflightInProgress ||
 		installation.Status.PreflightStatus == rhmiv1alpha1.PreflightFail {
-		return r.preflightChecks(installation, installType, configManager)
+		err := r.preflightChecks(installation, installType, configManager)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		return retryRequeue, nil
 	}
 
 	// If the CR is being deleted, handle uninstall and return
@@ -394,7 +398,7 @@ func (r *RHMIReconciler) Reconcile(ctx context.Context, request ctrl.Request) (c
 		installation.Status.ToVersion = version.GetVersionByType(installation.Spec.Type)
 		log.Infof("Setting installation.Status.ToVersion on initial install", l.Fields{"version": version.GetVersionByType(installation.Spec.Type)})
 		if err := r.Status().Update(context.TODO(), installation); err != nil {
-			return ctrl.Result{}, err
+			return retryRequeue, nil
 		}
 		metrics.SetVersions(string(installation.Status.Stage), installation.Status.Version, installation.Status.ToVersion, string(externalClusterId), installation.CreationTimestamp.Unix())
 		metrics.SetThreeScalePortals(nil, 0) // expose metric and set default value
@@ -762,14 +766,16 @@ func (r *RHMIReconciler) handleUninstall(installation *rhmiv1alpha1.RHMI, instal
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	if err := r.Client.List(context.TODO(), alerts, &k8sclient.ListOptions{
+	err = r.Client.List(context.TODO(), alerts, &k8sclient.ListOptions{
 		LabelSelector: ls,
-	}); err != nil {
+	})
+	if err != nil {
 		return ctrl.Result{}, err
 	}
 
 	for i := range alerts.Items {
-		if err := r.Client.Delete(context.TODO(), &alerts.Items[i]); err != nil {
+		err := r.Client.Delete(context.TODO(), &alerts.Items[i])
+		if err != nil {
 			return ctrl.Result{}, err
 		}
 	}
@@ -825,24 +831,26 @@ func (r *RHMIReconciler) handleUninstall(installation *rhmiv1alpha1.RHMI, instal
 	if len(installation.Finalizers) == 1 && installation.Finalizers[0] == deletionFinalizer {
 		log.Infof("Finalizers: ", l.Fields{"length": len(installation.Finalizers)})
 		// delete ConfigMap after all product finalizers finished
-		if err := r.Client.Delete(context.TODO(), &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: installationCfgMap, Namespace: installation.Namespace}}); err != nil && !k8serr.IsNotFound(err) {
+		err := r.Client.Delete(context.TODO(), &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: installationCfgMap, Namespace: installation.Namespace}})
+		if err != nil && !k8serr.IsNotFound(err) {
 			merr.Add(fmt.Errorf("failed to remove installation ConfigMap: %w", err))
 			installation.Status.LastError = merr.Error()
 			err = r.Client.Update(context.TODO(), installation)
 			if err != nil {
 				merr.Add(err)
 			}
-			return retryRequeue, merr
+			return ctrl.Result{}, merr
 		}
 
-		if err = r.handleCROConfigDeletion(*installation); err != nil && !k8serr.IsNotFound(err) {
+		err = r.handleCROConfigDeletion(*installation)
+		if err != nil && !k8serr.IsNotFound(err) {
 			merr.Add(fmt.Errorf("failed to remove Cloud Resource ConfigMap: %w", err))
 			installation.Status.LastError = merr.Error()
 			err = r.Update(context.TODO(), installation)
 			if err != nil {
 				merr.Add(err)
 			}
-			return retryRequeue, merr
+			return ctrl.Result{}, merr
 		}
 
 		installation.SetFinalizers(resources.Remove(installation.GetFinalizers(), deletionFinalizer))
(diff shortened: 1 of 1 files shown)


controllers/rhmi/rhmi_controller.go
```diff
@@ -338,26 +338,26 @@ func (r *RHMIReconciler) Reconcile(ctx context.Context, request ctrl.Request) (c
 
 	installType, err := TypeFactory(ctx, installation.Spec.Type, r.Client)
 	if err != nil {
-		return ctrl.Result{}, err
+		return retryRequeue, nil
```
KevFan, line 341: I'm curious as to why we are ignoring to return `err` in some places vs not in other places 🤔 

If we dont return the error, I believe this won't be logged by the reconciler which may a valid error to know where in the code it is failing for debug purposes (For example in this instance, we may fail to determine the platform type or an unknown installation type was used - https://github.com/integr8ly/integreatly-operator/blob/master/controllers/rhmi/types.go#L131-L160)

kramaranya: The decision to have the controller retriggered by an error in certain cases was made in order to promptly address important issues. However, in some cases where we are waiting for example for a resource, I’d to give it some time and requeue the controller. 

Agree, in this case with TypeFactory it makes sense to return an error. I'll review the code for any other missed error handling.

kramaranya: Updated 😌

controllers/rhmi/rhmi_controller.go
```diff
@@ -371,12 +371,18 @@ func (r *RHMIReconciler) Reconcile(ctx context.Context, request ctrl.Request) (c
...
+		}
 	}
 
 	// If the CR is being deleted, handle uninstall and return
 	if installation.DeletionTimestamp != nil {
-		return r.handleUninstall(installation, installType, request)
+		err := r.handleUninstall(installation, installType, request)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
```
KevFan, line 385:
```suggestion
		err := r.handleUninstall(installation, installType, request)
		if err != nil {
			return ctrl.Result{}, err
		}
		return retryRequeue, nil
```

Should we return if `err` is `nil`  also ? Otherwise I think the installation logic will be run during uninstallation 🤔

KevFan: Doesn't look to affect uninstallation though, so maybe not 🤔

kramaranya: Hmm, actually yes, the controller will not be requeued when `handleUninstall` returns nill, make sense to add `retryRequeue`. https://github.com/integr8ly/integreatly-operator/blob/bc737b41ae13ee7fdbfceb29abb3fe874bc63b46/controllers/rhmi/rhmi_controller.go#L374-L377 there as well..
Thanks for spotting it!

controllers/rhmi/rhmi_controller.go
```diff
@@ -842,30 +844,30 @@ func (r *RHMIReconciler) handleUninstall(installation *rhmiv1alpha1.RHMI, instal
...
 
 		log.Info("uninstall completed")
-		return ctrl.Result{}, nil
+		return nil
 	}
 
 	log.Info("updating uninstallation object")
 	// no finalizers left, update object
 	err = r.Update(context.TODO(), installation)
-	return retryRequeue, err
```
MStokluska, line 868: why is there no error handling on line 869

controllers/rhmi/rhmi_controller.go
```diff
@@ -995,7 +992,7 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
 			},
 		}
 		if exists, err := k8s.Exists(context.TODO(), r.Client, secret); err != nil {
-			return ctrl.Result{}, err
+			return err
```
MStokluska, line 995: please reformat to commonly used error handling

controllers/rhmi/rhmi_controller.go
```diff
@@ -1006,10 +1003,10 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
 			err = r.Status().Update(context.TODO(), installation)
 			if err != nil {
 				log.Infof("error updating status", l.Fields{"error": err.Error()})
-				return ctrl.Result{}, err
+				return err
 			}
 
-			return ctrl.Result{}, err
+			return nil
```
MStokluska, line 1009: So if secret is missing, we are going to continue with reconcile and by-pass the preflights. In here I would suggest to create your own error and return it instead.

controllers/rhmi/rhmi_controller.go
```diff
@@ -1067,10 +1064,9 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
 			err = r.Status().Update(context.TODO(), installation)
 			if err != nil {
 				log.Infof("error updating status", l.Fields{"error": err.Error()})
-				return result, err
+				return err
 			}
-
-			return result, nil
+			return nil
```
MStokluska, line 1069: same as above we need to fail here

controllers/rhmi/rhmi_controller.go
```diff
@@ -1098,17 +1094,17 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
 			err = r.Status().Update(context.TODO(), installation)
 			if err != nil {
 				log.Infof("error updating status", l.Fields{"error": err.Error()})
-				return result, err
+				return err
 			}
-			return result, err
+			return nil
```
MStokluska, line 1099: same as above

controllers/rhmi/rhmi_controller.go
```diff
@@ -1120,7 +1116,7 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
 	if err != nil {
 		log.Infof("error updating status", l.Fields{"error": err.Error()})
 	}
-	return result, nil
+	return nil
```
MStokluska, line 1119: same as above

controllers/rhmi/rhmi_controller.go
KevFan: > Hmm, actually yes, the controller will not be requeued when `handleUninstall` returns nill, make sense to add `retryRequeue`.
> 
> https://github.com/integr8ly/integreatly-operator/blob/bc737b41ae13ee7fdbfceb29abb3fe874bc63b46/controllers/rhmi/rhmi_controller.go#L374-L377
> 
> there as well..
> Thanks for spotting it!

Yes, I think it should be requeued at preflight checks on nil error also 👍

controllers/rhmi/rhmi_controller.go
```diff
@@ -817,7 +821,7 @@ func (r *RHMIReconciler) handleUninstall(installation *rhmiv1alpha1.RHMI, instal
 			if err != nil {
 				merr.Add(err)
 			}
-			return retryRequeue, nil
+			return nil
```
MStokluska, line 824: if we did return a Requeue - it means to me that at this point, the current controller loop should stop, and it should be requeued. By returning nil here we are saying "doesn't matter, just continue" - if we plan on changing logic like so, please ensure extended verification steps are in place to be sure it's working as expected.

controllers/rhmi/rhmi_controller.go
```diff
@@ -832,7 +836,7 @@ func (r *RHMIReconciler) handleUninstall(installation *rhmiv1alpha1.RHMI, instal
 			if err != nil {
```
MStokluska, line 836: can we have a look at above logic on 832 - if err > err != nil > !k8errIsNotFound looks a bit off.
If we are saying if err, I guess no point in checking if err != nil? WDYT? <not related to the PR but could be improved I think>

controllers/rhmi/rhmi_controller.go
```diff
@@ -979,10 +981,10 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
 		err := r.Status().Update(context.TODO(), installation)
 		if err != nil {
 			log.Infof("error updating status", l.Fields{"error": err.Error()})
-			return result, err
+			return err
 		}
 		log.Warning("preflight checks failed on useClusterStorage value")
-		return result, nil
+		return nil
```
MStokluska, line 987: I think in here, we should be returning custom error. Reason being is that we do not want to continue if useClusterStorage is set neither to false or true...

controllers/rhmi/rhmi_controller.go
kramaranya: It'll not skip it and continue - it will return nil and be requeued as on the line 387 the controller will be requeued if `handleUninstall` returns nil: https://github.com/integr8ly/integreatly-operator/blob/edbe2c72475b1b74ad724685fa540aabc301d366/controllers/rhmi/rhmi_controller.go#L383-L387

controllers/rhmi/rhmi_controller.go
MStokluska: Generally, anywhere where we return result previously, we need to return non-nil value to ensure the reconciler gets re-queued unless you are 100% that we do not need to re-queue but instead, can continue within same reconcile loop.

controllers/rhmi/rhmi_controller.go
KevFan: I dont think the reconcile will generally continue at the moment even without the custom error due to the following :
https://github.com/integr8ly/integreatly-operator/blob/edbe2c72475b1b74ad724685fa540aabc301d366/controllers/rhmi/rhmi_controller.go#L979-L980

https://github.com/integr8ly/integreatly-operator/blob/edbe2c72475b1b74ad724685fa540aabc301d366/controllers/rhmi/rhmi_controller.go#L372-L379

Though the logic probably should indicate an err rather than returning nil as you suggest 🤔

controllers/rhmi/rhmi_controller.go
```diff
@@ -842,30 +846,33 @@ func (r *RHMIReconciler) handleUninstall(installation *rhmiv1alpha1.RHMI, instal
...
 		}
 
 		if err := addon.UninstallOperator(context.TODO(), r.Client, installation); err != nil {
 			merr.Add(err)
-			return retryRequeue, merr
+			return merr
 		}
 
 		log.Info("uninstall completed")
-		return ctrl.Result{}, nil
```
MStokluska, line 862: It looks to me like on this line, when the uninstallation is completed, the intention was to stop the RHMI Controller entirely. By sending an empty Result + nil we are effectively saying - you're done, don't ever run again. 
This makes me think why was there an update on 871 at all. IMO the last step of this entire flow should be empty result and nil as it was but maybe I'm missing something. With your changes, it seems that we are going to be req-queueing the RHMI controller indefinitely.

controllers/rhmi/rhmi_controller.go
MStokluska: By saying reconcile will continue I mean it would return to rhmi controller and continue regardless of what's the story with preflight state, but Anna pushed other changes meanwhile so it's not relevant, but I think it would be better to return an error indicating what's going on rather than nil. 
Generally, I think it's better to be as informative as we can so if the failure is caused by an error, let's return an error.

PR #3302, opened 2023-07-06
MGDAPI-5890 changes to support 100mln redis bump

MStokluska:
# Issue link
https://issues.redhat.com/browse/MGDAPI-5890

# What
Bumped 100mln quota redis size to cache.m5.xlarge

# Verification steps
Must be done via OLM on CCS cluster with cluster storage set to false.
## Fresh install
- Via OLM, install 1.38.0 (you can use this index: 'quay.io/mstoklus/managed-api-service-index:1.38.0'
- Confirm that 100mln backend redis is now of cache.m5.xlarge size and that installation does not crash at any point

## Upgrade
- Via OLM, install rhoam 1.37.0 from this index: 'quay.io/mstoklus/managed-api-service-index:1.37.0' (it's matching prod)
- Quota must be 100mln
- Once installation completes, scale down RHOAM operator
- In backend redis CR add:
```
applyImmediately: true
maintenanceWindow: true
```
And change size to:
```
cache.m5.xlarge
```
- Wait for redis to successfully bump the instance on AWS
- Once redis reports completed, remove the applyImmediately flag from redis CR
- Scale back RHOAM
- Confirm that redis cr size has reverted to cache.m5.large but no action is taken on AWS side
- Apply the 1.38.0 index image from above (or build new index using this branch)
- Confirm that Redis CR size got bumped to m5.xlarge and no action is taken on aws side
- Confirm AWS backend worker redis remains at m5.xlarge

### DIFF
--- a/controllers/subscription/subscription_controller.go
+++ b/controllers/subscription/subscription_controller.go
@@ -19,8 +19,11 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/wait"
 
+	"github.com/integr8ly/cloud-resource-operator/apis/integreatly/v1alpha1"
 	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/apis/v1alpha1"
 	catalogsourceClient "github.com/integr8ly/integreatly-operator/pkg/resources/catalogsource"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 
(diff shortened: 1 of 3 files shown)


controllers/subscription/subscription_controller.go
```diff
@@ -115,9 +118,54 @@ func (r *SubscriptionReconciler) Reconcile(ctx context.Context, request ctrl.Req
...
 
+	// Remove code below post 1.38.0 delivery to production
+	// Temporary code to update backend redis 100mln to specific value
+	installation := &integreatlyv1alpha1.RHMI{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rhoam",
+			Namespace: "redhat-rhoam-operator",
+		},
+	}
+	err := r.Get(context.TODO(), k8sclient.ObjectKey{Name: installation.Name, Namespace: installation.Namespace}, installation)
```
trepel, line 129: This is fine I believe, just curious, why GetRhmiCr was not used to get RHMI CR here?

controllers/subscription/subscription_controller.go
```diff
@@ -115,9 +118,54 @@ func (r *SubscriptionReconciler) Reconcile(ctx context.Context, request ctrl.Req
...
+	if err != nil {
+		if !k8serr.IsNotFound(err) {
+			return ctrl.Result{}, err
+		}
+	}
+	// Only apply below logic for existing installations, we can't rely on toVersion not present due to possible race condition
+	// between sub and rhmi controller
+	if installation.Status.Version != "" {
```
trepel, line 137: Is it safe to assume that .Status is already in place at this point?

trepel:
Both the fresh install and upgrade worked as described. Code looks good to me too, just two questions but not a blocker for merging imo.

/lgtm

controllers/subscription/subscription_controller.go
MStokluska: It doesn't have to. I believe your concern is about nil pointer but even if status is not set, the code will not return a nil pointer.

controllers/subscription/subscription_controller.go
trepel: No need, it is a tmp code and it is clear what it is going on. I just wanted to make sure I am not missing something

controllers/subscription/subscription_controller.go
trepel: Ok, thanks. Yes, I was worried about the nil pointer.

PR #3308, opened 2023-07-10
MGDAPI-5874 Addition of PO resources stability e2e

Patryk-Stefanski:
# Issue link
[MGDAPI-5874](https://issues.redhat.com/browse/MGDAPI-5874)

# What
The addition of a new e2e test for resource stability, this tests checks that resources arent being continuously updated.  Threescale metrics monitor was used in this case.

# Verification steps
1. Checkout this branch 
2.  Install RHOAM
3.  Verify test passes by running 
```
INSTALLATION_TYPE=managed-api TEST="Verify package operator" make test/e2e/single
```
4. Verify test fails by running the command below, but when the `waiting 30 seconds` log shows update the `threescale-operator-controller-manager-metrics-monitor` in the `redhat-rhoam-3scale-operator` namespace. The test should fail
```
INSTALLATION_TYPE=managed-api TEST="Verify package operator" make test/e2e/single
```

### DIFF
--- a/test/common/verify_po_resource_stability.go
+++ b/test/common/verify_po_resource_stability.go
@@ -1,0 +1,55 @@
+package common
+
+import (
+	goctx "context"
+	"fmt"
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"time"
+)
+
+var (
+	threescaleOperatorNs = "redhat-rhoam-3scale-operator"
+	serviceMonitorName   = "threescale-operator-controller-manager-metrics-monitor"
+)
+
+func TestPackageOperatorResourceStability(t TestingTB, ctx *TestingContext) {
+	// Fetch the initial state of the resource
+	resource, err := getServiceMonitor(ctx, threescaleOperatorNs, serviceMonitorName)
+	if err != nil {
+		t.Fatalf("Failed to fetch resource: %v", err)
+	}
+
+	t.Log("Waiting 30 seconds")
+	time.Sleep(30 * time.Second)
+
+	// Update resource, will fail if resource has been modified after fetch
+	err = ctx.Client.Update(goctx.TODO(), resource)
+	if err != nil {
+		t.Fatalf("Failed to update resource: %v", err)
+	}
+
+}
+
+func getServiceMonitor(ctx *TestingContext, nameSpace, serviceName string) (*monitoringv1.ServiceMonitor, error) {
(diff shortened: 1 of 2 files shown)


test/common/verify_po_resource_stability.go
```diff
@@ -0,0 +1,55 @@
...
+	// Iterate over the service monitors to find the one with the specified name
+	for _, sm := range serviceMonitors.Items {
+		if sm.Name == serviceName {
+			return sm, nil
+		}
+	}
+
+	// If the specific service monitor is not found, return an error
+	return nil, fmt.Errorf("ServiceMonitor %s not found in namespace %s", serviceName, nameSpace)
+}
```
carlkyrillos, line 55: No pressure to accept this change but you could simplify things by getting the ServiceMonitor directly since we have the name and namespace:

```suggestion
func getServiceMonitor(ctx *TestingContext, namespace, serviceName string) (*monitoringv1.ServiceMonitor, error) {
	sm := &monitoringv1.ServiceMonitor{
		ObjectMeta: metav1.ObjectMeta{
			Name:      serviceName,
			Namespace: namespace,
		},
	}
	err := ctx.Client.Get(goctx.TODO(), k8sclient.ObjectKey{Name: sm.Name, Namespace: sm.Namespace}, sm)
	if err != nil {
		return nil, fmt.Errorf("ServiceMonitor %s not found in namespace %s", serviceName, namespace)
	}

	return sm, nil
}
```

test/common/verify_po_resource_stability.go
```diff
@@ -0,0 +1,55 @@
+package common
+
+import (
+	goctx "context"
+	"fmt"
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
```
carlkyrillos, line 6: You would also need to add an import for metav1 if changing the `getServiceMonitor()` function
```suggestion
	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

carlkyrillos (commented):
I added a few suggested changes. Also I can't get the test to fail after updating the ServiceMonitor, any idea what I'm doing wrong? @Patryk-Stefanski

Patryk-Stefanski:
hmm not to sure why its not failing, ill double check it tomorrow.

den-rgb:
should this be testing the other monitoring namespace?: 
image

den-rgb:
Test now fails when rhobs serviceMonitor is modified during test run.
/lgtm

PR #3331, opened 2023-07-27
MGDAPI-5929 - Use correct ingress router to determine IP for 3scale dns alerts

valerymo:
# Issue link
https://issues.redhat.com/browse/MGDAPI-5929
Use the correct ingress router to determine IP for 3scale dns alerts

# Verification steps
- Check that ***DNSByPass*** alerts are not firing in Custom Domain Rhoam installation
- Regression - check alerts in regular (Not custom domain) Rhoam  installation

- **References**:  
[Configuring custom domain](https://integreatly-operator.readthedocs.io/en/latest/products/custom_domain/)

### DIFF
--- a/pkg/products/threescale/reconciler.go
+++ b/pkg/products/threescale/reconciler.go
@@ -390,25 +390,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
 		return phase, err
 	}
 
-	ingressRouterService, err := customDomain.GetIngressRouterService(ctx, serverClient)
-	if err != nil || len(ingressRouterService.Status.LoadBalancer.Ingress) == 0 {
-		errorMessage := "failed to retrieve ingress router service"
-		r.log.Error(errorMessage, err)
-		phase = integreatlyv1alpha1.PhaseFailed
-		events.HandleError(r.recorder, installation, phase, errorMessage, err)
-		return phase, fmt.Errorf("%s: %v", errorMessage, err)
-	}
-
-	ips, err := customDomain.GetIngressRouterIPs(ingressRouterService.Status.LoadBalancer.Ingress)
-	if err != nil {
-		errorMessage := "failed to retrieve ingress router ips"
-		r.log.Error(errorMessage, err)
-		phase = integreatlyv1alpha1.PhaseFailed
-		events.HandleError(r.recorder, installation, phase, errorMessage, err)
-		return phase, fmt.Errorf("%s: %v", errorMessage, err)
-	}
-
-	phase, err = r.ping3scalePortals(ctx, serverClient, ips)
+	phase, err = r.ping3scalePortals(ctx, serverClient)
 	if err != nil || phase != integreatlyv1alpha1.PhaseCompleted {
 		errorMessage := "failed pinging 3scale portals through the ingress cluster router"
 		r.log.Error(errorMessage, err)
@@ -3308,7 +3290,7 @@ func (r *Reconciler) findCustomDomainCr(ctx context.Context, serverClient k8scli
 	return integreatlyv1alpha1.PhaseFailed, fmt.Errorf("finding CustomDomain CR failed: %v", err)
 }
 
-func (r *Reconciler) ping3scalePortals(ctx context.Context, serverClient k8sclient.Client, ips []net.IP) (integreatlyv1alpha1.StatusPhase, error) {
+func (r *Reconciler) ping3scalePortals(ctx context.Context, serverClient k8sclient.Client) (integreatlyv1alpha1.StatusPhase, error) {
 	format := "failed to retrieve %s 3scale route"
 	portals := map[string]metrics.PortalInfo{}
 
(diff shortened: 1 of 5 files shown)


pkg/products/threescale/reconciler.go
```diff
@@ -390,30 +390,46 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
 	}
 
-	ingressRouterService, err := customDomain.GetIngressRouterService(ctx, serverClient)
-	if err != nil || len(ingressRouterService.Status.LoadBalancer.Ingress) == 0 {
-		errorMessage := "failed to retrieve ingress router service"
-		r.log.Error(errorMessage, err)
-		phase = integreatlyv1alpha1.PhaseFailed
-		events.HandleError(r.recorder, installation, phase, errorMessage, err)
-		return phase, fmt.Errorf("%s: %v", errorMessage, err)
+	threescaleMasterRoute, err := r.getThreescaleRoute(ctx, serverClient, "system-master", nil)
```
briangallagher, line 393: We can not use the ingress-router from the the system-master route for the system provider and system-develoepr. The ingress router may be different on those

pkg/products/threescale/reconciler.go
```diff
@@ -390,30 +390,46 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
 
-	ips, err := customDomain.GetIngressRouterIPs(ingressRouterService.Status.LoadBalancer.Ingress)
-	if err != nil {
-		errorMessage := "failed to retrieve ingress router ips"
-		r.log.Error(errorMessage, err)
-		phase = integreatlyv1alpha1.PhaseFailed
-		events.HandleError(r.recorder, installation, phase, errorMessage, err)
-		return phase, fmt.Errorf("%s: %v", errorMessage, err)
+	listOfIngresServicesNames := []string{"router-default"}
+	if customDomainActive {
```
briangallagher, line 399: We should not only do this if custom domain active. There maybe another ingress router even if this is false. Probably unlikely but we should check anyway

pkg/products/threescale/reconciler.go
```diff
@@ -390,30 +390,46 @@ func (r *Reconciler) Reconcile(ctx context.Context, installation *integreatlyv1a
...
+		// get ingres custom domain service (LoadBalancer, router) name from 3scale route
+		if len(threescaleMasterRoute.Status.Ingress) > 1 {
```
briangallagher, line 401: We can just take the first item in the array. If there are multiple Ingress that just means the route is on multiple. The first in the list should always work

valerymo: the first item - is default . It's always "router-default". Here we are checking if we have more then one router (len > 1), and in this case we are taking the second one, that is CustomDomain

valerymo: The first in the list - is always working

briangallagher:
I think we should add the logic around [here](https://github.com/valerymo/integreatly-operator/blob/db756b1e77d41fc0bb3e9d0cce1864982751a59d/pkg/products/threescale/reconciler.go#L3342) 

After we get each route, look up the ingress router and then get the IP and use that for the ping.
We should assume that the ingress router could be different for each one of the 3scale routes - this is unlikely but possible.,

valerymo:
> I think we should add the logic around [here](https://github.com/valerymo/integreatly-operator/blob/db756b1e77d41fc0bb3e9d0cce1864982751a59d/pkg/products/threescale/reconciler.go#L3342)
> 
> After we get each route, look up the ingress router and then get the IP and use that for the ping. We should assume that the ingress router could be different for each one of the 3scale routes - this is unlikely but possible.,

OK, I will try change it. Will do on separate commit. Thank you

valerymo:
> I think we should add the logic around [here](https://github.com/valerymo/integreatly-operator/blob/db756b1e77d41fc0bb3e9d0cce1864982751a59d/pkg/products/threescale/reconciler.go#L3342)
> 
> After we get each route, look up the ingress router and then get the IP and use that for the ping. We should assume that the ingress router could be different for each one of the 3scale routes - this is unlikely but possible.,

Done. Logic was moved to ping3scalePortals. It impacted on unittest, but tests as seems to me do nothing almost, and I commented it; as result - small  (-0.06%) reducing of  coverage. Maybe we can accept it?

pkg/products/threescale/reconciler.go
valerymo: Logic was moved to ping3scalePortals

pkg/products/threescale/reconciler.go
valerymo: Logic was changed and moved ping3scalePortals

pkg/products/threescale/reconciler.go
valerymo: Logic was changed and moved to ping3scalePortals.

pkg/products/threescale/reconciler.go
```diff
@@ -3308,50 +3290,58 @@ func (r *Reconciler) findCustomDomainCr(ctx context.Context, serverClient k8scli
...
+			prefix = "3scale-admin."
+		}
+		route, err := r.getThreescaleRoute(ctx, serverClient, threeScaleService, func(r routev1.Route) bool {
+			return strings.HasPrefix(r.Spec.Host, prefix)
+		})
+		if err != nil || route == nil {
+			return integreatlyv1alpha1.PhaseFailed, fmt.Errorf(format, threeScaleService)
+		}
+		ingressRouters := []string{"router-default"}
+		if len(route.Status.Ingress) > 1 {
```
briangallagher, line 3316: I don't think we can assume that the first route will be the default one. On the latest OHSS customer issue this is not the case. The 3scale route only has one ingress on it and it's not the default one.

valerymo: before write it i looked at 3scale, and seems "router-default" is always present .. But I will look how to handle it

pkg/products/threescale/reconciler.go
```diff
@@ -3308,50 +3290,58 @@ func (r *Reconciler) findCustomDomainCr(ctx context.Context, serverClient k8scli
...
 	portals := map[string]metrics.PortalInfo{}
+	ips := make(map[string][]net.IP)
+	metricsLabelMap := make(map[string]string)
+	metricsLabelMap[labelRouteToSystemDeveloper] = metrics.LabelSystemDeveloper
+	metricsLabelMap[labelRouteToSystemProvider] = metrics.LabelSystemProvider
+	metricsLabelMap[labelRouteToSystemMaster] = metrics.LabelSystemMaster
+
+	for _, threeScaleService := range [3]string{labelRouteToSystemDeveloper, labelRouteToSystemProvider, labelRouteToSystemMaster} {
+		var prefix string
+		if threeScaleService == labelRouteToSystemDeveloper {
```
briangallagher, line 3304: We need to return to original filtering logic

pkg/products/threescale/reconciler_test.go
```diff
@@ -2948,6 +2964,21 @@ func TestReconciler_ping3scalePortals(t *testing.T) {
 			},
 		},
 	}
+	openshiftIngress := corev1.Service{
```
austincunningham, line 2967: @valerymo do we need to define the same object twice can we do without one of these.

austincunningham: never mind looks like they are in different functions

pkg/products/threescale/reconciler.go
valerymo: done, as suggested:  https://github.com/integr8ly/integreatly-operator/pull/3338

pkg/products/threescale/reconciler_test.go
valerymo: Thank you for help @austincunningham

carlkyrillos:
I reviewed the changed files and verified that the `DnsBypassThreeScaleUI` alerts don't fire when doing a custom domain installation.
/lgtm

carlkyrillos:
If we later want to configure the 3scale routes to only be serviceable by the custom domain IngressController,  this `oc` command can be used to update the default IngressController to filter out the 3scale routes:
```bash
oc patch ingresscontroller default -n openshift-ingress-operator --type=json -p '[{"op":"replace", "path":"/spec/routeSelector/matchExpressions", "value": [{"key": "zync.3scale.net/route-to", "operator": "DoesNotExist"}]}]'
```

CC: @briangallagher

austincunningham:
/lgtm
After reviewing the code changes, since the rebase and I agree the verification stands as there are no changes to the actual functional code. I will approve once the e2e pass

valerymo:
retested on custom domain  (cluster vmogilev01) and regression - for non custom domain (cluster vmogilev02). No DNS alerts

PR #3359, opened 2023-08-23
MGDAPI-5992 temporary workaround in tests

trepel:
# Issue link

https://issues.redhat.com/browse/MGDAPI-5992

# What

Temporary workaround for https://issues.redhat.com/browse/MGDAPI-5992. Rhsso- and UserSso- vailability2hto1dErrorBudgetBurn alerts will be ignored.

# Verification steps

Have a cluster with RHOAM installed. In case of no alerts firing the tests should just pass
`INSTALLATION_TYPE=managed-api DESTRUCTIVE=true LOCAL=false TEST="Verify Alerts are not firing during or after installation apart from DeadMansSwitch" make test/e2e/single`
`INSTALLATION_TYPE=managed-api DESTRUCTIVE=true LOCAL=false TEST="C01" make test/e2e/single`

I am not sure how to make **ErrorBudgetBurn alerts fire so I recommend to deploy workload-web-app and do some GET requests against the 3scale Product which should make ThreeScaleContainerHighMemory alert to fire. Then run 
`INSTALLATION_TYPE=managed-api DESTRUCTIVE=true LOCAL=false TEST="Verify Alerts are not firing during or after installation apart from DeadMansSwitch" make test/e2e/single`
It should fail.

Add "ScaleContainerHigh" into `ignoredAlertsPatterns` array and run it again. This time the test should pass.

If you are aware of a simpler way of making some alert fire (without making RHMI CR think that the installation is in progress) you can do the verification that way.

### DIFF
--- a/test/common/alerts_firing.go
+++ b/test/common/alerts_firing.go
@@ -3,6 +3,7 @@ package common
 import (
 	"encoding/json"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


martinhesko:
tested this on a cluster where `ThreeScaleContainerHighMemory` was firing, added it to the ignore list and both of the tests ignored it

PR #3361, opened 2023-08-23
MGDAPI-5996 Alert check not waiting for RHOAM installation to complete

trepel:
# Issue link

https://issues.redhat.com/browse/MGDAPI-5996

# What

We would like to use alerts-check.sh to monitor alerts during rhoam installation. In order to do that the script must not wait for RHOAM install to complete but only wait for prometheus to spin up.
I also slightly updated the output so that Openshift alerts and RHOAM alerts are not mixed together

# Verification steps

Have a cluster with RHOAM installed. Do `oc login ...` into the cluster. Run `./alerts-check.sh`. It should start watching alerts every 5 seconds. Then make the RHMI CR think that the installation is not complete, e.g.:
`oc scale deployment/threescale-operator-controller-manager-v2 -n redhat-rhoam-3scale-operator --replicas=0`
`oc scale dc/system-app -n redhat-rhoam-3scale --replicas=0`

Run `./alerts-check.sh` again. The script should behave the same as before.

Revert the scaling:
 `oc scale deployment/threescale-operator-controller-manager-v2 -n redhat-rhoam-3scale-operator --replicas=1`
`oc scale dc/system-app -n redhat-rhoam-3scale --replicas=3`

### DIFF
--- a/scripts/alerts-check.sh
+++ b/scripts/alerts-check.sh
@@ -22,10 +22,9 @@ ROUTE="http://localhost:9090/api/v1/alerts"
 NAMESPACE_PREFIX="${NAMESPACE_PREFIX:-$(oc get RHMIs --all-namespaces -o jsonpath='{.items[0].spec.namespacePrefix}')}"
 TOKEN=$(oc whoami --show-token)
 # wait for monitoring route to appear and have host populated
-# wait for rhoam install to be complete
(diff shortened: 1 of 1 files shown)


martinhesko:
scaled down the pods, checked with `oc get rhmi rhoam -n redhat-rhoam-operator -o json` and 3scale was marked as `failed` and installation as `in progress`, the alert check script was still working as expected

PR #3376, opened 2023-09-26
MGDAPI-5703 removing alerts installation, upgrade and missing metrics from openshift-monitoring namespace

austincunningham:
# Issue link

MGDAPI-5703 

# What

removing alerts installation, upgrade and missing metrics from openshift-monitoring namespace
# Verification steps

- install the rhoam addon via the ocm ui
- Check the rhoam alerts are in the openshift monitoring namespace in the openshift-monitoring prometheus ui. You can use port-forward to access through http://localhost:9090/alerts?search=rhoam
```bash
oc port-forward -ns openshift-monitoring service/prometheus-operated 9090:9090
```
image

- Patch the rhoam csv with this operator image `quay.io/austincunningham/managed-api-service: v1.39.0` 
- You should see the rhoam alerts are removed from the Prometheus monitoring. Only rhoam related alert present will be ElevatingClusterAdminRHOAMSRE which is unaffected by this change
image


controllers/rhmi/prometheusRules.go
```diff
@@ -19,98 +17,37 @@ var (
...
-				},
-			},
 		},
-		{
-			AlertName: fmt.Sprintf("%s-telemetry", installationName),
+	}
+	err = apiClient.Delete(ctx, upgradeAlert)
+	if err != nil {
+		return err
+	}
```
carlkyrillos, line 41:
```suggestion
	if err != nil && !k8serr.IsNotFound(err) {
		return err
	}
```
I think we should update the error check because I did see `not found` errors in the logs after the PrometheusRules had been deleted.

controllers/rhmi/prometheusRules.go
```diff
@@ -19,98 +17,37 @@ var (
...
 	}
-
-	return &resources.AlertReconcilerImpl{
-		ProductName:  "installation",
-		Installation: installation,
-		Log:          log,
-		Alerts:       alerts,
+	err = apiClient.Delete(ctx, metricMissingAlert)
+	if err != nil {
+		return err
```
carlkyrillos, line 50:
```suggestion
	if err != nil && !k8serr.IsNotFound(err) {
		return err
```
I think we should update the error check because I did see `not found` errors in the logs after the PrometheusRules had been deleted.

controllers/rhmi/prometheusRules.go
```diff
@@ -19,98 +17,37 @@ var (
...
-				},
-			},
-		},
-		{
-			AlertName: fmt.Sprintf("%s-missing-metrics", installationName),
+	}
+	err = apiClient.Delete(ctx, installationAlert)
+	if err != nil {
+		return err
+	}
```
carlkyrillos, line 31:
```suggestion
	if err != nil && !k8serr.IsNotFound(err) {
		return err
	}
```
I think we should update the error check because I did see `not found` errors in the logs after the PrometheusRules had been deleted.

carlkyrillos (commented):
@austincunningham I completed the verification steps and confirmed the alerts were removed. I do think some of the error checks need to be tweaked though - see comments.

controllers/rhmi/prometheusRules.go
austincunningham: yep agree

controllers/rhmi/prometheusRules.go
austincunningham: yep agree

carlkyrillos:
Reviewed latest changes
/lgtm

PR #3403, opened 2023-11-07
MGDAPI-6113 Fix removeGrafanaOperatorAlerts() to work for Internal rhoam  (rhmi)

valerymo:
# Issue link
Jira: https://issues.redhat.com/browse/MGDAPI-6113

# What
Fix removeGrafanaOperatorAlerts method to work for Internal rhoam (rhoami)

# Verification steps
Please see related PR: https://github.com/integr8ly/integreatly-operator/pull/3398

### DIFF
--- a/pkg/products/grafana/prometheusRules.go
+++ b/pkg/products/grafana/prometheusRules.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"github.com/integr8ly/integreatly-operator/pkg/resources"
+	"github.com/integr8ly/integreatly-operator/pkg/resources/k8s"
 	l "github.com/integr8ly/integreatly-operator/pkg/resources/logger"
 	monv1 "github.com/rhobs/obo-prometheus-operator/pkg/apis/monitoring/v1"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/products/grafana/prometheusRules.go
```diff
@@ -60,14 +61,22 @@ func (r *Reconciler) newAlertReconciler(logger l.Logger, installType string, nam
 }
 
 func (r *Reconciler) removeGrafanaOperatorAlerts(installType string, ctx context.Context, apiClient k8sclient.Client) error {
-	oboNamespace := "redhat-" + resources.InstallationNames[installType] + "-operator-observability"
+	installationName := resources.InstallationNames[installType]
```
MStokluska, line 64: This looks a bit overcomplicated, WDYT if in removeGrafanaOperatorAlerts we would accept "nsPrefix" string, and where the func is called we can get the nsPrefix from "installation.Spec.NamespacePrefix" then annotate the nsPrefix with "-operator-observability"

valerymo: sure, thanks a lot for this comment.  Updated.

valerymo:
@MStokluska  . Update retested on cluster vmo03. Old alerts - deleted. Thank you for review!

trepel:
I ran the upgrade scenario and it went ok.
/lgtm

PR #3404, opened 2023-11-07
MGDAPI-4918 Allow for specifying testing-idp password via env var

trepel:
# Issue link

https://issues.redhat.com/browse/MGDAPI-4918

# What

Populate Testing IDP password from env var preferably and only use DefaultPassword as a fallback if env var is not defined.

# Verification steps

- Install RHOAM and do "oc login" into the cluster
- "export TESTING_IDP_PASSWORD=yourpassword"
- "PASSWORD=$TESTING_IDP_PASSWORD DEDICATED_ADMIN_PASSWORD=$TESTING_IDP_PASSWORD ./scripts/setup-sso-idp.sh"

Execute some of the affected tests:
H07; H03; H11; H34; H35; C19; A16; "Verify Network Policy allows cross NS access to SVC"; B01B; H24; B04; B03; B08B;

Example:
 TESTING_IDP_PASSWORD=yourpassword INSTALLATION_TYPE=managed-api LOCAL=false TEST="A16" make test/e2e/single

Note that H03, H07 might fail since they are a bit flaky.

### DIFF
--- a/test/common/3Scale_user_promotion.go
+++ b/test/common/3Scale_user_promotion.go
@@ -64,7 +64,7 @@ func Test3ScaleUserPromotion(t TestingTB, ctx *TestingContext) {
 		t.Fatal(err)
 	}
 
-	err = loginToThreeScale(t, host, dedicatedAdminUser, DefaultPassword, TestingIDPRealm, httpClient)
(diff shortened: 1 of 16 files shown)


austincunningham:
# Verification
functional test pass 
image
And env var set
image

/lgtm

PR #3422, opened 2023-12-06
MGDAPI-6163 move grafana dep to RHOAM instead of PO temporarily

MStokluska:
# Issue link

# What

# Verification steps


MStokluska:
happy for prow checks to be overwritten due to the fact that we had 2 successfuly runs of them, the only added thing now are rbacs which aren't checked as part of local run (prow).

PR #3432, opened 2024-01-18
MGDAPI-6131 Requeued if pagerduty secret is missing

trepel:
# Issue link

https://issues.redhat.com/browse/MGDAPI-6131

# What

If pagerduty secret was missing (e.g. OCM/Hive was slow in creating it for whatever reason) the RHOAM installation got stuck. This fixes is by returing "requeue" result in case of error in getting the secret and in case of non-existence of the secret. This triggers another reconcile(s) later until OCM/Hive creates the secret.

# Verification steps

- Get an OSD cluster and `oc login...` into it
-  Checkout this PR
-  Prepare the cluster for rhoam installation: `make cluster/prepare/local`
- Remove the pagerduty secret from redhat-rhoam-operator namespace:
  - `oc delete secret redhat-rhoam-pagerduty -n redhat-rhoam-operator`
- Remove `cluster/prepare/pagerduty` make target from `code/run` make target in Makefile:
  - https://github.com/integr8ly/integreatly-operator/blob/master/Makefile#L157
- Run the operator locally: `make code/run`
- Watch the RHMI CR and wait for it to report preflight checks failing due to missing pagerduty secret
  - `oc get rhmi rhoam -n redhat-rhoam-operator -o json`
  - `.status.preflightMessage` to contain "Could not find redhat-rhoam-pagerduty secret in redhat-rhoam-operator namespace"
- Create a pagerduty secret: `make cluster/prepare/pagerduty`
- Watch the RHMI CR and wait for preflight checks to pass:
  - `oc get rhmi rhoam -n redhat-rhoam-operator -o json`
  - `.status.preflightMessage` to contain "preflight checks passed"

### DIFF
--- a/controllers/rhmi/rhmi_controller.go
+++ b/controllers/rhmi/rhmi_controller.go
@@ -795,7 +795,7 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
 				return ctrl.Result{}, err
 			}
 
-			return ctrl.Result{}, err
+			return result, nil
 		}
 		log.Infof("found required secret", l.Fields{"secret": secretName})
 		eventRecorder.Eventf(installation, "Normal", rhmiv1alpha1.EventPreflightCheckPassed,


controllers/rhmi/rhmi_controller.go
```diff
@@ -781,7 +781,7 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
 			},
 		}
 		if exists, err := k8s.Exists(context.TODO(), r.Client, secret); err != nil {
-			return ctrl.Result{}, err
+			return result, err
```
MStokluska, line 784: Just a side note.
This isn't really necessary. Reconcilers do re-trigger when:
- there's an error 
- when you tell them to in result{requeue}
Since in here, error is always there, the requeue shouldn't be happening because we are going to reconcile again due to error.

controllers/rhmi/rhmi_controller.go
```diff
@@ -792,10 +792,10 @@ func (r *RHMIReconciler) preflightChecks(installation *rhmiv1alpha1.RHMI, instal
 			err = r.Status().Update(context.TODO(), installation)
 			if err != nil {
 				log.Infof("error updating status", l.Fields{"error": err.Error()})
-				return ctrl.Result{}, err
+				return result, err
 			}
 
-			return ctrl.Result{}, err
+			return result, nil
```
MStokluska, line 798: Good spot! This is our offender :)

controllers/rhmi/rhmi_controller.go
trepel: Good point! I update both occurrences back to `ctrl.Result{}, err` since that makes more sense, thanks for explanation

PR #3437, opened 2024-02-14
MGDAPI-6243 Update A33 test to work with OSD v4.15

trepel:
# Issue link

https://issues.redhat.com/browse/MGDAPI-6243

# What

Change of section class in OSD v4.15 broke the test so the test now has to determine the cluster version first and populate the sectionSelector with proper value based on the cluster version. 

# Verification steps

Install RHOAM twice -> on OSD v4.15.0-rc.5 and OSD v4.14.10 (latest 4.14 at the time of writing this)
Run the A33 against both clusters, it should pass for each

```
LOCAL=false INSTALLATION_TYPE=managed-api TESTING_IDP_PASSWORD="HIDDEN" TEST=A33 make test/e2e/single
```

### DIFF
--- a/pkg/resources/cluster/cluster.go
+++ b/pkg/resources/cluster/cluster.go
@@ -108,5 +108,5 @@ func GetClusterVersion(cr *configv1.ClusterVersion) (string, error) {
 	if cr.Status.Desired.Version != "" {
 		return cr.Status.Desired.Version, nil
 	}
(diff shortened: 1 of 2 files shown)


carlkyrillos:
Reviewed code changes and confirmed that `A33` is passing on both OSD v4.14 and v4.15.

/lgtm

trepel:
Thanks @carlkyrillos !

/approve

PR #3466, opened 2024-06-05
MGDAPI-6378 Updated oc/ubi/centos-repo versions in test images

trepel:
# Issue link
https://issues.redhat.com/browse/MGDAPI-6378

# What
Version updates in test related Dockerfiles

# Verification steps

Eye review
(Optional) You can build the images and run the test suite using them against RHOAM but I did that using the pipeline so I think doing it again is optional


MStokluska:
@trepel does this need to go in? If so, can we please rebase

trepel:
> @trepel does this need to go in? If so, can we please rebase

Not necessarily since the important changes are in `master` bramch already. To be honest I can't recall now why I changed that builder image tags. The one that is used currently is based on Centos whereas the proposed one is based on RHEL.

PR #3499, opened 2024-09-09
MGDAPI-6437 redis service update

austincunningham:
# Issue link

https://issues.redhat.com/browse/MGDAPI-6437

# What

redis service update

# Verification steps
Hard to verify as all new redis will have the patch already applied.


austincunningham:
won't do

PR #3513, opened 2024-10-24
MGDAPI-6451 update ingress controller logic

MStokluska:
# Issue link
Fix an issue where ingressControllers spec.domain isn't actual domain but part of dns

# What
When coming from osd 4.13 to 4.14, openshift automatically creates ingressController of the same name as the custom domain and deprecates the custom domain. Our goal here is to find custom domain name and find ingressController of the same name -> then check the available condition of ingressController
On fresh installs, the spec.domain in ingressController is expected to be of the same value as routing subdomain in rhmi cr

# Verification steps
Already done on cluster with Tomas, please review code.


carlkyrillos:
Reviewed code changes and everything looks good
/approve

PR #3634, opened 2025-10-12
MGDAPI-6844 network policy test failing due to missing openshift-session-token

valerymo:
# Issue link
https://issues.redhat.com/browse/MGDAPI-6844

# What
Fix Tests: 
- Network Policy
- B03
- B04
Use an API Bearer token instead of Browser's cookie token.

**integreatly-operator-test-harness** Image change - ocp client (oc comman) added, as it required to get token (see commens below)

#### Approach for Token-Based Testing
We now use a single privileged bearer token for authentication, and impersonate the target user for authorization.
**Details**:
- We use one real bearer token with admin-level privileges (e.g., from OAUTH_BEARER_TOKEN or oc whoami -t).
(This can be a kubeadmin token or any cluster-admin token.)
- We do not mint individual tokens for developer or dedicated-admin users.
   Instead, we include impersonation headers in API requests:
       Impersonate-User: <target-username>
       Impersonate-Group: <group-name>
- The API server authenticates the request using the admin token, but evaluates RBAC permissions as the impersonated user and groups.
- We still run the OAuth web flow to ensure the test user exists and is reconciled. (However, the resulting cookie is not used for API authentication.)

**Examples**:
- Dedicated Admin Test:
Impersonate user with groups ["system:authenticated", "dedicated-admins"] → Should have permissions

- Developer Negative Test (B03):
Impersonate user with groups ["system:authenticated"] → Should receive 403 Forbidden

# Verification steps
- create OSD clister, install RHOAM (local), install IDP
- Run tests (required + some regression) : 
`TEST="Network Policy|B03|B04|B09|A16|H03|H07|H11" make test/e2e/single`

```
TEST="Network Policy|B03|B04|B09|A16|H03|H07|H11" make test/e2e/single
cd test && go clean -testcache && go test ./functional -ginkgo.focus="Network Policy|B03|B04|B09|A16|H03|H07|H11.*" -test.v -ginkgo.v -ginkgo.progress -timeout=80m
time="2025-10-15T11:27:18+03:00" level=info msg="calling reset on all prometheus gauge vectors" custom_metrics=StartGaugeVector
=== RUN   TestAPIs
Running Suite: Functional Test Suite - /Users/valerymogilevsky/go/integreatly-operator/test/functional
.....
.....
[ReportAfterSuite] PASSED [0.007 seconds]
------------------------------

Ran 8 of 55 Specs in 1110.990 seconds
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 47 Skipped

```

### DIFF
--- a/test/common/user_dedicated_admin_permissions.go
+++ b/test/common/user_dedicated_admin_permissions.go
@@ -4,7 +4,9 @@ import (
 	goctx "context"
 	"fmt"
 	"io/ioutil"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"strings"
+	"time"
 
 	integreatlyv1alpha1 "github.com/integr8ly/integreatly-operator/api/v1alpha1"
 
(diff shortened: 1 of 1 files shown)


Dockerfile.functional
```diff
@@ -36,6 +36,10 @@ RUN yum install -y wget \
     && rm -f google-chrome-stable_current_*.rpm \
     && yum clean all
 
+ENV OCP_VERSION=stable-4.19
```
austincunningham: Not sure I like adding the openshift version here. Leaves a dependency that will need to be maintained.  How about we use the latest instead https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/linux/oc.tar.gz

